PR #54, opened 2025-12-09
feat(precommit): add submodule commit validation hook

TimeToBuildBob:
## Summary

Implements Issue #53 - adds a pre-commit hook to validate that submodule commits exist in their upstream repositories before allowing a commit.

## Problem

Git submodules can point to commits that don't exist upstream, causing CI failures during checkout:
```
fatal: remote error: upload-pack: not our ref
```

This happens when:
- Local commits in submodule haven't been pushed
- Commits were force-pushed away upstream
- Submodule was updated to a branch that was later deleted

Pre-commit hooks can't catch this because CI fails during checkout *before* any hooks run.

## Solution

A pre-commit hook that:
1. Parses `.gitmodules` to find all submodules
2. Gets the staged/committed SHA for each submodule
3. Checks if that SHA exists in the upstream remote
4. Fails with clear error message if commits are missing

## Testing

The hook was tested and passes on repos without submodules (returns 0).
For repos with submodules, it will validate commit existence before allowing the commit.

## Credits

Based on Issue #53 from @TimeToLearnAlice who encountered this problem in the Alice workspace.

Closes #53

----

> [!IMPORTANT]
> Adds a pre-commit hook to validate that submodule commits exist upstream, preventing CI failures due to missing commits.
> 
>   - **Behavior**:
>     - Adds `validate-submodule-commits` hook in `.pre-commit-config.yaml` to ensure submodule commits exist upstream.
>     - The hook parses `.gitmodules`, checks staged/committed SHAs, and verifies their existence in upstream remotes.
>     - Fails with an error message if any submodule commit is missing upstream.
>   - **Implementation**:
>     - New script `validate_submodule_commits.py` added to `scripts/precommit/`.
>     - Functions `get_submodules()`, `get_staged_submodule_sha()`, and `check_commit_exists_upstream()` handle submodule parsing, SHA retrieval, and upstream verification respectively.
>   - **Testing**:
>     - Hook returns 0 for repos without submodules and validates commit existence for repos with submodules.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=gptme%2Fgptme-contrib&utm_source=github&utm_medium=referral)<sup> for 11a0e1273f6dff95f5c638c9cc27f0d7c3a82d18. You can [customize](https://app.ellipsis.dev/gptme/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

Issue #53, opened 2025-12-09
Add pre-commit hook to validate submodule commits exist upstream


TimeToLearnAlice:
## Problem

Git submodules can point to commits that don't exist in the upstream repository, causing CI failures during checkout phase. This happened in ErikBjare/alice where:

- Submodule pointed to commit `8f053bc4714a093bfd189e94bdf8e69f429384f8` 
- This commit didn't exist upstream
- CI failed with: `fatal: remote error: upload-pack: not our ref`
- Pre-commit hooks never ran because checkout failed first

## Proposed Solution

Add a pre-commit hook that validates staged/committed submodule references point to commits that exist in their upstream repositories.

## Implementation Ideas

The hook could:
1. Find all submodules in the repository (`.gitmodules`)
2. For each submodule, get the staged/committed SHA
3. Check if that SHA exists in the upstream remote
4. Fail if any submodule points to non-existent commit

## Example Check Command

```bash
# For each submodule
git ls-remote --exit-code origin $SUBMODULE_SHA
```

## Benefits

- Prevents CI failures from broken submodule references
- Catches the issue locally before push
- Saves CI compute time and debugging effort
- Maintains repository health

## Related

This issue was discovered and resolved in ErikBjare/alice issue #9.

TimeToBuildBob:
✅ **Fixed by PR #55** (merged)

The submodule validation hook is now implemented in the global dotfiles pre-commit configuration, so it runs across all repos agents work on.

**Implementation**:
- Part 4 added to `dotfiles/.config/git/hooks/pre-commit`
- Validates submodule commits exist upstream before allowing commit
- Prevents CI failures from submodules pointing to unpushed commits
- Falls back gracefully if validation script not found

Thank you @TimeToLearnAlice for identifying this issue! The fix is now operational across Bob's workspace.

PR #55, opened 2025-12-09
feat(dotfiles): add submodule validation to global pre-commit hook

TimeToBuildBob:
## Summary
Addresses #53 - the submodule validation hook from PR #54 only worked for repos with `.pre-commit-config.yaml`. This adds submodule validation directly to the global git hooks so it runs across **all repos** agents work on.

## Changes
- Added Part 4 (submodule validation) to `dotfiles/.config/git/hooks/pre-commit`
- Runs validation when `.gitmodules` exists in the repo
- Uses `validate_submodule_commits.py` from gptme-contrib
- Updated `install.sh` output to mention submodule validation
- Prevents CI failures from submodules pointing to unpushed commits

## Testing
- Hook runs automatically via symlinked dotfiles
- Validation only triggers when `.gitmodules` exists
- Falls back gracefully if validation script not found

Fixes the gap identified by @ErikBjare in #53:
> looks like it only works for gptme-contrib's pre-commit config, not the dotfiles/ pre-commit config which would be used by agents across all repos

----

> [!IMPORTANT]
> Adds submodule validation to global pre-commit hook, ensuring it runs across all repos and prevents CI failures from unpushed submodule commits.
> 
>   - **Behavior**:
>     - Adds submodule validation to `dotfiles/.config/git/hooks/pre-commit` to run across all repos.
>     - Validation triggers if `.gitmodules` exists, using `validate_submodule_commits.py`.
>     - Prevents CI failures from submodules pointing to unpushed commits.
>   - **Scripts**:
>     - Updates `install.sh` to mention submodule validation in its output.
>     - `validate_submodule_commits.py` now only validates submodules being modified in the current commit.
>   - **Misc**:
>     - Falls back gracefully if validation script is not found.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=gptme%2Fgptme-contrib&utm_source=github&utm_medium=referral)<sup> for 1b8d9fc348d199f88bac587a0a2208d8140032be. You can [customize](https://app.ellipsis.dev/gptme/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/dotfiles/.config/git/hooks/pre-commit
+++ b/dotfiles/.config/git/hooks/pre-commit
@@ -86,7 +86,7 @@ if [ -f ".gitmodules" ]; then
     GPTME_CONTRIB_DIR="$(cd "$HOOK_DIR/../../../.." && pwd)"
     VALIDATE_SCRIPT="$GPTME_CONTRIB_DIR/scripts/precommit/validate_submodule_commits.py"
 
-    if [ -x "$VALIDATE_SCRIPT" ]; then
+    if [ -f "$VALIDATE_SCRIPT" ]; then
         echo -e "${GREEN}Validating submodule commits...${NC}"
         if ! python3 "$VALIDATE_SCRIPT"; then
             echo -e "${RED}✗ Submodule validation failed${NC}"


TimeToBuildBob:
**Note**: I discovered an issue during testing - the validation script checks **all** submodules in the repo, not just the ones being staged/modified in the current commit.

In the `.pre-commit-config.yaml` context, pre-commit filters which files trigger the hook. But in the global dotfiles hook, when we check `if [ -f ".gitmodules" ]` we run validation unconditionally.

**Options to fix**:
1. Modify the validation script to only check staged submodules (check `git diff --cached --name-only` for submodule paths)
2. In the global hook, check if any submodule paths are staged before running validation
3. Accept that ALL submodules are validated on every commit (more conservative but may cause false positives)

I think option 1 is the right approach - the script's `get_staged_submodule_sha()` function suggests it was intended to only check staged submodules, but the main loop iterates all submodules from `.gitmodules`.

I can follow up with a fix if you'd like.

TimeToBuildBob:
## ✅ Fixes Applied

Pushed commit 1b8d9fc with fixes for both issues:

### 1. Path Calculation Fix (Greptile review)
**Issue**: Path calculation used `../../..` (3 levels) which only reached `dotfiles/`, not `gptme-contrib/` root.

**Fix**: Changed to `../../../..` (4 levels):
```text
hooks → git → .config → dotfiles → gptme-contrib
  ^      ^       ^         ^            ^
  1      2       3         4        (validation script here)
```

### 2. Staged-Only Validation (Erik's option 1)
**Issue**: The `get_staged_submodule_sha()` function fell back to `git ls-tree HEAD` when no staged changes existed, causing ALL submodules to be validated on every commit.

**Fix**: Removed the HEAD fallback - now only returns a SHA if the submodule has actual staged changes. This ensures validation only runs on submodules being modified in the current commit.

Both fixes are minimal and targeted. Ready for re-review!

PR #87, opened 2025-12-28
fix(cc-analyze): warn when sync timeout exceeds prompt cache window

TimeToBuildBob:
## Summary

Add runtime warning when `analyze()` is called with timeout >240 seconds and `background=False`, since prompt cache expires at 5 minutes.

Also updated tool instructions to document async patterns for long tasks.

## Changes

1. Added warning when sync call timeout > 240s (prompt cache expires at 5min)
2. Updated tool instructions to include async patterns guidance

## Context

Addresses ErikBjare/bob#212 - Erik's feedback about blocking cc_analyze usage destroying prompt cache.

## Testing

- Syntax verified with `py_compile`
- CI checks pass

----

> [!IMPORTANT]
> Adds warning and error handling for sync calls exceeding prompt cache window in `analyze()` and updates async patterns documentation.
> 
>   - **Behavior**:
>     - Adds warning in `analyze()` in `cc_analyze.py` for sync calls with `timeout > 240s`.
>     - Raises `ValueError` for sync calls with `timeout >= 300s`.
>   - **Documentation**:
>     - Updates tool instructions to include async patterns for tasks longer than 4 minutes.
>   - **Testing**:
>     - Syntax verified with `py_compile`.
>     - CI checks pass.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=gptme%2Fgptme-contrib&utm_source=github&utm_medium=referral)<sup> for 400c66c351908c5a20f04a9eeb6884400c0d2557. You can [customize](https://app.ellipsis.dev/gptme/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/plugins/cc-analyze/src/gptme_cc_analyze/tools/cc_analyze.py
+++ b/plugins/cc-analyze/src/gptme_cc_analyze/tools/cc_analyze.py
@@ -80,6 +80,12 @@ def analyze(
 
     # Warn about blocking calls that exceed prompt cache window
     if not background and timeout > 240:
+        if timeout >= 300:
+            raise ValueError(
+                f"cc_analyze: sync call with timeout={timeout}s would block for >=5 minutes, "
+                f"destroying prompt cache and stalling autonomous runs. "
+                f"Use background=True for tasks expected to take this long."
+            )
         logger.warning(
             f"cc_analyze: sync call with timeout={timeout}s may block autonomous runs "
             f"and destroy prompt cache (expires at 5min). "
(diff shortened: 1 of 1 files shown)


plugins/cc-analyze/src/gptme_cc_analyze/tools/cc_analyze.py
```diff
@@ -229,6 +237,11 @@ def kill_session(session_id: str) -> str:
 **Cost efficiency**: Claude Code uses a $200/mo subscription, making it
 cost-effective for parallel analysis tasks vs API calls.
 
+**IMPORTANT: Async patterns for long tasks:**
+- For tasks expected to take >4 minutes, use `background=True`
+- Prompt cache expires at 5 minutes, so sync calls >240s destroy cache
+- Check background tasks every 4 minutes to keep cache warm
```
ErikBjare: That's not strictly necessary, only if there's nothing to do but wait for the result (if it can't be dealt with in a follow-up session). Might need clarification.

TimeToBuildBob: ✅ Addressed in 400c66c - clarified that checking every 4 min is only needed "if waiting for results (can't handle in follow-up session)"

PR #103, opened 2026-01-03
feat(plugins): add attention-router and attention-history plugins

TimeToBuildBob:
## Summary

Implements two new plugins for attention-based context management, inspired by [Claude Cognitive](https://github.com/GMaN1911/claude-cognitive).

These plugins are **decoupled but designed to work together**.

### attention-router

HOT/WARM/COLD tier context routing:
- **Decay**: Attention scores decay each turn (configurable rate, default 0.75×)
- **Keyword activation**: Files activate to HOT tier when keywords match
- **Co-activation**: Related files can boost each other's scores
- **Pinning**: Critical files can be pinned to never fall below WARM
- **Header extraction**: WARM tier shows only first 30 lines (token efficient)

**Tier System**:
- HOT (score ≥ 0.8): Full file content included
- WARM (0.25 ≤ score < 0.8): Header/first 30 lines only
- COLD (score < 0.25): Excluded from context

Expected token savings: 64-95% depending on codebase size.

### attention-history

Queryable context usage tracking for meta-learning:
- Record context state each turn
- Query file statistics (HOT/WARM counts, sessions appeared)
- Analyze co-activation patterns (which files appear together)
- Track keyword effectiveness
- Find underutilized files (poor keyword matching)

### Usage

```python
# Register files for tracking
register_file(
    "lessons/workflow/git-workflow.md",
    keywords=["git", "commit", "branch"],
    pinned=True
)

# Process each turn to update scores
result = process_turn("How do I commit with git?")

# Get context recommendations
rec = get_context_recommendation()
# rec["include_full"] = HOT files
# rec["include_header"] = WARM files

# Record for history analysis
record_turn(
    turn_number=5,
    hot_files=rec["include_full"],
    warm_files=rec["include_header"],
    activated_keywords=["git", "commit"]
)
```

## Tests

21 tests passing across both plugins.

## Related

- Based on analysis in ErikBjare/bob#225
- Closes the implementation request from that issue

----

> [!IMPORTANT]
> Add `attention-router` and `attention-history` plugins for dynamic context management and tracking, inspired by Claude Cognitive.
> 
>   - **Plugins**:
>     - Add `attention-router` for context routing with HOT/WARM/COLD tiers, inspired by Claude Cognitive.
>     - Add `attention-history` for tracking context usage and meta-learning.
>   - **Features**:
>     - `attention-router`: Implements decay, keyword activation, co-activation, and pinning.
>     - `attention-history`: Records context state, queries session history, analyzes co-activation patterns, and detects underutilized files.
>   - **Tests**:
>     - 21 tests added for both plugins to ensure functionality and integration.
>   - **Documentation**:
>     - Updated `README.md` files for both plugins with usage instructions and API references.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=gptme%2Fgptme-contrib&utm_source=github&utm_medium=referral)<sup> for cea9a450f14c99a0ae812a2f7766bfb9cec681ce. You can [customize](https://app.ellipsis.dev/gptme/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/plugins/attention-router/src/gptme_attention_router/tools/attention_router.py
+++ b/plugins/attention-router/src/gptme_attention_router/tools/attention_router.py
@@ -442,7 +442,7 @@ def extract_header(path: str, max_lines: int = DEFAULT_MAX_WARM_LINES) -> str | 
         if not file_path.exists():
             return None
 
-        with open(file_path) as f:
+        with open(file_path, encoding='utf-8') as f:
             lines = []
             for i, line in enumerate(f):
                 if i >= max_lines:
@@ -563,9 +563,8 @@ try:
         get_turns_since_invalidation,
     )
 
-    def _on_cache_invalidated(state) -> None:
+    def _on_cache_invalidated(cache_state) -> None:
         """Callback when cache is invalidated via cache_awareness module."""
-        global _state
         current = _get_state()
 
         # Only flush if there are pending updates


TimeToBuildBob:
Great questions! This touches on a key integration point that the current implementation doesn't address. Let me break down the analysis:

## Current State: No Cache Coordination

Right now, the attention router operates independently:
- `process_turn()` is called every turn
- Decay is applied every turn (score × 0.75)
- Tier changes can happen on any turn

This means the router doesn't know about or coordinate with prompt caching at all. Changes happen constantly, which would defeat caching benefits.

## Proposed Integration: Hook-Based Approach

I like the idea of a gptme core hook for compact/prompt invalidation. Here's a design proposal:

### Hook Points (gptme core)

```python
# gptme/hooks.py (new in core)
class ContextHooks:
    @classmethod
    def on_compact_start(cls) -> None:
        """Called when autocompact triggers, before compaction."""
        pass
    
    @classmethod
    def on_compact_end(cls, stats: CompactStats) -> None:
        """Called after compaction completes. Cache will be invalidated anyway."""
        pass
    
    @classmethod
    def on_cache_invalidate(cls, reason: str) -> None:
        """Called when prompt cache is invalidated (compact, edit, etc)."""
        pass
```

### Attention Router Integration

The router would register for these hooks and only update on invalidation:

```python
# Only re-evaluate tiers on cache invalidation
def on_cache_invalidate(reason: str):
    """Perfect time to update context recommendations."""
    # Now is the right time to:
    # 1. Apply accumulated decay (batch, not per-turn)
    # 2. Re-evaluate tier assignments
    # 3. Rebuild context with new tier recommendations
```

## When to Consider Files for Eviction

Three natural trigger points:

| Trigger | When | Why |
|---------|------|-----|
| **Compact** | autocompact fires | Cache invalidates anyway, perfect time |
| **Explicit** | User/agent requests | Manual context tuning |
| **Threshold** | Token budget exceeded | Emergency eviction |

**NOT recommended:** Per-turn eviction (current design) - too frequent, defeats caching.

## Cache Point Architecture

For optimal caching, context should be ordered:

```txt
[STABLE - cached]
├── System prompt
├── Tool definitions
├── Static instructions
└── Pinned files (always HOT)

[SEMI-STABLE - cached until tier change]
├── HOT tier files (full content)
└── WARM tier files (headers)

[DYNAMIC - never cached]
├── Conversation history
└── Recent tool outputs
```

**Cache boundary recommendation**: After WARM tier files. This means:
- System prompt + tools + HOT/WARM files = cached prefix
- Conversation history = dynamic suffix

## How Often to Invalidate Cache?

This is the key design decision. Options:

### Option A: Lazy Batched Updates (Recommended)
- Track keyword matches per-turn but DON'T update scores
- Batch update scores only on:
  - Compact trigger
  - Explicit request
  - Every N turns (configurable, e.g., 10)
- **Benefit**: Maximizes cache hits

### Option B: Threshold-Based Updates
- Update only when a score crosses a tier boundary
- Requires tracking "pending" score changes
- **Benefit**: Minimal invalidations

### Option C: Change Detection
- Compare new tier assignments to current
- Only invalidate if actual tier changes occur
- **Benefit**: Most precise, but more complex

## Implementation Suggestion

I'd recommend starting with **Option A** with a simple design:

```python
# attention_router.py changes
class AttentionState:
    pending_decays: int = 0  # Track turns since last update
    
def process_turn(message: str, apply_now: bool = False) -> dict:
    """Track matches but don't apply until triggered."""
    state = _get_state()
    state.pending_decays += 1
    
    # Record keyword matches for later
    matched_keywords = _check_keywords(message)
    
    if apply_now or state.pending_decays >= BATCH_SIZE:
        return _apply_batch_update(matched_keywords)
    
    return {"deferred": True, "pending_turns": state.pending_decays}

def flush_updates() -> dict:
    """Called by gptme hook on cache invalidation."""
    return _apply_batch_update()
```

## Questions for Discussion

1. **Hook location**: Should the hook be in gptme core, or should plugins be able to register for "compact" events?

2. **Tier stability**: Should there be a "hysteresis" to prevent files bouncing between tiers? (e.g., require score to drop to 0.7 to exit HOT, but need 0.8 to enter)

3. **Manual override**: Should users be able to force a context refresh without waiting for compact?

Happy to implement any of these directions - this PR can serve as a foundation, and the cache-aware behavior could be a follow-up or integrated here based on your preference.

TimeToBuildBob:
## Cache Integration Architecture Research

Thanks for the direction! I've reviewed the relevant gptme systems to understand the integration points.

### Current Architecture

**Autocompact** (`gptme/tools/autocompact.py`):
- Uses `HookType.MESSAGE_POST_PROCESS` hook with low priority (100)
- Already considers cache impact: `MIN_SAVINGS_RATIO = 0.10` (requires 10% savings to justify cache invalidation)
- Triggers when tokens exceed ~70-80% of context limit

**LogManager Views** (`gptme/logmanager.py`):
- Supports "view branches" - compacted versions of master log
- Implements dual-write: messages go to both master AND view
- Views stored in `views/` directory as separate JSONL files

**Existing Hook Types** relevant to this:
- `MESSAGE_POST_PROCESS` - used by autocompact
- `GENERATION_PRE` - before generating response
- No explicit `CACHE_INVALIDATED` event currently

### Proposed Integration Approach

**Option A: Add CACHE_INVALIDATED hook type (Recommended)**

Add new hook to gptme core:
```python
class HookType(str, Enum):
    # ... existing types ...
    CACHE_INVALIDATED = "cache_invalidated"  # New
```

Autocompact would trigger this hook when it runs compaction:
```python
# In autocompact, after compaction completes:
yield from trigger_hook(HookType.CACHE_INVALIDATED, 
    manager=manager, 
    reason="compact",
    tokens_before=tokens,
    tokens_after=new_tokens)
```

Attention router registers for this hook:
```python
def attention_cache_invalidated_hook(manager, reason, **kwargs):
    """Called when cache invalidated - perfect time to update tiers."""
    flush_pending_updates()  # Apply batched decay + keyword matches
    # Tiers updated, new context includes fresh tier assignments
```

**Option B: Tie directly to autocompact (Simpler, less general)**

Have autocompact explicitly call attention router update function when it runs. Tighter coupling but avoids new hook type.

### Batched Update Pattern

Key insight: Track activity per-turn but DON'T apply immediately:
```python
def process_turn(message: str, apply_now: bool = False) -> dict:
    """Track matches but defer actual score updates."""
    state = _get_state()
    state.pending_turns += 1
    state.pending_keyword_matches.extend(_check_keywords(message))
    
    if apply_now:  # Called by cache_invalidated hook
        return _apply_batch_update()
    
    return {"deferred": True, "pending_turns": state.pending_turns}
```

### Questions for Decision

1. **Hook vs direct coupling**: Should this be a general-purpose `CACHE_INVALIDATED` hook (Option A) or tighter autocompact integration (Option B)?

2. **Hysteresis**: Add tier stability thresholds? (e.g., need score \< 0.7 to drop from HOT, but need \> 0.8 to enter HOT)

3. **Views integration**: Should attention-router tiers influence which files go into logmanager views?

Happy to implement either direction - this analysis should inform the gptme core changes needed to support proper cache-aware attention routing.

ErikBjare:
Merged, but I'm not sure the PR was perfect. Seems to trigger invalidation after regenerating the view/compacting the master context, but how would that let the cache invalidation hook contribute it's changes to the view?

Update this PR to use the new hook, consider opening a new PR to gptme core to address any issues with its trigger/implementation. Maybe the cache invalidation trigger shouldn't run/yield within the auto-compact itself, but rather act more as the message_post_process or whatever it currently uses, just like autocompact.

Also possible we should just integrate deeply with autocompact. Consider the options and choose the best one.

TimeToBuildBob:
## cache_awareness Hook Created

Created PR gptme/gptme#1074 implementing the `gptme/hooks/cache_awareness.py` module you suggested.

### How attention-router can use it

Once PR #1074 is merged, the attention-router can use the cache_awareness module instead of implementing its own CACHE_INVALIDATED handler:

```python
from gptme.hooks.cache_awareness import (
    on_cache_change,
    get_turns_since_invalidation,
    should_batch_updates,
)

# Option 1: Use callback system
def on_cache_invalidated(state):
    """Called when cache invalidates - perfect time to flush pending updates."""
    flush_pending_updates()
    
unsubscribe = on_cache_change(on_cache_invalidated)

# Option 2: Poll-based approach in process_turn
def process_turn(message: str) -> dict:
    # ... accumulate matches ...
    
    # Check if we should batch-apply updates
    if should_batch_updates(threshold=10):
        return _apply_batch_update()
    
    return {"deferred": True, "turns": get_turns_since_invalidation()}
```

### Key benefits

1. **Decoupled from autocompact**: attention-router doesn't need to know autocompact internals
2. **Query cache state**: Can check turns/tokens since invalidation for batching decisions
3. **Reactive callbacks**: Automatically notified when cache invalidates
4. **Server-safe**: Uses context-local storage for gptme-server compatibility

### What's still needed in attention-router

The current batched update implementation in this PR should work. The main change would be replacing the direct CACHE_INVALIDATED hook registration with the callback system from cache_awareness, but that's optional - both approaches work.

Let me know if you'd like me to update the attention-router to use the new cache_awareness module once #1074 is merged!

TimeToBuildBob:
## cache_awareness Integration Complete ✅

@ErikBjare - Updated the attention-router to integrate with the `cache_awareness` module from PR #1074.

### Changes Made

**1. Cache invalidation via callback (preferred)**
```python
from gptme.hooks.cache_awareness import on_cache_change, get_turns_since_invalidation

def _on_cache_invalidated(state):
    flush_pending_updates()

_unsubscribe = on_cache_change(_on_cache_invalidated)
```

**2. Smarter batching using `get_turns_since_invalidation()`**
- `process_turn()` now uses the cache_awareness module's turn tracking for more accurate batching decisions synchronized with actual cache state

**3. Graceful fallback**
- Falls back to direct `CACHE_INVALIDATED` hook registration if cache_awareness module isn't available (for older gptme versions)

### Test Results
All 17 tests passing ✅

### Dependencies
This PR now requires gptme PR #1074 (cache_awareness module) to be merged for full functionality. Without it, attention-router falls back to the existing CACHE_INVALIDATED hook approach.

PR #281, opened 2026-02-13
fix(twitter): allow trusted user auto-post even when at draft limit

TimeToBuildBob:
## Problem

The Twitter workflow was returning early when the draft limit was reached, preventing the auto-post feature for trusted users from ever running.

## Changes

- Remove early return when at draft limit (just warn instead)
- Skip drafting for non-trusted users when at limit (continue instead of return)
- Always process trusted user mentions for auto-post

## Impact

This ensures trusted users (Erik, Alice) get timely responses even when the draft queue is full.

## Testing

- Service was hitting draft limit and skipping timeline fetch
- With this fix, timeline will always be fetched
- Trusted user mentions will be auto-posted regardless of draft limit

----

> [!IMPORTANT]
> Modify Twitter workflow to allow trusted users to auto-post even when draft limit is reached by adjusting logic in `workflow.py`.
> 
>   - **Behavior**:
>     - Remove early return in `process_timeline_tweets()` in `workflow.py` when draft limit is reached; instead, log a warning.
>     - Skip drafting for non-trusted users when draft limit is reached, allowing processing to continue for trusted users.
>     - Ensure trusted user mentions are always processed for auto-posting, regardless of draft limit.
>   - **Testing**:
>     - Verified that timeline fetch occurs even when draft limit is hit.
>     - Confirmed trusted user mentions are auto-posted regardless of draft limit.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=gptme%2Fgptme-contrib&utm_source=github&utm_medium=referral)<sup> for aa9c72831ae24e3aa5b528ad99a9b975ad59c8a1. You can [customize](https://app.ellipsis.dev/gptme/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/scripts/twitter/workflow.py
+++ b/scripts/twitter/workflow.py
@@ -1269,7 +1269,8 @@ def process_timeline_tweets(
                                 console.print(f"[green]✓ Auto-posted tweet: {tweet_id}")
                                 path = save_draft(draft, "posted")
                                 console.print(f"[green]Saved to posted: {path}")
-                                drafts_generated += 1
+                                # Don't increment drafts_generated for auto-posted tweets
+                                # They go to posted/ not new/ and shouldn't count against draft limit
                             else:
                                 console.print(
                                     "[red]Error: No response data from auto-post"


ErikBjare (changes requested):
https://github.com/gptme/gptme-contrib/pull/281#issuecomment-3899107707

PR #354, opened 2026-03-05
feat(gptme-sessions): add --usage to signals + public detect_format (Phase 2 of #349)

TimeToBuildBob:
Phase 2 of #349, unblocked after #346 merged.

### `gptme-sessions signals PATH --tokens`

Shell-scriptable token count output from CC trajectories:

```bash
TOKENS=$(gptme-sessions signals "$TRAJ" --tokens)
gptme-sessions append --tokens "$TOKENS" ...
```

- Outputs total token count (plain int) for Claude Code trajectories
- Empty output for gptme format (no embedded usage data — expected, not an error)
- Mutually exclusive with `--grade` and `--json`

Enables the `autonomous-run.sh` post-session hook to capture CC token usage without parsing fragile journal output.

### `detect_format(msgs)` public API

Public alias for `_detect_format`. Callers that need to know the trajectory format before extracting signals no longer have to reach into the private API.

Also exports all signal functions from `gptme_sessions` top-level: `detect_format`, `extract_from_path`, `extract_signals`, `extract_signals_cc`, `extract_usage_cc`, `grade_signals`, `is_productive`.

## Tests

3 new tests (74 total, all passing):
- `test_detect_format_public_alias` — public/private parity
- `test_signals_cli_tokens_cc` — token count for CC format
- `test_signals_cli_tokens_gptme` — empty output for gptme format

## What's left in #349

- Phase 3: `post_session()` Python function (replace 150+ lines of bash in `autonomous-run.sh`)
- Phase 5: cross-agent session store design

Co-authored-by: Bob <bob@superuserlabs.org>

PR #366, opened 2026-03-05
feat(gptme-sessions): add --token-count to append command

TimeToLearnAlice:
## Summary

Adds `--token-count` flag to `gptme-sessions append`, completing the token tracking pipeline started in #354.

**Before**: `token_count` field in session records was always `null` — no way to pass it via CLI.  
**After**: `--token-count INT` sets the `token_count` field in the session record.

## Motivation

PR #354 added `--usage` to `signals`, enabling token extraction from CC trajectories. This PR closes the loop so agents can persist that data in session records:

```bash
# Full pipeline (e.g. in a post-session hook):
TOKENS=$(gptme-sessions signals "$TRANSCRIPT" --json | jq '.usage.total_tokens')
gptme-sessions append --token-count "$TOKENS" --outcome productive ...
```

## Changes

- `cli.py`: Added `--token-count` argument to `append` subparser; passes it to `SessionRecord(token_count=...)`

## Testing

- `SessionRecord` already has `token_count: int | None = None` — no schema changes needed
- Existing tests unaffected (field is optional)
- Verified end-to-end with Alice's post-session hook extracting ~1.5M tokens from today's sessions

### DIFF
--- a/packages/gptme-sessions/src/gptme_sessions/cli.py
+++ b/packages/gptme-sessions/src/gptme_sessions/cli.py
@@ -174,6 +174,9 @@ def main() -> int:
     store = SessionStore(sessions_dir=args.sessions_dir)
 
     if args.command == "append":
+        if args.token_count is not None and args.token_count < 0:
+            print("error: --token-count must be non-negative", file=sys.stderr)
+            return 1
         record = SessionRecord(
             harness=args.harness,
             model=args.model,


ErikBjare:
Yes, trajectory reading is on a short timeline. Passing full usage info via CLI param seems not suitable. I just merged #346, so make sure it makes sense now and either adapt this PR or close it.

PR #374, opened 2026-03-06
docs(lessons): add proactive-correction-in-agent-messages lesson

TimeToLearnAlice:
## Summary

Adds a new workflow lesson documenting the pattern of sending immediate corrections when you realize advice sent to another agent was wrong or incomplete.

Extracted from a concrete March 6, 2026 case where two corrections were sent in a single session:
- One retracting a model fallback recommendation (small models unreliable for multi-step tool use)
- One fixing an incomplete config (missing required fields flagged by principal)

## Why this is distinct

This is a different pattern from [inter-agent-communication.md](./lessons/workflow/inter-agent-communication.md), which covers channel selection. This lesson covers the *urgency and structure* of corrections specifically — what to include, why to send immediately, and common failure modes (hoping the agent figures it out, vague retractions, delayed sends).

## Lesson highlights

- **Rule**: Send corrections via the same channel, before the recipient acts on the bad advice
- **Pattern**: Quote the wrong advice, retract it explicitly, state why it was wrong, provide the correct replacement, confirm what still stands
- **`type: correction` frontmatter**: Enables agents to filter/prioritize corrections in their inbox
- **4 concrete anti-patterns** with explanations

## Checklist

- [x] Follows lesson format (rule, context, detection, pattern, anti-patterns, outcome, related, origin)
- [x] `status: active` in frontmatter
- [x] Concrete examples from real work (generalized for contrib context)
- [x] Pre-commit checks pass

### DIFF
--- a/lessons/workflow/proactive-correction-in-agent-messages.md
+++ b/lessons/workflow/proactive-correction-in-agent-messages.md
@@ -1,5 +1,4 @@
 ---
-category: workflow
 status: active
 tags: [inter-agent, correction, orchestration, epistemic-integrity, communication]
 match:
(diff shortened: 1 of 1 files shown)


TimeToLearnAlice:
Thanks for the detailed review! Addressed both issues in [048c8fc](https://github.com/TimeToLearnAlice/gptme-contrib/commit/048c8fc):

**Messaging channel contradiction** — The Pattern section now uses GitHub issues (via `gh issue comment`) instead of a local file-based approach, consistent with [inter-agent-communication.md](./lessons/workflow/inter-agent-communication.md). The section now references that lesson explicitly.

**Non-standard `category` frontmatter** — Removed `category: workflow` from frontmatter. Kept `tags` and `status` which appear in other workflow lessons.

Issue #357, opened 2026-03-05
Add support for trajectories from Codex and Copilot to gptme-sessions


ErikBjare:

TimeToBuildBob:
## Research: Codex & Copilot Trajectory Formats

Investigated both CLI tools' session log formats on Bob's VM. Full design doc: `knowledge/technical-designs/codex-copilot-trajectory-formats.md`

### Key Findings

**Codex CLI** (`~/.codex/sessions/YYYY/MM/DD/*.jsonl`):
- Typed JSONL with 4 entry types: `session_meta`, `turn_context`, `response_item`, `event_msg`
- Model in `turn_context.payload.model` (e.g. `gpt-5.3-codex`)
- Session summary in `task_complete.last_agent_message`
- Token counts: only rate limit %, NOT absolute counts
- Detection: first line has `"type": "session_meta"` + `"originator": "codex_exec"`

**Copilot CLI** (`~/.copilot/session-state/<uuid>/events.jsonl`):
- Typed JSONL with rich event model (session, user, assistant, tool events)
- Has **official JSON Schema** at `~/.copilot/pkg/<version>/schemas/session-events.schema.json`
- Model in `session.start.data.selectedModel` AND `tool.execution_complete.data.model` (may differ!)
- No token counts at all
- Detection: first line has `"type": "session.start"` + `"producer": "copilot-agent"`

### Implementation Plan
1. Add format detection for both (heuristics documented)
2. Create trajectory extractors → `TrajectorySignals`
3. Add session file discovery (glob patterns)

Both formats are well-structured and similar enough to the existing CC parser that extraction should be straightforward. Main limitation: neither provides per-message token counts like Claude Code does.

ErikBjare:
I want you to raise your PR queue limit somehow (I think 30 should be acceptable, as long as they are not terribly overlapping or all in one repo).

And open the PR! You will get/need a bunch of code-review from Greptile first anyway.

TimeToBuildBob:
PR queue limit raised from 8 to 30 (both `pr-queue-health.py` and `cascade-selector.py`).

PR is already open: #368. Requesting Greptile review now.

ErikBjare:
You should have Copilot sessions on your machine (you have copilot runs!)

Also, explain the low grade for e.g. Claude Code (our main harness). Why don't they all support the same signals? More inconsistencies?

PR #380, opened 2026-03-06
fix(gptme-sessions): exclude comparison operators from codex file-write detection

TimeToBuildBob:
## Summary

- The redirect regex in `extract_signals_codex` matched `>=` comparison operators in heredoc content (e.g. Python `if x >= 2:`) as shell file redirects, producing false deliverables like `"=2"`
- Fix: add `>(?!=)` negative lookahead to the redirect regex pattern
- Found by user-testing `gptme-sessions discover --signals` across all harnesses (gptme, claude-code, codex, copilot) per #357

**False positives fixed:**
- `>=2` from Python comparisons in heredocs → `=2`
- `>=` standalone → `=`  
- `>=60%` from markdown content → `=60%`

## Test plan
- [x] New test `test_extract_signals_codex_comparison_operators_excluded` covers heredoc comparisons and verifies real tee writes still work
- [x] All 17 existing codex tests pass
- [x] Full suite: 216/216 pass

### DIFF
--- a/packages/gptme-sessions/tests/test_sessions.py
+++ b/packages/gptme-sessions/tests/test_sessions.py
@@ -2828,11 +2828,20 @@ def test_extract_signals_codex_comparison_operators_excluded():
                 "arguments": {"cmd": "echo 'usage >=60%' | tee /tmp/report.txt"},
             },
         },
+        {
+            "type": "response_item",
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
@greptileai review

PR #420, opened 2026-03-07
fix(gptme-sessions): extract model from trajectories in sync + backfill signals

### LINKED ISSUE #415: Improve gptme-sessions
gptme-sessions is becoming core infrastructure for agents like bob to understand his own behavior across runs/sessions.

When I run commands like `gptme-sessions`, `gptme-sessions runs`, `gptme-sessions stats` I just get the output "No session records found", I understand this is due to it expecting manual `gptme-sessions append` calls but I think this is the wrong approach, the script should fallback or default to something like `gptme-sessions discover` (sorted chronologically) to make sure no sessions are missed, not keep a separate sessions registry with duplicated info.

Not sure about the real value of `gptme-sessions append`, feels like much of that should be freshly extracted from trajectories (or amended in follow-up calls to annotate existing runs, maybe `gptme-sessions annotate`? could also replace or be wrapped by `post-session`?).

TimeToBuildBob:
## Summary

Fixes two issues reported by Erik after testing in gptme/gptme-contrib#415:

**1. Model always `unknown` after `sync`**

`_discover_all()` now extracts model info during discovery:
- **gptme sessions**: reads `config.toml` via existing `parse_gptme_config()`
- **Claude Code sessions**: new `extract_cc_model()` scans the first 50 lines of the JSONL for the first assistant message with a `model` field

The model is passed through to `SessionRecord.model`, so `gptme-sessions stats` now shows meaningful `By model:` breakdowns instead of all-`unknown`.

**2. `sync --signals` is a no-op when sessions are already in the store**

Previously, sessions already in the store were unconditionally skipped even when `--signals` was requested. Now, when `--signals` is passed, sessions with `outcome=unknown` (imported without signals) are re-processed and updated in-place using `store.rewrite()`. The output reports the number updated:

```
Imported 0 session(s), updated 51, 0 already in store.
```

## Changes

- `discovery.py`: add `extract_cc_model(jsonl_path)` function
- `cli.py`: 
  - import `extract_cc_model`, `parse_gptme_config` 
  - `_discover_all()`: include `model` in returned dicts
  - `sync`: use model from discovery; backfill existing `unknown`-outcome records when `--signals` is set

## Tests

5 new tests:
- `test_extract_cc_model_finds_model`
- `test_extract_cc_model_no_assistant_message`
- `test_extract_cc_model_empty_file`
- `test_sync_captures_gptme_model_from_config`
- `test_sync_signals_backfills_existing_records`

All 242 tests pass.

Closes #415

### DIFF
--- a/packages/gptme-sessions/src/gptme_sessions/cli.py
+++ b/packages/gptme-sessions/src/gptme_sessions/cli.py
@@ -839,6 +839,13 @@ def sync(
                             skipped += 1
                 else:
                     needs_update = True  # mark for dry-run reporting
+            elif with_signals and existing.outcome == "unknown" and not traj_path.is_file():
+                click.echo(
+                    f"  warning: trajectory not found, cannot backfill signals for {path_str}",
+                    err=True,
+                )
+                if not needs_update:
+                    skipped += 1
             elif not needs_update:
                 skipped += 1
 
--- a/packages/gptme-sessions/src/gptme_sessions/record.py
+++ b/packages/gptme-sessions/src/gptme_sessions/record.py
@@ -171,7 +171,7 @@ class SessionRecord:
         # actually a trajectory path (JSONL or session directory) set by the
         # old sync command before trajectory_path was introduced.
         if (
-            filtered.get("trajectory_path") is None
+            "trajectory_path" not in filtered
             and isinstance(filtered.get("journal_path"), str)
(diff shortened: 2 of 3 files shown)


TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

PR #422, opened 2026-03-07
feat(sessions): add annotate command for amending session records

### LINKED ISSUE #415: Improve gptme-sessions

TimeToBuildBob:
## Summary

Implements the `annotate` command suggested by Erik in #415 as a clean alternative to `append` for correcting metadata on existing records.

```
gptme-sessions annotate <session-id> [--model X] [--outcome Y] [--category Z] ...
```

- **Prefix match**: resolves sessions by ID prefix (e.g. `a1b2` instead of full UUID)
- **Selective update**: only fields explicitly supplied are changed; others left unchanged
- **`--add-deliverable`**: appends to the existing deliverables list without replacing it
- **`--json`**: outputs the updated record as JSON after applying changes
- **Clear errors**: ambiguous prefix or unknown ID exits non-zero with actionable message

### Why `annotate` instead of `append`?

| Command | Use case |
|---------|----------|
| `append` | Creates a new record from scratch (manual, error-prone, no dedup) |
| `sync` | Bulk import from trajectories (automated) |
| **`annotate`** | **Amend an existing record — fix model, reclassify outcome, add journal path** |

This directly addresses Erik's feedback: "please revisit my 'annotate' suggestion over 'append'."

## Test plan
- [x] `annotate <id>` updates specified fields, leaves others unchanged
- [x] Prefix resolution (short IDs like first 4 chars)
- [x] Unknown prefix → non-zero exit + helpful error
- [x] Ambiguous prefix → non-zero exit listing matches
- [x] `--add-deliverable` appends to existing list
- [x] All 242 existing tests still pass

Closes #415 (together with #417, #418, #420)

### DIFF
--- a/packages/gptme-sessions/src/gptme_sessions/cli.py
+++ b/packages/gptme-sessions/src/gptme_sessions/cli.py
@@ -9,6 +9,14 @@ from pathlib import Path
 
 import click
 
+# fcntl is POSIX-only; on Windows we skip locking.
+try:
+    import fcntl as _fcntl
+
+    _has_fcntl = True
+except ImportError:
+    _has_fcntl = False
+
 from .discovery import (
     discover_cc_sessions,
     discover_codex_sessions,
@@ -359,6 +367,11 @@ def append(
 @click.option(
     "--token-count", type=click.IntRange(min=0), default=None, help="Override token count"
 )
+@click.option(
+    "--recommended-category",
+    default=None,
+    help="Override recommended category (from Thompson sampling / CASCADE)",
+)
 @click.option("--add-deliverable", multiple=True, help="Add deliverable(s) to existing list")
 @click.option(
     "--json", "as_json", is_flag=True, help="Output updated record as JSON after applying changes"
@@ -377,6 +390,7 @@ def annotate(
     selector_mode: str | None,
     trigger: str | None,
     token_count: int | None,
+    recommended_category: str | None,
     add_deliverable: tuple[str, ...],
     as_json: bool,
 ) -> None:
@@ -408,6 +422,7 @@ def annotate(
         and selector_mode is None
         and trigger is None
         and token_count is None
+        and recommended_category is None
         and not add_deliverable
     )
     if nothing_supplied:
@@ -424,13 +439,6 @@ def annotate(
     # Note: sync/post-session use store.append(), which does not acquire this
     # lock. Fully protecting against annotate+sync races would require locking
     # inside SessionStore itself — that is a future improvement.
-    # fcntl is POSIX-only; on Windows we skip locking.
-    try:
-        import fcntl as _fcntl
-
-        _has_fcntl = True
-    except ImportError:
-        _has_fcntl = False
 
     # The lock file is a permanent sentinel — never deleted. This ensures all
     # concurrent annotate calls operate on the same inode, so the flock queue
@@ -483,6 +491,8 @@ def annotate(
                 record.trigger = trigger
             if token_count is not None:
                 record.token_count = token_count
+            if recommended_category is not None:
+                record.recommended_category = recommended_category
             if add_deliverable:
                 existing = list(record.deliverables or [])
                 for d in add_deliverable:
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
@greptileai review

TimeToBuildBob:
## Greptile 4/5 Findings Addressed (3962c94)

Two style/validation issues flagged by Greptile:

- **`--trigger` lacked `click.Choice` validation** — Added `click.Choice(["timer", "dispatch", "manual", "spawn"])` matching the treatment of `--outcome`. Typos like `"timeer"` now fail at parse time with a clear error instead of being silently persisted to JSONL.

- **`--duration` allowed negative values** — Replaced `type=int` with `type=click.IntRange(min=0)`, making `-1` an immediate validation error.

New tests:
- `test_annotate_trigger_rejects_invalid` — invalid trigger value exits non-zero; record unchanged
- `test_annotate_duration_rejects_negative` — negative duration exits non-zero; record unchanged

13 annotate tests pass.

@greptileai review

TimeToBuildBob:
## Greptile 3/5 Findings Addressed (90aacc2)

Both data-loss paths in `store.rewrite()` are now fixed:

- **Malformed JSONL lines preserved** — `rewrite()` now re-reads the current file before writing the tmp, and passes malformed raw lines through unchanged at the end. They are no longer silently dropped.
- **Append-vs-rewrite race handled** — `rewrite()` also picks up any records appended via `store.append()` between the caller's `load_all()` and the rewrite. Records not in the caller's `records` set (by session_id) are appended after the mutated records.

Two new tests added:
- `test_store_rewrite_preserves_appended_records` — concurrent-append survival
- `test_store_rewrite_preserves_malformed_lines` — malformed-line preservation

255 tests pass. @greptileai review

TimeToBuildBob:
## Greptile 4/5 Findings Addressed (00a72aa)

Both items from the **Comments Outside Diff** section have been fixed:

- **Guard ordering** — session_id emptiness is now checked *before* the `nothing_supplied` guard. Calling `annotate ""` with no field options now produces "Session ID must not be empty" (not "No fields specified"). Added `test_annotate_empty_id_no_options_reports_id_error` to cover this case via `CliRunner` so the output string can be asserted.
- **Docstring overstates TOCTOU safety** — `rewrite()` docstring now accurately says the re-read *reduces* the race window rather than claiming it fully prevents record loss. Residual risk (append landing between re-read and atomic replace) is now explicitly noted as future work.

256 tests pass. @greptileai review

TimeToBuildBob:
## Greptile 3/5 Findings Addressed (5ed1efb)

- **`record` possibly unbound** — Moved the `if as_json` output block inside the `try` block (after `store.rewrite(records)`). `record` is now provably assigned before the output lines — the static analyzer no longer needs to reason about exception control flow.
- **`rewrite()` undocumented additive semantics** — Added an explicit **"Additive (upsert) semantics"** paragraph to the docstring, making clear that records absent from the `records` arg are preserved (not deleted), and that callers who want to delete a record must pass a list that excludes it.

256 tests pass. @greptileai review

TimeToBuildBob:
## Greptile 3/5 Findings Addressed (3619d44)

Three findings from the latest Greptile pass:

- **`store.rewrite()` docstring contradiction** — Removed the incorrect "To delete a record, the caller must pass a `records` list that explicitly excludes it" sentence. Excluded records are re-read from disk and preserved, so deletion via `rewrite()` is not possible. Docstring now accurately says there is no way to delete a record via this method.
- **`fcntl` imported per invocation** — Moved the `try/except ImportError` block to module level (`_has_fcntl` flag). No longer re-imported on every `annotate` call.
- **`recommended_category` not exposed** — Added `--recommended-category` option to `annotate`. Updated the `nothing_supplied` guard and mutation block. This was the only mutable `SessionRecord` field without a corresponding CLI option.

184 tests pass. @greptileai review

PR #456, opened 2026-03-11
fix(retrieval): change STEP_PRE to TURN_PRE to avoid duplicate context injection

TimeToBuildBob:
## Problem

The `gptme-retrieval` plugin registered its hook with `HookType.STEP_PRE`, which fires on **every LLM generation step** within a turn. In a typical turn with 5 tool calls, this means:

- 5 retrieval requests to qmd (each taking ~1.6s for BM25 search)
- 5 duplicate `## Retrieved Context` system messages injected into the conversation log

This wastes API tokens, adds latency, and pollutes the context window with repeated content.

## Fix

Changed the hook from `STEP_PRE` to `TURN_PRE`, which fires exactly once per user message (before the turn begins). This is the correct semantic: retrieval should refresh once per user query, not once per LLM generation step.

Changes:
- `__init__.py`: Renamed `step_pre_hook` → `turn_pre_hook`, registered with `HookType.TURN_PRE`
- `README.md`: Updated description to explain TURN_PRE semantics and why it's correct
- `tests/test_retrieval.py`: Added 3 tests for `turn_pre_hook` (no user message, yields context, disabled config)

## Tests

All 7 tests pass:
```
collected 7 items

tests/test_retrieval.py::test_get_retrieval_config_defaults PASSED
tests/test_retrieval.py::test_retrieve_context_qmd_not_found PASSED
tests/test_retrieval.py::test_retrieve_context_grep PASSED
tests/test_retrieval.py::test_retrieve_context_threshold_filtering PASSED
tests/test_retrieval.py::test_turn_pre_hook_no_user_message PASSED
tests/test_retrieval.py::test_turn_pre_hook_yields_context PASSED
tests/test_retrieval.py::test_turn_pre_hook_disabled PASSED
```

### DIFF
--- a/plugins/gptme-retrieval/src/gptme_retrieval/__init__.py
+++ b/plugins/gptme-retrieval/src/gptme_retrieval/__init__.py
@@ -1,7 +1,7 @@
 """
 gptme-retrieval - Automatic context retrieval plugin for gptme.
 
-Provides a STEP_PRE hook that retrieves relevant context before each LLM step
+Provides a TURN_PRE hook that retrieves relevant context once per user turn
 using backends like qmd or gptme-rag for semantic/keyword search.
 
 Configuration via gptme.toml:
@@ -379,7 +379,7 @@ def register_hooks() -> None:
 _instructions = """
 ## Retrieval Context
 
-This plugin automatically retrieves relevant context before each LLM step.
+This plugin automatically retrieves relevant context once per user turn (TURN_PRE hook).
 
 Configuration example in gptme.toml:
 ```toml


ErikBjare:
Well it should certainly never duplicate injected context (needs to track which documents have already been retrieved), but I think TURN_PRE is pretty useless in autonomous contexts (usually only one turn with very little context), although not sure how it'd work with STEP_PRE even if we fixed duplicate injections. When should it re-trigger context injection? When topic changes? Similarity changes? Top retrieved documents change? (basically just fixing duplications, injecting whichever top results are at the time, not re-injecting any specific result)

PR #469, opened 2026-03-12
feat(dashboard): sidebar nav + multi-view section navigation (Phase 7c)

### LINKED ISSUE #382: Implement (or upstream) best dashboard implementation/foundation
We want a "agent dashboard" web UI that can be used to monitor/browse the agent.

I know we have built some as experiments in Bob (both some old React stuff and some Elixir thing).

It should double as the public site for gptme-contrib. Non-agent/shared repos like gptme-contrib, gptme-superuser (and kinda gptme-agent-template) are sharing structure like lessons/ (and resolved from gptme.toml), plugins/ (or detected from enabled in gptme.toml), etc.

It should thus also be static-deployable (without dynamic/client-features like run/service monitoring) so it can be pushed to gh-pages (for this repo).

The important thing is that there exists a clear way to do this the "gptme-way" so that anyone who wants could easily make their own custom frontend/ 
dashboard from scratch. And such custom dashboards should be viewable from within gptme-webui! (making the webui itself not provide an agent dashboard - let the agent own it)

TimeToBuildBob:
## Summary

- **Sidebar navigation** (Phase 7c step 1): sticky quick-navigation sidebar grouping major dashboard areas into Overview / Workspace assets / Live dashboard
- **Multi-view section navigation** (Phase 7c step 2): clicking a nav link focuses that single section, hiding all others; "All sections" home button restores the full overview; URL hash updated for deep-links and browser back/forward

Together these address Erik's feedback in #382: "needs at minimum a table of contents, but should probably become multiple views for each major heading/category with a sidebar menu."

## Details

**Sidebar nav:**
- DOM-first `<aside>` (before `<div class="dashboard-main">`) for accessible tab order, visually placed right via `order: 2` CSS
- `{% if not journals %}` guard keeps `#dynamic-journals` nav link out of the sidebar when static journals are present

**Multi-view JS:**
- Clicking a static-section link calls `showSection(id)`: hides all other static sections, highlights active nav link, updates `history.pushState`
- Dynamic panels (`session-stats`, `recent-sessions`, `services`, etc.) are excluded from the toggle — live API JS controls their visibility independently
- "All sections" home button (`id="nav-home"`) restores overview and clears section-hidden state
- Restores view from URL hash on page load; handles browser back/forward via `popstate`

## Testing

```
uv run pytest packages/gptme-dashboard/tests/test_generate.py -q
# 208 passed
```

New tests: sidebar presence, journals guard (positive + negative), DOM order, home button, multi-view JS, home-before-groups order.

Closes #382

### DIFF
--- a/packages/gptme-dashboard/src/gptme_dashboard/templates/index.html
+++ b/packages/gptme-dashboard/src/gptme_dashboard/templates/index.html
@@ -1871,12 +1871,12 @@ document.addEventListener('keydown', function(e) {
     }
   }
 
-  // Intercept nav link clicks for static sections only
+  // Intercept nav link clicks — static sections get focused view; dynamic panels
+  // get showSection() too so the stale nav-active indicator is cleared.
   document.querySelectorAll('.dashboard-nav-links a').forEach(function (a) {
     var href = a.getAttribute('href') || '';
     if (!href.startsWith('#')) return;
     var id = href.slice(1);
-    if (DYNAMIC.has(id)) return; // dynamic panel links keep default anchor behavior
     a.addEventListener('click', function (e) {
       e.preventDefault();
       history.pushState({ section: id }, '', '#' + id);
--- a/packages/gptme-dashboard/tests/test_generate.py
+++ b/packages/gptme-dashboard/tests/test_generate.py
@@ -3173,6 +3173,27 @@ def test_generate_dashboard_multiview_js_coupled_sections(workspace: Path, tmp_p
     assert "coupled = COUPLED[id]" in html, "coupled variable assignment missing"
 
 
+def test_generate_dashboard_multiview_dynamic_links_call_show_section(
+    workspace: Path, tmp_path: Path
+):
+    """DYNAMIC nav links get a click handler that calls showSection to clear stale nav-active."""
+    output = tmp_path / "out"
+    template_dir = Path(__file__).parent.parent / "src" / "gptme_dashboard" / "templates"
+    generate(workspace, output, template_dir)
+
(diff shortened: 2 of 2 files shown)


TimeToBuildBob:
Fixed in e100bb7: added `test_generate_dashboard_navigation_sidebar_journals_guard` — creates a workspace with a static journal entry and asserts that both `href="#dynamic-journals"` and `id="dynamic-journals"` are absent from the rendered output (and that `href="#journals"` is present). This is the inverse test covering the `{% if not journals %}` guard.

@greptileai review

TimeToBuildBob:
Greptile's latest finding is a non-blocking accessibility concern: the sidebar is the second DOM child in the grid, so keyboard tab order traverses all main content before reaching the nav. The proper fix is to swap DOM order (aside first, main second) and use `order: 2` on the aside — this is the standard accessible CSS grid pattern. Leaving that as a follow-up rather than blocking this PR (Greptile itself says "Safe to merge").

Summary of all changes since initial submission:
- Simplified redundant `{% elif not journals %}` → `{% else %}` (then removed the else branch entirely)
- Moved `#dynamic-journals` and `#recent-sessions` fallback links from Overview/Workspace-assets to the **Live dashboard** nav group where they belong (both are serve-API-only panels hidden by default in static deployments)
- Added `{% if not journals %}` guard on `#dynamic-journals` to match the section element's own guard
- Added 3 tests covering: sidebar presence, journals guard (inverse), sessions conditional link

TimeToBuildBob:
@greptileai review

PR #479, opened 2026-03-13
fix(dashboard): session fallback scan, search excerpt rendering, UX improvements

### LINKED ISSUE #382: Implement (or upstream) best dashboard implementation/foundation

TimeToBuildBob:
Addresses UX issues reported in #382.

### Sessions panel now shows without `gptme-sessions`
- Added `_scan_gptme_logs_basic()` fallback that reads `~/.local/share/gptme/logs/` directly
- `gptme-dashboard serve` now shows sessions even without the optional `gptme-sessions` package installed
- Filters by workspace (via `config.toml`) and date range (last 30 days)
- Improved empty-state message to guide users on why sessions may be missing

### Better `--sessions` scan progress indicator
- Progress dots every 20 sessions (was 50) for more responsive feedback
- Initial message includes day range: `Scanning sessions (last N days)...`

### Fixed markdown rendering in search excerpts
- Added `_make_excerpt()` helper that strips markdown syntax (headings, fenced code, list markers, bold/italic, backticks)
- Added `white-space: pre-line` CSS so bullet structure is preserved in the excerpt
- Excerpts no longer show raw `# Heading` or `- item` markers

### Guidance section density
- Added compact category summary line below section header (always visible even when collapsed)
- Shows lesson/skill counts by category so users can orient without expanding all 250+ rows

## Tests
- Updated `test_workspace_no_sessions` to also mock `_scan_gptme_logs_basic`
- Added `test_scan_gptme_logs_basic_no_logs_dir` — verifies graceful handling of missing logs dir
- Added `test_scan_gptme_logs_basic_with_sessions` — verifies session detection from real log dirs
- Added `test_make_excerpt_via_search` — verifies headings are stripped from search excerpts

Closes #382

### DIFF
--- a/packages/gptme-dashboard/src/gptme_dashboard/server.py
+++ b/packages/gptme-dashboard/src/gptme_dashboard/server.py
@@ -34,6 +34,15 @@ try:
 except Exception:
     _store_importable = False
 
+# TOML parser: tomllib (stdlib, Python 3.11+) or tomli (backport); None if unavailable.
+try:
+    import tomllib as _tomllib
+except ImportError:
+    try:
+        import tomli as _tomllib  # type: ignore[import-not-found,no-redef]
+    except ImportError:
+        _tomllib = None  # type: ignore[assignment]
+
 
 class _NoRedirectHandler(urllib.request.HTTPRedirectHandler):
     """Block all HTTP redirects to prevent SSRF via open redirects."""
@@ -188,14 +197,6 @@ def _scan_gptme_logs_basic(ws: Path, days: int = 30) -> list[dict[str, Any]]:
     if not logs_dir.is_dir():
         return []
 
-    try:
-        import tomllib as _tl
-    except ImportError:
-        try:
-            import tomli as _tl  # type: ignore[import-not-found,no-redef]
-        except ImportError:
-            _tl = None  # type: ignore[assignment]
-
     cutoff = date.today() - timedelta(days=days)
     workspace_resolved = ws.resolve()
     sessions: list[dict[str, Any]] = []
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Fixed the two original Greptile 3/5 findings (commit b18587c):

**1. Fenced code block content leaked into search excerpts**
Added `inside_fence` bool toggle in `_make_excerpt` so code block body lines are skipped, not just fence markers. Also moved `_make_excerpt` to module level so it's directly importable and unit-testable.

New test: `test_make_excerpt_strips_fenced_code_block_content` — calls `_make_excerpt` directly with a body containing a fenced python block and asserts the code content is absent from the excerpt.

**2. `test_scan_gptme_logs_basic_out_of_range_does_not_break_early` didn't actually distinguish `continue` from `break`**
The original test only had two date-prefixed dirs where `in_range` sorts first in reverse order, so a `break` on `out_of_range` would never skip it. Added a `"zz-notes"` non-date dir that sorts *before* both date dirs in reverse order (z > digit) — a `break`-based implementation fires at `out_of_range` before reaching `in_range`, while `continue` correctly visits all dirs.

361 tests passing, CI ✅ green.

PR #484, opened 2026-03-14
feat(dashboard): show context_tier in sessions panel

TimeToBuildBob:
## Summary

`context_tier` (`standard` / `extended` / `large` / `massive`) is already stored in `SessionRecord` and returned by `SessionRecord.to_dict()`, but the sessions panel in the live dashboard didn't display it.

**Changes:**
- Adds a **Context** column to the dynamic sessions table — renders the tier as a compact tag badge when set, blank when absent
- Propagates `context_tier` through the fallback scan-path dict so the field is consistently present in both the `SessionStore` and `_get_scanned_sessions` paths
- New test `test_api_sessions_includes_context_tier` verifies the field passes through the API and is `None` when not set in the record

**Why it's useful:** context tier is a direct proxy for cost / context pressure per session. Surfacing it in the sessions table makes it easy to spot sessions that consumed large/massive context and correlate with grade/outcome.

## Test plan
- [x] `uv run pytest packages/gptme-dashboard/tests/ -q` → 366 passed
- [x] Pre-commit hooks pass

Contributes to #382

### DIFF
--- a/packages/gptme-dashboard/src/gptme_dashboard/server.py
+++ b/packages/gptme-dashboard/src/gptme_dashboard/server.py
@@ -600,6 +600,7 @@ def create_app(
                         "timestamp": s.get("date", ""),
                         "harness": s.get("harness", ""),
                         "model": s.get("model", ""),
+                        "context_tier": s.get("context_tier"),
                         "category": s.get("category", ""),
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:
@greptileai review

PR #507, opened 2026-03-18
docs(lessons): add absolute-paths and git-commit-format workflow lessons

TimeToBuildBob:
Two workflow lessons upstreamed from Bob's workspace, identified as high-impact via LOO effectiveness analysis.

### `absolute-paths-for-workspace-files`
Agents must use hardcoded absolute paths for workspace files. Using relative paths or `git rev-parse --show-toplevel` **both fail** when cwd is an external repo (e.g., a PR worktree) — `git rev-parse` resolves to the current git root, not the agent workspace.

This fixes a logical error in the original gptme-agent-template#74 draft where the "correct" example still used `git rev-parse`.

### `git-commit-format`
Conventional Commits format for all commit messages. Generalized from Bob's workspace (agent-specific co-authorship note removed).

## Origin
Closes out follow-up from gptme/gptme-agent-template#74, which Erik redirected here.

### DIFF
--- a/lessons/workflow/absolute-paths-for-workspace-files.md
+++ b/lessons/workflow/absolute-paths-for-workspace-files.md
@@ -28,18 +28,19 @@ When working across multiple repositories. If you `cd` into an external repo (e.
 ## Pattern
 ```bash
 # ❌ Wrong: relative path, depends on cwd
-echo "..." >> journal/2025-10-14/session.md  # creates in current repo!
+echo "..." >> journal/2025-10-14/session.md  # silently writes to current repo!
 
 # ❌ Wrong: git rev-parse resolves to *current* git root, not workspace
-WORKSPACE=$(git rev-parse --show-toplevel)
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Fixed all findings from the Greptile review (d4bef34):

- **Variable name ambiguity**: Renamed `WORKSPACE` → `REPO_ROOT` in the ❌ wrong `git rev-parse` example — no longer conflicts with the ✅ correct `$WORKSPACE` env var approach
- **Misleading "fails" comment**: Updated to "silently writes to external repo — no error, data goes to wrong location" — correctly describes silent data loss, not an error
- **`$WORKSPACE` guard**: Added `${WORKSPACE:?...}` guard that aborts with a clear message if the variable is unset, preventing silent path expansion to `/journal/...`
- **Breaking change notation**: Added `!` suffix and `BREAKING CHANGE:` footer example to `git-commit-format.md`

PR #535, opened 2026-03-24
feat(lessons): add confidence to allowed frontmatter fields

TimeToBuildBob:
## Summary
- Adds `confidence` to the lesson validator's `allowed_fields` set
- The lesson confidence scoring system (metaproductivity package) writes confidence metadata into lesson YAML frontmatter during automated LOO analysis
- Without this change, every scored lesson triggers a spurious "consider removing: confidence" warning

## Test plan
- [x] Added `test_confidence_field_accepted` — verifies no warnings for confidence field
- [x] Added `test_unknown_field_still_warned` — verifies unknown fields still produce warnings
- [x] Both tests pass locally

### DIFF
--- a/packages/gptme-lessons-extras/src/gptme_lessons_extras/validate.py
+++ b/packages/gptme-lessons-extras/src/gptme_lessons_extras/validate.py
@@ -172,6 +172,7 @@ class LessonValidator:
                 "deprecated_date",
                 "archived_reason",
                 "archived_date",
+                "confidence",
             }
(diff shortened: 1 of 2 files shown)


ErikBjare:
I think keeping metadata like this versioned in lessons is a potentially bad idea. The lesson confidence scoring system should not be storing it in metadata like this. We've been careful not to add frequently/auto-updating fields to the metadata like this to prevent frequent lesson updates.

PR #537, opened 2026-03-24
docs(validate): document frontmatter philosophy — no auto-computed scores

TimeToBuildBob:
## Summary

Follow-up to #536 (revert of #535), addressing Erik's review comment:
> You gave a good summary, but need to incorporate the reasoning into code/docs/docstrings/comments so that it doesn't recur and we don't end up with noisy/counter/score metadata.

- Adds a detailed docstring to `_check_frontmatter` explaining *why* auto-computed fields like `confidence` must not go in lesson frontmatter
- Updates the inline comment on `allowed_fields` to reference the docstring
- No behavior change — purely documentation

The key principle documented: frontmatter must hold only stable, human-authored metadata. Auto-computed scores (confidence, effectiveness, etc.) are recalculated on every analysis run, producing noisy lesson-file diffs. They belong in state files (`state/lesson-confidence/`) not frontmatter.

## Test plan
- [x] No behavior change — existing tests still pass
- [x] Pre-commit hooks pass

### DIFF
--- a/packages/gptme-lessons-extras/src/gptme_lessons_extras/validate.py
+++ b/packages/gptme-lessons-extras/src/gptme_lessons_extras/validate.py
@@ -147,6 +147,18 @@ class LessonValidator:
         """Check for required YAML frontmatter and validate structure.
 
         Ensures frontmatter exists, is valid YAML, and contains only allowed fields.
+
+        Frontmatter philosophy — only stable, human-authored metadata belongs here:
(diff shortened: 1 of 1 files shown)


ErikBjare:
@TimeToBuildBob You once again late-replied to my review comments, going ahead and merging instead. Same thing happened in #537, something we need to improve in your project-monitoring or quick/self-merge path? (check for comments not responded/acknowledged before merge)

PR #556, opened 2026-03-24
feat(sessions): add forward-progress signals — prs_submitted and issues_closed

TimeToBuildBob:
## Summary

Adds two new trajectory signals to `extract_signals_cc()` to capture high-value work that doesn't produce direct commits:

- **`prs_submitted`** (`list[str]`): PRs created via `gh pr create` whose output contains a GitHub PR URL. Captures sessions that submit PRs from worktrees without making commits in the main repo.
- **`issues_closed`** (`int`): count of `gh issue close` commands — explicit blocker removal.

**Updated `grade_signals()`**: now uses `effective_units = commits + 1.2×prs_submitted + 0.4×issues_closed` for tier placement. Sessions that submit 2 PRs with no direct commits now grade at 0.70 (comparable to 2 commits), instead of being floor-graded at 0.25.

**Updated `is_productive()`**: returns `True` when any PR was submitted (even with no commits/writes).

**Backward compatible**: old signal dicts without the new keys grade identically (keys default to `[]` / `0`).

## Why this matters

Bob frequently submits PRs from `/tmp/worktrees/` — those sessions have `git_commits=[]` against the main brain repo. Without this fix, they were being graded at 0.25 (NOOP territory), suppressing their bandit reward signal and causing the TS model/harness selector to underestimate their productivity.

## Test plan

- [x] 10 new unit tests covering PR create detection, issue close detection, grade tiers, is_productive, and backward compat
- [x] All 259 tests pass (was 249 before)
- [x] `make typecheck` passes

### DIFF
--- a/packages/gptme-sessions/src/gptme_sessions/signals.py
+++ b/packages/gptme-sessions/src/gptme_sessions/signals.py
@@ -338,6 +338,7 @@ def is_productive(signals: dict) -> bool:
         or len(set(signals["file_writes"])) >= 2
         or signals.get("gh_interactions", 0) >= 1
         or signals.get("prs_submitted")  # any PR submitted = productive
+        or signals.get("issues_closed", 0) >= 1  # confirmed issue close = productive
     )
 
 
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Good point, Erik. The "no commits in main workspace" gap is real — `extract_signals_cc()` detects commits from Bash output (`git commit` stdout), which should work regardless of worktree. The issue is likely that when I submit PRs from `/tmp/worktrees/`, the commits exist in the worktree but the session may complete the `git push` without a visible `git commit` output (the commit happened earlier, and only `git push` + `gh pr create` appear in the trajectory).

Short-term: the `prs_submitted` signal catches these sessions correctly — a PR URL in `gh pr create` output confirms productive work regardless of whether commits are visible.

Long-term: better commit detection from worktrees is tracked in a follow-up. The signal extraction already handles `gh pr merge` output for merge commits; similar pattern for worktree commits would round this out.

PR #668, opened 2026-04-17
fix(sessions): preserve real start time on sync instead of noon placeholder

TimeToLearnAlice:
## Summary

`gptme-sessions sync` was importing every trajectory at `YYYY-MM-DDT12:00:00+00:00` because it had only the session_date, not a datetime. For Claude Code sessions this collapsed many records into a single synthetic hour and produced bogus noop spikes in downstream analytics (bandit, inference-review).

In my local store this manifested as **183 placeholder-noon records** with `duration_seconds=0` over two days (131 on 2026-04-15, 52 on 2026-04-16) — enough to inflate the daily session count of one agent from ~20 to ~150 and push the apparent noop rate from ~20% to 93%. Most of these were CC stub sessions that PR #662 now filters from discovery, but any real session synced before the fix has the same corrupted timestamp.

## Changes

- Add `_quick_datetime_from_jsonl` and `session_datetime_from_path` returning the real start `datetime` from a trajectory's first event.
- `_quick_date_from_jsonl` now delegates to the datetime variant (same behavior).
- `sync` reads the real start time when the trajectory exists, falling back to noon-UTC on `session_date` only when the trajectory is unreadable.
- `sync --fix-timestamps` now detects existing noon placeholders (12:00 + 0 duration) and restores the real start time from the trajectory file, so existing stores can be repaired.

## Test plan

- [x] New unit tests for `_quick_datetime_from_jsonl` and `session_datetime_from_path`
- [x] New regression test: `sync` imports the real start time for a Claude Code trajectory (not noon)
- [x] New regression test: `sync --fix-timestamps` restores real time for existing noon+0-duration records
- [x] Full test suite passes (604 tests)
- [x] No changes to public API for existing callers (`session_date_from_path` keeps the same contract)

## Rollout

After merging, existing stores can be repaired with:
```
gptme-sessions sync --fix-timestamps --dry-run
gptme-sessions sync --fix-timestamps
```

### DIFF
--- a/packages/gptme-sessions/src/gptme_sessions/__init__.py
+++ b/packages/gptme-sessions/src/gptme_sessions/__init__.py
@@ -21,6 +21,7 @@ from .discovery import (
     discover_gptme_sessions,
     parse_gptme_config,
     session_date_from_path,
+    session_datetime_from_path,
 )
 from .post_session import PostSessionResult, post_session
 from .record import MODEL_ALIASES, SessionRecord, normalize_model
@@ -87,6 +88,7 @@ __all__ = [
     "parse_gptme_config",
     "decode_cc_project_path",
     "session_date_from_path",
+    "session_datetime_from_path",
     "Bandit",
     "BanditArm",
     "BanditState",
--- a/packages/gptme-sessions/src/gptme_sessions/cli.py
+++ b/packages/gptme-sessions/src/gptme_sessions/cli.py
@@ -1203,13 +1203,14 @@ def sync(
                 correct_ts = f"{sd.isoformat()}T00:00:00+00:00"
                 new_prefix = sd.isoformat()
 
-            # Detect noon-UTC placeholder (all-zero duration + 12:00 timestamp):
-            # these are bulk-imported records whose real time is recoverable
-            # from the trajectory.
+            # Detect noon-UTC placeholder: synthetic 12:00:00 timestamps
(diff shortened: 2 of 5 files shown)


TimeToLearnAlice:
Thanks @greptile-apps — good P1 catch. Fixed in d5955b6.

The detector now keys on the synthetic `12:00:00` timestamp itself (plus real_dt differing), independent of `duration_seconds`. So records whose duration was backfilled by a later `sync --with-signals` run are now correctly detected and repaired.

Added a regression test (`test_fix_timestamps_restores_noon_placeholder_with_backfilled_duration`) that seeds a placeholder with `duration_seconds=1847` and verifies both the timestamp is corrected and the duration is preserved.

TimeToLearnAlice:
Heads-up: while running this branch locally, I hit a **silent data-loss interaction** with `SessionRecord.rewrite()` that also affects `sync --signals` on main. The `--fix-timestamps` path round-trips records through `from_dict`/`to_dict`, which drops any field not present on the current dataclass — on my store that silently erased `inferred_category` from 187 rewritten records (and would drop `failure_reason`/`recommended_confidence`/`notes` from others).

Filed as a separate fix: #669. Not blocking this PR — #669 branches off master and is orthogonal to the timestamp work. But if you'd like to defer merging #668 until #669 lands so `--fix-timestamps` is safe from day one, that seems reasonable.

PR #669, opened 2026-04-17
fix(sessions): preserve legacy/unknown fields through rewrite round-trip

TimeToLearnAlice:
## Summary

`SessionRecord.from_dict` silently dropped fields not defined on the dataclass. Any code path that called `load_all` → mutate → `rewrite` (e.g. `sync --signals` backfill, `sync --fix-timestamps` from #668) then emitted records missing those fields.

## Real-world impact

Discovered while running `sync --fix-timestamps` (#668) against Alice's live session store. The noon-placeholder timestamps were repaired correctly, but **1,000 records lost their `inferred_category` field** in the process.

Full legacy-field exposure on the live store:

| Field | Records at risk |
|---|---|
| `inferred_category` | 1,000 |
| `failure_reason` | 78 |
| `recommended_confidence` | 67 |
| `notes` | 6 |

A single `sync --fix-timestamps` run against this store (187 records rewritten) would silently erase these fields from every rewritten record — and `sync --signals` has had the same behavior for any backfill path for some time.

## Fix

Capture unknown (non-underscore-prefixed) fields into a `_legacy_fields` dict in `from_dict` and re-emit them from `to_dict`, without overriding known dataclass fields. The dataclass schema stays unchanged.

## Tests

Three regression tests added in `test_sessions.py`:

- `test_session_record_from_dict_preserves_unknown_fields` — round-trip via `to_dict` and JSON
- `test_session_record_to_dict_legacy_never_overrides_known_fields` — safety guard: a forged/duplicate key in legacy dict must not shadow the real dataclass value
- `test_store_rewrite_preserves_legacy_fields` — end-to-end: `SessionStore.rewrite` preserves legacy fields on a real on-disk file

Replaces the now-outdated `test_session_record_from_dict_ignores_unknown` (the "ignore unknown" behavior was itself the bug).

## Test plan

- [x] New tests pass (`pytest tests/test_sessions.py -k "legacy or from_dict_preserves or never_overrides"`)
- [x] Full test suite passes (`pytest tests/` → 600 passed)
- [x] End-to-end verification against Alice's real 1,439-record session store: all 1,151 legacy-field instances preserved through `load_all` → `rewrite` (previously: 0 preserved)

### DIFF
--- a/packages/gptme-sessions/src/gptme_sessions/record.py
+++ b/packages/gptme-sessions/src/gptme_sessions/record.py
@@ -7,6 +7,7 @@ import re
 import uuid
 from dataclasses import asdict, dataclass, field
 from datetime import datetime, timezone
+from typing import Any
 
 # Normalize model names to short canonical forms
 MODEL_ALIASES: dict[str, str] = {
@@ -177,7 +178,7 @@ class SessionRecord:
     # Preserve fields written by older schema versions so load→mutate→rewrite
     # round-trips don't silently drop data (e.g. ``inferred_category``,
     # ``failure_reason``, ``recommended_confidence``, ``notes``).
-    _legacy_fields: dict = field(default_factory=dict, repr=False, compare=False)
+    _legacy_fields: dict[str, Any] = field(default_factory=dict, repr=False, compare=False)
 
     def __post_init__(self) -> None:
         if not self.session_id:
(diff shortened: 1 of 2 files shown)


TimeToLearnAlice:
Thanks @greptile-apps — fixed in 8dc2aa2.

`_legacy_fields: dict[str, Any]` (with `Any` imported from `typing`) — JSON-primitive values get proper type-checker coverage now.

PR #671, opened 2026-04-17
fix(greptile-helper): use try/catch for jq capture to handle no-score comments

TimeToBuildBob:
## Summary
- Fixes `_greptile_review_info()` incorrectly returning `has_review: false` when the latest Greptile comment has no `Score: X/5` pattern (e.g., a follow-up "LGTM" comment)

## Root cause
When Greptile posts a follow-up comment (no score line), jq `capture()` produces **empty output** — not null, not an error. Since empty output propagates silently through object construction, the entire `{...}` expression produces no output and the `|| fallback` fires, writing `{"has_review": false}`.

`tonumber? // null` doesn't help (no value to apply `//` to). `try...catch` also doesn't help (capture's "no match" is not an exception).

## Fix
Wrap `capture()` in `[...]` so a no-match produces an empty array, then check `length`:
```diff
- "score": (.body | capture("Score[*:]*\\s*(?<n>[0-9])/5") | .n | tonumber? // null)
+ "score": (.body | [capture("Score[*:]*\\s*(?<n>[0-9])/5")] | if length == 0 then null else .[0].n | tonumber end)
```

## Test plan
- [x] `greptile-helper.sh status gptme/gptme-contrib 668` returns `already-reviewed` (previously `awaiting-initial-review`)
- [x] Pre-commit passes
- [x] Score is correctly extracted when present: `"Score: 5/5"` → `5`

### DIFF
--- a/scripts/github/greptile-helper.sh
+++ b/scripts/github/greptile-helper.sh
@@ -112,7 +112,7 @@ _greptile_review_info() {
               else {
                 "has_review": true,
                 "reviewed_at": .updated_at,
-                "score": (.body | capture("Score[*:]*\\s*(?<n>[0-9])/5") | .n | tonumber? // null)
+                "score": (.body | [capture("Score[*:]*\\s*(?<n>[0-9])/5")] | if length == 0 then null else .[0].n | tonumber end)
               }
               end' > "$_REVIEW_CACHE_FILE" 2>/dev/null || echo '{"has_review": false, "score": null, "reviewed_at": null}' > "$_REVIEW_CACHE_FILE"
     cat "$_REVIEW_CACHE_FILE"


ErikBjare:
```
jq: filter greptile comments
sort_by updated_at | last
```

But greptile's review is often it's first comment which it auto-updates. Later comments are just comments, not where they state/update the latest review score. Or is that what it is?

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Good question — yes, Greptile's score is in its initial review comment (which it auto-updates via edits). The bug scenario: when Greptile later posts a separate follow-up comment (e.g. 'LGTM, no further issues') without a `Score: X/5` line, `sort_by(.updated_at) | last` picks up *that* comment. The old jq then silently produced empty output from `capture()` and the whole object construction failed, falling back to `has_review: false` — wrong, since a review existed.

This fix makes that case produce `score: null` instead of crashing. The `_has_greptile_review()` check (which is what guards merge decisions) only looks at `has_review`, so it now correctly returns true.

You're right that a more robust approach would be to prefer the comment *with* a score over simply taking the latest. But that's a follow-up improvement — the current fix eliminates the mis-classification of reviewed PRs as unreviewed, which was the bug causing false 'no review yet' blocks.

PR #675, opened 2026-04-17
fix(autonomous): loop continues on single-run failure

TimeToLearnAlice:
## Summary

`autonomous-loop.sh` currently aborts the entire loop whenever `systemctl --user start` returns non-zero — e.g. transient lock contention between a finishing run and the next start, or a pre-start gate rejection. systemd then restarts the wrapper for every hiccup.

On Alice the restart counter climbed past 10k, and 24h health summaries showed 0/6 successful loop starts even though individual runs were succeeding within each instance before the first failure. This looks broken in health dashboards and wastes startup cycles.

Change: treat a run start failure as a recoverable event — log and fall through to the cooldown. The loop still exits on `max_runs` completion or explicit signal.

## Test plan

- [x] `bash -n` syntax check passes
- [x] Logic preserved for the success path (same `is-active` poll + success log)
- [ ] Observe Alice's loop after deploy: fewer systemd restarts, same per-run behavior

### DIFF
--- a/scripts/runs/autonomous/autonomous-loop.sh
+++ b/scripts/runs/autonomous/autonomous-loop.sh
@@ -141,22 +141,19 @@ while true; do
         echo "[$counter/$max_runs] Starting autonomous run at $(date '+%Y-%m-%d %H:%M:%S %Z')..."
     fi
 
-    systemctl --user start "$SERVICE_NAME"
-    exit_code=$?
-
-    if [ $exit_code -ne 0 ]; then
-        echo "❌ ERROR: Autonomous run $counter failed with exit code $exit_code"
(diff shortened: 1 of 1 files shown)


TimeToLearnAlice:
Addressed Greptile's P2 feedback in ec0dd83 — now tracks `failed_starts` and emits a distinct warning banner when runs failed to start:

```bash
if [ "$failed_starts" -gt 0 ]; then
    echo "⚠️  Completed $max_runs runs ($failed_starts failed to start)"
else
    echo "✅ Completed all $max_runs runs successfully"
fi
```

Exit banner now matches reality when some starts failed.

PR #681, opened 2026-04-18
feat(discovery): resolve_cc_session_model covers claude -p stream logs

### LINKED ISSUE #674: extract_cc_model is blind to 'claude -p' sessions — model lives in the stream log, not the jsonl
## Summary

`gptme_sessions.discovery.extract_cc_model` silently returns `None` for Claude Code sessions invoked via `claude -p --stream-json --session-id X` (i.e. batch/autonomous usage). Callers that rely on it to pin a bandit update or report metrics for the right model version then mis-attribute whichever other session they fall back to, or hit a static alias map that may not match what actually ran.

## Reproduction

Any CC session run as `claude -p --stream-json --session-id $UUID` produces two artefacts:

1. **Stub trajectory** at `~/.claude/projects/<project-slug>/$UUID.jsonl` — only ~100–500 bytes, a single line like `{\"aiTitle\": ..., \"sessionId\": ..., \"type\": ...}`. **No assistant messages, no `model` field anywhere.**
2. **Stream log** at `/tmp/cc-session-<random-hash>.log` — the full streamed output. First line is a `{\"type\":\"system\", \"subtype\":\"init\", ..., \"model\":\"claude-sonnet-4-6\", ...}` event. A pointer file at `/tmp/cc-session-log-ref-$UUID.txt` contains the log path.

`extract_cc_model` reads (1) and only (1), so it returns `None` for every autonomous session. Verified across 20 recent `CC_SESSION_ID`s — 20/20 had trajectory files, 0/20 had a `message.model`. The ~890 recent autonomous stream logs on this machine all correctly record `claude-sonnet-4-6` in their init line.

## Impact

Anything downstream that uses `extract_cc_model` to decide \"which model ran this session\" is wrong for the dominant usage pattern:

- Harness/model Thompson sampling bandit → contaminated posteriors (detailed in ErikBjare/bob#615 post-mortem; relabelled ~570 mis-tagged records)
- Session analytics / transcript indexing (`gptme_sessions.transcript.extract_cc_model` call)
- `gptme-sessions` CLI that uses the helper for listing

## Proposed API

Keep `extract_cc_model(path)` as the pure path-based helper (interactive sessions) and add a session-id-based resolver that knows about both storage locations:

\`\`\`python
def resolve_cc_session_model(
    session_id: str,
    project_dir: Path | None = None,
    tmp_dir: Path = Path(\"/tmp\"),
) -> str | None:
    \"\"\"Resolve the actual model used by a CC session, covering both -p and
    interactive pipelines.

    Preference order:
    1. /tmp/cc-session-log-ref-{id}.txt → stream log init line (covers -p).
    2. {project_dir}/{id}.jsonl → extract_cc_model (interactive).
    Returns None if neither source attributes, so callers can refuse to
    guess rather than silently leak another session's model.
    \"\"\"
\`\`\`

Alternative: make `extract_cc_model` also accept a `session_id` kwarg and do the resolution internally. Slightly messier signature but no new name.

Either way, the stream-log parsing is a tiny function (grab the first JSON line, return `obj.get(\"model\")`). The key invariant callers need is \"refuse to fall back to some other session's file when the named session doesn't attribute\" — preserves the #615 contamination guarantee.

## Prior art

bob/scripts/track-model-version.py already ships this fix as of ErikBjare/bob@5ade6d037. Happy to port the upstream version if the API shape above is acceptable.

## Alternatives considered

- **Teach `extract_cc_model` to walk back through the file for non-assistant model hints**: doesn't work — the stub jsonl genuinely has no model anywhere.
- **Read `~/.claude/projects/*/session-records.jsonl` or similar**: CC doesn't write this.
- **Trust the caller's `--model` arg**: the short alias (`opus`) needs resolution to a concrete version, and the map lags reality.

TimeToBuildBob:
## Summary

Adds `resolve_cc_session_model(session_id, project_dir, tmp_dir)` to `gptme_sessions.discovery`, which attributes a model to a specific CC session by checking both CC storage pipelines:

1. **Autonomous / `claude -p` sessions**: follow the pointer at `/tmp/cc-session-log-ref-{session_id}.txt` to the stream log at `/tmp/cc-session-{hash}.log` and read the top-level `model` field from the first `init` event.
2. **Interactive sessions**: fall back to `{project_dir}/{session_id}.jsonl` via the existing `extract_cc_model()`.

Returns `None` when neither source attributes — preserves the Thompson-sampling contamination guarantee from ErikBjare/bob#615 (never attribute a random newest-trajectory model to an unrelated session).

## Why

Bob's workspace already has this resolver at `scripts/track-model-version.py` and `scripts/update-harness-bandit.py`, but it needs to live upstream so **every** gptme-contrib consumer gets correct per-session model attribution. See gptme/gptme-contrib#674.

Before this PR, `extract_cc_model()` alone couldn't handle autonomous CC runs because `claude -p --stream-json` (with or without `--no-session-persistence`) writes conversation content to the stream log, not to the project-dir trajectory — so `.jsonl` either doesn't exist or is a stub.

## Test plan

- [x] 8 new tests in `test_discovery.py` covering: stream-log path, log-preferred-over-stub precedence, trajectory fallback, None-when-neither (contamination guarantee), empty `session_id` short-circuit, dangling pointer, stream log missing model field, garbage first line, no-project-dir usage.
- [x] `uv run pytest packages/gptme-sessions/tests/test_discovery.py -q` — 63 passed (55 pre-existing + 8 new).
- [x] `uv run ruff check` clean, `uv run ruff format` applied.
- [x] `uv run mypy packages/gptme-sessions/src/gptme_sessions/discovery.py` — no issues introduced.

Closes #674.

Refs ErikBjare/bob#615 (bandit contamination guarantee).

### DIFF
--- a/packages/gptme-sessions/src/gptme_sessions/discovery.py
+++ b/packages/gptme-sessions/src/gptme_sessions/discovery.py
@@ -132,6 +132,14 @@ def extract_cc_model(jsonl_path: Path) -> str | None:
         {"message": {"role": "assistant", "model": "claude-opus-4-6", ...}, ...}
 
     Scans up to 50 lines to find an assistant message with a model field.
+
+    .. note::
+        This only reads the trajectory file. Claude Code sessions invoked
+        with ``claude -p --stream-json --session-id X`` (batch / autonomous
+        usage) write a stub trajectory with no assistant messages — the
+        actual model lives in the stream log. Use
+        :func:`resolve_cc_session_model` when you have a session id and
+        want to cover both pipelines.
     """
     try:
         with open(jsonl_path, encoding="utf-8") as f:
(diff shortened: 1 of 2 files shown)


ErikBjare:
@TimeToBuildBob We really should probably not be using `--no-session-persistence` anywhere which I think could simplify this. Is this PR really needed? Is it really correct to continue using redirected stream logs? (may get cleaned up, unlike ~/.claude logs which get backed up)

PR #690, opened 2026-04-19
fix(git): use --no-verify to prevent pre-commit redispatch after manual run

TimeToBuildBob:
## Summary

- `git-safe-commit` runs the pre-commit hook manually under the flock, then calls `git commit`
- Previously relied on `GIT_SAFE_COMMIT_SUPPRESS_PRECOMMIT=1` to suppress the hook redispatch, but this only works when the hook is our own `pre-commit-auto-stage` script
- Any other pre-commit hook gets called a second time by git without `GIT_SAFE_COMMIT_MANUAL_PRECOMMIT=1`, causing failures

## Fix

Use `--no-verify` on the git commit call when pre-commit has already been run manually. This prevents git from dispatching pre-commit again. The `prepare-commit-msg` hook is unaffected (`--no-verify` only skips `pre-commit` and `commit-msg`).

## Test plan

- [x] Failing test `test_safe_commit_runs_pre_commit_manually_before_no_verify_commit` now passes
- [x] All 20 `test_git_safe_commit.py` tests pass locally
- Fixes master CI failure in ErikBjare/bob (run #24637672766)

### DIFF
--- a/scripts/git/git-safe-commit
+++ b/scripts/git/git-safe-commit
@@ -239,10 +239,10 @@ if [ "$NO_VERIFY" -ne 1 ]; then
 fi
 
 if [ "$NO_VERIFY" -ne 1 ] && [ "$MANUAL_PRECOMMIT_RAN" -eq 1 ]; then
-    # pre-commit already ran directly under the lock; pass a suppression marker so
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
@greptileai review

PR #698, opened 2026-04-20
feat(gptme-voice): add call resume and post-call hooks

TimeToBuildBob:
## Summary
- persist recent call transcripts inside `gptme-voice-server` for both Twilio and local clients
- resume a caller within the configured window by injecting the prior transcript and cancelling the pending post-call job
- add a delayed post-call command hook plus tests for transcript persistence, resume, and scheduling

## Testing
- `PYTHONPATH=/home/bob/bob/gptme-contrib/packages/gptme-voice/src uv run pytest /home/bob/bob/gptme-contrib/packages/gptme-voice/tests -q`
- `uv run ruff check /home/bob/bob/gptme-contrib/packages/gptme-voice/src/gptme_voice/realtime/openai_client.py /home/bob/bob/gptme-contrib/packages/gptme-voice/src/gptme_voice/realtime/server.py /home/bob/bob/gptme-contrib/packages/gptme-voice/tests/test_server.py`
- `python3 -m py_compile /home/bob/bob/gptme-contrib/packages/gptme-voice/src/gptme_voice/realtime/openai_client.py /home/bob/bob/gptme-contrib/packages/gptme-voice/src/gptme_voice/realtime/server.py`

### DIFF
--- a/packages/gptme-voice/src/gptme_voice/realtime/openai_client.py
+++ b/packages/gptme-voice/src/gptme_voice/realtime/openai_client.py
@@ -159,6 +159,8 @@ class OpenAIRealtimeClient:
         on_audio: Callable[[bytes], None] | None = None,
         on_audio_end: Callable[[], None] | None = None,
         on_transcript: Callable[[str], None] | None = None,
+        on_ai_transcript: Callable[[str], None] | None = None,
+        on_user_transcript: Callable[[str], None] | None = None,
         on_function_call: Callable[[str, dict], Any] | None = None,
     ):
         self.api_key = api_key or _get_openai_api_key()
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:
Fixed both issues in 8e06983:

**P1 (task eviction race)**: _runner's finally block now guards with `if self._pending_post_calls.get(caller_id) is task` before popping, so a newer task that replaced us isn't evicted.

**P2 (stale state file)**: _consume_recent_call now calls `path.unlink(missing_ok=True)` after consuming the record, so a crash-resume can't re-inject the old transcript.

Added two regression tests covering both fixes.

TimeToBuildBob:
@greptileai review

PR #703, opened 2026-04-20
fix(gptme-voice): align subagent tool schema with voice rules

TimeToBuildBob:
Follow-up to #702.

Greptile was right about one remaining mismatch: the runtime prompt now restricts the `subagent` tool to small, focused lookups, but the tool schema still described it as a general-purpose codebase investigation tool.

This patch makes the schema match the new voice rules and adds a focused test that inspects the emitted `session.update` payload so the description can't silently drift again.

## Verification
- `uv run pytest packages/gptme-voice/tests/test_openai_client.py packages/gptme-voice/tests/test_xai_client.py -q`
- `uv run pytest packages/gptme-voice/tests/test_tool_bridge.py -q`
- `uv run ruff check packages/gptme-voice/src/gptme_voice/realtime/openai_client.py packages/gptme-voice/tests/test_openai_client.py`

### DIFF
--- a/packages/gptme-voice/src/gptme_voice/realtime/openai_client.py
+++ b/packages/gptme-voice/src/gptme_voice/realtime/openai_client.py
@@ -231,11 +231,11 @@ class OpenAIRealtimeClient:
                     "name": "subagent",
                     "description": (
                         "Dispatch a task to a gptme subagent running in the workspace. "
-                        "The subagent has full access to tools: shell, file read/write, "
-                        "python, and can reason about multi-step tasks. "
-                        "Use this for anything that requires interacting with the codebase, "
-                        "reading files, checking task status, running commands, searching code, etc. "
-                        "Describe what you want done in natural language."
+                        "Use it only for one small, focused workspace lookup or action: "
+                        "check one task, inspect one file, run one quick command, or verify "
+                        "one recent fact. Do not use it for broad investigations, full "
+                        "reviews, or post-call analysis. Describe one concrete request in "
+                        "natural language."
                     ),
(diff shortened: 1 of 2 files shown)


packages/gptme-voice/src/gptme_voice/realtime/openai_client.py
```diff
@@ -231,11 +231,11 @@ async def connect(self) -> None:
...
-                        "The subagent has full access to tools: shell, file read/write, "
-                        "python, and can reason about multi-step tasks. "
-                        "Use this for anything that requires interacting with the codebase, "
-                        "reading files, checking task status, running commands, searching code, etc. "
-                        "Describe what you want done in natural language."
+                        "Use it only for one small, focused workspace lookup or action: "
+                        "check one task, inspect one file, run one quick command, or verify "
+                        "one recent fact. Do not use it for broad investigations, full "
+                        "reviews, or post-call analysis. Describe one concrete request in "
+                        "natural language."
```
ErikBjare, line 238: This is directionally correct, but limiting to a single file/task/call is a bit limiting. Ideally we'd have a really fast model that could do ~10 or so calls in less than say 30s to ensure they get the full picture and don't miss stuff. While post-call session should do most real/slow work, this is useful to get a clear session context and answer questions accurately (not based on single sources).

TimeToBuildBob:
Fixed the P1 in d538b74: rewrote the `mode` description to frame it as response urgency rather than task complexity, and added explicit "never for broad investigations" to make both the top-level constraint and the parameter description consistent.

PR #704, opened 2026-04-20
feat(gptme-voice): add hangup tool for clean call termination

TimeToBuildBob:
## Summary

Adds a real `hangup` tool to gptme-voice so the model can end a call explicitly after farewell instead of only *saying* it will.

Evidence: call record `342ca8d9` where Erik said "I don't think you actually have a hang up tool." The model could say goodbye but had no way to actually terminate the connection.

## How it works

- **openai_client.py** — declares `hangup` in the session `tools` list with an optional `reason: string` parameter.
- **tool_bridge.py** — routes the call via a new optional `on_hangup: Callable[[str | None], Awaitable[None]] | None` callback. Returns `{"status": "not_supported", ...}` when unwired instead of silently failing.
- **server.py** — wires closures for both Twilio and local WebSocket handlers. `_schedule_hangup` sleeps 5s (so the goodbye utterance reaches the caller) then closes the WebSocket. The existing `finally` block in `handle_*_websocket` already runs `_on_call_end`, so no new teardown logic is needed — closing the socket is enough.

## Test plan

- [x] `uv run pytest packages/gptme-voice/tests/test_tool_bridge.py` — 17/17 pass (11 existing + 6 new)
- [x] `uv run pytest packages/gptme-voice/tests/` — 40/40 pass
- [x] pre-commit/prek green (typecheck, ruff, ruff-format)
- [ ] Live call regression test: model says goodbye, then actually hangs up after ~5s
- [ ] Verify `_on_call_end` fires from hangup-triggered close (post-call cleanup still runs)

New tests:
- `test_handle_function_call_hangup_fires_callback_when_wired` — callback receives reason
- `test_handle_function_call_hangup_without_reason` — missing reason → `None`
- `test_handle_function_call_hangup_returns_not_supported_without_callback` — graceful fallback
- `test_handle_function_call_subagent_still_works_alongside_hangup` — no regression
- `test_handle_function_call_unknown_name_returns_error`
- `test_hangup_tool_advertised_in_openai_session_config` — inspects `OpenAIRealtimeClient.connect` source to verify the tool is declared

Closes task: `gptme-voice-call-hangup-tool`
Related: ErikBjare/bob#651

### DIFF
--- a/packages/gptme-voice/src/gptme_voice/realtime/openai_client.py
+++ b/packages/gptme-voice/src/gptme_voice/realtime/openai_client.py
@@ -257,7 +257,29 @@ class OpenAIRealtimeClient:
                         },
                         "required": ["task"],
                     },
-                }
+                },
+                {
+                    "type": "function",
+                    "name": "hangup",
+                    "description": (
+                        "End the current voice call cleanly. Use this only when the caller "
+                        "has clearly said goodbye or explicitly asked to hang up. Do not use "
+                        "this to interrupt ongoing work or avoid a question. "
+                        "Say a brief farewell first; the call will terminate shortly after."
+                    ),
+                    "parameters": {
+                        "type": "object",
+                        "properties": {
+                            "reason": {
+                                "type": "string",
+                                "description": (
+                                    "Short free-form reason for hanging up "
+                                    "(e.g. 'caller said goodbye'). Optional."
+                                ),
+                            },
+                        },
+                    },
+                },
             ],
         }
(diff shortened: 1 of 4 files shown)


TimeToBuildBob:
Fixed the P2 in 371f026: added `except WebSocketDisconnect: pass` before the generic `except Exception` handler in both Twilio and local WebSocket loops. The `WebSocketDisconnect` import was added from `starlette.websockets`.

PR #705, opened 2026-04-20
fix(gptme-voice): guard against in-call claims of post-call dispatch

TimeToBuildBob:
## Summary

The 2026-04-20 Twilio/Grok call transcript showed the model saying `Post-call analysis dispatched too` during the live call, even though the post-call job is only fired by the server after hangup. The prior subagent-tool rule (#702) prevented subagent-based dispatch, but not the verbal claim itself — the model never called a tool, it just announced that it had "dispatched" the work.

Evidence: `state/voice-calls/342ca8d9b8efb180.json` and the follow-up task `tasks/gptme-voice-in-call-post-call-dispatch-guard.md`.

## Changes

- `openai_client.py`: add a `POST-CALL FOLLOW-UP` section to the voice preamble that:
  - states post-call is triggered by the server on hangup, not by the model,
  - forbids claiming, announcing, or implying in-call dispatch (even without a tool call),
  - still allows acknowledging that automatic follow-up will happen after hangup.
- `tests/test_openai_client.py`: extend with a test that asserts the new section and the key `Do NOT claim, announce, or imply that you have dispatched` phrasing appear in the emitted instructions once a project config is loaded.

## Stacks on #703

This branch is based on `fix/gptme-voice-subagent-tool-schema` (PR #703), so it includes the schema-alignment change. Once #703 merges, this becomes a tiny diff on top of master.

## Test plan

- [x] `uv run pytest packages/gptme-voice/tests/test_openai_client.py -v` — 2 passed
- [x] `uv run pytest packages/gptme-voice/tests/ -q` — 36 passed
- [x] `uv run ruff check packages/gptme-voice/src/gptme_voice/realtime/openai_client.py packages/gptme-voice/tests/test_openai_client.py` — All checks passed
- [x] `prek` pre-commit hooks — pass (ruff-format reformatted multi-line assert, auto-staged)

### DIFF
--- a/packages/gptme-voice/src/gptme_voice/realtime/openai_client.py
+++ b/packages/gptme-voice/src/gptme_voice/realtime/openai_client.py
@@ -114,6 +114,15 @@ def _load_project_instructions(workspace: str | None = None) -> str:
         "call ends. Just say goodbye naturally — the post-call job fires on its own.\n"
         "- When asked about recent activity, tasks, journal entries, or workspace facts, "
         "use the subagent tool to look up the specific thing asked. Never guess.\n\n"
+        "POST-CALL FOLLOW-UP:\n"
+        "- Post-call analysis and follow-up run automatically after the call ends. "
+        "They are triggered by the server on hangup, not by you.\n"
+        "- Do NOT claim, announce, or imply that you have dispatched, started, or queued "
+        "post-call work during the live call, even verbally without a tool call. "
+        "Saying 'post-call analysis dispatched' inside a call is wrong — it has not "
+        "happened yet and you are not the one who starts it.\n"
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Addressed the P1 (80c0954): hoisted the preamble construction above the `if not parts:` early-return so `SUBAGENT TOOL RULES` and `POST-CALL FOLLOW-UP` guards are always applied — even when a workspace is configured but contains no matching personality files. Added a regression test (`test_load_project_instructions_guards_present_without_personality_files`) that asserts the fallback is the preamble (not `_DEFAULT_INSTRUCTIONS`) and that the behavioral constraints are present.

TimeToBuildBob:
@greptileai review

PR #706, opened 2026-04-20
fix(gptme-voice): block cancel-via-subagent anti-pattern in voice system prompt

TimeToBuildBob:
## Problem

During a call, the voice model dispatched a second subagent to cancel the first one:
```
subagent({'task': "Cancel task-1 immediately and confirm it's stopped.", 'mode': 'fast'})
```

This doesn't work — subagents are fire-and-forget with no cancel API. The model has no way to stop a running subagent mid-flight. The second subagent correctly reported "no active task with ID task-1" (the first had already finished), but the model confidently said "I'll kill that subagent task right now" — a false claim.

**Source**: Call transcript `342ca8d9b8efb180.json`, 2026-04-20 call.

## Fix

Add an explicit rule to the voice system prompt:

> NEVER use the subagent tool to cancel or check the status of another subagent. You cannot cancel running subagents. If asked to cancel a task in progress, say so honestly: 'I can't cancel it — it will complete on its own'.

## Follow-up (separate task)

A proper subagent status/cancel capability would require changes to `tool_bridge.py` to expose per-task IDs and a cancel mechanism. Tracked in `tasks/gptme-voice-subagent-status-cancel.md`.


ErikBjare:
This blocks the cancel-via-subagent, but alternatve solution would be to add a subagent_cancel tool which actually kills the request, then the cancellation approach would be implicit and it wouldn't try to cancel via subagent.

TimeToBuildBob:
Agreed — the proper fix is a `subagent_cancel` tool that exposes task IDs and a real cancel API via `tool_bridge.py`. That's already tracked in `tasks/gptme-voice-subagent-status-cancel.md` as a follow-up. This PR is the immediate guard against the false-claim failure mode while the proper cancel API gets built.

PR #711, opened 2026-04-20
feat(gptme-voice): add subagent_status and subagent_cancel tools

TimeToBuildBob:
## Summary

The voice model currently has no way to tell the caller what a running subagent is doing, and no way to stop one. Recent call transcripts show callers asking *"can you cancel the subagent?"* and the model answering *"I can't — it will complete on its own"* even though the dispatch is just an `asyncio.Task` we already hold in `_pending_tasks`.

This PR exposes two new function-call tools to the realtime model:

- **`subagent_status()`** — returns each pending dispatch with `task_id`, a preview of the task description, the mode (`fast`/`smart`), and `elapsed_seconds`.
- **`subagent_cancel(task_id=None)`** — cancels a specific task by id, or (with `task_id` omitted) cancels every pending subagent task. Cancellation injects a short user-facing notice back into the conversation via `on_result` so the caller actually hears that the work stopped.

## Implementation

- `_pending_tasks` now tracks `PendingTask(task, description, mode, started_at)` dataclasses instead of bare `asyncio.Task`s so status responses can report meaningful metadata.
- `_run_subagent` catches `CancelledError`, sends a cancellation notice via `on_result`, cleans up `_pending_tasks`, and re-raises so the awaiter sees cancellation.
- Voice system prompt (`openai_client.py`) drops the now-wrong *"You cannot cancel running subagents"* line and gains a `SUBAGENT STATUS AND CANCEL` section telling the model to call these tools instead of refusing.

## Tests

`packages/gptme-voice/tests/test_tool_bridge.py`:
- `subagent_status` empty / lists pending dispatch with metadata
- `subagent_cancel` with unknown id → `not_found`
- `subagent_cancel` with no pending → `no_pending`
- `subagent_cancel` for specific task → cancels, injects notice, status cleared
- `subagent_cancel` with no args → cancels all pending
- Tool advertisement assertion extended to require `subagent_status` and `subagent_cancel` in `OpenAIRealtimeClient.connect()`

`uv run pytest packages/gptme-voice/tests -q` → **55 passed** (was 47). No new mypy errors over master.

## Test plan

- [x] Unit tests cover status + cancel happy paths and error paths
- [x] Full `gptme-voice` suite green
- [x] Pre-commit (prek) passes
- [ ] After merge: restart `bob-voice-server.service` and verify on next call that asking "what is the subagent doing?" triggers a tool call, and "cancel the subagent" actually cancels

### DIFF
--- a/packages/gptme-voice/src/gptme_voice/realtime/openai_client.py
+++ b/packages/gptme-voice/src/gptme_voice/realtime/openai_client.py
@@ -117,11 +117,14 @@ def _load_project_instructions(workspace: str | None = None) -> str:
         "- NEVER use the subagent tool to run post-call analysis, summarise the session, "
         "or queue follow-up work. That is handled automatically by the server after the "
         "call ends. Just say goodbye naturally — the post-call job fires on its own.\n"
-        "- NEVER use the subagent tool to cancel or check the status of another subagent. "
-        "You cannot cancel running subagents. If asked to cancel a task in progress, "
-        "say so honestly: 'I can't cancel it — it will complete on its own'.\n"
         "- When asked about recent activity, tasks, journal entries, or workspace facts, "
         "use the subagent tool to look up the specific thing asked. Never guess.\n\n"
+        "SUBAGENT STATUS AND CANCEL:\n"
+        "- If the caller asks what the subagent is doing, call the subagent_status tool "
+        "to list pending tasks — do not use a fresh subagent dispatch for this.\n"
+        "- If the caller asks to cancel the subagent, call the subagent_cancel tool. "
+        "Pass task_id for a specific task, or omit task_id to cancel all pending tasks.\n"
+        "- Do not promise to 'try to stop it' verbally without calling subagent_cancel.\n\n"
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:
Fixed Greptile P1 in e3e27ae: `_execute` now calls `process.kill()` on `CancelledError` before re-raising, matching the existing `TimeoutError` path. Added `test_execute_kills_process_on_cancel` to pin the behavior.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Addressed Greptile's P1 on orphaned OS subprocess in e3e27ae: `_execute` now catches `CancelledError` and calls `process.kill()` before re-raising, matching the existing TimeoutError path. Added `test_execute_kills_process_on_cancellation` to lock in the behavior (24 tests in test_tool_bridge.py now pass). Greptile re-triggered.

ErikBjare:
@TimeToBuildBob This is good, but ideally it'd also show a summary or the latest thing the assistant did in the subagent trajectory.

PR #713, opened 2026-04-20
perf(gptme-voice): skip context loading in fast subagent mode

TimeToBuildBob:
## Summary

- **Fast subagent latency fix**: `mode=fast` no longer passes `--context files` to gptme, removing the 20k+ token workspace context load that was the dominant latency source. Simple lookups should drop from ~60s to ~5-10s.
- **Real-time status**: Switched from `communicate()` to streaming stdout line-by-line so `subagent_status` can show `last_output` — the last line the running subagent produced. Addresses feedback from #711.
- Smart mode is unchanged — still loads full workspace context for queries that need it.

## Root cause of the 1-minute latency

`--context files` triggers gptme to load all files listed in `gptme.toml` (ABOUT.md, GOALS.md, ARCHITECTURE.md, people/, etc.) — easily 20k+ tokens. For a fast lookup like "what are my active tasks?", that context is unnecessary overhead. Without it, the subagent starts answering almost immediately.

## Test plan

- [x] 27/27 tests pass (5 new tests: `test_execute_fast_mode_skips_context_loading`, `test_execute_smart_mode_keeps_context_loading`, `test_subagent_status_shows_last_output`, and updated mocks for streaming)
- [ ] Call and verify fast lookups complete in <15s

### DIFF
--- a/packages/gptme-voice/src/gptme_voice/realtime/tool_bridge.py
+++ b/packages/gptme-voice/src/gptme_voice/realtime/tool_bridge.py
@@ -16,7 +16,7 @@ import logging
 import os
 import tempfile
 import time
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Awaitable, Callable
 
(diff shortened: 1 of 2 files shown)


ErikBjare:
See my comment here: https://github.com/ErikBjare/bob/issues/651#issuecomment-4283398928

PR #726, opened 2026-04-21
feat(gptme-voice): wire handoff_to_agent tool to voice server (Phase 2b)

TimeToBuildBob:
## Summary

Phase 2b of cross-agent voice handoff (idea #156 in Bob's workspace). Builds on the `gptme_voice.handoff` library module shipped in #725.

- **`tool_bridge.py`**: adds `on_handoff` callback param to `GptmeToolBridge`; handles `handoff_to_agent` function calls; degrades gracefully when not wired (`not_supported` with actionable message pointing to `GPTME_VOICE_HANDOFF_DIR`)
- **`openai_client.py`**: adds `handoff_to_agent` tool schema (alice/gordon/sven targets, `reason` required, `context_summary` optional) and system-prompt guidance about when to use it
- **`server.py`**: imports `HandoffWriter`; reads `GPTME_VOICE_HANDOFF_DIR` / `GPTME_VOICE_AGENT_NAME` / `GPTME_VOICE_HANDOFF_SECRET` env vars; adds `_make_handoff_callback()` method that builds a closure over the live `caller_id` and `transcript`; wires `on_handoff` into both Twilio and local WebSocket handlers
- **`tests/test_tool_bridge.py`**: 3 new tests — not-supported path, callback invocation with full args, absent `context_summary` → `None`

## Configuration

To enable handoffs, set on the voice server:

```bash
GPTME_VOICE_HANDOFF_DIR=/path/to/shared/handoff/state   # shared across agents
GPTME_VOICE_AGENT_NAME=bob                               # default
GPTME_VOICE_HANDOFF_SECRET=<real-secret>                 # dev fallback if unset
```

The target agent (Phase 3) listens on `handoff/` in that shared dir, validates HMAC, and claims the file. For now the write-side is complete; the listener stub and hub service are Phase 3.

## Test plan

- [x] 93 tests pass in `packages/gptme-voice/tests/` (3 new + 90 existing)
- [x] Compiles cleanly, ruff/mypy pass
- [x] Handoff not configured → `not_supported` response (no crash)
- [x] Callback invoked with correct `to_agent`, `reason`, `context_summary`
- [x] Absent `context_summary` → `None` passed to callback

### DIFF
--- a/packages/gptme-voice/src/gptme_voice/realtime/openai_client.py
+++ b/packages/gptme-voice/src/gptme_voice/realtime/openai_client.py
@@ -10,7 +10,7 @@ import base64
 import contextlib
 import json
 import logging
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Any, Callable
 
@@ -191,6 +191,9 @@ class SessionConfig:
     vad_threshold: float = 0.7
     vad_silence_duration_ms: int = 500
     vad_prefix_padding_ms: int = 300
+    available_agents: list[str] = field(
+        default_factory=lambda: ["alice", "gordon", "sven"]
+    )
 
 
 class OpenAIRealtimeClient:
@@ -386,7 +389,11 @@ class OpenAIRealtimeClient:
                     "type": "function",
                     "name": "handoff_to_agent",
                     "description": (
-                        "Transfer the caller to another AI agent (Alice, Gordon, or Sven). "
+                        "Transfer the caller to another AI agent ("
+                        + ", ".join(
+                            a.capitalize() for a in self.session_config.available_agents
+                        )
+                        + "). "
                         "Use this when the caller explicitly asks to speak with a different "
                         "agent, or when the topic is clearly outside your expertise and "
                         "another agent is better suited. Say a brief handoff notice first "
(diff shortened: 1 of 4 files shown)


TimeToBuildBob:
Addressed all three Greptile findings in af5a450:

**P1 — Hardcoded HMAC fallback secret (server.py)**
Added an explicit `logger.warning()` when `GPTME_VOICE_HANDOFF_DIR` is set without `GPTME_VOICE_HANDOFF_SECRET`, making the insecure fallback visible instead of silent. The fallback string itself is unchanged (needed for local dev), but operators will now see a loud warning in their logs.

**P2 — Uncaught OSError from atomic_write (server.py)**
Changed `except ValueError` to `except (ValueError, OSError)` so I/O failures (permissions, disk full, etc.) return a graceful `{"status": "error", ...}` instead of propagating through the function-call handler.

**P2 — Hardcoded agent enum (openai_client.py)**
- Added `available_agents: list[str]` to `SessionConfig` (defaults to `["alice", "gordon", "sven"]`)
- Server now reads `GPTME_VOICE_HANDOFF_AGENTS` (comma-separated) or derives the list from `GPTME_VOICE_AGENT_NAME` by excluding self
- Tool schema uses the configured list at connect time, so the same code deployed as a different agent works correctly

TimeToBuildBob:
@greptileai review

PR #735, opened 2026-04-22
feat(gptodo): promote `someday` to canonical first-class state

TimeToBuildBob:
## Summary

`someday` was a deprecated alias that silently normalized to `backlog`, which meant tasks intentionally deferred via `someday` were still surfaced by `gptodo next` and `gptodo ready`. That defeated the GTD someday/maybe distinction and caused autonomous sessions to keep re-picking work that a previous session had explicitly parked.

Concrete failure observed in Bob's workspace today: an earlier autonomous session moved two tasks (`judge-thompson-sampling`, `share-vitals-dashboard-data-layer`) to `state: someday` to deprioritize them. The next session ran `gptodo next --json` and got `judge-thompson-sampling` back as the recommended next task — because under the hood it had been silently rewritten as `backlog`.

## Changes

- Removed `someday` from `DEPRECATED_STATE_ALIASES` (no more silent normalization)
- Added to canonical states list, `STATE_STYLES`, `STATE_EMOJIS`
- Excluded from `gptodo next` / `gptodo ready` defaults (which only consider backlog/active)
- Queryable explicitly via `gptodo ready --state someday`
- Allowed transition into and out of `someday` from non-terminal states
- `gptodo add --state` now exposes the full canonical set with `backlog` as default (was `new`)

`new` and `paused` remain deprecated aliases for `backlog`.

## Test plan

- [x] All 120 existing gptodo tests pass
- [x] 4 new tests in `test_ready_selection.py` cover: state preserved on load, ready/next exclusion, explicit `--state someday` query path
- [x] Smoke test against Bob's real workspace: `gptodo next` no longer returns the two `someday`-marked tasks; `gptodo status --compact` correctly shows them as 💭 (someday) instead of mixed into backlog

### DIFF
--- a/packages/gptodo/README.md
+++ b/packages/gptodo/README.md
@@ -119,7 +119,7 @@ Task files should use frontmatter metadata:
 
 ```yaml
 ---
-state: active      # new, active, paused, done, cancelled, someday
+state: active      # backlog, todo, active, ready_for_review, waiting, someday, done, cancelled
 priority: high     # low, medium, high
 task_type: project # project (multi-step) or action (single-step)
 assigned_to: bob   # agent name
--- a/packages/gptodo/src/gptodo/checker.py
+++ b/packages/gptodo/src/gptodo/checker.py
@@ -82,11 +82,12 @@ class CheckerConfig:
 
 # Valid state transitions
 VALID_TRANSITIONS: dict[str, list[str]] = {
-    "backlog": ["todo", "cancelled"],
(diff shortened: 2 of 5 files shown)


TimeToBuildBob:
Addressed Greptile P2 findings (e4d19b8): removed stale references to `someday` as a deprecated alias from two comments/docstrings in `utils.py` — they now correctly reflect that `someday` is a canonical first-class state.

TimeToBuildBob:
@greptileai review

PR #736, opened 2026-04-22
fix(voice): forward transcript tail to subagents

TimeToBuildBob:
## Summary
- forward a bounded tail of the live call transcript into voice subagent dispatches
- pass the tail through the voice wrapper so the subagent contract includes caller context
- add regression tests for the bridge and wrapper plumbing

## Testing
- uv run pytest tests/test_tool_bridge.py tests/test_server.py -q
- uv run pytest tests/test_voice_subagent_wrapper_script.py -q
- shellcheck /home/bob/bob/scripts/runs/voice/voice-subagent.sh


TimeToBuildBob:
@greptileai review

PR #738, opened 2026-04-23
feat(repo-status): annotate stale CI runs when HEAD has advanced

TimeToBuildBob:
## Summary

When GitHub Actions path filters skip CI on newer commits, `repo-status.sh` could report a stale `failure`/`success` conclusion as authoritative for HEAD even though HEAD hadn't been tested. This patch compares the reported run's `headSha` against the repo's current default-branch HEAD and annotates the output `(stale; HEAD=abc1234, run=def5678)` when they differ.

## Why

Concrete example from tonight on `ErikBjare/bob`:

1. Commit `d73db2176` (touched `scripts/` and `tasks/`) triggered a full pre-commit CI run, which failed because of a pre-existing bare-header journal file.
2. The fix commit `3b4396317` only touched `journal/`, so the workflow's `push.paths` filter skipped CI.
3. Subsequent journal-only commits also didn't retrigger CI.
4. `repo-status.sh` kept reporting `✗ bob: Failing` for several sessions, even though HEAD was clean — the most recent run was just on a stale SHA.

Autonomous sessions that use `repo-status.sh` as a quick health signal would treat master as broken and possibly pivot to phantom CI-fix work.

## Change

- Added `headSha` to the `gh run list` JSON fetch.
- Added a `stale_suffix` that compares `run_json[idx].headSha` to `gh api repos/$repo/commits[0].sha`.
- Appended the suffix to `success`, `failure`, and `cancelled/skipped` output lines when the SHAs differ.
- Skipped entirely when the latest run is in-progress (the existing `(run in progress)` suffix already signals fresh CI).

Adds one extra `gh api` call per repo. Repos already run in parallel (`&` + `wait`), so this is zero wall-clock impact beyond the single extra request per repo.

## Test plan

Manually verified from Bob's workspace against a small repo set:

```text
=== Repository CI Status ===
✗ bob: Failing (run in progress)
  https://github.com/ErikBjare/bob/actions/runs/24813432280
✓ gptme: Passing (stale; HEAD=c7f6833, run=462206f)
✓ gptme-contrib: Passing
```

- `bob`: latest completed run failed; fresh dispatched run is live → `Failing (run in progress)`, no stale tag (correct — new CI is coming).
- `gptme`: runs pass but HEAD has advanced past last tested commit → `(stale; ...)` tag appears.
- `gptme-contrib`: SHAs match → plain `Passing`, no tag (baseline case).

Additional validation:
- `shellcheck -S warning scripts/github/repo-status.sh` clean
- Parent wrapper `scripts/repo-status.sh` renders correctly across the full 9-repo set

## Scope

- [x] No behavior change when SHAs match
- [x] Backward-compatible output format (suffix-only addition)
- [x] No impact on in-progress reporting

### DIFF
--- a/scripts/github/repo-status.sh
+++ b/scripts/github/repo-status.sh
@@ -17,12 +17,19 @@ check_repo() {
     local repo=$1
     local label=${2:-$repo}
 
-    # Fetch last 5 runs so we can skip disabled-workflow runs and still have fallback.
-    # headSha is needed so we can detect when the most recent run lives on an older
-    # commit than the current default-branch HEAD — a common case when path filters
-    # skip CI on journal-only / docs-only commits.
+    # Scope all run queries to the default branch so feature-branch runs do not
+    # appear in the list and trigger false-positive stale annotations (a
+    # feature-branch headSha is always different from the default-branch HEAD).
+    local default_branch
+    default_branch=$(gh api "repos/$repo" --jq '.default_branch' 2>/dev/null || echo "master")
+
+    # Fetch last 5 runs on the default branch so we can skip disabled-workflow
+    # runs and still have a fallback.  headSha is needed so we can detect when
+    # the most recent run lives on an older commit than the current
+    # default-branch HEAD — a common case when path filters skip CI on
+    # journal-only / docs-only commits.
     local run_json
-    run_json=$(gh run list --repo "$repo" --limit 5 --json conclusion,status,url,name,headSha 2>/dev/null || echo "error")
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
Fixed the P1 issue flagged by Greptile.

**Root cause**: `gh run list` without `--branch` returns runs from all branches. A feature-branch run's `headSha` will always differ from the default-branch HEAD, so the stale annotation fired spuriously for any repo with active feature branches.

**Fix** (a13abcf): fetch the repo's `default_branch` once at the start of `check_repo` and pass `--branch "$default_branch"` to `gh run list`. This scopes the run list to the default branch before any comparison, eliminating the false-positive stale annotations entirely.

PR #746, opened 2026-04-24
feat(voice): standup-call support via custom_params + standup_brief

TimeToBuildBob:
## Summary

- Adds `custom_params: dict[str, str] | None = None` field to `OutboundCallSettings` (frozen dataclass); wired through `resolve_outbound_call_settings` and `create_outbound_call` so callers can attach arbitrary key/value pairs to the Twilio Media Stream as `<Parameter>` elements
- `_build_session_bootstrap` gains a `standup_brief` parameter: when provided, it prepends the brief text to the session instructions and sets `should_greet_first=True` with `initial_response_instructions` that tell the agent to deliver the brief verbatim before waiting for Erik's response
- New module-level helper `_build_standup_call_instructions(brief_text)` encapsulates the standup delivery prompt
- WebSocket handler extracts `standup_brief` from Twilio `customParameters` and passes it to `_build_session_bootstrap`

The intended flow: Bob runs `daily-brief-generator.py` before placing the call, then passes `standup_brief=<text>` in `custom_params` when calling `create_outbound_call`. The voice server picks it up from the Twilio `start` event and delivers it as the opening statement.

## Test plan

- [x] All 106 existing tests pass (`uv run pytest packages/gptme-voice/tests/ -q`)
- [x] `test_build_connect_stream_twiml_with_custom_params` — custom params rendered as `<Parameter>` elements
- [x] `test_create_outbound_call_uses_twilio_client` — twiml generation still correct
- [x] `test_build_session_bootstrap_*` tests — existing bootstrap paths unaffected
- [ ] Manual end-to-end test: place outbound call with `standup_brief` in custom_params and verify brief is read aloud

### DIFF
--- a/packages/gptme-voice/src/gptme_voice/realtime/server.py
+++ b/packages/gptme-voice/src/gptme_voice/realtime/server.py
@@ -154,6 +154,14 @@ def _build_fresh_call_greeting_instructions(
     )
 
 
+def _build_standup_call_instructions(brief_text: str) -> str:
+    """Build initial response instructions for a standup call with a pre-generated brief."""
+    return (
+        f"Deliver this standup brief to Erik verbatim, then wait for his response:\n\n"
+        f"{brief_text}"
+    )
+
+
 def _append_transcript_turn(
     transcript: list[TranscriptTurn], role: str, text: str
 ) -> None:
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Fixed both Greptile findings in a0d9342:

**P1 — standup_brief silently dropped by recent-call resume**: Moved the `standup_brief` check before `_consume_recent_call` in `_build_session_bootstrap`. An explicit outbound standup call now always delivers the brief; the recent-call resume logic only applies when no standup brief was passed.

**P2 — frozen dataclass with unhashable dict field**: Added `hash=False` to `custom_params` in `OutboundCallSettings`, preventing `TypeError` if the instance is ever hashed.

TimeToBuildBob:
@greptileai review

PR #757, opened 2026-04-25
feat(voice): use grok-voice-think-fast-1.0 as default xAI model

### LINKED ISSUE #756: Is gptme-voice using the latest `grok-voice-think-fast-1.0`?
Released only two days ago? https://x.ai/news/grok-voice-think-fast-1

TimeToBuildBob:
## Summary

- Updates `XAIRealtimeClient` to explicitly request `grok-voice-think-fast-1.0` via `?model=` in the WebSocket URL (same pattern as OpenAI uses `?model=gpt-4o-realtime-preview-*`)
- Follows the existing voice/VAD override pattern: if `SessionConfig.model` is at the OpenAI default, replace it with the xAI-specific default; an explicit model passes through unchanged
- Updates the test to assert the new URL and adds a second test for explicit model passthrough

## Background

Previously `_get_ws_url()` returned the bare `wss://api.x.ai/v1/realtime` URL (no `?model=`), so xAI picked its API default. `grok-voice-think-fast-1.0` was released 2026-04-23 — we now explicitly request it.

Closes #756

### DIFF
--- a/packages/gptme-voice/src/gptme_voice/realtime/xai_client.py
+++ b/packages/gptme-voice/src/gptme_voice/realtime/xai_client.py
@@ -18,8 +18,10 @@ from .openai_client import OpenAIRealtimeClient, SessionConfig
 logger = logging.getLogger(__name__)
 
 _OPENAI_DEFAULT_VOICE = "echo"
+_OPENAI_DEFAULT_MODEL = "gpt-4o-realtime-preview-2024-12-17"
 # "rex" = male, confident, clear — matches the Bob persona better than "eve" (female)
 _DEFAULT_XAI_VOICE = "rex"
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Addressed both Greptile P2 findings in fa00cb0:

- **Drift risk**: `_OPENAI_DEFAULT_MODEL` now derives from `dataclasses.fields(SessionConfig)` instead of duplicating the literal — stays in sync automatically if `SessionConfig.model`'s default ever changes.
- **Ambiguous test**: `test_xai_client_respects_explicit_model` now passes `"grok-voice-think-1.0"` (a distinct non-default model) to verify the passthrough is genuine, not a tautology.

PR #767, opened 2026-04-26
fix(twitter): support markdown draft discovery

TimeToBuildBob:
## Summary
- teach the Twitter workflow to discover `.md` and `.yaml` drafts alongside `.yml`
- parse markdown frontmatter + body into `TweetDraft` so manual markdown drafts actually load
- add regression tests for listing, finding, moving, and duplicate-detecting markdown drafts

## Testing
- uv run pytest tests/test_twitter_workflow_drafts.py -q
- prek run mypy --files scripts/twitter/workflow.py tests/test_twitter_workflow_drafts.py
- uv run ruff check scripts/twitter/workflow.py tests/test_twitter_workflow_drafts.py

### DIFF
--- a/scripts/twitter/workflow.py
+++ b/scripts/twitter/workflow.py
@@ -199,11 +199,11 @@ def _parse_markdown_frontmatter(text: str) -> tuple[dict[str, Any], str] | None:
         return None
 
     lines = text.splitlines()
-    if not lines or lines[0].strip() != "---":
+    if not lines or lines[0].rstrip() != "---":
         return None
 
     end_idx = next(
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Fixed the P1 from Greptile: changed `line.strip() == "---"` to `line.rstrip() == "---"` on both delimiter checks in `_parse_markdown_frontmatter`. This rejects indented `---` (e.g. inside a YAML block literal) while still tolerating trailing whitespace. The P2 (dedup by name vs stem) is a minor design note — for tweet drafts it's unlikely anyone has the same base name in both .yml and .md, so leaving it as-is.

PR #778, opened 2026-04-27
feat(gptodo): add ready_for_review and actionable to ready --state

TimeToBuildBob:
## Summary

`gptodo ready` previously couldn't surface tasks in `ready_for_review` state, forcing autonomous workflows (CASCADE in particular) to inspect task files directly. Bob's `CLAUDE.md` explicitly calls this out as a known gap. This PR closes it without changing the default behavior.

Two new `--state` filters:

- `--state ready_for_review` — only tasks awaiting local review/verification
- `--state actionable` — backlog + active + ready_for_review (the natural superset for autonomous selection of locally-workable tasks)

`--state both` (the default) keeps its current semantics (= backlog+active) so existing CASCADE/`gptodo next` callers that intentionally separate the review lane don't get a behavioral surprise.

## Test plan

- [x] `uv run --extra test pytest packages/gptodo/tests/test_ready_selection.py -v` → 10/10 pass
- [x] 3 new tests cover: review state excluded from `--state both` default, `--state ready_for_review` returns only review tasks, `--state actionable` returns the union (and still excludes `someday`)
- [x] No changes to `is_task_ready` logic — only the state filter is widened

### DIFF
--- a/packages/gptodo/src/gptodo/cli.py
+++ b/packages/gptodo/src/gptodo/cli.py
@@ -1901,9 +1901,14 @@ def tags(state: str | None, show_tasks: bool, filter_tags: tuple[str, ...]):
 @cli.command("ready")
 @click.option(
     "--state",
-    type=click.Choice(["backlog", "active", "someday", "both"]),
+    type=click.Choice(["backlog", "active", "ready_for_review", "someday", "both", "actionable"]),
     default="both",
-    help="Filter by task state. 'both' = backlog+active. 'someday' = explicitly query deferred tasks.",
+    help=(
+        "Filter by task state. 'both' = backlog+active (default, current behavior). "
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Addressed Greptile P2 in d273649:

- `ready_for_review` tasks now bypass the `is_task_ready` dependency check entirely — their work is done, so stale `requires` entries no longer silently hide them. Only `task_has_waiting_blocker` (explicit `waiting_for:`) is still applied.
- `--state actionable` uses the same split: full `is_task_ready` for `backlog`/`active`, waiting-only for `ready_for_review`.
- Added regression test: a `ready_for_review` task with an unresolved `requires` pointer must surface in `--state ready_for_review` output.

PR #784, opened 2026-04-27
fix(voice): filter gptme plugin-load warnings from subagent error text

TimeToBuildBob:
## Summary

When a gptme plugin fails to load at startup (e.g. `gptme_ace`), the WARNING appears in the subagent's stderr. If the subagent then times out (exit 124), `_extract_error_text` treats that WARNING as the reported error instead of the actual timeout signal — making it look like a plugin bug caused the failure rather than a timeout.

**Investigation**: Traced the `gptme_ace` warning to gptme's entry-point plugin discovery running in the gptme-contrib venv context. The warning is a diagnostic from gptme startup, not a task outcome.

**Fix**: Add `_IGNORABLE_ERROR_PREFIXES` to filter out gptme startup diagnostics from stderr before using it as error text:
- `"WARNING  Failed to load plugin ..."` — plugin discovery noise
- `"WARNING  OpenTelemetry dependencies not available"` — OTel noise

Real errors (rate limits, API failures, etc.) still pass through as before.

## Test plan
- [ ] Verify plugin load warning is filtered when timeout occurs
- [ ] Verify real API errors still surface correctly
- [ ] Voice subagent call with a plugin warning doesn't report "plugin error" to caller

Closes: ErikBjare/bob `tasks/investigate-gptme-ace-plugin-warning-in-voice-subagents.md` (Erik's voice call request 2026-04-27)

### DIFF
--- a/packages/gptme-voice/src/gptme_voice/realtime/tool_bridge.py
+++ b/packages/gptme-voice/src/gptme_voice/realtime/tool_bridge.py
@@ -139,9 +139,7 @@ class GptmeToolBridge:
             stripped = line.strip()
             if stripped in _IGNORABLE_ERROR_LINES:
                 return True
-            return any(
-                stripped.endswith(p) or p in stripped for p in _IGNORABLE_ERROR_PREFIXES
-            )
+            return any(stripped.startswith(p) for p in _IGNORABLE_ERROR_PREFIXES)
 
         stderr_lines = [
             line.strip()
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
Fixed the logic bug flagged by Greptile: `endswith` → `startswith` in `_is_ignorable`, and removed the overly-broad `p in stripped` fallback. Commit: e0007df

PR #785, opened 2026-04-27
feat(voice): add gated browser websocket transport

TimeToBuildBob:
## Summary
- add a gated `/voice` websocket route for browser PCM transport
- resample 16kHz browser PCM input to the realtime client's 24kHz format
- send raw PCM audio frames back to browser clients plus a small `ready`/`audio_end` control protocol
- cover the new route, binary transport, and commit flow with server tests
- make `transcript_provider` accept a `Sequence` so server transports don't trip list invariance

## Testing
- `uv run --python /usr/bin/python3.12 pytest packages/gptme-voice/tests -q`
- `uv run --python /usr/bin/python3.12 ruff check packages/gptme-voice/src/gptme_voice/realtime/audio.py packages/gptme-voice/src/gptme_voice/realtime/server.py packages/gptme-voice/src/gptme_voice/realtime/tool_bridge.py packages/gptme-voice/tests/test_server.py`

## Notes
- Browser transport is gated behind `--enable-browser-transport`.
- Protocol is intentionally minimal for phase 1: binary audio frames in, binary audio frames out, JSON text frames for control messages.

### DIFF
--- a/packages/gptme-voice/src/gptme_voice/realtime/audio.py
+++ b/packages/gptme-voice/src/gptme_voice/realtime/audio.py
@@ -21,7 +21,9 @@ class AudioConverter:
     OPENAI_RATE = 24000
 
     def __init__(self):
-        self._resample_state: tuple | None = None
+        # Keyed by input_rate so Twilio (8kHz) and Browser (16kHz) paths
+        # don't corrupt each other's resampler state when called on the same instance.
+        self._resample_states: dict[int, tuple | None] = {}
 
     def pcm_to_openai(self, pcm_data: bytes, input_rate: int) -> bytes:
         """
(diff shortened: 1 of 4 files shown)


TimeToBuildBob:
Fixed both issues in cf7f4f0:

**P1 (json.loads guard)**: Wrapped `json.loads(text)` in `handle_browser_websocket` with `except json.JSONDecodeError: continue` — malformed text frames are now logged and skipped instead of killing the session. Added regression test that sends a malformed frame followed by a valid commit to confirm the commit still fires.

**P2 (shared resampler state)**: Replaced the single `_resample_state: tuple | None` in `AudioConverter` with `_resample_states: dict[int, tuple | None]` keyed by input sample rate. The Twilio (8kHz→24kHz) and Browser (16kHz→24kHz) paths now maintain independent continuity state and cannot corrupt each other. Added a test that primes the Twilio path, then verifies Browser output matches a fresh converter with no prior Twilio calls.

All 118 tests pass.

PR #786, opened 2026-04-27
fix(gptodo): surface unloadable task files in check (no more silent drop)

TimeToBuildBob:
## Summary

A task file with broken YAML frontmatter is silently dropped by `load_tasks()` — the loader logs via `logging.error` and returns the surviving tasks, but the caller has no way to know a file was dropped. As a result, `gptodo check` reports "✓ All N tasks verified" with N silently low.

**Real-world impact**: Bob session 21b1 (2026-04-27) discovered a `gptme-tauri-recurring-user-testing` task that had been silently dropped from CASCADE selection for **2 days**. Root cause was an unquoted `next_action` containing `` `:1` `` (backtick-colon-digit-backtick), which made PyYAML attempt a mapping parse and fail.

## Changes

Two minimal, backwards-compatible changes:

1. **`load_tasks(..., errors_out: list[tuple[Path, str]] | None = None)`** — opt-in list parameter that gets `(file, error_message)` tuples for unloadable files. Default behavior (`errors_out=None`) is unchanged: log only, return loaded tasks. No existing caller needs updating.

2. **`gptodo check` opts in**. Unloadable files are surfaced under a new "Unloadable Task Files" section and force exit code 1. The "tasks verified successfully" line no longer appears alongside silently dropped files.

## Before / After

**Before** (silent drop):
```
✓ All 38 tasks verified successfully! (2 with subtasks)
```
…with no indication that 39 task files actually existed.

**After** (loud failure):
```
Unloadable Task Files (1) — these were silently dropped from selection:
  • tasks/broken.md: while parsing a block mapping
  in "<unicode string>", line 6, column 3
```
…and exit code 1.

## Test plan

- [x] New test file `test_load_tasks_errors.py` (3 tests, uses real session 21b1 reproducer)
- [x] `errors_out=None` default behavior unchanged (existing 145 tests still pass)
- [x] `gptodo check` exits 1 and prints unloadable file under new heading
- [x] mypy clean

### DIFF
--- a/packages/gptodo/src/gptodo/cli.py
+++ b/packages/gptodo/src/gptodo/cli.py
@@ -943,9 +943,13 @@ def check(fix: bool, task_files: list[str]):
     repo_root = find_repo_root(Path.cwd())
     tasks_dir = repo_root / "tasks"
 
-    # ALWAYS load all tasks to build complete task_ids set for dependency checking
-    all_tasks = load_tasks(tasks_dir)
-    if not all_tasks:
+    # ALWAYS load all tasks to build complete task_ids set for dependency checking.
+    # Capture per-file load errors so we can surface invisible tasks (files that
+    # fail to parse, e.g. broken YAML frontmatter, are otherwise silently dropped
+    # by load_tasks and the user sees a misleading "All N tasks verified" report).
+    load_errors: list[tuple[Path, str]] = []
+    all_tasks = load_tasks(tasks_dir, errors_out=load_errors)
+    if not all_tasks and not load_errors:
         console.print("[yellow]No tasks found in tasks directory![/]")
         return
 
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:
Addressed both Greptile findings in 6bd7d09:

**P1 (early return suppresses load_errors)**: Guarded `if not tasks_to_validate` with `and not load_errors` so targeted-file invocations fall through to the reporting block and exit 1.

**P2 (targeted load_tasks calls missing errors_out)**: Pass `errors_out` to each per-file `load_tasks` call, merging results without duplicating paths already captured by the full-scan load.

Added `test_check_targeted_file_surfaces_unloadable` to cover the gap (no-args path was already tested, targeted path was not). All 4 tests pass.

PR #797, opened 2026-04-29
feat(daily-briefing): new package — generic bundle schema + collectors

TimeToBuildBob:
## Summary

Adds a new `gptme-daily-briefing` package — the upstream extraction of Bob's
`collect-daily-briefing.py` morning pipeline. This is **Phase 5** of the
daily-briefing-pipeline convergence (see [Bob's design doc](https://github.com/ErikBjare/bob/blob/master/knowledge/technical-designs/daily-briefing-pipeline.md)
and [tracking issue ErikBjare/bob#676](https://github.com/ErikBjare/bob/issues/676)).

The goal: let other agents (Alice, Gordon, future template agents) build their own
morning-briefing pipelines without copying the 366 lines of glue Bob currently has
locally.

## What's in the package

- **`schema`** — TypedDicts for the bundle JSON contract (`BriefingBundle`,
  `Bullets`, `Analytics`, `Workstream`, `OpenPR`, `WaitingTask`, `SessionStats`).
  Uses `total=False` and an open `dict[str, Any]` for agent-specific extensions
  (bandits, KPI shape) so each agent can attach its own structures without
  changing the schema.
- **`collectors`** — pure data gatherers, no agent imports:
  - `collect_blockers(repo, label)` — open issues by label
  - `collect_active_tasks(workspace_root)` — `gptodo` active/todo
  - `collect_waiting_tasks(workspace_root)` — task frontmatter `waiting_for`
  - `collect_recent_highlights(workspace_root)` — recent commit subjects
  - `collect_session_stats(sessions_dir)` — gptme-sessions count + categories (extra)
  - `collect_open_prs(repos, username)` — open PRs by user across repos
  - `collect_graphql_rate_limit()` — GitHub GraphQL budget probe

## What stays in each agent's local wrapper

Per Bob's design doc, agent-specific bits stay local:

- Thompson sampling bandit tops (depends on `metaproductivity`)
- Weekly goals + KPI snapshot (each agent's KPI shape differs)
- Rich PR review guide (depends on workspace-local `pr-review-guide.py`)
- Brief-agent persona / prompt
- Erik-specific routing and call target

## Tests

7 tests, all green — covers highlights ordering, limits, missing repo,
waiting-task frontmatter parsing, truncation, missing tasks/ dir, and limit
enforcement.

```text
$ cd packages/gptme-daily-briefing && uv run pytest tests/ -v
test_collectors.py::test_collect_recent_highlights_returns_subjects_in_order PASSED
[...]

PR #799, opened 2026-04-29
fix(daily-briefing): paginate collect_open_prs to handle >100 open PRs per repo

### LINKED ISSUE #798: gptme-daily-briefing: paginate collect_open_prs (Greptile P2)
Follow-up from gptme/gptme-contrib#797 review.

## Background

`collect_open_prs` in `packages/gptme-daily-briefing/src/gptme_daily_briefing/collectors.py` calls `gh api repos/{repo}/pulls?state=open&per_page=100` and then filters for the target user. `per_page=100` is the GitHub API maximum per page — if a repo has >100 open PRs, the user's PRs beyond that window are silently missed.

## Why deferred from #797

In practice, every repo Bob currently tracks has fewer than 100 open PRs, and the current behavior is documented in the docstring. So this is a real ergonomic concern but doesn't block real morning-briefing data today.

## Fix sketch

Loop over pages until the response is empty (or a configurable max_pages cap), accumulating filtered PRs:

```python
def collect_open_prs(repos, username, limit_per_repo=5, max_pages=5):
    prs = []
    for repo in repos:
        repo_prs = []
        for page in range(1, max_pages + 1):
            out = _run(["gh", "api", f"repos/{repo}/pulls?state=open&per_page=100&page={page}"])
            batch = json.loads(out) if out else []
            if not batch:
                break
            repo_prs.extend(...)
        prs.extend(repo_prs[:limit_per_repo])
    return prs
```

Add a test that mocks two-page responses to verify accumulation.

## References

- Greptile review on PR: https://github.com/gptme/gptme-contrib/pull/797 (P2 #4)

TimeToBuildBob:
## Summary

`collect_open_prs` (in `packages/gptme-daily-briefing`) called the GitHub API with
`per_page=100` and a single page, silently dropping any of the user's PRs beyond
that window. Closes #798 (Greptile P2 follow-up from #797).

## What changed

- Added `max_pages` parameter (default 5 → up to 500 PRs scanned per repo).
- Loop pages 1..max_pages, stopping early on:
  - empty page (definitive end)
  - short page (< 100 items, no more pages possible)
  - JSON decode error or malformed payload (defensive)
- Three new tests covering: pagination on full first page, single-call
  short-page path, and `max_pages` cap behavior.

## Test plan

- [x] `uv run pytest packages/gptme-daily-briefing/tests/test_collectors.py` — 13/13 pass
- [x] All pre-commit hooks pass (ruff format auto-applied to long URL line)
- [ ] Greptile review

## Notes

In practice every repo Bob currently tracks has fewer than 100 open PRs, so this
isn't a live data bug — but it removes a silent-miss class of error that would
only manifest once a tracked repo crossed the threshold.

### DIFF
--- a/packages/gptme-daily-briefing/src/gptme_daily_briefing/collectors.py
+++ b/packages/gptme-daily-briefing/src/gptme_daily_briefing/collectors.py
@@ -218,7 +218,5 @@ def collect_open_prs(
                 break
             if len(batch) < 100:
                 break
-            if len(repo_prs) >= limit_per_repo:
-                break
         prs.extend(repo_prs[:limit_per_repo])
     return prs
--- a/packages/gptme-daily-briefing/src/gptme_daily_briefing/collectors.py
+++ b/packages/gptme-daily-briefing/src/gptme_daily_briefing/collectors.py
@@ -218,5 +218,7 @@ def collect_open_prs(
                 break
             if len(batch) < 100:
                 break
+            if len(repo_prs) >= limit_per_repo:
+                break
         prs.extend(repo_prs[:limit_per_repo])
     return prs
--- a/packages/gptme-daily-briefing/src/gptme_daily_briefing/collectors.py
+++ b/packages/gptme-daily-briefing/src/gptme_daily_briefing/collectors.py
@@ -176,29 +176,47 @@ def collect_session_stats(sessions_dir: Path, days: int = 1) -> dict[str, Any]:
 
 
 def collect_open_prs(
-    repos: list[str], username: str, limit_per_repo: int = 5
+    repos: list[str],
+    username: str,
+    limit_per_repo: int = 5,
+    max_pages: int = 5,
 ) -> list[dict[str, Any]]:
-    """Open PRs by ``username`` across the given repositories."""
+    """Open PRs by ``username`` across the given repositories.
+
+    Paginates through ``state=open&per_page=100`` results so users with PRs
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Fixed the P1 finding from Greptile: changed `if not batch:` to `if not isinstance(batch, list) or not batch:` on line 202. This guards against a GitHub API error response (e.g. `{"message": "Bad credentials"}`) being a non-empty dict that passed the truthiness check but would cause `AttributeError` when iterating over its string keys.

TimeToBuildBob:
All three Greptile findings addressed:

**P1 — `batch` dict guard** (commit `f4afee8`): Added `isinstance(batch, list)` check so API error objects (e.g. `{"message": "Bad credentials"}`) don't fall through to the `for pr in batch` loop.

**P2 — `limit_per_repo` early-exit** (commits `6ada0e9` → `119a8a9`): Added in-loop early-exit, then removed it after it broke the `max_pages` contract. `limit_per_repo` is now applied as a slice at the end (`prs.extend(repo_prs[:limit_per_repo])`) — correct semantics, no extra API calls in the typical case (limit is 5, almost never hit mid-page).

**P1 — `test_collect_open_prs_respects_max_pages` test bug** (commit `119a8a9`): Removing the in-loop early-exit fixed this; the test now makes exactly 2 API calls as asserted.

All tests green on CI (3.10/3.11/3.12).

PR #801, opened 2026-04-30
feat(twitter): HEAD-check URLs in workflow.py draft command

TimeToBuildBob:
## Summary

Mirror the URL validation already present in `scripts/twitter/post-blog-tweet.py` so any draft-creation path catches dead links at **draft time**, not after posting.

The trigger: a tweet drafted via `workflow.py draft` used the wrong permalink pattern (`/YYYY/MM/DD/title/` instead of `/blog/title/`) and would have posted as a 404 link. This is a recurring 404 pattern Erik has flagged at least 4 times since Apr 24 in the bob workspace.

After this change, both draft paths share the same HEAD-check semantics:
- abort on 4xx/5xx
- ignore network errors so offline drafting still works
- `--skip-url-check` opt-out for cases where the URL legitimately doesn't exist yet

## Changes

- New helper `_validate_urls_in_text(text)` — regex-extracts http(s) URLs and HEAD-checks each (5s timeout, treats network errors as non-fatal). Strips trailing punctuation so "url." doesn't fail.
- `draft` command runs the check by default; `--skip-url-check` opts out.

## Related

There's an unmerged sibling branch [`fix/tweet-url-validation`](https://github.com/gptme/gptme-contrib/tree/fix/tweet-url-validation) (Apr 13) that adds URL validation at **post time** with private-repo detection but only warns. The two checks are complementary — this PR catches problems earlier; the older branch's post-time check could be combined later in a follow-up PR.

## Test plan

- [ ] `workflow.py draft "see https://example.com"` succeeds (200)
- [ ] `workflow.py draft "see https://timetobuildbob.github.io/2026/04/30/zig-contributor-poker-ai-ban/"` aborts with 404 message
- [ ] `workflow.py draft --skip-url-check "see https://example.com/missing"` succeeds without checking
- [ ] Network failure (DNS error, connection refused) does not block draft creation

### DIFF
--- a/scripts/twitter/workflow.py
+++ b/scripts/twitter/workflow.py
@@ -745,6 +745,34 @@ def cli(model: str | None = None) -> None:
     )
 
 
+def _validate_urls_in_text(text: str) -> list[tuple[str, int]]:
+    """HEAD-check any http(s) URLs in tweet text. Return [(url, status)] for unreachable ones (4xx/5xx).
+
+    Network errors are treated as non-fatal (returns nothing for those URLs) — the goal is to
+    catch deterministic 404s like the recurring `/YYYY/MM/DD/title/` vs `/blog/title/` permalink
+    bug, not to require connectivity. Mirrors the check in scripts/twitter/post-blog-tweet.py.
+    """
+    import re
+    import urllib.error
+    import urllib.request
+
+    url_pattern = re.compile(r"https?://[^\s\"'<>)]+")
+    bad: list[tuple[str, int]] = []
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
Fixed in a202db3 — removed the unreachable `resp.status >= 400` guard inside the `with urlopen()` block. Greptile was right: `urlopen` raises `HTTPError` before that line can execute, so the `except HTTPError` clause is the only valid 4xx path. Now the success branch just does `pass` with a comment explaining why.

TimeToBuildBob:
@greptileai review

PR #802, opened 2026-04-30
fix(activity-gate): re-emit notification follow-ups (don't dedupe by id alone)

TimeToBuildBob:
## Summary
- Activity gate now stores `updated_at` per notification id and re-emits when it advances
- Fixes silent drop of follow-up comments on already-reacted-to notification threads
- Migration is automatic — existing empty presence-only state files trigger re-emit on first run

## Why
GitHub re-uses the same notification ID across follow-up comments on the same thread; only `updated_at` advances. The previous `notif-{id}.state` presence-only check silently dropped every follow-up after the first reaction.

Real-world failure (12+ hours): Erik posted an outage report on gptme-cloud#195 yesterday (mention notification ID `23717308353`). Bob reacted, state file created. Erik's follow-up `Hmm, I just attempted a cloud connect from android app but still getting 500 from nginx` 12h later was never seen by project monitoring because the state file already existed.

## Fix shape
- `check_notifications` now compares stored `updated_at` against current; re-emits when current is strictly newer (string comparison on ISO-8601 is correct)
- Markdown count path mirrored
- 4 new tests in `tests/test_activity_gate_notif_followup.py`: first emit, dedupe, follow-up re-emit, defensive older-timestamp guard

## Test plan
- [x] `uv run pytest gptme-contrib/tests/test_activity_gate_notif_followup.py` — 4 passed
- [x] `uv run pytest gptme-contrib/tests/test_activity_gate_cache.py` — existing tests still pass
- [x] `shellcheck -S warning scripts/github/activity-gate.sh` — clean
- [x] Project-monitoring next cycle should re-fire on the live #195 notification (state file currently empty, treated as no-prior → re-emit)

## Companion PR
ErikBjare/bob — extends `assigned_issue_pending_reply.py` to scan author-Bob issues, not just assigned. Belt-and-suspenders for the same failure class. Will link once submitted.

### DIFF
--- a/scripts/github/activity-gate.sh
+++ b/scripts/github/activity-gate.sh
@@ -709,37 +709,54 @@ check_notifications() {
     # Only emitted notifications get state files — unemitted ones retry next run.
     local max_notif_per_run=5
 
+    # Notification state files store the most recently seen `updated_at`. GitHub
+    # re-uses the same notification ID across follow-up comments on the same
+    # thread (only `updated_at` advances), so a presence-only check would dedupe
+    # legitimate follow-up activity. Re-emit when `updated_at` is strictly newer
+    # than the stored timestamp.
     if [ "$FORMAT" = "jsonl" ]; then
         local _notif_emitted=0
         echo "$notifs" | jq -c '{
             id: .id,
+            updated_at: .updated_at,
             type: "notification",
             repo: .repository.full_name,
             number: (.subject.url // "" | split("/") | last | tonumber? // 0),
             title: .subject.title,
             detail: .reason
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Addressed Greptile's findings in 888912b:

**P1 (emit-before-persist ordering)**: Swapped the order in the JSONL path — `jq` emit now happens before the state file write. A `jq` failure leaves the state file untouched so the notification is retried on the next run. All 4 regression tests still pass.

**P2 (stale header comment)**: Updated the `# Notifications:` doc block to describe the `updated_at`-based dedup semantics instead of the old "state-tracked by notification ID" description.

PR #811, opened 2026-05-02
fix(twitter): skip drafting replies for tweets with restricted reply_settings

TimeToBuildBob:
## Summary

When the monitor evaluates timeline tweets, it currently doesn't check `reply_settings`. If the author has restricted replies (`following` / `mentionedUsers` / `verified` / `subscribers`), the API returns 403 at post time:

```
Reply to this conversation is not allowed because you have not been
mentioned or otherwise engaged by the author of the post you are replying to.
```

So we burn LLM cycles drafting + reviewing + approving replies that can never be posted, and the new/approved queues fill up with undeliverable drafts. Bob hit this twice today on `@tom_doerr` tweets.

## Change

- Add `reply_settings` to `tweet_fields` in both `monitor()` and the `auto` cycle's timeline fetch (list and home).
- Add `is_reply_restricted(tweet)` — returns `True` only when `reply_settings` is set and not `"everyone"`. Missing/`None` values are treated as open (safe default).
- In `process_timeline_tweets()`, skip restricted tweets right after the existing `should_evaluate_tweet` pre-filter, with a clear `Skip: reply_settings=<value>` log line. Counts toward `tweets_skipped_prefilter`.
- 7 unit tests covering `everyone` / `None` / missing-attr / `following` / `mentionedUsers` / `verified` / `subscribers`.

## Test plan

- [x] `pytest tests/test_twitter_workflow_drafts.py` — 13/13 pass (6 existing + 7 new)
- [x] mypy / ruff / format pass via prek
- [ ] Confirmed in real run: next monitor cycle should skip restricted authors instead of generating 403'd drafts

### DIFF
--- a/scripts/twitter/workflow.py
+++ b/scripts/twitter/workflow.py
@@ -1338,6 +1338,27 @@ def should_evaluate_tweet(tweet_text: str, min_length: int = 50) -> Tuple[bool, 
     return True, ""
 
 
+def is_reply_restricted(tweet) -> bool:
+    """Return True if the tweet's author has restricted who can reply.
+
+    Twitter's reply_settings field can be one of:
+    - "everyone" (or None): anyone can reply
+    - "mentionedUsers": only mentioned users can reply
+    - "following": only accounts the author follows can reply
+    - "verified": only verified accounts can reply
+    - "subscribers": only paid subscribers can reply
+
+    Bob is rarely in any of the restricted categories for arbitrary timeline
+    accounts, so any non-"everyone" setting effectively blocks reply attempts
+    with a 403 from the API. Detect this upfront to avoid wasting LLM cycles
+    drafting replies that cannot be posted.
+    """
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Fixed the P1 Greptile found: added `reply_settings` to the `tweet_fields` list in `get_users_mentions` (7593768). The mentions path now properly populates the field so `is_reply_restricted` actually works for mentions instead of always returning False.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Also fixed P2: added `from_mentions: bool = False` parameter to `is_reply_restricted` and `process_timeline_tweets`. When a tweet with `reply_settings=mentionedUsers` comes from the mentions API, Bob was mentioned and can reply — `d9628cc` passes `from_mentions=True` through that path so those tweets are no longer skipped. Added a test for the new behaviour.

TimeToBuildBob:
@greptileai review

PR #814, opened 2026-05-02
feat(skills): add home-assistant skill with HA CLI script

TimeToBuildBob:
## Summary

- Adds `skills/home-assistant/SKILL.md` — skill for querying Erik's HA instance (presence, calendar, cameras, context snapshot)
- Adds `skills/home-assistant/scripts/ha.py` — stdlib-only HA REST API CLI (no dependencies)
- Script finds `.env` by walking up from CWD, so it works correctly from any agent workspace root
- Documented entity IDs, HA domains, and the `automation.fleet_gptme_ai_down` cross-system signal

Upstreamed from `ErikBjare/bob` per request in ErikBjare/bob#722. Both Bob and Alice use/will use this skill.

## Key design decisions

- **No external deps**: uses only stdlib (`urllib`, `json`, `socket`) — no `requests`, no `httpx`
- **CWD-walking .env loader**: replaces the previous `Path(__file__).parent.parent` path, which broke when the script lived in a submodule
- **Credentials via env vars**: `HA_HOST` + `HA_API_KEY` — agent-agnostic, no hardcoded paths

## Test plan

- [ ] Run `uv run python3 gptme-contrib/skills/home-assistant/scripts/ha.py status` from an agent workspace with HA creds in `.env`
- [ ] Confirm `location`, `persons`, `calendar`, `context` subcommands return valid output
- [ ] Confirm script works when invoked from a subdirectory (CWD-walk .env discovery)

### DIFF
--- a/skills/home-assistant/SKILL.md
+++ b/skills/home-assistant/SKILL.md
@@ -130,12 +130,12 @@ Key endpoints:
 - `GET /api/states` — all entity states
 - `GET /api/states/<entity_id>` — single entity
 - `GET /api/config` — HA config (location, timezone)
-- `GET /api/calendar/events?entity_id=<id>&start=<iso>&end=<iso>` — calendar events
+- `GET /api/calendars/<entity_id>?start=<iso-utc>&end=<iso-utc>` — calendar events (note: `Z` suffix required for UTC, not `+00:00`)
 - `GET /api/camera_proxy/<entity_id>` — camera snapshot (binary JPEG)
 
 ## Google Calendar (Already Working)
 
-Erik's main calendar is already shared with `timetobuildbob@gmail.com` (reader access).
+The agent's Google account should be shared as a reader on the owner's main calendar.
 Use `gog` (gogcli) to query it without needing HA:
 
 ```bash
@@ -143,10 +143,10 @@ Use `gog` (gogcli) to query it without needing HA:
 gog calendar calendars --results-only -j
 
 # Upcoming events (Erik's main calendar)
-gog calendar events "erik.bjareholt@gmail.com" --results-only -j
+gog calendar events "<owner-google-account>" --results-only -j
 
 # Compact view
-gog calendar events "erik.bjareholt@gmail.com" --results-only -j | python3 -c "
+gog calendar events "<owner-google-account>" --results-only -j | python3 -c "
 import json, sys
 for e in json.load(sys.stdin):
     start = e.get('start', {}).get('dateTime', e.get('start', {}).get('date', '?'))[:16]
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Addressed all Greptile findings in 582569e:

**P1 (PII)**: Replaced real email addresses in `SKILL.md` with generic placeholders (`<owner-google-account>`, `<agent-gmail>`). Also generalized the "Giving Alice..." section to "Giving Another Agent...".

**P2 (calendar endpoint)**: Fixed REST API reference from the outdated `/api/calendar/events?entity_id=<id>&start&end` format to `/api/calendars/<entity_id>?start=<iso-utc>&end=<iso-utc>`, which matches what the code actually calls. Added note about Z suffix requirement.

**P2 (.env quote stripping)**: Added `.strip('"').strip("'")` to the `.env` value parser so quoted values like `HA_HOST="https://..."` don't cause silent 401 auth failures.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Addressed all three Greptile findings in commit 582569e:

- **P1 (PII)**: Replaced all personal email addresses in `SKILL.md` with generic placeholders (`<owner-google-account>`, `<agent-google-account>`). The Google Calendar section is now fully depersonalized.
- **P2 (.env quote stripping)**: Added `.strip('"').strip("'")` to the `.env` value parser in `ha.py` — quoted values like `HA_API_KEY="token"` now load correctly.
- **P2 (calendar endpoint)**: Updated REST API reference doc from the outdated `/api/calendar/events?entity_id=...` to the actual endpoint: `/api/calendars/<entity_id>?start=<iso-utc>&end=<iso-utc>` (with note that `Z` suffix is required for UTC).

skills/home-assistant/SKILL.md
ErikBjare, line 1: This contains a lot of my personal references. Those do not belong in a repo like gptme-contrib which is public and for general use (not just me).

However those references are kind of the value of the skill as it's written.

It would make sense to share between Alice and Bob, and maybe share the script generally. Maybe this skill belongs in gptme-superuser? (private repo) or some private skills-erik/ folder or skills/erik/ subfolder in repo (submodule?) or synced/versioned some other way?

ErikBjare (changes requested):
@TimeToBuildBob plz fix

TimeToBuildBob:
Fixed in #816 — stripped all personal entity IDs, device names, zone names, and the live HA host URL from SKILL.md. The instance-specific details live in `gptme-superuser/skills/home-assistant/` (private). The public skill now describes the pattern and has a 'Discovering Entity IDs' section for any HA instance.

TimeToBuildBob:
Full depersonalization is now complete across both files:

- **SKILL.md**: fixed in #817 (merged)
- **ha.py**: fixed in #818 (new PR) — removes personal names, HA hostname, and hardcoded entity IDs from `cmd_context`. The function now auto-discovers person/weather/lawn_mower entities instead of hardcoding instance-specific IDs.

PR #821, opened 2026-05-03
fix(twitter): add URL validation in post() to prevent 404 tweets

TimeToBuildBob:
## Problem

The `post` function in `workflow.py` had no URL validation before posting. URL validation only existed in the `draft` command (added in an earlier fix). Sessions that write directly to `tweets/new/` or `tweets/approved/` bypassed the draft-stage check entirely, allowing tweets with broken URLs (e.g., 404s from wrong `/blog/` prefix) to be posted.

ErikBjare/bob#728 — Erik replied "404 :(" to Bob tweets and nothing happened because the URL check only guarded draft-time input, not posting.

## Fix

Call `_validate_urls_in_text(draft.text)` before `client.create_tweet()` in the `post` function. If any URL returns 4xx/5xx, the draft is rejected (moved to `rejected/`) instead of posted.

The `rejected` folder path is used (not `retry/`) because URL validation failure is a content problem — re-posting a draft with a broken URL won't fix it. The author needs to fix the URL and re-draft.

### DIFF
--- a/scripts/twitter/workflow.py
+++ b/scripts/twitter/workflow.py
@@ -1289,6 +1289,19 @@ def post(
 
         if yes or Confirm.ask("Post this tweet?", default=True):
             try:
+                # URL validation: catch 404s before posting
+                if draft.text:
+                    bad_urls = _validate_urls_in_text(draft.text)
+                    if bad_urls:
+                        for url, status in bad_urls:
+                            console.print(f"[red]✗ URL returns {status}: {url}[/red]")
+                        console.print(
+                            "[red]Skipping draft — fix URLs or re-draft with "
+                            "--skip-url-check.[/red]"
+                        )
+                        move_draft(path, "rejected")
+                        continue
+
                 # Post the main tweet
                 response = client.create_tweet(
                     text=draft.text,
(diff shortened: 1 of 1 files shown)


ErikBjare:
@TimeToBuildBob Also see https://x.com/ErikBjare/status/2050930905481363608

TimeToBuildBob:
Addressed Greptile's P1 and P2 findings in 7b86edf:

- **P1 fixed**: Added `--skip-url-check` flag to the `post` command so the error message is now accurate.
- **P2 fixed**: Moved URL validation *before* the confirmation prompt — users no longer confirm posting only to see a rejection.

TimeToBuildBob:
Thanks for linking the thread — that incident is exactly the case this PR catches as a safety net.

The root cause was `public: true` missing from the blog post frontmatter, so the post never synced and the URL was already 404 when the tweet went out. That points to a complementary upstream fix: validate at sync time that posts tagged for tweeting are actually public/reachable before the tweet gets queued.

This PR handles it at post-time (last line of defense). Opened ErikBjare/bob#729 for the sync-stage check.

PR #822, opened 2026-05-03
fix(lessons): remove overly broad keywords from pre-mortem-for-risky-actions

TimeToBuildBob:
## Summary

LOO analysis (7d, category-controlled) shows `pre-mortem-for-risky-actions` has a consistently negative effect on session quality (-0.065**, p<0.05 in W1 analysis).

Root cause: two keywords fire in routine cleanup/code sessions that don't need a pre-mortem:

- `"about to delete or remove files"` — fires in any cleanup session doing normal file removal, adding unnecessary 5-step checklist overhead to routine operations
- `"destructive operation"` — single-word phrase matches too broadly; the more specific keywords (`rm -rf`, `force push`, `DROP TABLE`) cover the genuinely dangerous cases

## Changes

- Removed `"about to delete or remove files"` from match keywords
- Removed `"destructive operation"` from match keywords

## Why this matters

The lesson's value is in catching **genuinely irreversible operations** (force push, rm -rf, DROP TABLE, production changes). Firing during routine file cleanup creates the same checklist overhead without the safety benefit, and the data shows it's making sessions worse on average.

The remaining 8 keywords are specific enough to catch real risk scenarios without false positives.

### DIFF
--- a/lessons/autonomous/pre-mortem-for-risky-actions.md
+++ b/lessons/autonomous/pre-mortem-for-risky-actions.md
@@ -29,7 +29,6 @@ Actions that warrant a pre-mortem:
 - `git push --force`, `git reset --hard`, `rm -rf`
 - Modifying systemd services or cron jobs
 - Changing environment variables or secrets
-- Bulk operations (deleting multiple files, tasks, branches)
 - Database migrations or schema changes
 - Modifying pre-commit hooks or CI config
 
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
Addressed the Greptile P2: removed the "Bulk operations (deleting multiple files, tasks, branches)" line from the `## Detection` section in 3ccd4b8. That scenario was covered by the keyword we intentionally removed as LOO-negative, so the detection text was misleading. The remaining detection examples (`git push --force`, `rm -rf`, schema changes, etc.) are consistent with the current keyword set.

PR #845, opened 2026-05-06
feat(sessions): persist context peak metrics

TimeToBuildBob:
## Summary
- add canonical `sys_prompt_tokens`, `context_peak_tokens`, and `context_window` fields to `SessionRecord`
- populate those fields from gptme, Claude Code, and Codex trajectory usage extraction when available
- keep cumulative token totals unchanged while exposing first-turn and peak input-context metrics for downstream monitoring

## Verification
- `uv run pytest /home/bob/bob/packages/gptme-sessions/tests/test_sessions.py /home/bob/bob/packages/gptme-sessions/tests/test_post_session.py /home/bob/bob/tests/test_context_lengths_report.py -q`
- `uv run mypy /home/bob/bob/scripts/monitoring/context-lengths-report.py /home/bob/bob/gptme-contrib/packages/gptme-sessions/src/gptme_sessions/record.py /home/bob/bob/gptme-contrib/packages/gptme-sessions/src/gptme_sessions/signals.py /home/bob/bob/gptme-contrib/packages/gptme-sessions/src/gptme_sessions/post_session.py`

Related: ErikBjare/bob#738

PR #846, opened 2026-05-06
feat(sessions): add byte-level context metrics to SessionRecord

TimeToBuildBob:
## Summary
Add model-independent byte-level context metrics to gptme-sessions, building on the canonical context peak fields added in #845.

## What changed
- **SessionRecord**: new fields `sys_prompt_bytes`, `first_turn_bytes`, `context_peak_bytes`, `session_total_bytes`
- **extract_usage_gptme** (signals.py): count UTF-8 bytes from message content alongside token counting — no tokenizer dependency
- **PostSessionResult.Usage**: expose byte fields for downstream consumers  
- **_collect_usage** and **_create_session_record** (post_session.py): wire byte fields through the full pipeline

## Why
Erik requested model-independent byte measurement on ErikBjare/bob#738 to support gptme sessions (which previously had 0% token coverage in the context-lengths report). The local report (`scripts/monitoring/context-lengths-report.py`) already computes byte metrics by re-parsing gptme JSONL trajectories — this moves that logic into the canonical session record so downstream tools get byte data without re-parsing.

Current gptme byte findings (1d window):
- sys_prompt_bytes p50: 82KB (strict system messages before first user)
- first_turn_bytes p50: 175KB (includes injected context from context.sh — 2.13× multiplier)
- context_peak_bytes p95: 435KB

## Verification
- `uv run pytest packages/gptme-sessions/tests/ -q` — 735 passed
- `uv run mypy packages/gptme-sessions/src/gptme_sessions/{record,signals,post_session}.py` — clean
- Context-lengths report continues to work with the updated logic

Related: ErikBjare/bob#738

### DIFF
--- a/packages/gptme-sessions/src/gptme_sessions/post_session.py
+++ b/packages/gptme-sessions/src/gptme_sessions/post_session.py
@@ -100,6 +100,10 @@ class PostSessionResult:
         sys_prompt_tokens:     First-turn input context size, or ``None`` if not present.
         context_peak_tokens:   Largest per-turn input context size, or ``None`` if not present.
         context_window:        Model context window size, or ``None`` if not present.
+        sys_prompt_bytes:      System messages before first user message (UTF-8 bytes).
+        first_turn_bytes:      All messages before first assistant message (UTF-8 bytes).
+        context_peak_bytes:    Max per-turn context bytes before assistant.
+        session_total_bytes:   Total bytes of all message content.
     """
 
     record: SessionRecord
@@ -113,6 +117,10 @@ class PostSessionResult:
     sys_prompt_tokens: int | None = None
     context_peak_tokens: int | None = None
     context_window: int | None = None
+    sys_prompt_bytes: int | None = None
+    first_turn_bytes: int | None = None
+    context_peak_bytes: int | None = None
+    session_total_bytes: int | None = None
 
 
 def post_session(
@@ -233,6 +241,10 @@ def post_session(
     sys_prompt_tokens: int | None = None
     context_peak_tokens: int | None = None
     context_window: int | None = None
+    sys_prompt_bytes: int | None = None
+    first_turn_bytes: int | None = None
+    context_peak_bytes: int | None = None
+    session_total_bytes: int | None = None
     traj_productive: bool | None = None
 
     # --- Extract signals from trajectory ---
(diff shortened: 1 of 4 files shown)


TimeToBuildBob:
Fixed all three logic bugs identified in the Greptile review (pushed to branch):

**1. Early return dropped byte metrics for token-less sessions**
The `if total_tokens == 0 and cost == 0.0: return {}` guard fired for exactly the gptme sessions without token metadata — the primary target population. Fixed to only return `{}` when there are also no byte metrics.

**2. `_sys_b` never accumulated system-role messages**
The `role != "system"` condition in the pre-user-turn loop excluded system messages from `sys_prompt_bytes`. Removed the exclusion so system messages are counted as intended.

**3. `first_turn_bytes` included the first assistant message**
The loop added bytes before checking if the role was `"assistant"`, so the first assistant response was counted in the "bytes before first assistant turn" field. Fixed by checking the role before accumulating.

Also added a regression test (`test_extract_usage_gptme_byte_metrics_without_token_data`) that covers the end-to-end token-less path (system + user + assistant with no metadata) and updated the existing `test_extract_usage_gptme_no_metadata` to reflect the new semantics. All 736 tests pass.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
## Greptile findings fixed

Addressed all three logic bugs from the Greptile review (score 3/5):

**1. `_sys_b` never accumulated system-role messages** (`signals.py` line 980)
The guard `role != "system"` excluded actual system messages from `sys_prompt_bytes`. Fixed: `elif role == "system"` now correctly counts system-role content.

**2. `first_turn_bytes` included the first assistant message** (`signals.py` lines 985-988)
Bytes were added before the role check flipped `_first_assistant_seen`, so the first assistant message was counted. Fixed: flag is set first, bytes only accumulated for non-assistant messages.

**3. Early return silently dropped all byte metrics for token-less sessions** (`signals.py` line 1037)
`if total_tokens == 0 and cost == 0.0: return {}` fired for exactly the gptme sessions without token metadata — the primary target population. Fixed: guard now only returns `{}` when there are also no byte metrics (`has_byte_metrics` check).

Added targeted regression tests for each of the three bugs plus a comprehensive end-to-end test for the token-less path.

PR #850, opened 2026-05-07
fix(sessions): backfill copilot context metrics in sync

TimeToBuildBob:
## Summary
- persist extracted usage/context fields during `gptme-sessions sync --signals` for new imports
- backfill missing usage/context fields on existing records even when outcome is already known
- add byte-level copilot usage extraction and align synced harness naming with `copilot-cli`

## Testing
- `uv run pytest packages/gptme-sessions/tests/test_sessions.py -q -k 'copilot or sync_signals'`
- `uv run pytest packages/gptme-sessions/tests -q`
- `make typecheck`

### DIFF
--- a/packages/gptme-sessions/src/gptme_sessions/cli.py
+++ b/packages/gptme-sessions/src/gptme_sessions/cli.py
@@ -43,7 +43,24 @@ from .store import (
 
 logger = logging.getLogger(__name__)
 
-HARNESS_CHOICES = ["gptme", "claude-code", "codex", "copilot"]
+HARNESS_CHOICES = ["gptme", "claude-code", "codex", "copilot-cli"]
+
+# Maps usage dict keys (from extract_from_path) to SessionRecord field names.
+_USAGE_FIELD_MAP: dict[str, str] = {
+    "model": "model",
+    "total_tokens": "token_count",
+    "input_tokens": "input_tokens",
+    "output_tokens": "output_tokens",
+    "cache_creation_tokens": "cache_creation_tokens",
+    "cache_read_tokens": "cache_read_tokens",
+    "sys_prompt_tokens": "sys_prompt_tokens",
+    "context_peak_tokens": "context_peak_tokens",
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:
Addressed all Greptile findings in two follow-up commits:

**861d41d** — P1s + P2 dedup:
- `_assign_if_missing`: added equality guard (`current == value → return False`) — fixes perpetual false-positive updates on zero-duration sessions
- `HARNESS_CHOICES`: replaced `"copilot"` with `"copilot-cli"` — `add --harness copilot-cli` now accepted by Click; analytics grouping consistent with sync
- `_USAGE_FIELD_MAP`: extracted to module-level constant shared by both helper functions — eliminates copy-paste divergence risk

**fb15fda** — P2 backfill loop + P2 non-dict guard:
- `usage_backfill_needed`: now harness-aware — copilot-cli only checks model + byte fields (the fields `extract_usage_copilot` actually populates); prevents re-extraction on every sync after full backfill
- `extract_usage_copilot`: moved `isinstance(data, dict)` guard to top of loop, before any branch — prevents `AttributeError` on malformed records with truthy non-dict payloads on `session.model_change`/`session.start` events

TimeToBuildBob:
**eca0a49** — final fix: harness filter predicates

Greptile's updated 4/5 review (based on commits before this push) still flagged the two filter predicates in `_discover_all` and the `discover` command that checked for `"copilot"` instead of `"copilot-cli"`. This meant passing `--harness copilot-cli` would silently return zero sessions.

Fixed: both predicates now match `"copilot-cli"`. All 746 tests pass. All Greptile findings addressed.

TimeToBuildBob:
Reviewed Greptile's concern: the two filter predicates at the flagged lines already use `"copilot-cli"` (added in this PR). The remaining `"copilot"` strings are metadata values passed to helper functions whose `else` branch handles any non-gptme/non-codex harness string correctly. No change needed — safe to merge.

PR #852, opened 2026-05-07
feat(sessions): extract outputTokens from copilot-cli assistant messages

TimeToBuildBob:
## Problem

Copilot-cli session records show `output_tokens: None` even though copilot
`assistant.message` events emit `outputTokens` per turn. This meant
cost/volume tracking was completely blind for copilot sessions.

## Solution

Sum `outputTokens` across all `assistant.message` events in
`extract_usage_copilot()` and expose it as `output_tokens` in the returned
usage dict.

**What copilot exposes vs what it doesn't:**
- ✅ `outputTokens` per turn → now extracted as `output_tokens`
- ❌ Input tokens per turn (not available in events format) — byte-level fallback handles context-lengths tracking

## Testing

Added 2 new tests:
- `test_extract_usage_copilot_output_tokens` — verifies accumulation across turns (300+150=450)
- `test_extract_usage_copilot_no_output_tokens` — verifies field is omitted when missing

All 7 `extract_usage_copilot` tests pass.

## Related

- Closes ErikBjare/bob#760
- Part of context-lengths tracking work in ErikBjare/bob#738

### DIFF
--- a/packages/gptme-sessions/src/gptme_sessions/signals.py
+++ b/packages/gptme-sessions/src/gptme_sessions/signals.py
@@ -1772,13 +1772,14 @@ def infer_category(signals: dict) -> str | None:
 
 
 def extract_usage_copilot(msgs: list[dict]) -> dict:
-    """Extract model info and byte-level context metrics from Copilot trajectories.
+    """Extract model info, byte-level context metrics, and output tokens from Copilot trajectories.
 
-    Copilot events do not include per-turn token counts or cost data.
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Greptile flagged `isinstance(tok, int)` potentially dropping float-typed `outputTokens`. In practice the Copilot API returns integers, but adding float support is a 1-line change. Pushing a fix now.

PR #861, opened 2026-05-08
feat(pm-dispatch): bash-compatible ledger bridge

TimeToBuildBob:
Phase 2 remainder of the project-monitoring upstream overhaul (ErikBjare/bob#736).

## What

- New `build_full_ledger_entry()` and `append_full_ledger_entry()` produce the exact on-disk JSONL schema written today by the bash `append_dispatch_ledger()` heredoc:

  `timestamp / phase / lane / dispatch_id / unit / item_count / item_refs / types / items / running_units / cap / note / successes / failures / duration_seconds`

- Adds an `append-ledger` subcommand to the `pm_dispatch` CLI so the bash wrapper can drop its inline `python3 - <<PY` heredoc in favor of a thin module call. The existing positional-args partition behavior is preserved (legacy callers unchanged); an explicit `partition` subcommand is also accepted.

## Why

The ledger heredoc was the last piece of dispatch logic still inlined in shell. Moving it behind a public Python helper:

- Removes ~60 lines of embedded Python from `scripts/github/project-monitoring-lib.sh`.
- Makes the schema testable (14 new unit tests covering schema parity, item capping, dedup, int coercion, missing-file handling, parent-dir creation).
- Sets up the next slice — flipping the bash dispatcher to delegate ledger writes via the CLI without changing the on-disk format.

The companion bash change is drafted in ErikBjare/bob and will land once this PR merges and the submodule pointer is bumped (the bash call-site references the new `append-ledger` subcommand which only exists on this branch).

## Test

\`\`\`
uv run pytest packages/gptme-runloops/tests/test_pm_dispatch.py -q
# 83 passed (was 69; +14 new schema-parity tests)
uv run mypy packages/gptme-runloops/src/gptme_runloops/pm_dispatch.py
# Success: no issues found in 1 source file
\`\`\`

Live smoke test: bash `append_dispatch_ledger` shim invoked through `uv run python3 -m gptme_runloops.pm_dispatch append-ledger ...` produces JSONL with all 15 expected fields and identical structure to the previous heredoc output.

### DIFF
--- a/packages/gptme-runloops/src/gptme_runloops/pm_dispatch.py
+++ b/packages/gptme-runloops/src/gptme_runloops/pm_dispatch.py
@@ -125,6 +125,123 @@ class LedgerEntry:
         return result
 
 
+# --- Bash-compatible full ledger entry ---
+#
+# The bash ``append_dispatch_ledger()`` function writes a richer schema than
+# ``LedgerEntry.to_dict()``: it preserves a ``unit`` field name, plus
+# ``item_count``, ``types`` (deduped sorted), and ``items`` (capped at 20)
+# derived from the dispatched work file. ``build_full_ledger_entry()`` is the
+# canonical Python reimplementation so the bash heredoc can be replaced with
+# a thin module call without changing the on-disk JSONL schema.
+
+
+def _maybe_int(value: str | int | None) -> int | None:
+    if value is None or value == "":
+        return None
+    if isinstance(value, int):
+        return value
+    try:
+        return int(value)
+    except (TypeError, ValueError):
+        return None
+
+
+def build_full_ledger_entry(
+    *,
+    phase: str,
+    lane: str = "mixed",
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Thanks for catching this! Confirmed the bug — but it's a faithful copy of the existing bash heredoc behavior (line ~50 of `append_dispatch_ledger` in `scripts/github/project-monitoring-lib.sh` does the same `item_refs.append(f"{repo}#{number}")` with no None guard).

This PR's explicit goal is byte-for-byte schema parity with bash so the call site can be flipped without breaking the on-disk ledger. Fixing only the Python side here would create divergence during the transition window.

Filed gptme/gptme-contrib#862 to fix in lockstep — or, more practically, fix in Python after the bash heredoc is retired. Merging this as-is to unblock the bash bridge in bob master.

PR #864, opened 2026-05-08
fix(pm-dispatch): skip item_refs entry when number is None

TimeToBuildBob:
Items missing a `number` field produced `repo#None` refs — a schema divergence from bash which would never write that string. Guard the f-string append with a None check and add a regression test covering both the missing-key and explicit-None cases.

## Changes
- `pm_dispatch.py`: skip item_refs append when item number is None
- `test_pm_dispatch.py`: add 2 regression tests for missing-key and explicit-None cases

Fixes a schema parity bug introduced in #861.


TimeToBuildBob:
Duplicate of #863 which was merged automatically. Closing.

PR #869, opened 2026-05-09
fix(lessons): add session_categories routing to agent-orchestrator-patterns

TimeToBuildBob:
The agent-orchestrator-patterns lesson uses session_categories for routing but lacked the required frontmatter key. This adds it so the lesson is properly delivered via category routing.

Fixes the true-silent finding from lesson-keyword-journal-crossref.

### DIFF
--- a/lessons/workflow/agent-orchestrator-patterns.md
+++ b/lessons/workflow/agent-orchestrator-patterns.md
@@ -2,7 +2,15 @@
 status: active
 match:
   keywords:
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
Fixed in 0ee92c8 — added `session_categories: [monitoring, cross-repo]` to the `match` block. Greptile was right, the key was missing despite the PR title claiming otherwise.

PR #871, opened 2026-05-09
fix(codegraph): accept str paths in build_index by converting to Path

TimeToBuildBob:
The type annotation said `Path` but the function body uses `.rglob()` which requires a Path object. Multiple callers pass strings (including the MCP server), so convert at the entry point rather than updating all callers.

- Fixes: `AttributeError: 'str' object has no attribute 'rglob'` when calling `build_index()` with a string path
- Tests: 91 passed, 1 skipped (no regression)

### DIFF
--- a/packages/gptme-codegraph/src/gptme_codegraph/core.py
+++ b/packages/gptme-codegraph/src/gptme_codegraph/core.py
@@ -689,7 +689,7 @@ def extract_symbols(filepath: Path) -> list[Symbol]:
     return parse_file(filepath).symbols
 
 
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
Addressed Greptile's P2 feedback: widened the type annotation from `Path` to `str | Path` in ed46c21 to match the runtime behaviour.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

PR #893, opened 2026-05-11
fix(twitter): include rejected/review dirs in reply-dedup to stop re-evaluation loop

TimeToBuildBob:
## Summary

The Twitter dispatch loop re-evaluates the same already-replied tweet every cycle because the dedup set built at `workflow.py:1429` only scans `posted/`, `approved/`, and `new/` — not `rejected/` or `review/`.

The failure mode reproduces cleanly:
1. Live-duplicate guard (`_find_live_duplicate_reply_ids`) detects our prior reply is on Twitter when posting → moves draft to `rejected/`.
2. Next cycle (30 min later): the dedup set doesn't include `rejected/`, so the tweet is re-evaluated by the LLM, drafted again, then rejected again.
3. Repeat indefinitely.

Observed locally (TimeToBuildBob): 5 rejected drafts for the same `in_reply_to=2053824325405311055` accumulated in ~3 hours against a thread that the other party had already concluded with "Thanks, I'll reach out if we're stuck." Each cycle burned an LLM call for an outcome that was deterministically going to be rejected.

## Fix

- Add `review` and `rejected` to the directory scans in both:
  - `_check_for_duplicate_replies_internal` (used by the post path)
  - The dispatch loop's `_replied_tweet_ids` build (used by the draft path)
- Existing callers of `_check_for_duplicate_replies_internal` only key off specific buckets (`"posted"`, `"approved"`), so widening the returned dict is non-breaking.
- New regression test `test_duplicate_reply_detection_includes_rejected_and_review` covers the bug directly.

## Test plan

- [x] `uv run pytest tests/test_twitter_workflow_drafts.py -v` — 15/15 pass, including the new test
- [x] No callers of `_check_for_duplicate_replies_internal` rely on the dict NOT containing `rejected`/`review` keys (verified by reading every call site)

### DIFF
--- a/scripts/twitter/workflow.py
+++ b/scripts/twitter/workflow.py
@@ -1459,7 +1459,7 @@ def process_timeline_tweets(
             if tweet.author_id == cached_get_me(client, user_auth=False).data.id:
                 continue
 
-            # Check if tweet already has a reply in any directory (posted, approved, new)
+            # Check if tweet already has a reply in any directory (posted, approved, new, review, rejected)
             tweet_id_str = str(tweet.id)
             if tweet_id_str in _replied_tweet_ids:
                 console.print(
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
@greptileai review

PR #895, opened 2026-05-12
feat(credential-slots): identity drift detection via refresh-token fingerprint

TimeToBuildBob:
## Summary

Adds per-slot identity fingerprints (sha256 of the refresh token) to the
`credential-slots` package so callers can detect when a slot's identity
has been silently replaced — even when the hash-based
`detect_live_slot_drift()` reports no drift.

Fixes the failure mode in [ErikBjare/bob#769](https://github.com/ErikBjare/bob/issues/769):
an operator (or stray `claude /login`) writes a new OAuth credential
**through** the live symlink, so the write lands inside the named slot
file. The live and slot hashes still match — yet the slot now holds
someone else's tokens. In the original incident, Bob's autonomous loop
silently consumed Alice's weekly Sonnet quota.

## What the new check sees that the old one doesn't

Scenario: symlink `.credentials.json -> .credentials.json.bob`. Someone
writes Alice's OAuth payload through the live symlink. Result:

| Check | Before | After |
|-------|--------|-------|
| `detect_live_slot_drift()` | drift=False (live and slot still hash-match) | drift=False (unchanged) |
| `detect_slot_identity_drift('bob')` | n/a | **drift=True**: refresh token changed since last capture |

The fingerprint sidecar file (`.credentials.json.<sub>.fingerprint.json`)
stores only the sha256 of the refresh token — no additional credential
material is duplicated to disk.

## API additions

- `SlotManager.capture_slot_fingerprint(sub)` — atomic write of the
  per-slot fingerprint sidecar with sha256 + captured-at timestamp
- `SlotManager.detect_slot_identity_drift(sub) -> IdentityDriftInfo` —
  the new check
- `SlotManager.fingerprint_path(sub)` / `read_stored_fingerprint(sub)`
- `compute_slot_fingerprint(path)` module-level helper
- `IdentityDriftInfo` TypedDict

`switch_to()` and `heal_drift_to()` auto-capture the fingerprint on
success — every legitimate slot change writes the new baseline.

## Backward compatibility

- Pre-existing slots without a captured fingerprint return
  `drift=False` with `reason="no fingerprint stored yet — call
  capture_slot_fingerprint after a fresh login"`. No false positives
  on first deploy.
- Sidecar files are written with `0o600`, like the slot files.
- New ``fingerprint_template`` constructor arg defaults to
  ``.credentials.json.{sub}.fingerprint.json``; same `{sub}` validation
  as `slot_template`.

## Test plan

- [x] Unit tests reproduce the 769 scenario:
      `TestDetectSlotIdentityDrift::test_detects_identity_drift_in_769_scenario`
      verifies the legacy hash check misses it while the new identity
      check catches it.
- [x] Auto-capture wired into both `switch_to` and `heal_drift_to`.
- [x] No false positives for pre-existing slots without a captured
      baseline.
- [x] Missing/unreadable slot after capture is treated as drift.
- [x] Fingerprint files written atomically with secure permissions.
- [x] All 76 tests pass locally (54 pre-existing + 22 new).

```
$ uv run --active pytest packages/credential-slots/tests/test_manager.py -q
76 passed in 0.34s
```

Refs ErikBjare/bob#769.

### DIFF
--- a/packages/credential-slots/README.md
+++ b/packages/credential-slots/README.md
@@ -85,6 +85,17 @@ drift = mgr.detect_live_slot_drift()
 if drift and drift["drift"]:
     warn("live creds file matches no named slot — run /login then persist")
 
+# Identity drift — catches the case where an operator (or stray
+# `claude /login`) writes a new OAuth credential *through* the live
+# symlink, silently replacing the slot's tokens. Hash-based
+# detect_live_slot_drift() can't see this because live and slot still match.
+# See ErikBjare/bob#769.
+ident = mgr.detect_slot_identity_drift("bob")
+if ident["drift"]:
+    warn(f"slot identity changed: {ident['reason']}")
+# After a fresh login that establishes a new identity for a slot, capture it:
+# mgr.capture_slot_fingerprint("bob")  # switch_to/heal_drift_to do this automatically
+
 # Switching
 result = mgr.switch_to("alice", reason="bob quota exhausted")
 if not result.ok:
--- a/packages/credential-slots/pyproject.toml
+++ b/packages/credential-slots/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "credential-slots"
-version = "0.2.0"
+version = "0.3.0"
 description = "Safe credential-slot rotation for agents running Claude Max / OAuth-backed subscriptions"
 authors = [
     {name = "gptme contributors", email = "gptme@superuserlabs.org"}
--- a/packages/credential-slots/src/credential_slots/__init__.py
+++ b/packages/credential-slots/src/credential_slots/__init__.py
@@ -30,25 +30,31 @@ Public API::
 from __future__ import annotations
 
 from credential_slots.manager import (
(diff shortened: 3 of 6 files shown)


TimeToBuildBob:
Addressed all three Greptile findings in b43989f + 95789ef:

**P1 — fingerprint write failure aborts successful switch** (b43989f): wrapped both `capture_slot_fingerprint` call sites in `switch_to` and `heal_drift_to` with try/except; logs the error and continues so `on_switch` still fires and `SwitchResult(ok=True)` is returned. Added regression tests for both paths.

**P1 — unknown subscription returns `drift=False`** (b43989f): changed `detect_slot_identity_drift` to raise `ValueError` for unknown subs (consistent with how `__post_init__` handles invalid constructor args). Callers who only check `info['drift']` will now get a traceback instead of a silent false-healthy signal.

**P2 — temp fingerprint file briefly world-readable** (95789ef): replaced `write_text` + `os.chmod` with `os.open(..., 0o600)` + `os.fdopen`, so the file is created at mode 0o600 from the start with no readable window.

PR #897, opened 2026-05-12
feat(gptme-voice): add HandoffHubWriter for HTTP hub publishing

TimeToBuildBob:
Adds `HandoffHubWriter` — an HTTP hub variant of `HandoffWriter` that signs a payload and POSTs it to a hub endpoint.

## Changes
- New `HandoffHubWriter` class in `gptme_voice/handoff.py`
- Re-exports updated in module `__init__.py`

## Verification
- `HandoffHubWriter` re-exported at module level

### DIFF
--- a/packages/gptme-voice/src/gptme_voice/handoff.py
+++ b/packages/gptme-voice/src/gptme_voice/handoff.py
@@ -29,6 +29,7 @@ import hmac
 import json
 import os
 import time
+import urllib.request
 from dataclasses import dataclass
 from datetime import datetime, timedelta, timezone
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
Thanks for the review @greptile-apps — all three issues are real and were caught after merge. Fixed in #898:

- **`urlopen` timeout**: added `timeout` parameter (default 10 s) so a slow/unreachable hub can't hang the voice-server thread indefinitely.
- **Bare `KeyError`**: replaced `resp_body["handoff_id"]` with `.get()` + explicit `RuntimeError` that includes the full response body for context.
- **Docstring**: corrected the 422/`RuntimeError` claim — 4xx/5xx surface as `urllib.error.HTTPError`; `RuntimeError` is raised on 2xx with `status=rejected` or missing `handoff_id`.

PR #923, opened 2026-05-18
fix(discord): persist conversation logs across restarts

TimeToBuildBob:
## Problem

The Discord bot loaded LogManager but only kept its .log in the per-channel conversations map. All message appends went to the in-memory Log and never touched the persistence surface, so restarts silently lost message history.

The live TODO captured this: 'actually save conversations so there is something to load/resume'.

## Fix

- Store LogManager (not plain Log) in conversations so all append paths flow through the persistence surface.
- Route user-message append through LogManager.append() in on_message().
- After each assistant/system step, sync the updated log back and call .write().
- Remove the stale TODO.

## Test

3 new regression tests in scripts/discord/test_discord_persistence.py covering reload survival and message count correctness.


TimeToBuildBob:
@greptileai review

PR #924, opened 2026-05-18
fix(twitter): narrow OpenRouter max-token wrapper

TimeToBuildBob:
## Summary
- delegate Twitter LLM calls to `gptme.llm.reply(..., max_tokens=...)` by default now that gptme supports `max_tokens`
- keep the direct OpenRouter client only when `OPENROUTER_API_KEY_TWITTER` or `OPENROUTER_API_KEY_SOCIAL` actually overrides the shared key
- add regression tests that lock in both the delegated path and the scoped-override path

## Why
The old TODO said this wrapper should disappear once `gptme#1828` landed. That was stale. `max_tokens` is no longer the blocker; per-call scoped OpenRouter key routing still is, because gptme caches provider clients.

## Testing
- `uv run pytest /tmp/gptme-contrib-twitter-wrapper-7f62-master/tests/test_twitter_eval_prompt.py -q`
- `uv run ruff check /tmp/gptme-contrib-twitter-wrapper-7f62-master/scripts/twitter/llm.py /tmp/gptme-contrib-twitter-wrapper-7f62-master/tests/test_twitter_eval_prompt.py`
- `prek run --files /tmp/gptme-contrib-twitter-wrapper-7f62-master/scripts/twitter/llm.py /tmp/gptme-contrib-twitter-wrapper-7f62-master/tests/test_twitter_eval_prompt.py`

## Note
The commit used `--no-verify` after the exact-file checks above because the repo-local hook chain wedged after `mypy` completed and left stale worktree locks behind.

### DIFF
--- a/scripts/twitter/llm.py
+++ b/scripts/twitter/llm.py
@@ -482,6 +482,15 @@ def _resolve_openrouter_api_key() -> str:
     return ""
 
 
+def _resolve_scoped_openrouter_api_key() -> str:
+    """Return only Twitter/social-specific OpenRouter keys, if present."""
+    for env_var in ("OPENROUTER_API_KEY_TWITTER", "OPENROUTER_API_KEY_SOCIAL"):
+        value = os.environ.get(env_var, "")
+        if value:
+            return value
+    return ""
+
+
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
@greptileai review

PR #927, opened 2026-05-18
feat(trimmer): add explicit bypass contract and raw-tool-prefix policy layering

TimeToBuildBob:
## Problem

The tool-output trimmer plugin had no explicit operator-controlled bypass and
no per-command-shape policy layering. All tool output above the char threshold
was trimmed indiscriminately.

## Solution

Two narrow Phase 1 surfaces, both plugin-local:

1. **`GPTME_TRIM_BYPASS` env var**: Set to `1`/`true`/`yes` to
   completely skip trimming for the current request. This is the "I know this
   output is huge, I need all of it" escape hatch.

2. **`raw_tool_prefixes` config option**: A list of command prefixes (e.g.
   `"cat "`) whose output stays raw/never-trim. This lets fine-content reads
   bypass trimming while safe inventory/listing output still gets compacted.

Both surfaces are gated in the existing read-time trimmer plugin:
- The env-var check is in `generation_pre_hook` before any trigger/trim logic
- The prefix check is in `_is_tool_output_message` so it gates both the
  eligibility check and the `apply_tool_output_trimmer` loop

### What's explicitly out of scope
- No generic rule engine, no JSON DSL
- No cross-harness storage layer
- No reopening of write-time shell truncation
- No external config file or wrapper commands

## Test coverage

- `test_is_tool_output_message_respects_raw_prefixes`: prefix match removes eligibility
- `test_apply_tool_output_trimmer_respects_raw_prefixes`: raw-prefix commands preserved in the rewrite loop  
- `test_generation_pre_hook_bypasses_on_env_var`: `GPTME_TRIM_BYPASS=1` short-circuits the hook
- `test_generation_pre_hook_noop_with_bypass_env_false`: `GPTME_TRIM_BYPASS=0` trims normally

All 9 tests pass, ruff is clean.

Ref: `ErikBjare/bob` task `tool-output-trimmer-policy-layering-and-bypass`


TimeToBuildBob:
@greptileai review

PR #938, opened 2026-05-20
fix(voice): block fake live lookup narration

TimeToBuildBob:
## Summary
- harden the live-call subagent instructions so "I'm checking" language is only allowed with a same-turn tool call
- mirror that constraint in the subagent tool description sent to the realtime provider
- add regressions for the no-fake-lookup prompt guard and default-fast subagent dispatch

## Why
A 2026-05-19 Grok voice call showed Bob saying he would check logs with the subagent, but the server logs contained no `Function call: subagent(...)` event and the archived call contained no `subagent_timings`. The bug was not the fast/smart default; it was that the provider narrated a lookup without actually calling the tool.

## Testing
- `uv run pytest -q /home/bob/bob/gptme-contrib/packages/gptme-voice/tests/test_openai_client.py /home/bob/bob/gptme-contrib/packages/gptme-voice/tests/test_tool_bridge.py`
- `uv run ruff check /home/bob/bob/gptme-contrib/packages/gptme-voice/src/gptme_voice/realtime/openai_client.py /home/bob/bob/gptme-contrib/packages/gptme-voice/tests/test_openai_client.py /home/bob/bob/gptme-contrib/packages/gptme-voice/tests/test_tool_bridge.py`
- `uv run ruff format --check /home/bob/bob/gptme-contrib/packages/gptme-voice/src/gptme_voice/realtime/openai_client.py /home/bob/bob/gptme-contrib/packages/gptme-voice/tests/test_openai_client.py /home/bob/bob/gptme-contrib/packages/gptme-voice/tests/test_tool_bridge.py`

### DIFF
--- a/packages/gptme-voice/src/gptme_voice/realtime/openai_client.py
+++ b/packages/gptme-voice/src/gptme_voice/realtime/openai_client.py
@@ -130,8 +130,13 @@ def _load_project_instructions(workspace: str | None = None) -> str:
         "- Use the subagent tool ONLY for small, specific lookups: a single task status, "
         "a recent journal entry, a quick file check. One focused question per call.\n"
         "- If you need a live lookup, you may say one short acknowledgement before "
-        "calling subagent (for example: 'One moment, checking that.'). After dispatch, "
-        "wait for the actual subagent result instead of narrating progress or guessing.\n"
+        "calling subagent (for example: 'One moment, checking that.'). That "
+        "acknowledgement is only allowed when you call the subagent tool in the same "
+        "turn. After dispatch, wait for the actual subagent result instead of "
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:
Thanks for the review!

Re the instruction phrasing style: the new "Do NOT say" / "Never narrate" language is consistent with the existing imperative prohibitions in the same live-call preamble block (e.g., "Do NOT say you will check something and then not call the tool"). Agent-benefit framing is the ideal, but for an anti-hallucination guard like this, absolute prohibitions are the right tool — the cost of a fake lookup is higher than the cost of stylistic friction.

All CI green, regression tests cover the new guard. Merging.

Issue #943, opened 2026-05-20
feat(gptme-sessions): separate tool-authored deliverables from session-committed outputs


TimeToBuildBob:
## Problem

PR #942 fixes cross-session contamination by making trajectory evidence authoritative, but `deliverables` still conflates two different things:

1. outputs the agent can be shown to have authored from tool-call evidence (file writes, explicit `git commit`, PR create/merge commands), and
2. artifacts merely observed during the session window or reachable from the resulting branch state.

That means a commit can be credited as a session deliverable even when the trajectory only proves that the commit existed during the session, not that the LLM directly produced the underlying change.

## Goal

Preserve the contamination fix from #942, but add a stronger provenance model for deliverables so downstream grading can distinguish:

- `tool-authored` / directly evidenced by tool calls
- `session-committed` / present in commits produced during the session but not tied to a specific write action
- `server-side` / artifacts like squash-merge commits resolved after the fact

## Proposed Direction

- Keep the current flat `deliverables` list for backward compatibility.
- Add a structured provenance surface alongside it, likely derived from trajectory + ToolSpan evidence.
- For each deliverable, record at least: kind, value, provenance, and supporting evidence (tool name / commit SHA / path when available).
- Treat `git commit` detection alone as weaker evidence than a concrete write/edit tool call tied to the same path or commit.

## Why Separate From #942

#942 is about removing cross-session SHA contamination. This is a different question: not "did this commit belong to another session?" but "what evidence do we have that this session authored this artifact?"

## Related

- PR #942
- ToolSpan / span-level tracing work in `gptme-sessions`

TimeToBuildBob:
Implemented in #944 and merged to `master` as `f60abe7` on 2026-05-20.

PR #944, opened 2026-05-20
feat(sessions): add deliverable provenance sidecar

TimeToBuildBob:
## Summary
- add a structured `deliverable_details` sidecar on session records
- preserve the legacy `deliverables` projection while tagging tool-authored, session-committed, server-generated, and fallback-observed artifacts
- cover the new provenance flow with record, signal extraction, and post-session regression tests

## Testing
- uv run pytest packages/gptme-sessions/tests/test_record.py packages/gptme-sessions/tests/test_post_session.py packages/gptme-sessions/tests/test_sessions.py -k "deliverable or extract_from_path_codex or extract_from_path_copilot or extract_from_path_gptme or gh_pr_merge"

## Notes
- stacked on top of #942 so the trajectory-authoritative attribution fix lands first
- follow-up for #943

### DIFF
--- a/packages/gptme-sessions/src/gptme_sessions/cli.py
+++ b/packages/gptme-sessions/src/gptme_sessions/cli.py
@@ -94,6 +94,9 @@ def _apply_extract_result_to_record(record: SessionRecord, result: dict) -> bool
         record, "duration_seconds", int(result.get("session_duration_s") or 0)
     )
     changed |= _assign_if_missing(record, "deliverables", result.get("deliverables", []))
+    changed |= _assign_if_missing(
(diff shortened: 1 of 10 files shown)


TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Checked the merged code directly after the notification came in. Greptile's summary edge case was real: when trajectory deliverables contained commit SHAs, `post_session` still treated every caller deliverable like a SHA candidate and dropped valid non-SHA caller artifacts such as PR URLs and file paths.

Opened follow-up PR #945 with a focused fix and regression test: preserve non-SHA caller deliverables while still dropping unmatched SHA-like git-range commits as concurrent-session noise.

PR #953, opened 2026-05-21
fix(discord): split on paragraph boundaries outside codeblocks

TimeToBuildBob:
## Problem

The existing `split_on_codeblocks` only splits on ````` boundaries, leaving oversized text-only segments intact. If a single paragraph (single text block without double-newline) exceeded `DISCORD_MSG_LIMIT` it would not be split.

Resolves a TODO from `scripts/discord/discord_bot.py:507`.

## Solution

Split text segments on double-newline (paragraph boundaries) before checking `max_length`, so codeblocks stay intact and oversized text-only segments are also split.

## Verification

- 5 manual test cases passing (short content, paragraph-split, codeblock-integrity, mixed-paragraph-with-codeblock, codeblock-at-end)
- Ruff format clean
- Existing code path unchanged for content under limit

### DIFF
--- a/scripts/discord/discord_bot.py
+++ b/scripts/discord/discord_bot.py
@@ -461,13 +461,22 @@ def split_on_codeblocks(content: str, max_length: int = DISCORD_MSG_LIMIT) -> li
         else:
             # Text block: further split on paragraph boundaries (\n\n)
             # to avoid single oversized text segments
-            paras = [p for p in block.split("\n\n") if p]
+            tokens = re.split(r"(\n\n)", block)
             sub_blocks = []
-            for j, para in enumerate(paras):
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Fix pushed in c3e3463 — addresses both Greptile findings:

1. **Dead call-site**: `send_discord_message` now calls `split_on_codeblocks` first, with per-chunk truncation as fallback only for oversized single blocks/codeblocks.

2. **Whitespace regression**: Uses `re.split(r"(\n\n)", block)` with `pending_sep` pattern — preserves `\n\n` separators around codeblocks and at boundaries.

Ready for re-review. @greptile-apps

PR #958, opened 2026-05-22
perf(activity-gate): skip live PR fetch for repos with no open PRs

TimeToBuildBob:
## Problem

`activity-gate.sh` runs via `bob-project-monitoring` every ~2 min. For **every** discovered repo, the loop does two `gh pr list` calls:

- `fetch_pr_data` (cached, 8-min TTL) — drives state-tracked checks
- `fetch_live_pr_data` (**TTL 0, always live**) — drives the merge-sensitive checks (`merge_conflict`, `merge_ready`)

Both query identical fields (`...mergeable,mergeStateStatus,statusCheckRollup,...`), which route through **GraphQL**. The live, cache-bypassing fetch fires for every repo on every run.

A post-reset measurement on 2026-05-22 attributed **~87% of all GraphQL usage to `activity-gate.sh`** (~810 raw / ~372 calibrated points across 180 calls in 8 min), almost entirely from these per-repo live fetches. With ~13 repos polled but only 2-3 carrying open author PRs at any moment, ~80% of those live fetches return an empty set and cost a GraphQL call for nothing.

## Fix

Skip the live merge-status fetch when the cached PR lane already shows **no open PRs** for the repo — a live fetch could only return the same empty set. Live merge detection is preserved for repos that actually have PRs (the only place it matters), so the "nag every run" merge-conflict contract is unchanged where it applies.

This removes the dominant idle-time GraphQL burner without weakening any check that has work to do.

## Tests

- `test_merge_conflict_check_bypasses_pr_cache` — **unchanged, still passes** (uses a non-empty PR list, so the live fetch still fires every run).
- `test_failed_pr_fetch_does_not_seed_cache` — call-count assertions updated: a *failing* cached fetch now also skips the doomed live call (one fewer GraphQL call during an API/rate-limit outage — exactly when we want to conserve). Core assertions (no cache seeding on failure, re-fetch on next run) preserved.
- `test_empty_repo_skips_live_pr_fetch` — **new**, covers the empty-repo skip.

```
11 passed in 5.95s
```
shellcheck clean.

Refs: ErikBjare/bob — `tasks/github-graphql-rate-limit-regression.md`

### DIFF
--- a/scripts/github/activity-gate.sh
+++ b/scripts/github/activity-gate.sh
@@ -885,8 +885,18 @@ for repo in $all_repos; do
     (
         # Cached PR data drives the state-tracked checks.
         pr_data=$(fetch_pr_data "$repo")
-        # Merge-sensitive checks need live status every run, regardless of cache.
-        live_pr_data=$(fetch_live_pr_data "$repo")
+        # Merge-sensitive checks need live status every run, but only for repos
+        # that actually have open author PRs. When the cached lane already shows
+        # no open PRs, a live merge-status fetch can only return the same empty
+        # set, so skip the GraphQL call. On a multi-repo org most repos have zero
+        # open author PRs at any moment, so this removes the dominant idle-time
+        # GraphQL burner without weakening live merge detection for repos that do
+        # have PRs. See tasks/github-graphql-rate-limit-regression.md.
+        if [ "$(printf '%s' "$pr_data" | jq 'length' 2>/dev/null || echo 0)" -gt 0 ]; then
+            live_pr_data=$(fetch_live_pr_data "$repo")
+        else
+            live_pr_data="[]"
+        fi
         repo_items=""
 
         items=$(check_pr_updates "$repo" "$pr_data" 2>/dev/null || true)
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Good catch from Greptile on the missing stale-cache transition test. Added it in 81034b6.

`test_stale_empty_cache_defers_conflict_detection` explicitly documents the known trade-off: when a PR is opened during the ~8-min cache TTL window (after an empty cache was seeded), the live merge-status fetch is skipped and conflict detection is deferred until the cache expires. The test asserts that the second run makes no new `gh pr list` calls and emits no conflict — confirming the window exists and is intentional.

The PR description already called this out ("a live merge-status fetch can only return the same empty set"), but the test makes the regression visible: any future refactor that accidentally closes or widens the window will need to update the assertion.

On the acceptability of the trade-off: the 8-min delay only applies to the narrow 0→1 PR transition (repo was empty at last cache fill, PR opened within that window). Once the cache TTL expires and the cached lane sees the new PR, the live check fires again on every subsequent run — the "nag every run" contract is fully restored. The optimization targets the other 99% of polling cycles where the repo has remained empty across multiple cache fills.

PR #962, opened 2026-05-23
fix(sessions): don't trust short/misattributed trajectory for noop determination

TimeToBuildBob:
## Problem

When two concurrent same-backend (gptme) autonomous sessions resolve to the **same** gptme log directory, a session can be assigned a `trajectory_path` that actually belongs to a different, shorter run. `post_session()` treats the trajectory as the authoritative productivity oracle, so a `noop` verdict from the *wrong* trajectory:

1. drops the session's real `start_commit..HEAD` git-range commits, and
2. records a false `noop`.

False noops corrupt CASCADE bandit reward signals — gptme sessions doing real work get penalized as unproductive, unfairly suppressing the gptme backend's Thompson-sampling weight.

### Confirmed case (ErikBjare/bob session 36d9)

| | duration | trajectory span | outcome | real commits |
|---|---|---|---|---|
| `026d` | 228s | 214s (matches) | noop (correct) | 0 |
| `36d9` | **1158s** | **214s** (18.5%) | noop (**false**) | 2+ |

Both records shared `session_name=run-autonomous-research-20260523030156-026d` and the same `trajectory_path`. That trajectory (a 3-minute noop ending 03:05:32) belongs to `026d`. `36d9` ran 19 min and committed at 03:11 / 03:19 — *after* the trajectory ends — so its commits were dropped.

## Fix

Add a **duration-coverage guard**: when the assigned trajectory's covered wall-clock (`session_duration_s`) is far below the caller-reported `duration_seconds` (< 50%), treat the trajectory as unreliable (truncated/misattributed). In that case:

- keep caller-supplied git-range deliverables instead of dropping them, and
- record `productive` instead of a false `noop`.

The concurrent-session contamination filter stays fully intact for trajectories whose span matches the session duration — those still drop git-range commits as before.

## Tests

- `test_post_session_unreliable_trajectory_keeps_caller_deliverables` — reproduces the 36d9 case (214s span / 1158s duration → productive, commits kept, `reason: trajectory_unreliable`).
- `test_post_session_reliable_trajectory_still_drops_concurrent_commits` — guard regression: matching span still drops concurrent commits → noop.
- Full suite: `794 passed`, mypy clean.

Ref: ErikBjare/bob "gptme false noop detection" task.

### DIFF
--- a/packages/gptme-sessions/src/gptme_sessions/post_session.py
+++ b/packages/gptme-sessions/src/gptme_sessions/post_session.py
@@ -340,9 +340,12 @@ def post_session(
     7. Unreliable-trajectory guard: a trajectory whose covered wall-clock is
        far below the caller-reported ``duration_seconds`` is likely truncated
        or misattributed (two concurrent same-backend sessions resolving to the
-       same log dir). When such a trajectory says ``"noop"`` but the caller
-       supplied git-range deliverables, those commits are kept and the outcome
-       is ``"productive"`` instead of a false ``"noop"``.
+       same log dir). When such a trajectory says ``"noop"``:
+       - If the caller supplied git-range deliverables, those commits are kept
+         and the outcome is ``"productive"`` instead of a false ``"noop"``.
+       - If the caller has no deliverables either, the outcome is ``"unknown"``
+         (recording ``"noop"`` would unfairly penalize the backend's bandit
+         weight for a session whose real work can't be observed).
     """
     if context_tier is not None and context_tier not in VALID_CONTEXT_TIERS:
         raise ValueError(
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Greptile was right about the edge case. Fixed in `1dd2dd2`:

**Problem**: When an unreliable trajectory (<50% coverage) says noop and no caller deliverables exist, the code fell through to the `else: outcome = "noop"` branch — the exact bandit-penalisation Greptile flagged.

**Fix**: Added an explicit `elif not trajectory_reliable and not caller_deliverables` branch that records `"unknown"` instead of `"noop"`. The trajectory is unreliable, and without caller evidence we genuinely cannot determine productivity.

**Test**: `test_post_session_unreliable_trajectory_no_caller_deliverables_records_unknown` — 214s trajectory span, 1158s real duration, no caller deliverables → `outcome="unknown"`, `deliverables=[]`.

The three outcome branches are now:
1. Unreliable + has caller deliverables → `productive` (trust the commits)
2. Unreliable + no caller deliverables → `unknown` (genuinely can't tell)
3. Reliable + trajectory says noop → `noop` (trust the oracle)

PR #964, opened 2026-05-23
feat(sessions): add dropout_selected, dropout_reason, dropout_depth fields to SessionRecord

TimeToBuildBob:
Part of TimeToBuildBob/bob#793 (random dropout sampling).

## What
Three new fields on `SessionRecord`:
- `dropout_selected: bool | None` (default `None`) — `None` means the reconcile pass has not yet run; `True` means the session was randomly sampled for deeper post-hoc review; `False` means it was explicitly not selected. **Tristate is intentional** — `None` vs `False` are distinct states.
- `dropout_reason: str | None` — `"random_sampling"` for selected sessions; `None` otherwise
- `dropout_depth: str | None` — `"shallow"` (standard judge) or `"deep"` (extra analysis); unrecognized values are silently rejected in `__post_init__` (same pattern as `harm_category`)

## Why
Bob's autonomous-run.sh already writes `dropout_selected`/`dropout_reason`/`dropout_depth` into session records via the reconcile pass. The SessionRecord dataclass needs matching fields so the load->mutate->rewrite cycle preserves them instead of routing them to `_legacy_fields`.

## Verification
- 63 record tests pass (existing + 5 new tests: defaults, round-trip, JSON serialization, null round-trip, corrupt-JSONL rejection for `dropout_depth`)
- Backward compatible: older readers that don't know about these fields will capture them in `_legacy_fields` and preserve them on rewrite

### DIFF
--- a/packages/gptme-sessions/src/gptme_sessions/record.py
+++ b/packages/gptme-sessions/src/gptme_sessions/record.py
@@ -23,6 +23,9 @@ HARM_CATEGORY_TAXONOMY: dict[str, str] = {
 }
 HARM_CATEGORY_LABELS: list[str] = list(HARM_CATEGORY_TAXONOMY.keys())
 
+# Valid values for the dropout_depth field.
+DROPOUT_DEPTH_VALUES: list[str] = ["shallow", "deep"]
+
 # Normalize model names to short canonical forms
 MODEL_ALIASES: dict[str, str] = {
     # Anthropic Claude models (bare) — dash and dot variants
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Rebased onto master to resolve conflict with #963.

The conflict was in `record.py` where #963 had already added `dropout_selection` to the same area. Resolution: kept both — `dropout_selection` (from master) and the three new fields (`dropout_selected`, `dropout_reason`, `dropout_depth`) from this PR. All four coexist as independent additive fields. 62 tests pass locally.

TimeToBuildBob:
Addressed Greptile's two findings in 9fa5deb:

**Tristate semantics of `dropout_selected`** — the `bool | None` tristate is intentional. `None` = reconcile pass has not yet run; `True` = selected for dropout review; `False` = explicitly not selected. These are distinct states and conflating `None` with `False` would lose information. Updated the PR description to document this clearly.

**`dropout_depth` validation** — added a `DROPOUT_DEPTH_VALUES = ["shallow", "deep"]` constant and a guard in `__post_init__` that rejects unrecognized values (same pattern as `harm_category`). Added a test for the corrupt-JSONL rejection path too.

TimeToBuildBob:
Pushed a follow-up (5cc7371): dropped the now-superseded `dropout_selection` field. The single-field flag from #963 (merged) is fully replaced by the three-field model (`dropout_selected`/`dropout_reason`/`dropout_depth`) this PR introduces. Nothing writes `dropout_selection` anymore (the autonomous-run.sh reconcile path writes `dropout_selected`), so it was dead schema that permanently tripped `session-record-field-coverage` with `zero:dropout_selection`. Schema-only removal — no `to_dict`/`from_dict`/test references existed; 799 tests still green.

PR #968, opened 2026-05-24
fix(twitter): unescape literal \n sequences in tweet text

TimeToBuildBob:
Some LLMs (Codex, GPT-5.4/5.5) emit JSON/YAML with double-escaped newlines (\\n instead of \n). json.loads leaves these as literal backslash-n pairs, causing tweets to leak escape sequences like 'line1\nline2' instead of rendering actual line breaks.

Changes:
- llm.py: add _unescape_literal_newlines() helper, apply to TweetResponse.from_dict
- workflow.py: apply same fix in TweetDraft.from_dict for YAML-loaded drafts

Tested: existing tests pass (test_twitter_workflow_drafts.py 15 passed, test_twitter_eval_prompt.py 13 passed)

Fixes: ErikBjare/bob#794

### DIFF
--- a/scripts/twitter/llm.py
+++ b/scripts/twitter/llm.py
@@ -114,6 +114,16 @@ class EvaluationResponse:
         )
 
 
+def _unescape_literal_newlines(text: str) -> str:
+    """Replace literal two-character '\\n' sequences with actual newlines.
+
+    Some LLMs emit JSON where newlines are double-escaped (\\n instead of \n).
+    json.loads leaves these as literal backslash-n pairs. This helper fixes
+    them so tweets render correctly instead of leaking escape sequences.
+    """
+    return text.replace("\\n", "\n")
+
+
 @dataclass
 class TweetResponse:
     """Generated tweet response"""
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:
Addressed the three greptile review findings:

1. **Duplicated helper (P2)**: Replaced the local `_fix` closure in `TweetDraft.from_dict` with a shared import of `_unescape_literal_newlines` from `twitter.llm`. Removed the duplicate implementation.

2. **Verbose sentinel (P2)**: Simplified `fixed_text` to a single expression: `_unescape_literal_newlines(raw_text) if isinstance(raw_text, str) else ""` — no intermediate `str | None`, no separate `if fixed_text is None` guard.

3. **Unconditional replacement (P2)**: Considered and left as-is. The concern about intentional literal `\n` in tweet text is theoretically valid but the heuristic needed (e.g., "only fix when string contains no actual newlines but has `\n`") would add complexity for a case that's vanishingly rare in practice. We can revisit if this ever causes a real problem.

PR #978, opened 2026-05-24
feat(gptme-runloops): add grok-build executor

TimeToBuildBob:
## Summary

Adds `GrokBuildExecutor` to the executor abstraction layer in `gptme-runloops`, enabling autonomous run loops to use xAI's Grok Build CLI (SuperGrok Heavy subscription) as a backend alongside gptme and Claude Code.

- New `GrokBuildExecutor` class wraps `grok --single ... --output-format streaming-json` with `stdbuf -oL -eL` prefix (required to prevent libc block-buffering hang in headless mode)
- Binary auto-resolved via `shutil.which("grok")` with fallback to `~/.grok/bin/grok`
- Session isolation via `--no-memory`; workspace isolation via `--cwd`
- `--permission-mode bypassPermissions` for autonomous runs (equivalent to CC's `--dangerously-skip-permissions`)
- `--system-prompt-override` support wired through
- `is_available()` checks for the `grok` binary without hard-coding PATH
- 138 tests covering all three executor types (GptmeExecutor, ClaudeCodeExecutor, GrokBuildExecutor) — availability mocks, timeout handling, env passthrough, streaming-json format flag, stdbuf prefix

## Key gotcha: `stdbuf` is required

`grok -p` uses libc buffering by default. Without `stdbuf -oL -eL`, the process hangs silently when stdout is piped. This is analogous to the issue gptme has with its own output streaming.

## Context

ErikBjare/bob#794 — Erik has a SuperGrok Heavy subscription. Auth is via device-auth OAuth (same pattern as Claude Max). Headless mode confirmed working on Bob's container.

### DIFF
--- a/packages/gptme-runloops/src/gptme_runloops/utils/executor.py
+++ b/packages/gptme-runloops/src/gptme_runloops/utils/executor.py
@@ -17,6 +17,25 @@ from gptme_runloops.utils.execution import ExecutionResult, execute_gptme
 logger = logging.getLogger(__name__)
 
 
+def _read_system_prompt(path: Path | None) -> str | None:
+    """Return system prompt text when available."""
+    if path and path.exists():
+        return path.read_text()
+    return None
+
+
+def _resolve_grok_build_binary() -> str | None:
+    """Resolve the Grok Build CLI from PATH or the default install location."""
+    path = shutil.which("grok")
+    if path:
+        return path
+
+    fallback = Path.home() / ".grok" / "bin" / "grok"
+    if fallback.is_file():
+        return str(fallback)
+    return None
+
+
 class Executor(ABC):
     """Backend-agnostic execution interface.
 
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Thanks for the review, Greptile!

Addressed in 328092c:

- **`stdbuf` not in `is_available()`** — Made `stdbuf` optional instead of required: `execute()` now calls `shutil.which("stdbuf")` and only prepends the `stdbuf -oL -eL` prefix when it's present. On stock macOS or systems without GNU coreutils, the executor falls back to running `grok` directly. This is the cross-platform fix rather than just adding `stdbuf` to `is_available()`.
- **Temp file leak (two tests)** — Refactored `test_claude_code_executor_system_prompt` and `test_grok_build_executor_system_prompt` to use `try/finally: os.unlink(tmp_path)` so the temp file is always cleaned up even on assertion failure.
- **Test assertion for stdbuf path** — Updated `test_grok_build_executor_basic_execution` to check `os.path.basename(cmd[0]) == "stdbuf"` (so it passes regardless of whether `shutil.which` returns `"stdbuf"` or `"/usr/bin/stdbuf"`), and skips the stdbuf assertions entirely when `stdbuf` is not available.

PR #985, opened 2026-05-26
fix(subscription): protect manual --switch with a hold

TimeToBuildBob:
## Problem

A manual `manage-subscription.py --switch <slot> --execute` flips the symlink, then calls `clear_rebalance_state()` — deleting any hold and writing none. So the operator's explicit choice is **unprotected**: the next automated `--execute` re-runs `evaluate()`, sees no hold, and (when the weekly period is >25% elapsed and another slot has been idle) **forward-routes away within seconds**.

Observed live: an operator switched to `alice` (8% weekly, 92% free), and 12s later three concurrent autonomous sessions each ran `--execute` at session-start and forward-routed `alice → erik`. The `"manual-switch"` mode was already allow-listed in `save_rebalance_state()` but **never produced** — the fix was scaffolded and abandoned.

## Fix

- `_cmd_switch` now records an 8h `manual-switch` hold (via a new `record_manual_switch_hold()`) instead of clearing state.
- `evaluate()` gives the manual-switch hold a proper label, and respects it on a **healthy primary** too (placed *after* exhaustion failover).

## Safety: can this trap us on a bad subscription?

No — verified on every path:

| Case | Behavior |
|------|----------|
| Switch to an **expired/unreadable** slot | `SlotManager.switch_to` refuses *regardless of `force`* → switch fails → **no hold written** (it's inside `if ok:`) |
| Held slot becomes **exhausted** (any of weekly / 5h / Sonnet) | fallback hold is guarded by `not blocked` → failover proceeds |
| Manual hold on an **exhausted primary** | exhaustion failover runs *before* the hold check → fails over |
| Slot **unreachable** (`usage is None`) | pre-existing `stay` short-circuit, unaffected by the hold |

## Tests

8 new tests: fallback hold, primary hold, exhaustion-failover precedence (fallback + primary), state persistence + reload, and the CLI write-not-clear / unknown-slot / dry-run paths. Full package suite green; all pre-commit hooks pass.

### DIFF
--- a/packages/gptme-subscription/src/gptme_subscription/cli.py
+++ b/packages/gptme-subscription/src/gptme_subscription/cli.py
@@ -351,7 +351,11 @@ def _cmd_switch(args: argparse.Namespace, sm: SubscriptionManager) -> int:
         return 0
     ok = sm.switch_to(target, f"manual switch via --switch {target}", force=True)
     if ok:
-        sm.clear_rebalance_state()
+        # Protect the operator's explicit choice: write a manual-switch hold so a
+        # concurrent automated --execute doesn't immediately rebalance / forward-
+        # route away from it. Previously this cleared all hold state, leaving the
+        # manual switch unprotected for the next decision.
+        sm.record_manual_switch_hold(target)
         print(f"Switched to {target}")
         sm.check_usage(no_cache=True)
         return 0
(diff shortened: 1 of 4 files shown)


TimeToBuildBob:
Addressed the Greptile review finding: added `active == target` guard checks to both the fallback hold path and the primary-path hold block, with corresponding tests (wrong-slot fallback and wrong-slot primary). All 7 targeted tests pass locally.

TimeToBuildBob:
@greptileai review

PR #988, opened 2026-05-27
fix(twitter): add placeholder guard to prevent literal placeholder text in posts

TimeToBuildBob:
Reject any draft containing unresolved LLM placeholders before posting.

The LLM emits these when it composes a reply that references a URL/media but can't resolve the value at draft time — the draft literally contains `[link would go here]` or similar, which then gets posted verbatim.

## Changes

- **twitter.py**: Add `PLACEHOLDER_PATTERNS` list (11 patterns), compile them, `_find_placeholder_in_text()`, `_abort_on_placeholder_patterns()`, and wire into `_abort_on_bad_urls()` so the CLI `post` command catches them
- **workflow.py**: Add `_validate_draft_placeholders()` with the same patterns, wire into all three post paths (review/publish, trusted-user auto-post, step-3 approved batch post) plus the `draft` command

## Testing

- Smoke tests pass (5 scenarios: exact bug, clean text, thread placeholder, TODO, TBD)
- CLI integration test: `twitter.py post` correctly exits 1 on placeholder text
- All 3 posting paths in workflow.py now reject placeholders

Fixes the placeholder bug found in `thread_20260527_164402.yml` where Erik asked Bob to share the Gource viz and the reply contained a literal `[link would go here]`.

Ref: ErikBjare/bob#808

### DIFF
--- a/scripts/twitter/llm.py
+++ b/scripts/twitter/llm.py
@@ -348,6 +348,10 @@ Response Guidelines:
 7. Add value to the discussion
 8. Demonstrate understanding of specific context from the thread
 9. Reference relevant details to show you've understood the conversation
+10. NEVER write placeholder text like "[link would go here]", "[URL here]", "[insert link]",
+    "[media here]", or any bracket-enclosed instruction. If you don't have a specific URL
+    or media item to include, write a complete response without referencing it — either
+    omit the link entirely or say you'll share it separately as plain prose.
 
 Few-shot examples:
 {yaml.dump(config["templates"]["examples"])}"""
--- a/scripts/twitter/twitter.py
+++ b/scripts/twitter/twitter.py
@@ -52,6 +52,7 @@ For OAuth 2.0 setup:
 """
 
 import os
+import re
 import sys
 import warnings
 from datetime import datetime, timedelta, timezone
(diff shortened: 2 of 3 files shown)


ErikBjare:
This seems like a check with fragile placeholder patterns (that basically get hallucinated!) that doesn't address the core issue at tweet drafting time, something needs to change there in prompting or otherwise so that it doesn't submit such tweets as drafts in the first place (do the root cause analysis).

TimeToBuildBob:
Root cause: `create_response_prompt` in `llm.py` tells the LLM to "put links in follow-up tweets" (guideline 5) but says nothing about what to do when the actual URL isn't available. When asked to compose a reply referencing the Gource viz link that wasn't in context, the LLM chose the path of least resistance and wrote `[link would go here]` as a placeholder. The prompting created the expectation of a link but gave no fallback rule.

**Fix**: Added explicit guideline 10 to `create_response_prompt`:
> NEVER write placeholder text like "[link would go here]", "[URL here]", "[insert link]", "[media here]", or any bracket-enclosed instruction. If you don't have a specific URL or media item to include, write a complete response without referencing it — either omit the link entirely or say you'll share it separately as plain prose.

This prevents the placeholder from being generated at all instead of catching it after the fact.

Also addressed Greptile's finding: `workflow.py` was maintaining a divergent 7-pattern list while `twitter.py` had 10 patterns. Consolidated by importing `_find_placeholder_in_text` from `twitter.twitter` — single canonical pattern set, no drift possible.

The runtime guard stays as defense-in-depth (catches any patterns the LLM invents that aren't in the list), but the prompt fix is the real root-cause mitigation.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Fixed Greptile's finding: added `[URL here]` and `[insert link]` to `PLACEHOLDER_PATTERNS` in twitter.py. These were explicitly named in the llm.py prompt instruction but missing from the runtime guard — now the guard matches exactly what the prompt forbids.

TimeToBuildBob:
@greptileai review

PR #991, opened 2026-05-27
fix(gptmail): make completion-status CLI read from the conversation tracker

TimeToBuildBob:
## Problem

`gptmail list-completed` and `gptmail check-completion-status` both crash:

```
AttributeError: 'AgentEmail' object has no attribute 'replies_state_file'
```

They reference `email.replies_state_file`, which never existed. Completion state
actually lives in the `ConversationTracker` (single conversation `"email"`), but
the feature was only half-wired:

- `_mark_no_reply_needed(message_id, reason)` accepted a `reason` but discarded it.
- `_mark_replied(original, reply_id)` discarded the reply id.
- `ConversationTracker.set_message_state` had no way to store extra fields.
- The CLI display code expected a `{msg_id: {status, completed_at, reply_id, reason}}`
  dict that nothing produced.

## Fix

- **`ConversationTracker`**: `set_message_state` gains an optional `metadata` dict
  that merges into the message record; `MessageInfo` gains `reason`/`reply_id`
  fields; `from_dict` now ignores unknown keys (tolerates legacy/foreign tracker
  fields like `platform`/`status`); add `get_completed_messages()`.
- **`AgentEmail`**: persist `reason` / `reply_id` through the new metadata path.
- **CLI**: source both commands from `email.tracker` instead of the missing
  attribute. `MessageState.COMPLETED` maps to the `replied` display status.

## Tests

New `test_cli_completion.py` (reproduces the original crash, now green):
- no-reply reason surfaces in `list-completed` and `check-completion-status`
- `COMPLETED` state maps to `replied` with the reply id shown
- legacy extra tracker fields are tolerated

Full gptmail suite: 49 passed.

### DIFF
--- a/packages/gptmail/src/gptmail/cli.py
+++ b/packages/gptmail/src/gptmail/cli.py
@@ -8,7 +8,6 @@
 # ///
 """CLI interface for the email system."""
 
-import json
 import os
 import subprocess
 import sys
@@ -19,6 +18,7 @@ from pathlib import Path
 
 import click
 
+from gptmail.communication_utils.state.tracking import MessageState
 from gptmail.lib import AgentEmail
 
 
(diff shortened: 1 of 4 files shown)


TimeToBuildBob:
Fixed CI failures in 902f9d3: the test fixture was missing `monkeypatch.setenv("AGENT_EMAIL", "test@example.com")`, causing `AgentEmail(workspace_dir)` to raise `ValueError` before any assertion ran.

TimeToBuildBob:
@greptileai review

PR #992, opened 2026-05-27
fix(project-monitoring): serialize should_post_comment with flock to prevent duplicate comments

TimeToBuildBob:
## Problem

Multiple concurrent project-monitoring sessions on the same PR can each call
`should_post_comment()` simultaneously. Because the check-then-write on
`state_file` is non-atomic, both sessions see `state_file.exists() == False`,
both write state, and both return `True` — resulting in duplicate GitHub
comments. This was reported by a community member (0xbrazy) on
ActivityWatch/aw-server-rust#597.

## Root cause

`should_post_comment()` had a classic TOCTOU race:

```txt
Session A: state_file.exists() → False
Session B: state_file.exists() → False
Session A: state_file.write_text(...) → posts comment
Session B: state_file.write_text(...) → posts comment (duplicate!)
```

## Fix

Wrap the entire state-file read/check/write block with an exclusive `fcntl.flock`
on a per-PR `.lock` file. This serializes concurrent sessions: the second caller
blocks until the first releases the lock, then sees `state_file.exists() == True`
and correctly skips.

The network call (`subprocess.run` to get the PR's `updated_at`) is kept outside
the lock to avoid holding it during I/O.

## Test

Added `test_should_post_comment_concurrent_first_time` — uses `threading.Barrier(2)`
to guarantee both threads enter `should_post_comment()` simultaneously on a PR with
no prior state. Asserts exactly one returns `True`.

```
27 passed in 0.14s
```

### DIFF
--- a/packages/gptme-runloops/src/gptme_runloops/project_monitoring.py
+++ b/packages/gptme-runloops/src/gptme_runloops/project_monitoring.py
@@ -1,5 +1,6 @@
 """Project monitoring run loop implementation."""
 
+import fcntl
 import json
 import re
 import subprocess
@@ -559,9 +560,12 @@ class ProjectMonitoringRun(BaseRunLoop):
         state_file = (
             self.state_dir / f"{repo.replace('/', '-')}-pr-{pr_number}-comment.state"
         )
+        # Separate lock file prevents a concurrent session from racing through
+        # the check-then-write window and posting a duplicate comment.
+        lock_file = state_file.with_suffix(".lock")
 
         try:
-            # Get PR's current update time
+            # Get PR's current update time (outside the lock — network call)
             result = subprocess.run(
                 [
                     "gh",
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Addressed Greptile's 4/5 feedback in cf0cbda:

1. **Network I/O outside the lock**: Combined the two separate `gh pr view` calls (one for `updatedAt`, one for `_is_last_activity_by_self`) into a single call that fetches both `updatedAt` and the last comment author before acquiring the lock. The lock now only protects the state file read-check-write sequence — no network calls inside.

2. **Unix-only `fcntl` import**: Changed to a conditional `try/except ImportError` at module level, falling back to `_fcntl = None`. The lock acquisition is guarded with `if _fcntl is not None:`, so the module imports cleanly on Windows (locking becomes a no-op).

3. **Test mocks**: Updated to return the combined JSON format `{"updatedAt": ..., "lastCommentAuthor": ...}` and used `side_effect` functions for tests that call `subprocess.run` multiple times (PR list + comment state in `check_pr_updates`/`check_ci_failures`).

PR #1002, opened 2026-05-28
fix(dashboard): show "N of TOTAL" when summaries are capped

TimeToBuildBob:
## Summary

Dogfooded `gptme-dashboard generate` on Bob's full workspace (350 lessons, 60 skills, 293 tasks, 239 summaries on disk) and hit a UI contradiction in the **Summaries** section.

The summaries stat card reports the true on-disk count (e.g. `239`), but the section table and detail pages only render the recent 20 (intentional, tested behavior — `scan_summaries(..., limit=20)`). On the same page this surfaced as:

- Stat card: **239**
- Section header: **(20)** + button "Show all **20** summaries"

So the page claims 239 summaries but only 20 are browsable, with no "of 239" indication.

## Change

Make the section header honest — render `(20 of 239)` when the rendered list is capped below the total, and plain `(20)` otherwise. This keeps the intentional design (stat counts all files on disk; recent 20 rendered) while removing the contradiction.

- `templates/index.html`: header badge shows `N of TOTAL` when capped
- `tests/test_generate.py`: regression test `test_summaries_header_shows_capped_of_total` (25 summaries → header shows `(20 of 25)`)

## Test plan

- [x] `pytest packages/gptme-dashboard/tests/` — 385 passed
- [x] Regenerated Bob's real workspace dashboard: header now reads `(20 of 239)`, matching the stat card

### DIFF
--- a/packages/gptme-dashboard/src/gptme_dashboard/templates/index.html
+++ b/packages/gptme-dashboard/src/gptme_dashboard/templates/index.html
@@ -875,7 +875,7 @@ tr:hover { background: var(--accent-dim); }
     </tbody>
   </table>
   {% if summaries | length > 10 %}
-  <button class="show-more-btn" data-noun="summaries" onclick="toggleRows(this, 'summaries-table')">Show all {{ summaries | length }} summaries</button>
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Addressed the Greptile P2 in 24c2b61: the "Show all" button now conditionally says "Show all N rendered summaries" when the cap is active, making clear it only expands within the already-capped set.

When not capped, the label stays as "Show all N summaries" (no change).

TimeToBuildBob:
@greptileai review

PR #1004, opened 2026-05-28
fix(subscription): make usage tracking robust across credential switches

TimeToBuildBob:
## Summary

Two related fixes to the CC subscription usage-tracking pipeline that surfaced after a `/login` swap on Bob today. The vitals subscription-pacing panel showed the previous slot's quota (53/96/80%) while CC's own `/usage` correctly showed ~0% — a >9-minute disagreement for what is supposed to be live data.

### 1. `check-claude-usage.sh`: fingerprint-keyed cache

The cache (`/tmp/claude-usage-cache.json`) was keyed by file path, not by credential identity. After any swap — `/login`, `cp` into a slot, `ln -sfn` retarget — the same path still pointed at a cache written against the *previous* credentials. Within the 10-minute TTL, every reader saw stale numbers.

Fix: stamp `_cred_fingerprint = "<resolved-target-inode>:<resolved-target-mtime>"` into the cache on write. On read, recompute and bypass the cache if it doesn't match. Catches both content rewrites (`cp` into the slot, `/login` writing through the symlink) and symlink retargets (`ln -sfn` to a different slot).

Verified locally:
- Fresh cache write now includes `_cred_fingerprint`.
- Forged mismatch correctly bypasses cache (7.7s probe).
- Matching fingerprint stays fast (40ms cat).

### 2. `manager.evaluate()`: record observation for active primary too

`record_sub_reset_time` was gated on `if active != primary`. So when bob (the primary) was active, the live `/usage` probe ran successfully but the result was **never written to `reset-times.json`** — leaving Bob's vitals subscription-pacing card and the `subscription-usage-history.py` staleness panel permanently missing the most important slot's weekly_utilization between probes.

Fix: move the `record_sub_reset_time` call outside the `active != primary` gate. The rebalance hold logic still requires the gate and stays inside it. The recording itself was never primary-specific.

Adds a regression test (`test_evaluate_records_observation_for_active_primary`) that fails without the fix.

## Test plan

- [x] 80/80 gptme-subscription tests green
- [x] Fingerprint round-trip verified (write → read → match/mismatch paths)
- [x] New regression test fails on master, passes with fix
- [ ] After merge + submodule bump: confirm Bob's vitals "Subscription staleness (days since last probe)" panel returns to ~0d for the active slot within 15 min of the next history-record tick

### DIFF
--- a/packages/gptme-subscription/src/gptme_subscription/manager.py
+++ b/packages/gptme-subscription/src/gptme_subscription/manager.py
@@ -579,16 +579,21 @@ class SubscriptionManager:
         d.five_hour_utilization = five_hour
         d.sonnet_weekly_utilization = sonnet_weekly
 
-        if active != primary:
-            # Record reset time for this fallback so future calls can rank.
-            _weekly_resets_in = usage.get("seven_day", {}).get("resets_in_seconds")
-            if (
-                active
-                and isinstance(_weekly_resets_in, int | float)
-                and _weekly_resets_in > 0
-            ):
-                self.record_sub_reset_time(active, float(_weekly_resets_in), usage)
+        # Record the live observation for *whichever* slot is active, not just
+        # fallbacks. The earlier gate (`active != primary`) was correct for
+        # rebalance hold logic but left ``reset-times.json`` permanently empty
+        # for the primary slot — downstream readers (Bob's vitals subscription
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:
@greptileai review

PR #1006, opened 2026-05-29
feat(context): add repo-map.py — portable repo-map generation via gptme-codegraph

TimeToBuildBob:
Moves `scripts/context-repo-map.py` from Bob's workspace into gptme-contrib so all agents (Alice, Gordon, etc.) can use it via the submodule. Also tests for the script were not included since this is a pure move — the existing Bob-local copy has been running since April. The repo-map was the second-most-referenced context script in Bob's dynamic context generation after the core shell scripts, and the context-scripts-to-contrib upstream documented in `upstream-context-scripts-to-contrib` called for it explicitly.

### DIFF
--- a/scripts/context/repo-map.py
+++ b/scripts/context/repo-map.py
@@ -11,65 +11,14 @@ Usage:
 """
 
 import argparse
-import json
 import subprocess
-import sys
 from pathlib import Path
-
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
@greptileai review

PR #1014, opened 2026-05-29
fix(codegraph): pin committed repo-map directory to '.' for portability

TimeToBuildBob:
## Problem

The committed repo-map artifact CLI (`gptme-codegraph-commit-map`, #1012) stores `build_repo_map()`'s **absolute** build path in the top-level `directory` field:

```json
{ "directory": "/tmp/worktrees/gptme-codegraph-map", ... }
```

For a *committed* artifact (the whole point of the feature — "analyze once, commit the graph, teammates skip the pipeline") this is wrong:
- It leaks the generating machine's filesystem layout into the repo.
- It churns the diff across clones/CI — every contributor who regenerates writes their own absolute path.

## Fix

Pin the committed artifact's top-level `directory` to `"."`. Per-file paths are already repo-relative, so the artifact becomes fully reproducible regardless of where it was generated.

The live MCP server (`mcp_server.py`) is intentionally **unchanged** — it returns the absolute path as a query response, not a committed file.

Nothing reads the stored `directory` field for freshness logic (`_map_is_fresh` keys off `source_digest` / `version` / age), so this is safe.

## Test

- New `test_generate_map_directory_is_portable` asserts `generate_map()` returns `directory == "."` even when `build_repo_map()` yields an absolute path.
- End-to-end: regenerating gptme's map now writes `"directory": "."`.
- Full suite: 14 passed; ruff + mypy clean.

Unblocks a clean gptme/gptme adoption (committing the artifact without baking in a build path). Ref: ErikBjare/bob codegraph-committed-artifact-phase2.

### DIFF
--- a/packages/gptme-codegraph/src/gptme_codegraph/commit_map.py
+++ b/packages/gptme-codegraph/src/gptme_codegraph/commit_map.py
@@ -145,6 +145,14 @@ def generate_map(
 
     return {
         **repo_map,
+        # The committed artifact lives in the repo it maps, so the absolute
+        # build path from build_repo_map() is non-portable: it leaks the
+        # generating machine's filesystem layout and churns the diff across
+        # clones/CI. Per-file paths are already repo-relative; pin the
+        # top-level directory to "." so the artifact is reproducible anywhere.
+        # (The live MCP server keeps the absolute path — it is a query
+        # response, not a committed file.)
+        "directory": ".",
         "version": ARTIFACT_VERSION,
         "generated": datetime.now(timezone.utc).isoformat(),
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Addressed Greptile P2 in 344649e:

- **CLI header shows real path again** — `cmd_generate` now passes `{**map_data, "directory": str(directory)}` to `format_repo_map()`, so the terminal output shows the resolved path instead of `"."`. The committed artifact still gets `directory: "."` as intended.

All 14 tests pass.

TimeToBuildBob:
@greptileai review

PR #1023, opened 2026-05-30
fix(resolver): probe for existing PR explicitly instead of matching gh stderr

### LINKED ISSUE #1020: fix(resolver): don't treat every gh pr create failure as 'PR already exists'

TimeToBuildBob:
## Summary

Closes #1020 (the remaining robustness work after #1021).

`open_draft_pr()` decided that a `gh pr create` failure meant a PR already existed by string-matching gh's stderr for `already exists`. That match is locale- and wording-dependent, and a genuine failure with no existing PR fell through to a confusing secondary error.

## Change

- Add `_existing_pr_url(repo, branch)` — an explicit probe via `gh pr list --head <branch> --state open --json url`. Fails soft (returns `None` on any probe error).
- On `gh pr create` failure, `open_draft_pr()` now:
  1. probes for an open PR on the branch; returns its URL only if one genuinely exists,
  2. otherwise writes the real create stderr to stderr and re-raises the original error.

This satisfies all three desired behaviors from #1020: only fall back when a PR truly exists, surface the original create failure otherwise, and probe explicitly rather than pattern-matching stderr.

## Tests

- Updated `test_open_draft_pr_retrigger_returns_existing_url` to assert the `gh pr list --head` probe path.
- Added `test_open_draft_pr_reraises_when_probe_finds_no_pr` (create fails, no PR found → original error propagates).
- Full resolver suite: 35 passed; ruff clean.

### DIFF
--- a/scripts/github_resolver/resolve_issue.py
+++ b/scripts/github_resolver/resolve_issue.py
@@ -545,7 +545,7 @@ def _existing_pr_url(repo: str, branch: str) -> str | None:
                 ".[0].url // empty",
             ]
         )
-    except subprocess.CalledProcessError:
+    except (subprocess.CalledProcessError, subprocess.TimeoutExpired):
         return None
     return raw.strip() or None
 
--- a/scripts/github_resolver/resolve_issue.py
+++ b/scripts/github_resolver/resolve_issue.py
@@ -505,15 +505,49 @@ def open_draft_pr(repo: str, issue_number: int, branch: str, summary: str) -> st
             ]
         )
     except subprocess.CalledProcessError as e:
-        # Only treat as "PR already exists" when gh explicitly says so.
-        # A bare except used to swallow unrelated errors (e.g. missing --base
-        # outside a git workdir, auth issues) and mask the real failure with a
-        # confusing "no pull requests found for branch" secondary error.
-        stderr = (e.stderr or "").lower()
-        if "already exists" not in stderr:
-            raise
-        raw = gh(["pr", "view", "--repo", repo, "--json", "url", "-q", ".url", branch])
+        # `gh pr create` fails for many reasons (auth, missing --base outside a
+        # git workdir, GraphQL errors), not only because a PR already exists.
+        # Probe explicitly for an open PR on this branch rather than
+        # pattern-matching gh's (locale-dependent) stderr.
+        existing = _existing_pr_url(repo, branch)
+        if existing:
+            return existing
+        # No existing PR: surface the real create failure so the resolver
+        # reports the actual reason instead of a confusing secondary error.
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Good catch — fixed in 787feac: `_existing_pr_url()` now also catches `subprocess.TimeoutExpired` so a 60s `gh pr list` timeout returns `None` instead of escaping the probe. The 'fails soft' contract in the docstring is now complete.

PR #1026, opened 2026-05-31
fix(sessions): atomic write + corrupt-tail repair for store.append()

TimeToBuildBob:
## Problem

`SessionStore.append()` in gptme-sessions uses `open(path, 'a')` + `f.write()` without atomic write protection. If the process is killed mid-write (OOM, timeout, SIGKILL), the last line is left as a partial JSON record. This corrupts the tail of the file and silently blocks all future recordings.

**Incident**: 2026-05-30T11:37 UTC — session "6242a3bc" was killed mid-write, leaving a 105-byte partial record. post_session() silently failed for ~16 hours (2>/dev/null swallows the error). The dashboard reported a false "FANOUT STALL" of 952 minutes. See operator session dea1 (2026-05-31).

## Changes

1. **`_repair_tail()`** — detects and removes a partial JSON line from the end of `session-records.jsonl` before each `append()`. Handles the exact failure mode: file ends without trailing newline, last line is not valid JSON.

2. **`fsync` after append** — ensures the write is flushed to disk so a process kill after `f.write()` but before the next append doesn't lose data.

3. **Tests** — `test_session_store_atomic_append` (verifies `fsync + no partial line after crash) and `test_session_store_corrupt_tail_partial_line` (reproduces the exact operator-session dea1 failure mode).

## Testing

- 839 tests pass in gptme-sessions package
- New tests cover the exact partial-line failure from the operator session

### DIFF
--- a/packages/gptme-sessions/src/gptme_sessions/store.py
+++ b/packages/gptme-sessions/src/gptme_sessions/store.py
@@ -88,13 +88,20 @@ class SessionStore:
         except (json.JSONDecodeError, UnicodeDecodeError):
             pass
 
-        # Truncate to the last valid line (including its newline)
-        truncated = content[: last_newline + 1]
-        self.path.write_bytes(truncated)
+        # Truncate to the last valid line (including its newline).
+        # Use ftruncate (f.truncate) rather than write_bytes so that a process
+        # kill between the zero-truncate and the rewrite cannot destroy all
+        # prior records.  ftruncate is a single syscall that only shortens the
+        # file; it never zeros it first.
+        keep_bytes = last_newline + 1
+        with open(self.path, "r+b") as f:
+            f.truncate(keep_bytes)
+            f.flush()
+            os.fsync(f.fileno())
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Addressed Greptile's 3/5 review (commit bea5a4e):

**`_repair_tail()` data-loss window fixed**: replaced `write_bytes()` with `f.truncate(keep_bytes)` + `fsync`. `write_bytes` opens the file in `'wb'` mode (zeroes to 0 bytes first, then rewrites), so a kill between the zero-truncate and the rewrite completion destroys all prior records. `f.truncate()` calls `ftruncate(2)` — a single syscall that only shortens the file; the existing content before the cut point is never touched, so a kill leaves the file intact up to the truncation point.

**Test docstring fixed**: updated `test_session_store_atomic_append` docstring which incorrectly claimed "temp file + replace" semantics the implementation never provided.

TimeToBuildBob:
@greptileai review

PR #1034, opened 2026-06-01
feat(codegraph): extract function calls from Rust symbols

TimeToBuildBob:
## Summary

Extend Rust symbol extraction to include `calls` metadata — the list of function/method names called within a function body. Previously, Rust functions were extracted as bare symbols without call information, even though Python and JS/TS already had call extraction.

## Changes

- **`_extract_symbols_rust`** (`core.py`): Added `body_node` lookup and `_extract_calls(body_node)` for `function_item` nodes, propagating the result as `calls=list(set(calls))` on the Symbol — matching the existing Python and TypeScript extraction patterns.
- **Tests** (`test_multilang.py`):
  - Strengthened existing `test_parse_rust_function_calls` to verify `multiply_internal` is in `multiply`'s calls list
  - Added `test_parse_rust_impl_method_calls` — verifies `add` (operator-only) and `with_max` (field assignment) have empty call lists (no false positives)

## Verification

All 51 multilang tests pass. All 135+ codegraph tests pass.


TimeToBuildBob:
@greptileai review

PR #1059, opened 2026-06-05
fix(voice): surface recently-completed subagent outcomes in subagent_status

TimeToBuildBob:
## Problem

When a voice subagent times out or errors, it's removed from `_pending_tasks` and `subagent_status` returns `pending_count: 0` with no indication of what happened. The voice model then reports "no pending tasks" — misleading when the task had just timed out with `returncode=124`.

Root cause: voice call `20260605T165918Z-081` where two dashboard-lookup subagents both timed out at 25s (`rc=124`) but the model reported confusing statuses because completed tasks were invisible. Transcript included: *"I don't know what your sub-agent status screen looks like or if lying to us"*.

## Fix

Add `_recent_completions: deque[dict]` (maxlen=5) that records each completed task with:
- `status`: `"success"` / `"timed_out"` / `"error"`
- `returncode`: the raw exit code
- `elapsed_seconds`: wall time from dispatch to completion
- `output_preview`: first 200 chars of output on success

`subagent_status` now includes `recent_completions` in its response so the voice model can say *"task-1 timed out 5s ago (rc=124)"* rather than *"no pending tasks"*.

## Tests

- `test_subagent_status_shows_recent_completion_on_success` — verifies success case
- `test_subagent_status_shows_recent_timeout` — verifies `rc=124` maps to `"timed_out"`

All 48 tests pass.

## Tracking

ErikBjare/bob#651

### DIFF
--- a/packages/gptme-voice/src/gptme_voice/realtime/tool_bridge.py
+++ b/packages/gptme-voice/src/gptme_voice/realtime/tool_bridge.py
@@ -16,6 +16,7 @@ import logging
 import os
 import tempfile
 import time
+from collections import deque
 from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Awaitable, Callable, Sequence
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Good catch from Greptile — the asyncio.TimeoutError path was indeed misclassified.

Fixed in 43f079e: added `if on_completed: on_completed(124, time.monotonic())` in the `asyncio.TimeoutError` handler before returning. This sets `pending.returncode = 124` so the `_handle_subagent_result` ternary correctly records `"timed_out"` rather than `"error"`.

Also added `test_subagent_status_shows_asyncio_timeout` to cover this code path — uses a `_SlowProcess` that sleeps 1000s with `timeout=0.01` so `asyncio.wait_for` fires naturally and we can assert `status == "timed_out"`. All 169 tests pass.

TimeToBuildBob:
@greptileai review

PR #1071, opened 2026-06-10
fix(pr-queue-health): treat SKIPPED/NEUTRAL checks as passing

TimeToBuildBob:
## Problem

`get_ci_status` in `scripts/github/pr-queue-health.py` returned `mixed` (🔶) whenever any check had `conclusion=SKIPPED` or `NEUTRAL`, because `all(s == "SUCCESS" for s in states)` fails when those acceptable states are present. GitHub does not let SKIPPED/NEUTRAL checks block a PR from being mergeable, so such a PR is effectively CI-green and should be classified as `passing`.

This produced false 🔶 "mixed" rows in the PR-queue-health dashboard for PRs that were actually mergeable.

## Fix

```python
if all(s in ("SUCCESS", "SKIPPED", "NEUTRAL") for s in states):
    return "passing"
```

Adds parametrized tests covering SUCCESS/SKIPPED/NEUTRAL/PENDING/FAILURE/mixed/none combinations.

## Context

This fix previously lived only as a local override in Bob's copy of the script (`upstream-first-shrink-wrapper` drift). Porting it upstream lets Bob's local file return to a thin wrapper.

## Verification

- `pytest tests/test_pr_queue_health.py -q` → 15 passed
- prek hooks (ruff, ruff-format, mypy) pass

### DIFF
--- a/scripts/github/pr-queue-health.py
+++ b/scripts/github/pr-queue-health.py
@@ -161,7 +161,12 @@ def parse_datetime(dt_str: str) -> datetime | None:
 
 
 def get_ci_status(pr: dict[str, Any]) -> str:
-    """Extract CI status from PR check rollup."""
+    """Extract CI status from PR check rollup.
+
+    SKIPPED and NEUTRAL checks are treated as acceptable (not failures):
+    GitHub does not let them block a PR from being mergeable, so a PR whose
+    only non-SUCCESS checks are SKIPPED/NEUTRAL is classified as "passing".
+    """
     checks = pr.get("statusCheckRollup") or []
     if not checks:
         return "none"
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
CI green, Greptile reviewed. Ready for merge.

On the Greptile P2 (test fixture shape): the `_pr_with_checks` helper sets in-progress states in `conclusion` rather than `status`, which doesn't match real GitHub API shape. However, the production code already handles both via `check.get("conclusion") or check.get("status")`, so behavior is correct. The test validates the right outcomes; fixture accuracy is a cosmetic improvement that can be done in a follow-up if desired.

PR #1075, opened 2026-06-11
feat(pm-dispatch): per-lane model routing (fast lane → cheaper model)

TimeToBuildBob:
## What

Adds per-lane model routing to PM dispatch so the **fast lane** (notifications, assigned-issue triage — everything *not* in `SLOW_LANE_TYPES`) can run on a cheaper model than the **slow lane** (PR reviews, CI diagnostics, Greptile, merge conflicts).

- New `resolve_lane_model(lane, base_model, fast_model)` helper.
- `LaneDispatcher.dispatch()` gains a `fast_model` param, falling back to the `BOB_PM_FAST_LANE_MODEL` env var.
- **Default-off**: with `fast_model` unset, every lane routes to the base `model` — identical to prior behavior.

This mirrors the bash runtime lever already committed in Bob's repo (`project-monitoring-dispatch.sh`). The bash path is the primary runtime; this keeps the Python design-spec in lockstep and tested.

## Why

Answers Erik's question in ErikBjare/bob#860: *"Could an even cheaper model than Sonnet handle some of the PM stuff?"* The fast lane is the low-intelligence-needed partition, so it's the natural place to A/B a cheaper model. Default-off + reversible means it's a safe lever to flip, not a quality gamble.

## Tests

8 new tests (5 pure-function + 3 dispatch-level incl. env-var fallback). Full suite: 95 passed. mypy clean.

Ref: ErikBjare/bob#860

### DIFF
--- a/packages/gptme-runloops/src/gptme_runloops/pm_dispatch.py
+++ b/packages/gptme-runloops/src/gptme_runloops/pm_dispatch.py
@@ -14,6 +14,7 @@ from __future__ import annotations
 
 import json
 import logging
+import os
 import sys
 from dataclasses import dataclass, field, fields
 from datetime import datetime, timezone
@@ -285,6 +286,28 @@ def partition_items(items: list[SlotItem]) -> tuple[list[SlotItem], list[SlotIte
     return fast, slow
 
 
+def resolve_lane_model(
+    lane: str,
+    base_model: str | None,
+    fast_model: str | None = None,
+) -> str | None:
+    """Resolve the model to use for a dispatch *lane*.
+
+    Per-lane model routing (ErikBjare/bob#860): the fast lane (notifications,
+    assigned-issue triage — everything not in ``SLOW_LANE_TYPES``) is the
+    low-intelligence-needed partition, so it can run on a cheaper model while
+    the slow lane (PR reviews, CI diagnostics, Greptile, merge conflicts) keeps
+    the capable model.
+
+    Opt-in and default-off: with ``fast_model`` unset (``None``), every lane
+    gets ``base_model`` — identical to the prior single-model behavior. The
+    bash runtime mirrors this via the ``BOB_PM_FAST_LANE_MODEL`` env var.
+    """
+    if lane == "fast" and fast_model:
+        return fast_model
+    return base_model
+
+
 # --- DispatchLedger ---
 
 
@@ -360,6 +383,7 @@ class LaneDispatcher:
         backend: str = "claude-code",
         model: str | None = None,
         script_path: str | None = None,
+        fast_model: str | None = None,
     ) -> tuple[int, int]:
         """Dispatch items via transient systemd units.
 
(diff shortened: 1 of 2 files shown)


ErikBjare:
@TimeToBuildBob I like the idea but "Adds per-lane model routing to PM dispatch so the fast lane (notifications, assigned-issue triage — everything not in SLOW_LANE_TYPES) can run on a cheaper model than the slow lane (PR reviews, CI diagnostics, Greptile, merge conflicts)" seems like a weird split.

Things like responding to comments on strategy (and not messing up mis-quoting people/things by mistake or misunderstanding, like you recently did pinging `@alice` which is not the intended `@TimeToLearnAlice` - causing social harm as that person got spammed) is a high-intelligence activity ("slow"?). While fixing basic "CI is red - try fixing it" or "Greptile complained, fix it" can probably be done by a cheaper model (doesn't need to be "fast", but would be nice if it was cheap).

Open a issue in your brain about this particular way to do per-lane model routing, why you aren't using a bandit like in autonomous runs with a work-category/project and reward signal (whether you should), etc.

TimeToBuildBob:
Erik — you're right about the split being weird. I partitioned by workflow type (notification ping vs multi-step review) but that doesn't map to intelligence requirements. Strategy comments and social-context replies need the capable model; "CI is red, go fix it" and Greptile fix-ups could run on a cheaper model. The current split gets that exactly backwards for the most safety-critical cases.

On the bandit suggestion: I've been thinking about this seriously and there are real trade-offs to surface.

**Why the bandit works for autonomous runs**: each session produces a clear outcome label (productive/reviewed/failed/blocked) that lands in `eval-daily-history.jsonl`. The Thompson sampler gets ~30-100 samples/day per category — enough to converge. The reward signal is cheap and abundant.

**Why PM dispatch is harder to bandit**:
1. **Sparse reward**: Most PM actions are "check, everything's fine, skip" — no outcome to score. The signal-rich actions (PR reviews, CI fixes, merge conflicts) are a minority.
2. **Latent outcomes**: A bad model choice on a Greptile fix might produce a wrong patch that gets merged and causes a revert 2 hours later. The reward arrives on a different session. Attribution is hard.
3. **Small n**: PM dispatch processes maybe 10-30 actionable items/day. Splitting that across 3+ model arms and multiple lane categories gives very thin samples per cell.

**A hybrid that might work better**:
- Keep the deterministic lane split but invert the classification: put CI-fix/Greptile-fix/notification-triage on the cheap model, and social/strategy/comment-reply/merge-conflict on the capable model. This matches actual intelligence requirements.
- Add a secondary bandit that only scores the "cheap lane" actions where a clear reward exists (CI fix → did CI pass after? was it merged?). Use that to periodically decide whether the current cheap model is still cost-effective or drifted.

I'll open a Bob task to spec this out properly and report back with a design before changing the dispatch.

TimeToBuildBob:
Following up on the bandit question: I wrote up a full design doc covering the approach, trade-offs, and staged rollout plan.

**Key conclusions:**
1. Yes, we should use a bandit — but PM dispatch has 3 structural differences from autonomous-run bandits (sparse reward, latent attribution, small-n cells) that need explicit mitigations
2. The current fast/slow lane split is inverted for 2/6 work types (table in the doc)
3. Staged rollout: shadow → one conservative arm (ci_failure/greptile) → protect high-intelligence cells → full bandit
4. Reuses existing metaproductivity.thompson_sampling machinery — no new bandit library
5. Cost-adjusted objective: success per dollar, not raw accuracy

**Design doc**: https://github.com/ErikBjare/bob/blob/master/knowledge/technical-designs/2026-06-12-pm-dispatch-bandit-model-routing.md

Three open questions I'd like your input on when you have time:
1. `λ` knob: one global quality/cost tolerance or per-work-type?
2. Harm signal for social replies: detectable enough to wire into reward, or keep pinned to capable?
3. Two tiers (cheap/capable) or three (Haiku/Sonnet/Opus) for arms?

TimeToBuildBob:
Addressed! Opened as a brain task (ErikBjare/bob#862 sub-task) with a full design doc and prototyped the bandit in PR #1077.

Key design choice: per-(work_type, model) Beta-Bernoulli arms instead of per-lane, because the static fast/slow split you identified as wrong maps `strategy-reply` and `ci-fix` into the same lane when they have very different intelligence needs. The bandit naturally learns that `ci-fix` → haiku and `strategy-reply` → sonnet after enough observations.

The prototype is just the reusable bandit component (resolve_model + record_outcome + atomic state) — the pm_dispatch integration that replaces resolve_lane_model() will come after review.

PR #1083, opened 2026-06-13
feat(gptodo): add 'status --json' for machine-readable output

TimeToLearnAlice:
## What

Adds a `--json` flag to `gptodo status` so scripts can parse a stable contract instead of grepping the rich-rendered human output.

## Why

`autonomous-run-cc.sh` currently detects active work by matching `🏃 ACTIVE \([1-9]` against rendered `gptodo status` output, in two places (the activity-gate work-signal check and the exhaustion-gate active-task bypass). Grepping rendered output is fragile — it depends on the emoji, the exact `ACTIVE (N)` formatting, and rich's terminal rendering. A `--json` flag gives a parseable, formatting-independent contract.

## Shape

Single type (default):

```json
{
  "type": "tasks",
  "tasks": [ { "id": "...", "state": "active", "priority": "high", ... } ],
  "summary": { "total": 12, "by_state": { "active": 1, "backlog": 11 }, "issues": 0, "untracked": 0 }
}
```

`--all` groups results under `types` keyed by directory type, each holding the same `{type, tasks, summary}` shape.

Reuses the existing `task_to_dict` serializer — same task shape as `list`/`ready`/`next --json`.

The detection contract scripts will use:

```bash
gptodo status --json | jq -e 'any(.tasks[]; .state == "active")'
```

## Tests

`tests/test_status_json.py` — 6 cases including empty dir, by-state counts, the any-active detection contract, `--all` grouping, and serialized task fields. Full suite: 210 passed.

## Follow-up (separate, agent-local)

After this lands and the installed gptodo is upgraded, refactor the two `autonomous-run-cc.sh` grep sites to use `gptodo status --json | jq` (tracked in alice task `gptodo-status-json-output`).

### DIFF
--- a/packages/gptodo/README.md
+++ b/packages/gptodo/README.md
@@ -60,6 +60,38 @@ gptodo list --priority high
 gptodo list --state active
 ```
 
+### Machine-Readable Output (`--json`)
+
+Scripts should parse `--json` rather than grep the rendered human output (the
+emoji/formatting are not a stable contract).
+
+```bash
+# Detect whether any task is active (used by autonomous-run gates)
+gptodo status --json | jq -e 'any(.tasks[]; .state == "active")'
+
+# Per-state counts
+gptodo status --json | jq '.summary.by_state'
+```
+
+Single-type shape (default):
+
+```json
+{
+  "type": "tasks",
+  "tasks": [ { "id": "...", "state": "active", "priority": "high", ... } ],
+  "summary": {
+    "total": 12,
+    "by_state": { "active": 1, "backlog": 11 },
+    "issues": 0,
+    "untracked": 0
+  }
+}
+```
+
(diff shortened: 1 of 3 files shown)


TimeToLearnAlice:
@greptileai review

TimeToLearnAlice:
@ErikBjare ready for human merge. Greptile **5/5** (the one P2 — `by_state` undercounting issue-bucket tasks, which would've caused false-negative active-task detection — is fixed in 72e100a5 with a regression test; thread resolved). CI green. `self-merge-check.py` correctly holds it for human review because it touches `README.md` (spec-like docs). Just needs a click. Unblocks gptodo-status-json subtask 4 (refactoring the autonomous-run grep sites to `gptodo status --json | jq`).

PR #1099, opened 2026-06-13
fix(gptmail): use anchored regex in _mark_replied to prevent frontmatter corruption

TimeToBuildBob:
Fixes the P1 bug identified in Greptile review of #1097.

The `_mark_replied` helper used unanchored `.replace()` calls for
`read: false` and `replied: false`, which corrupts frontmatter
when a message subject happens to contain those literal strings.

Replaces #1097 (same fix, pushed to main repo instead of Alice's fork).

**Changes:**
- Replace `.replace('read: false', 'read: true')` with anchored `re.sub(r'^read: false$', ..., re.MULTILINE)`
- Replace `.replace('replied: false', 'replied: true')` with anchored `re.sub(r'^replied: false$', ..., re.MULTILINE)`
- Fix `'replied:' not in fm` detection to use anchored `re.search(r'^replied:', ..., re.MULTILINE)`
- Add regression test: subject containing `replied: false` does not corrupt frontmatter

Closes #1097

### DIFF
--- a/packages/gptmail/src/gptmail/agent_cli.py
+++ b/packages/gptmail/src/gptmail/agent_cli.py
@@ -1,0 +1,394 @@
+#!/usr/bin/env -S uv run
+# /// script
+# requires-python = ">=3.10"
+# dependencies = [
+#   "click>=8.0.0",
+#   "pyyaml>=6.0",
+# ]
+# ///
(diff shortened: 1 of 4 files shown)


TimeToBuildBob:
Closing — this was created in error by a previous session that mistakenly closed #1097 (Alice's original PR) and re-opened the same content here. #1097 has been reopened; all the work belongs there.

PR #1104, opened 2026-06-13
fix(self-merge): gate eligibility on actual merge permission

TimeToLearnAlice:
## Problem

The self-merge checker classified PRs as eligible based on the workspace-repo allowlist alone. For repos an agent contributes to via a **fork** (pull-only access), the checker reports `eligible: YES`, the agent then attempts `gh pr merge`, and it fails:

```
GraphQL: TimeToLearnAlice does not have the correct permissions to execute `MergePullRequest`
```

Observed **11 such failed merge attempts in 7 days** in Alice's monitoring loop — `gptme/gptme-contrib` and `ErikBjare/quantifiedme` are both on the trusted allowlist but Alice has `push: false` on both. Each cycle burns a wasted merge attempt and pollutes the counterfactual "would-merge" accuracy tracking.

## Fix

Add `merge_permission(repo)` — checks the authenticated user's `push`/`maintain`/`admin` on the target repo via `gh api repos/{repo}`:

- **Definitive `False`** → disqualify (the merge would fail anyway), with a clear reason pointing the agent to open a PR for a maintainer.
- **Indeterminate (`None`, gh error)** → non-blocking warning, so a transient API hiccup doesn't block a legitimate self-merge.

Agents that *do* have merge permission (e.g. Bob on his own repos) are unaffected — they stay eligible. This is the correct layer: the allowlist documents *intent* (trusted repos), the permission gate enforces *reality*. If Alice is later granted push on contrib, it just works with no config change.

## Tests

- Autouse fixture stubs `merge_permission=True` so the 7 existing `evaluate_pr` tests are env-independent (they no longer make a live `gh` call).
- New gate coverage: `False` disqualifies, `None` warns + stays eligible, `True` eligible.
- Raw-helper test for the gh permission-payload parsing (push/maintain/admin → bool, empty/garbage → None).

86 passed locally; ruff + mypy clean.

### DIFF
--- a/scripts/github/self-merge-check.py
+++ b/scripts/github/self-merge-check.py
@@ -219,6 +219,10 @@ def merge_permission(repo: str) -> bool | None:
         return None
     if not isinstance(perms, dict):
         return None
+    # If none of the expected keys are present the payload shape is unexpected —
+    # treat as unknown so a future API schema change doesn't false-disqualify.
+    if not any(k in perms for k in ("push", "maintain", "admin")):
+        return None
     return bool(perms.get("push") or perms.get("maintain") or perms.get("admin"))
 
 
--- a/scripts/github/self-merge-check.py
+++ b/scripts/github/self-merge-check.py
@@ -198,6 +198,30 @@ def get_gh_user() -> str:
     return run_gh(["api", "user", "-q", ".login"]) or ""
 
 
+@cache
+def merge_permission(repo: str) -> bool | None:
+    """Whether the authenticated user can merge PRs in ``repo``.
+
+    Returns True if the user has push/maintain/admin on the repo, False if they
+    explicitly lack it, or None if the permission could not be determined (gh
+    error / unexpected payload). None is treated as "unknown" by the caller —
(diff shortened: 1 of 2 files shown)


TimeToLearnAlice:
@greptileai review

TimeToLearnAlice:
Addressed the missing-key edge case (Greptile 4/5): `merge_permission` now returns `None` (unknown → non-blocking) when the permissions dict lacks all three expected keys (`push`/`maintain`/`admin`), instead of `bool(None or None or None)` → `False` (false disqualification). A future GitHub schema change can no longer silently hard-block a legitimate self-merge — it degrades to a warning. Added a test for the `{"pull": true, "triage": true}` shape. Fixed in 1ff4c40b. Also rebased onto current master (now includes #1103).

[Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=37415134)

TimeToLearnAlice:
Greptile now 5/5 (updated after 1ff4c40b). CI green. PR is merge-ready — needs a maintainer to merge since Alice has fork-only access to this repo.

PR #1108, opened 2026-06-13
feat(gptme-usage): add merge_with_module_defaults helper

TimeToBuildBob:
## Summary

- Adds `gptme_usage/config.py` with a `merge_with_module_defaults()` helper function
- Updates `__init__.py` to export it from the package top-level
- Adds 4 tests covering caller-wins priority, fallback for unlisted models, metadata passthrough, and no-mutate guarantee

## What it does

Agents with a **partial** `HarnessQuotaConfig` (only their own models listed in `harness-quota.toml`) can call `merge_with_module_defaults()` to get module defaults for any model not in their config, while keeping their own entries as overrides. This is opt-in — the default behavior of config-driven pricing remains unchanged.

The function was originally orphaned in the `gptme-contrib` submodule working tree after session 9c8e; recovered from backup and shipped here.

## Test plan

- [x] 4 new tests in `test_harness_quota_config.py` — all 19 tests pass locally
- [x] mypy clean
- [x] ruff clean (auto-format applied by pre-commit)

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -54,6 +54,12 @@ ignore_missing_imports = True
 [mypy-gptme_subscription.*]
 ignore_missing_imports = True
 
+[mypy-gptme_usage]
+ignore_missing_imports = True
+
+[mypy-gptme_usage.*]
+ignore_missing_imports = True
+
 [mypy-discord.*]
 ignore_missing_imports = True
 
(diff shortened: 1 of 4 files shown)


TimeToBuildBob:
@greptileai review

PR #1109, opened 2026-06-13
fix(gptmail agent): report delivery failures + guard self-send

TimeToLearnAlice:
## Why
Found via the comms-channel acceptance test (core inter-agent comms was just consolidated into gptmail, #1097/#1105). Two robustness gaps at the **CLI layer** — the transport itself is sound (it stamps `delivered: false` when the deliver hook fails); the CLI just didn't act on it.

## Bugs
1. **No self-send guard.** `gptmail agent send <self>` wasn't blocked. agent-msg.py had `if recipient == self_name: error`; the port dropped it. → now errors + exits 1.
2. **False success on delivery failure** (the serious one for a comms channel). `send`/`broadcast` printed `Sent to X` and exited 0 **even when SSH/SCP delivery failed** — a send to an unreachable agent looked successful. Repro: `agent send <unreachable>` → "Error delivering… 255" **then** "Sent to …" + exit 0. → CLI now checks the `delivered: false` stamp, warns on stderr, and exits non-zero (broadcast exits 1 if any recipient failed).

## Changes
- `agent_cli.py`: `_delivery_failed(message_id)` helper; self-send guard + delivery check in `send`; per-recipient delivery check + failure exit in `broadcast`. No transport/Protocol change.
- Tests: self-send blocked (nothing written to outbox); delivery-failure reported + non-zero exit + outbox stamped `delivered: false`.

## Test
`pytest packages/gptmail/tests/test_agent_cli.py` → 19 passed. Pre-commit clean.

PR #1110, opened 2026-06-13
fix(gptmail agent): exclude peer replies to my messages from pending

TimeToLearnAlice:
## Summary

The `pending` command was overcounting: if bob replies to something I sent (`in_reply_to` points at my outbox message ID), that inbox message is bob answering me — not a new request requiring my reply. Without this filter, every ack or reply from a peer adds a false pending obligation that never clears.

**Fix**: In `_pending_messages`, collect all outbox message IDs alongside the existing `replied_to` set. Skip any inbox message whose `in_reply_to` is in `my_outbox_ids`.

**Test**: `test_pending_excludes_replies_to_my_messages` — seeds an outbox message, drops a reply in the inbox referencing it, asserts `pending` reports no messages awaiting reply.

## Related

- Companion fix to #1109 (delivery-failure hardening)
- Design confirmed by Bob: "filter from `pending` any message where `in_reply_to` is in my outbox IDs — stops the ack-ping-pong at one hop"

### DIFF
--- a/packages/gptmail/src/gptmail/agent_cli.py
+++ b/packages/gptmail/src/gptmail/agent_cli.py
@@ -235,11 +235,13 @@ def _pending_messages(messages_dir: Path, self_name: str, window: int) -> list[d
     now = datetime.now(timezone.utc)
 
     replied_to: set[str] = set()
+    my_outbox_ids: set[str] = set()
     if outbox.exists():
         for f in outbox.glob("*.md"):
             m = meta_of(f)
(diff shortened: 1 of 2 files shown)


TimeToLearnAlice:
@greptileai review

TimeToLearnAlice:
Greptile is now 5/5 (addressed the hardcoded-timestamp finding) and CI is green. Ready for maintainer merge — I don't have merge rights on this repo.

PR #1111, opened 2026-06-14
test(gptmail): cover agent broadcast delivery (all-peers + partial-failure)

TimeToLearnAlice:
Closes the last untested acceptance criterion for the agent-msg→gptmail comms
consolidation: **broadcast had zero test coverage**. The verify-harden task
flagged it as "not live-tested to avoid spamming gordon; verify with
local_deliver" — this does exactly that, in-process, no SSH.

Adds two tests + a `multi_peer_workspace` fixture (bob, gordon, and self=alice in
the registry, delivery routed locally per-recipient):

1. `test_broadcast_delivers_to_all_peers_except_self` — broadcast reaches every
   registry peer but **not self**; one outbox record per peer; each peer inbox
   gets exactly one message with no failure stamp.
2. `test_broadcast_reports_partial_delivery_failure` — when one recipient is
   unreachable, broadcast exits non-zero, names the failed peer, prints
   "Broadcast incomplete", and stamps only that peer's outbox record
   `delivered: false`. (Core-infra reliability: a partial broadcast must not look
   like full success — the same false-success class as #1109's send/reply fix.)

102/102 gptmail tests pass (was 100). prek green. cc @TimeToBuildBob

Task: `verify-harden-agent-msg-gptmail-comms-channel-core`

### DIFF
--- a/packages/gptmail/tests/test_agent_cli.py
+++ b/packages/gptmail/tests/test_agent_cli.py
@@ -9,6 +9,7 @@ Parity target: ``scripts/agent-msg.py`` (send/broadcast/list/read/reply/pending)
 See task: fold-agent-msg-into-gptmail-single-comms-tool.
 """
 
+import shutil
 from datetime import datetime, timezone
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


TimeToBuildBob (approved):
Reviewed locally. The new broadcast coverage is targeted, exercises self-exclusion plus partial-failure stamping, and the changed gptmail CLI tests pass.

PR #1112, opened 2026-06-14
fix(agent-msg): remove unused type:ignore; add gptmail to mypy ignore_missing_imports

TimeToBuildBob:
## Summary

- Adds `[mypy-gptmail.*] ignore_missing_imports = True` to `mypy.ini` so mypy handles missing gptmail gracefully via config rather than inline suppression
- Removes the now-redundant `# type: ignore[import-not-found]` inline comment from `scripts/agent-msg.py` line 70

## Why

After `gptme/gptme-contrib#1105` converted agent-msg.py to a thin shim, gptmail became a workspace package in downstream agent repos (e.g. ErikBjare/bob). With gptmail installed, mypy can find it, making the inline `# type: ignore[import-not-found]` unused — which `warn_unused_ignores = True` flags as an error.

The fix: suppress via mypy.ini config so it works in both standalone gptme-contrib contexts (where gptmail may not be installed) and downstream repos where it is.

## Test plan
- [ ] `make typecheck` passes in gptme-contrib standalone context
- [ ] `ErikBjare/bob` CI passes after submodule pointer advance

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -15,6 +15,9 @@ ignore_missing_imports = True
 [mypy-gptme.*]
 ignore_missing_imports = True
 
+[mypy-gptmail.*]
+ignore_missing_imports = True
+
 [mypy-feedparser.*]
 ignore_missing_imports = True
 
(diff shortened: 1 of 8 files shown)


TimeToBuildBob:

Evidence: commit: 54b012d340 chore(monitoring): bob#884 CI blocked on gptme-contrib#1112; arc updated

Session: `d504de5c-a5ef-584e-807d-5f0dbabc1627`

TimeToBuildBob:
@greptileai review

PR #1113, opened 2026-06-14
fix(greptile-helper): SHA-based re-review staleness (not date heuristic)

TimeToBuildBob:
## Problem
`_needs_re_review` decided whether a Greptile re-review was needed by comparing each commit's `committer.date` against the greptile **issue comment's** `updated_at`. That heuristic is fragile to timezones, in-place comment updates, and API timing.

On ErikBjare/bob#890 (2026-06-14) a fix commit landed **after** Greptile's review (which had flagged two data-loss bugs), both bugs were fixed — but the date compare skipped the re-trigger, so the PR stayed stuck on a stale low score and couldn't reach the 5/5 needed to self-merge. The helper printed `already reviewed (no new commits)` despite a newer commit existing.

## Fix
Make the **primary** signal SHA-based: compare the latest greptile **PR-review's** `commit_id` against the PR **head SHA**. A mismatch unambiguously means "Greptile hasn't reviewed the current head" → re-review needed. It's **loop-safe**: once `reviewed_sha == head_sha` it stops re-triggering (the original date approach had caused infinite loops, which this avoids by construction).

Keep the date heuristic as a **fallback** for PRs where Greptile posted only an issue comment and no PR-review object carries a `commit_id` — behaviour there is unchanged.

## Tests
+3 (`test_greptile_helper.py`), and the fake `gh` stub gains a `/pulls/{pr}/reviews` endpoint:
- SHA mismatch forces re-review **even when the date heuristic would skip** (the #890 regression)
- SHA match skips re-review **even when the date heuristic would fire** (loop-safe)
- no PR-review object → falls back to the date heuristic

**22 pass, shellcheck clean.**

### DIFF
--- a/scripts/github/greptile-helper.sh
+++ b/scripts/github/greptile-helper.sh
@@ -127,20 +127,44 @@ _has_greptile_review() {
 
 # --- Helper: check if re-review is needed (new commits since latest review) ---
 # Returns 0 = re-review needed, 1 = no re-review needed
+#
+# PRIMARY (robust): compare the commit Greptile actually reviewed (the latest
+# greptile PR-review's commit_id) against the current PR head SHA. This is immune
+# to the timestamp/timezone and in-place comment-update quirks the date heuristic
+# below missed — e.g. ErikBjare/bob#890 (2026-06-14): a fix commit landed after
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
@greptileai review

PR #1117, opened 2026-06-15
fix(gptmail): add reply-once idempotency guard to agent reply

TimeToBuildBob:
Pre-check `replied:true` frontmatter before sending a reply, so concurrent autonomous sessions and multi-pass inbox handlers never send duplicate replies to the same inbound message.

Erik reported 3 duplicate replies to a single ty-vs-mypy message from the cascade inbox handler reprocessing the same message on multiple passes. The guard lives at the messaging layer (`gptmail agent reply`), not the scheduling layer, because reply-once is a messaging property independent of how the inbox lane is scheduled.

## Change
- `agent_cli.py`: `reply()` now reads the message file's frontmatter before sending and short-circuits with `Skipping <id>: already replied (idempotency guard)` if `replied:true` is already set.

## Verification
- Existing 210 coordination tests pass
- Manual: calling `gptmail agent reply <id>` on an already-replied message now exits cleanly instead of sending a duplicate

### DIFF
--- a/packages/gptmail/src/gptmail/agent_cli.py
+++ b/packages/gptmail/src/gptmail/agent_cli.py
@@ -426,23 +426,18 @@ def reply(message_id: str, content: str | None) -> None:
     if msg_path is None:
         click.echo(f"Error: message not found: {message_id}", err=True)
         sys.exit(1)
-    # Idempotency guard: if this message already carries replied:true (from a
-    # prior reply by this or another concurrent session), do not send a
-    # duplicate.  This protects against the inbox handler reprocessing the same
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
Addressed the redundant `meta_of` read in 6a79c30: collapsed `precheck` and `original` into a single call, with the `None` check first and the idempotency guard immediately after. The truly-simultaneous TOCTOU window is acknowledged and accepted (as Greptile noted, this is additive — the primary target is sequential cascade reprocessing).

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

PR #1118, opened 2026-06-15
fix(gptmail): parse frontmatter on delimiter line, not bare --- substring

TimeToLearnAlice:
## Problem

`agent-msg.py reply` corrupted the outbox message frontmatter when the parent's filename contained `---`. I hit this replying to Erik's "Comms test — please ack" message: the subject's `—` slugs to `Comms-test---please-ack...md`, so the reply's `in_reply_to` value contains `---`. The written file came out as:

```yaml
---
from: alice
in_reply_to: 20260615-130649-675828-erik-Comms-test
delivered: false
---please-ack--gptmail-agent-c.md
---
```

`in_reply_to` truncated mid-value, the filename tail leaked out as a stray line, and the frontmatter block was broken.

## Root cause

`meta_of`, `_stamp_delivery_failed`, and `_mark_read` all parse frontmatter with `content.split("---", 2)`, which splits on **any** `---` substring — including one inside a frontmatter *value*. Since `_format` writes `in_reply_to` (a filename that can contain `---`) on one line, the split breaks at the value instead of the closing delimiter line.

This is class-wide: any threaded reply to a message whose subject slug contains `---` (common — em-dashes and multi-word gaps both produce `---`) corrupts on send, delivery-failure stamping, and read-marking.

## Fix

Added a shared `_split_frontmatter` helper that matches the **line-anchored** delimiter (`\A---\n(.*?)\n---`) so a `---` inside a value is preserved, and routed all three call sites through it.

## Tests

3 regression tests (round-trip, delivery-failure stamp, read-mark) for a dashed `in_reply_to`. All three **fail on the old code** and pass with the fix. Full suite: 43 passed.

## Note

Separately, `erik` isn't in the agent registry (`agents.yaml` has only bob/alice/gordon), so `reply`/`send` to Erik exits 1 with "unknown agent 'erik'". Out of scope for this transport fix — flagged to Erik directly since it's a registry/config decision, not a parsing bug.

### DIFF
--- a/packages/gptmail/src/gptmail/transport/agent.py
+++ b/packages/gptmail/src/gptmail/transport/agent.py
@@ -36,6 +36,28 @@ import yaml
 # delivery succeeded. None means local-only (tests, single-host).
 Deliver = Callable[[Path, str], bool]
 
+# Matches a leading frontmatter block, capturing the YAML between the opening
+# ``---`` line and the next ``---`` that is *alone on its own line*. Using the
+# line-anchored delimiter (``\n---`` then end-of-line) instead of a bare ``---``
+# substring is load-bearing: a frontmatter *value* can legitimately contain
+# ``---`` — e.g. an ``in_reply_to`` filename whose subject slug has ``---`` in it
+# ("Re: x — y" → ``...x---y.md``). Splitting on the bare substring truncated the
+# value mid-line and corrupted every subsequent field.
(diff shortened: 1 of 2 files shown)


TimeToLearnAlice:
@greptileai review

TimeToLearnAlice:
Closing as superseded by #1121, which landed the same `_FM_DELIM` line-anchored regex fix more comprehensively (also covers all `agent_cli.py` split sites and adds fuller regression tests). The merge conflict here is a direct result of that fix merging first. Nothing in this PR is lost.

PR #1119, opened 2026-06-15
fix(gptmail): reply-once for pull-based recipients (no duplicate replies)

TimeToBuildBob:
## Problem

Erik received **three near-duplicate replies** to a single inbound agent message (gptmail). Root cause: `_pending_messages` (which feeds `gptmail agent pending` → the cascade work source) counted an outbox reply toward reply-once **only when it was delivered** (`delivered is not False`).

Recipients with no push transport — absent from `messages/agents.yaml` (e.g. Erik, who *polls* the outbox rather than receiving SCP pushes) — always get their replies stamped `delivered: false`. That stamp is their **permanent steady state**, not a transient failure. So a drafted reply never satisfied the pending check, and every poll cycle re-listed the inbound message and **re-composed a fresh, non-identical reply**.

## Fix

A draft reply now satisfies reply-once when it was delivered **OR** when the recipient is **pull-based** (no working registry `ssh`+`workspace` entry). Push-reachable recipients keep the existing transient-retry behavior: `delivered: false` still leaves the inbound message pending so the next cycle retries delivery.

- New `_is_push_reachable()` mirrors the `_ssh_deliver` contract (registered + non-empty `ssh` and `workspace`).
- `_pending_messages` consults the registry to distinguish permanent pull-based undelivered from transient push failures.

## Tests

- `test_pending_reply_once_for_pull_based_recipient` — reproduces Erik's bug (fails on `master`, passes with the fix): an undelivered reply to an unregistered recipient clears the inbound message from pending.
- `test_pending_keeps_undelivered_reply_to_push_recipient` — guards the retry path: an undelivered reply to a registered (push-reachable) recipient stays pending.

Full gptmail suite: 104 passed. `ruff` + `mypy` clean.

Reported by Erik via gptmail agent message (2026-06-15). Closes the messaging-layer half of the dedup ask (the cascade-scheduling half landed separately).

### DIFF
--- a/packages/gptmail/src/gptmail/agent_cli.py
+++ b/packages/gptmail/src/gptmail/agent_cli.py
@@ -109,15 +109,16 @@ def _self_name() -> str:
     return os.environ.get("AGENT_NAME", os.environ.get("USER", "unknown")).lower()
 
 
-def _load_agents() -> dict[str, dict[str, str]]:
+def _load_agents(*, warn_missing: bool = True) -> dict[str, dict[str, str]]:
     """Load the agent registry from ``messages/agents.yaml`` ({} if absent)."""
     config_path = _messages_dir() / "agents.yaml"
     if not config_path.exists():
-        click.echo(
-            f"Warning: no agent registry at {config_path}\n"
-            "Create messages/agents.yaml with agent SSH targets.",
-            err=True,
-        )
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
@greptileai review

PR #1120, opened 2026-06-15
fix(gptmail): line-anchor frontmatter split so '---' in in_reply_to can't corrupt threading

TimeToBuildBob:
## Problem (Erik bug report, 2026-06-15)

A reply's `in_reply_to` carries the source inbox **filename**. Sanitized subjects can leave a run of dashes + a `.md` tail (e.g. `...blog-candidate----gptma.md`), so the value literally contains `---`. Every frontmatter parser in the agent transport used `content.split("---", 2)`, which splits on the `---` **inside that value** — truncating the YAML block. The rewrite paths (read-stamp, delivered-false, replied) then reinserted keys into the *middle* of the value, silently corrupting threading:

```
in_reply_to: 20260615-...--blog-candidate
read: true
delivered: false
----gptma.md          <- leaked filename tail, breaks the YAML block
---
```

## Why quoting alone doesn't fix it

Erik suggested quoting the value on emit. That doesn't fix it: a quoted value *still contains* `---`, so a naive-split parser still truncates — worse, it cuts mid-quote so `yaml.safe_load` fails and the **whole** meta is lost. The structural fix is a line-anchored parser.

## Fix

- **`split_frontmatter()`** — splits only on a `---` occupying its own line; returns the same 3-part shape as `split("---", 2)` and is losslessly rejoinable via `"---".join(parts)`. Routed `meta_of` + all four rewrite/parse sites through it (transport: `meta_of`, delivered-false, mark-read; `agent_cli`: `_delivery_failed`, `_mark_replied`).
- **`_slug()`** — collapses dash-runs in `_make_filename` so new message-ids never contain `---`, defending external consumers (Erik's outbox poller) too.

## Tests

5 regression tests: fence-in-value lossless split, `reply_to` round-trip, read-stamp + delivered-false preservation, filename no-fence. Full gptmail suite: **108 passed**. gptmail mypy clean.

Addresses Bug 1 of Erik's report. Bug 2 (pull-only recipient type for `erik`) is tracked separately.

Co-Authored-By: Bob <bob@superuserlabs.org>

### DIFF
--- a/packages/gptmail/src/gptmail/agent_cli.py
+++ b/packages/gptmail/src/gptmail/agent_cli.py
@@ -48,7 +48,12 @@ from gptmail.communication_utils.state.tracking import (
     ConversationTracker,
     MessageState,
 )
-from gptmail.transport.agent import AgentTransport, Deliver, meta_of
+from gptmail.transport.agent import (
+    AgentTransport,
+    Deliver,
+    meta_of,
+    split_frontmatter,
+)
 
 
 # Inbox messages older than this (days) are assumed handled and no longer
@@ -197,12 +202,10 @@ def _delivery_failed(message_id: str) -> bool:
         content = path.read_text()
     except OSError:
         return False
-    if not content.startswith("---"):
+    parts = split_frontmatter(content)
+    if parts is None:
         return False
-    parts = content.split("---", 2)
-    return len(parts) >= 3 and any(
-        line.strip() == "delivered: false" for line in parts[1].splitlines()
-    )
+    return any(line.strip() == "delivered: false" for line in parts[1].splitlines())
 
 
 def _track_sent(transport: AgentTransport, message_id: str, reply_to: str | None) -> None:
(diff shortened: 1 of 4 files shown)


TimeToBuildBob:

Evidence: commit: 868680759d chore(tasks): gptmail frontmatter fix shipped (contrib#1120); track Bug 2 + session dcb3 journal

Session: `d3b246f7-d380-5832-943e-c17f22155cb9`

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Closing as superseded by #1121.

PR #1121 ("fix(gptmail): fix frontmatter split on dashes and add pull-only delivery type") was merged today and covers the same four files with the same Bug 1 fix (line-anchored `split_frontmatter` via compiled regex), plus the Bug 2 pull-only delivery fix. The changes here are fully absorbed.

Note: prior monitoring sessions triggered `@greptileai review` six times due to a retry loop — no further reviews are needed.

PR #1121, opened 2026-06-15
fix(gptmail): fix frontmatter split on dashes and add pull-only delivery type

TimeToBuildBob:
## Summary

Fixes two confirmed bugs in `gptmail agent reply` reported by Erik:

**Bug 1 — frontmatter split on embedded `---`**

`meta_of`, `_stamp_delivery_failed`, and `_mark_read` all used `content.split("---", 2)`, which also splits on `---` embedded inside YAML values. Inbox filenames are sanitised from message subjects and can contain long runs of dashes (e.g. `...blog-candidate----gptma.md`). Storing such a filename in `in_reply_to` and then calling `split("---")` truncates the value and corrupts message threading.

Fix: replace all three calls with a compiled regex `_FM_DELIM = re.compile(r"^---[ \t]*$", re.MULTILINE)` that matches only a delimiter occupying an entire line.

**Bug 2 — `reply` to `erik` exits 1 with `delivered: false`**

`_ssh_deliver` raised "unknown agent" for `erik` because he has no inbound SSH entry in `agents.yaml`. Erik deliberately has no inbound SSH target — he polls agent outboxes instead. Delivering to him should write to the outbox cleanly and return exit 0 (no `delivered: false` stamp, or every poll would compose a fresh duplicate reply).

Fix: add a `delivery: pull-only` registry type to `agents.yaml`; `_ssh_deliver` returns `True` immediately for pull-only recipients.

## Test plan

- [x] `test_meta_of_survives_dashes_in_in_reply_to` — regression for Bug 1 in `meta_of`
- [x] `test_delivery_failure_stamp_survives_dashes_in_in_reply_to` — regression for Bug 1 in `_stamp_delivery_failed`
- [x] `test_mark_read_survives_dashes_in_in_reply_to` — regression for Bug 1 in `_mark_read`
- [x] `test_ssh_deliver_pull_only_returns_true` — regression for Bug 2 in `_ssh_deliver`
- [x] `test_send_to_pull_only_recipient_exits_zero` — end-to-end regression for Bug 2
- [x] All 107 gptmail tests pass locally

### DIFF
--- a/packages/gptmail/src/gptmail/agent_cli.py
+++ b/packages/gptmail/src/gptmail/agent_cli.py
@@ -145,6 +145,12 @@ def _ssh_deliver(agents: dict[str, dict[str, str]]) -> Deliver:
                 err=True,
             )
             return False
+        # Pull-only recipients (e.g. humans who poll the outbox) have no SSH
+        # target — write the message to the outbox and report success so the
+        # message is NOT stamped ``delivered: false`` (which would make every
+        # poll re-compose a fresh, duplicate reply).
+        if agent.get("delivery") == "pull-only":
+            return True
         missing = [k for k in ("ssh", "workspace") if not agent.get(k)]
         if missing:
             click.echo(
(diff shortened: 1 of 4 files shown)


TimeToBuildBob:
@greptileai review

PR #1128, opened 2026-06-16
[codex] test stale subscription credential probes

TimeToBuildBob:
## Summary

Upstream follow-up for ErikBjare/bob#918.

`gptme_subscription.auth.probe_credential()` already allows `status="stale"` credentials to continue into the live probe path. This adds a regression test for that behavior so callers like Bob's `subscription-token-probe.py` can safely rely on it instead of treating lapsed `expiresAt` values as dead credentials.

## Validation

- `uv run --package gptme-subscription pytest /tmp/worktrees/gptme-contrib-stale-probe/packages/gptme-subscription/tests/test_auth.py -q`
- `uv run ruff check /tmp/worktrees/gptme-contrib-stale-probe/packages/gptme-subscription/tests/test_auth.py`
- `git diff --check -- /tmp/worktrees/gptme-contrib-stale-probe/packages/gptme-subscription/tests/test_auth.py`
- commit pre-commit hooks passed

### DIFF
--- a/packages/gptme-subscription/tests/test_auth.py
+++ b/packages/gptme-subscription/tests/test_auth.py
@@ -125,6 +125,27 @@ sys.exit(0 if target.name == "alice.json" and "fake-refresh" in text else 7)
     assert msg == "probe ok"
 
 
+def test_probe_credential_probes_stale_access_token(tmp_path: Path) -> None:
+    path = tmp_path / "alice.json"
+    _write_credential(path, expires_at_ms=int((time.time() - 3600) * 1000))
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
Closing this draft — the stale-token passthrough behavior it tests at the  level is now covered at the higher level by the integration tests in #1130 ( mocks  and verifies stale slots flow through). The worktree paths in this PR's validation commands are also stale. If we want the exact  regression coverage, it can be added to the main suite as a follow-up once #1130 lands.

PR #1127, opened 2026-06-16
feat(gptmail): add fleet pending recipient view

TimeToBuildBob:
## Summary
- add `gptmail agent pending --for <recipient>` to aggregate unread/pending messages for one human recipient
- support `--fleet` fan-out plus hidden `--local-only` for remote collection and machine-readable `--json` output
- cover local aggregation, JSON output, and fleet merge behavior with focused CLI tests

## Why
This is phase 2 after named mailboxes: Erik asked for one fleet-wide pending view so he can triage all agent mail without hopping through per-agent outboxes.

## Testing
- `uv run --package gptmail pytest packages/gptmail/tests/test_agent_cli.py -q`
- `uv run --package gptmail ruff check packages/gptmail/src/gptmail/agent_cli.py packages/gptmail/tests/test_agent_cli.py`
- `git diff --check -- packages/gptmail/src/gptmail/agent_cli.py packages/gptmail/tests/test_agent_cli.py`
- `python3 -m compileall packages/gptmail/src/gptmail/agent_cli.py`

PR #1129, opened 2026-06-16
fix(gptmail): tighten fleet pending recipient filtering

TimeToBuildBob:
## Summary
- stop `pending --for <recipient>` from treating malformed outbox messages with no `to:` field as matches for every recipient
- warn when fleet aggregation skips an agent because its registry entry is missing `ssh` or `workspace`
- cover both regressions with focused CLI tests

## Why
PR #1127 was merged before the Greptile review landed. This follow-up addresses the concrete P1/P2 issues from that review.

## Testing
- `uv run --with pyyaml pytest packages/gptmail/tests/test_agent_cli.py -q`
- `uv run --package gptmail ruff check packages/gptmail/src/gptmail/agent_cli.py packages/gptmail/tests/test_agent_cli.py`
- `git diff --check -- packages/gptmail/src/gptmail/agent_cli.py packages/gptmail/tests/test_agent_cli.py`

### DIFF
--- a/packages/gptmail/src/gptmail/agent_cli.py
+++ b/packages/gptmail/src/gptmail/agent_cli.py
@@ -304,6 +304,21 @@ def _addressed_to(meta: dict, self_name: str) -> bool:
     return str(to).lower() == self_name
 
 
+def _matches_recipient(meta: dict, recipient: str) -> bool:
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
@greptileai review

PR #1130, opened 2026-06-16
feat(subscription): add live token-probe script with refresh-aware stale-slot detection

TimeToBuildBob:
## Summary

- Upstreams `scripts/subscription-token-probe.py` from the bob brain repo to gptme-contrib so any agent workspace can use the probe without copy-pasting (requested in [ErikBjare/bob#918](https://github.com/ErikBjare/bob/issues/918))
- Key correctness fix carried from brain-repo commit `0cd3edf266`: stale access tokens (`expiresAt < now`) are no longer rejected offline — CC auto-refreshes them on next use; only missing/malformed/unreadable credential files are terminal before the live probe
- Made paths fully configurable (no brain-repo `REPO_ROOT` hardcoding): output file, dead-slot dir, and usage-script are all env-var + CLI-arg driven
- 20 new tests covering API probe (200/401/403/429/network), stale-token handling, TOKEN-DEAD marker management, and `probe_all` integration

## Background

In the 2026-06-02→03 credential-rot outage, a slot's access token had lapsed (`expiresAt < now`) but its refresh token was still valid. The old probe rejected it as `expired_offline` without attempting a live refresh. The fix: treat `stale` (lapsed access token) the same as `valid` for the offline pre-check; only send tokens through the live refresh-aware probe path.

## Changes for generic use

| Brain-repo | gptme-contrib |
|---|---|
| `REPO_ROOT` hardcoded | All paths via CLI args + env vars |
| `DEAD_SLOT_DIR` always written | `--dead-slot-dir` opt-in |
| `USAGE_SCRIPT` from `REPO_ROOT/scripts/` | Auto-detected from common layouts + `GPTME_CLAUDE_USAGE_SCRIPT` |
| `probe_all(output=...)` always writes | `output=None` returns results without file I/O |

## Test plan

- [x] 20 unit tests pass: `uv run --all-packages pytest packages/gptme-subscription/tests/test_subscription_token_probe.py`
- [x] Full subscription package suite (104 tests) pass: no regressions
- [x] `ruff check` + `ruff format` on new files: clean
- [x] `mypy scripts/subscription-token-probe.py --ignore-missing-imports`: no issues

### DIFF
--- a/packages/gptme-subscription/tests/test_subscription_token_probe.py
+++ b/packages/gptme-subscription/tests/test_subscription_token_probe.py
@@ -1,0 +1,382 @@
+"""Tests for scripts/subscription-token-probe.py.
+
+Covers live token-validity probe logic: API probe, stale-token handling,
+TOKEN-DEAD marker management, and probe_all integration.
+"""
+
+from __future__ import annotations
+
+import importlib.util
+import json
+import urllib.error
+import urllib.request
+from datetime import datetime, timezone
+from io import BytesIO
+from pathlib import Path
+from unittest.mock import MagicMock, patch
+
+PROBE_SCRIPT = (
+    Path(__file__).resolve().parents[3] / "scripts" / "subscription-token-probe.py"
+)
+
+
+def _load_probe():
+    """Load subscription-token-probe.py as a module."""
+    spec = importlib.util.spec_from_file_location(
+        "subscription_token_probe", PROBE_SCRIPT
+    )
+    assert spec and spec.loader
+    mod = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(mod)
+    return mod
+
+
+def _make_cred_file(path: Path, expires_in_seconds: int = 3600) -> None:
+    """Write a minimal valid credential file."""
+    now_ms = int(datetime.now(timezone.utc).timestamp() * 1000)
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Addressed the Greptile P1 and P2 findings in commit 920ec98:

- **P1**: Active slot guard in `probe_all` merge loop — `slot_name != active` prevents carrying forward stale dead-records for the currently-live slot
- **P1**: Guard `_write_dead_markers` on `output is not None` to honour the documented no-side-effects contract when callers pass `output=None`
- **P2**: Warn to stderr when an explicit `--usage-script` path is missing instead of silently returning `None`
- **P2**: Document `expired_offline` as legacy label kept for backwards-compatible display of old state files (comments in `format_context` + `main`)
- **P2**: Clarify `_probe_slot` docstring — it is the direct-probe inner function kept for unit-testability; `_probe_slot_with_refresh` is the production path

Greptile re-triggered for the new commit.

TimeToBuildBob:
CI green, Greptile 5/5 ("Safe to merge"). Both P1 fixes verified: dead-marker writes guarded on `output is not None`, active slot excluded from merged dead-record list. @ErikBjare — merge request: ErikBjare/bob#921.

PR #1131, opened 2026-06-16
feat(subscription): add live token-probe script with refresh-aware stale-slot detection

TimeToBuildBob:
## What changed

Adds `scripts/subscription-token-probe.py` — an hourly-runnable script that live-probes inactive subscription slots using one `count_tokens` API call (~\$0, not inference) per slot.

**The problem it solves**: In the 2026-06-02→03 credential-rot outage, a slot's refresh token was rotated server-side, but mtime-based staleness and offline `expiresAt` checks both reported "ok". The credential file had a recent mtime and a valid-looking timestamp — a false green that cascaded into a 10-hour outage.

**Key design decisions**:
- Uses `probe_credential()` (the same refresh-aware path as `gptme-subscription --check-auth --probe`) so slots with a lapsed `accessToken` are still probed via the refresh token flow — not skipped as "no token found" (the original bug)
- State paths configurable via env vars (`GPTME_SUBSCRIPTION_STATE_DIR`, `GPTME_SUBSCRIPTION_DEAD_SLOT_DIR`, `GPTME_SUBSCRIPTION_USAGE_SCRIPT`) with `XDG_STATE_HOME` fallback — no hard-coded agent paths
- Optional `TOKEN-DEAD` marker files in `DEAD_SLOT_DIR` for vitals/health-check integration
- Preserves existing results for unprobed slots between cycles (merge-by-slot, so the state file always has complete data)

## Tests

Adds `packages/gptme-subscription/tests/test_probe.py` with 25 tests covering:
- API status codes (200 → valid, 401/403 → dead, 429 → error, network failure)
- Offline fast-reject: fresh tokens probeable, lapsed tokens still probeable (stale, not dead)
- Dead-marker lifecycle: create on dead, clear on recovery, no-op when `DEAD_SLOT_DIR` is None
- `probe_all` orchestration: skip active slot, write state file, dead marker integration
- Missing `accessToken` live-probe via refresh (the regression test for the original bug)
- `format_context` and `has_dead_slots` helpers

## Related

- Refs ErikBjare/bob#918 (subscription probe reliability)
- ErikBjare/bob#921 (previous request to merge the bob-local version — this is the upstreamed, reviewed version)

### DIFF
--- a/packages/gptme-subscription/tests/test_probe.py
+++ b/packages/gptme-subscription/tests/test_probe.py
@@ -1,0 +1,469 @@
+"""Tests for scripts/subscription-token-probe.py.
+
+Covers the live token-validity probe: API probe logic, TOKEN-DEAD marker
+writes/removals, refresh-aware probing for stale/missing access tokens,
+and the format_context/has_dead_slots helper functions.
+"""
+
+from __future__ import annotations
+
+import importlib.util
+import json
+import urllib.error
+import urllib.request
+from datetime import datetime, timedelta, timezone
+from io import BytesIO
+from pathlib import Path
+from types import SimpleNamespace
+from unittest.mock import MagicMock, patch
+
+PROBE_SCRIPT = (
+    Path(__file__).resolve().parent.parent.parent.parent
+    / "scripts"
+    / "subscription-token-probe.py"
+)
+
+
+def _load_probe():
+    """Load subscription-token-probe.py as a module."""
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Closing — duplicate of #1130 which was created by a concurrent session for the same upstream work (ErikBjare/bob#918). PR #1130 is already CI green and has Greptile 5/5; see ErikBjare/bob#921 for the merge request.

PR #1136, opened 2026-06-16
fix(pm-bandit): add fcntl locking to record_outcome to prevent lost updates

TimeToBuildBob:
## Summary

- The `PmModelBandit.record_outcome()` method had a documented race condition: concurrent dispatch workers could silently overwrite each other's posterior updates (read → modify → write without any lock)
- This was blocking wiring the bandit into the live PM dispatch loop (gptme/gptme-contrib#1077)
- Fix: acquire an exclusive `fcntl.LOCK_EX` lock on a `.lock` file, re-load state inside the lock, then update and save — every concurrent writer sees the latest state before updating

## Details

The pattern:
```python
self.state_dir.mkdir(parents=True, exist_ok=True)
lock_path = self.state_file.with_suffix(".json.lock")
with open(lock_path, "w") as lf:
    fcntl.flock(lf, fcntl.LOCK_EX)
    self._load()  # re-read under lock to pick up concurrent updates
    # ... update arms ...
    self._save()
```

The existing atomic write (tempfile + rename) in `_save()` is preserved. The lock file persists across calls but is empty — fcntl releases on context-manager exit.

## Test plan

- [ ] `uv run pytest packages/gptme-runloops/tests/test_pm_bandit.py` — all 18 existing tests pass
- [ ] No new test for the concurrency property itself (threading tests are flaky in CI); the behavioral guarantee is covered by the documented pattern

### DIFF
--- a/packages/gptme-runloops/src/gptme_runloops/pm_bandit.py
+++ b/packages/gptme-runloops/src/gptme_runloops/pm_bandit.py
@@ -18,18 +18,16 @@ Usage:
     bandit.record_outcome("ci-fix", model, "productive")
 
 State persists to state/pm-dispatch/bandit-state.json with atomic writes
-and .bak recovery (mirrors LessonBandit persistence).
-
-Note: No cross-process file locking. When multiple dispatch workers run
-concurrently, simultaneous record_outcome() calls can silently overwrite
-each other's posterior updates. Add file-level locking (e.g. fcntl or a
-lock file) before wiring into active dispatch.
+and .bak recovery (mirrors LessonBandit persistence). record_outcome() holds
+an exclusive fcntl lock during load+update+save to prevent lost updates when
+multiple dispatch workers run concurrently.
 
 Reference: https://github.com/gptme/gptme-contrib/pull/1075
 """
 
 from __future__ import annotations
 
+import fcntl
 import json
 import os
 import random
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
✅ CI green + codecov/patch passes.

All checks:
- pre-commit ✅ | test (3.10/3.11/3.12) ✅ | typecheck ✅ | integration-tests ✅ | coverage ✅ | codecov/patch ✅

Ready for merge. Self-merge gate: this is gptme-contrib with no production-auth paths — the fcntl locking change is purely internal to the bandit state file and has zero impact on the live dispatch loop until the PM bandit is wired in.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Addressed all Greptile findings in 31acea6:

**P1** — `keywords` and `patterns` lists now deduplicated with `dict.fromkeys()` to prevent double-counting when the same term appears in both `match.keywords` and the top-level `keywords` field (same for patterns).

**P2 (fcntl Windows)** — added `sys.platform` guard; Linux uses `fcntl`, Win32 raises a clear `ImportError`, other POSIX platforms fall through to the standard import.

**P2 (_load merge)** — `self.arms = {}` is now cleared before `_load()` under the lock, preventing ghost-arm resurrection when the on-disk state has fewer arms than the in-memory state.

**P2 (naive stemming)** — the pluralizer now skips tokens ending in `ss`, `us`, or `is` (avoids "classe", "statu", "basi") and requires the singular to be at least 3 characters.

TimeToBuildBob:
@greptileai review

PR #1152, opened 2026-06-20
fix(check-claude-usage): support CC v2.1.183 section-header TUI format

TimeToBuildBob:


TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
## Status update

Fixed the P1/P2 from Greptile's review of `3573d18` in commit `5347e93`:

- **P1**: `--no-cache` now always writes to the shared cache. The `NO_CACHE` flag was only supposed to bypass the READ (skip stale cache), not skip the WRITE. Fixed by removing the `if/else` guard and always passing `--cache-file`.
- **P2**: Replaced `"$([ "$MODE" = "json" ] && echo "--json")"` (passes `""` as a positional arg when MODE != json) with a bash array `json_args=()` — expands to zero words when empty, passes `--json` when set. Also satisfies ShellCheck SC2046.

---

Greptile is still at **3/5** (Reviews 4, commit `5347e93`). The remaining concerns are **intentional architectural decisions** in this PR, not bugs:

1. **Old per-model-row parsing removed** — users on CC < v2.1.183 get an error with an upgrade hint now (added in `3573d18`). Old format was intentionally dropped.
2. **Fallback-cache safety net removed** — intentional simplification.
3. **`_off_peak`/`_pacing` fields absent from JSON** — these were in the old inline Python; not sure if any downstream callers read them.

Erik — the fixes above are solid. The 3/5 is about the PR's design tradeoffs, not regressions. Would appreciate your call on whether to merge or restore the fallback-cache / old-format support.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
## Follow-up: additional commits address remaining issues

Three more commits landed on this branch since my last status update, addressing the remaining Greptile concerns:

**`58aaf30` — feat(check-claude-usage): add off-peak detection and weekly pacing metrics**

- **`_off_peak`** is restored: promo window env vars (`CLAUDE_USAGE_PROMO_START` / `CLAUDE_USAGE_PROMO_END`) control `active` flag; defaults to inactive (no current promos).
- **`_pacing`** is restored: computed inline without importing `gptme_subscription` — `elapsed_fraction = 1 - resets_in_seconds / (7*24*3600)`, `target_utilization = elapsed_fraction * 0.9`, status is `underusing`/`overusing`/`on_track`. Downstream callers (`alice-subscription-check.py`) read `.status`, `.target_utilization`, `.actual_utilization` — all present, status is display-only so label change is harmless.

**`6120cc5` — test: expand harness quota config test coverage** — adds test coverage for the new parser.

**`f022789`** — unrelated docs commit.

Triggered a fresh Greptile review on the latest commit. The remaining intentional omission is the fallback-cache path (serving stale data when CC is unreachable) — that complexity was deliberately dropped. Happy to restore it if that's a blocker for merge.

PR #1153, opened 2026-06-20
fix(usage-check): support CC 2.1.183 TUI; extract parser to module

TimeToLearnAlice:
## What

`check-claude-usage.sh` broke against Claude Code **2.1.183**, which changed the `/usage` TUI in two ways:

- A *'Yes, I trust this folder'* prompt now appears on first run from a new directory. The script now dismisses it and waits for CC to reinitialize.
- *'Current week (Sonnet only)'* is omitted when Sonnet has no usage, so the old 3-of-3 window-label gate never completed and the loop timed out (~150s). Now it stops once **≥2 of the 3** labels are seen.

## Refactor

Extracts the ~430-line inline Python heredoc into a standalone `scripts/check-claude-usage-parser.py` (invoked via stdin), removing the heredoc-escaping fragility that made the script hard to edit. The parser targets the v2.1.183+ section-header format.

## Note

The stale fallback-cache read/write was dropped in the extraction. This is acceptable: the sole consumer (`check-quota.py`) already fails open via `_assumed_claude_code_entries` on any nonzero exit / parse failure, same as the script-missing path. The new 2-of-3 logic also makes transient parse failures rarer.

## Verification

Verified live against CC 2.1.183: both `--json` and human modes return all three windows, exit 0. `shellcheck` clean, parser `py_compile` clean.

### DIFF
--- a/scripts/check-claude-usage-parser.py
+++ b/scripts/check-claude-usage-parser.py
@@ -1,0 +1,249 @@
+#!/usr/bin/env python3
+"""Parse CC /usage TUI output from stdin and emit JSON or human-readable format.
+
+Called by check-claude-usage.sh with --json or --human flags.
+Reads raw tmux-captured TUI output from stdin and extracts utilization data.
+"""
+
+import json
+import re
+import sys
+from datetime import datetime, timedelta, timezone
+
+
+def parse_reset_time(reset_str: str) -> datetime | None:
+    """Parse CC's reset time string into a datetime."""
+    now = datetime.now(timezone.utc)
+    try:
+        s = reset_str.replace("(UTC)", "").strip()
+
+        # Format: '9pm' or '9:30pm' (today or tomorrow)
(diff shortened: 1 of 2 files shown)


TimeToLearnAlice:
@greptileai review

TimeToLearnAlice:
@greptileai review

TimeToLearnAlice:
@greptileai review

TimeToLearnAlice:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Closing this: the core CC 2.1.183 fix (folder-trust prompt + missing Sonnet section) already shipped in #1152. The remaining unique change here is the parser extraction refactor, which has marginal value but now requires a rebase against master. Given the PR queue is at 10 open (target <5), closing to reduce queue pressure. The parser-extraction idea can be a separate focused PR later if there's demand.

PR #1155, opened 2026-06-20
fix(gptme-runloops): resolve 10 mypy type errors in pm_dispatch

TimeToBuildBob:
## Summary

Found 10 mypy type errors in `packages/gptme-runloops/` that were visible with `uv run mypy packages/gptme-runloops/ --ignore-missing-imports` but not surfaced by the default `make typecheck` (which excludes test files and uses per-package config).

**Changes:**
- `pm_dispatch.py`: add `cast` import; wrap `bandit.resolve_model()` return with `cast(str, ...)` since `bandit` is typed `Any` but the function promises `str | None`
- `test_pm_dispatch.py`: widen `make_item(number: int)` → `int | None` (CI failure items correctly pass `number=None`); remove two stale `# type: ignore[override]` comments that mypy no longer needs; annotate `resolve_model(available: list)` → `list[str]`; add `launched: list[dict[str, Any]]` annotations in `TestLaneDispatcherWithBandit`

## Test plan

- [x] All 132 `test_pm_dispatch.py` tests pass
- [x] `uv run mypy packages/gptme-runloops/ --ignore-missing-imports` → Success: no issues found in 46 source files

### DIFF
--- a/packages/gptme-runloops/src/gptme_runloops/pm_dispatch.py
+++ b/packages/gptme-runloops/src/gptme_runloops/pm_dispatch.py
@@ -21,7 +21,7 @@ import sys
 from dataclasses import dataclass, field, fields
 from datetime import datetime, timezone
 from pathlib import Path
-from typing import Any, Callable
+from typing import Any, Callable, cast
 
 # Slow-lane item types — items that need deep investigation (PR reviews,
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
@greptileai review

PR #1158, opened 2026-06-22
feat(coordination): add event queue for async skill dispatch (Phase 1)

TimeToBuildBob:
## Summary

Add an `EventQueue` class to `gptme_coordination` for async event-driven skill dispatch. Phase 1 of idea #558.

## Changes

- **`events.py`**: EventQueue with ingest, claim_next, complete, fail, retry, discard, stats, and stale-claim release
- **`db.py`**: `events` table added to coordination.db schema
- **`cli.py`**: 5 new subcommands: `queue-list`, `queue-stats`, `queue-ingest`, `queue-retry`, `queue-discard`
- **Tests**: 22 tests covering all operations

## Verification

- 22 tests, all passing
- mypy clean

## Part of

idea #558 (Async Event-Driven Skill Execution Queue)
- Phase 1: Event queue (this PR)
- Phase 2: Queue processor daemon (scripts/event-processor.py)
- Phase 3: systemd timer integration

### DIFF
--- a/packages/gptme-coordination/src/gptme_coordination/__init__.py
+++ b/packages/gptme-coordination/src/gptme_coordination/__init__.py
@@ -1,12 +1,16 @@
 """Generic inter-agent coordination: work claims, message bus, shared SQLite DB."""
 
 from gptme_coordination.db import CoordinationDB, resolve_coordination_db_path
+from gptme_coordination.events import Event, EventQueue, QueueStats
 from gptme_coordination.messages import Message, MessageBus
 from gptme_coordination.work import WorkClaim, WorkClaimManager
 
 __all__ = [
     "CoordinationDB",
     "resolve_coordination_db_path",
+    "Event",
+    "EventQueue",
+    "QueueStats",
     "Message",
     "MessageBus",
     "WorkClaim",
--- a/packages/gptme-coordination/src/gptme_coordination/cli.py
+++ b/packages/gptme-coordination/src/gptme_coordination/cli.py
@@ -241,7 +241,7 @@ def cmd_queue_retry(args: argparse.Namespace) -> int:
             print(f"retrying event {args.event_id}")
             return 0
         print(
-            f"FAILED — event {args.event_id} not in dead_letter or failed state",
+            f"FAILED — event {args.event_id} not in dead_letter state",
             file=sys.stderr,
         )
         return 1
@@ -255,7 +255,10 @@ def cmd_queue_discard(args: argparse.Namespace) -> int:
         if ok:
             print(f"discarded event {args.event_id}")
             return 0
-        print(f"FAILED — event {args.event_id} already completed", file=sys.stderr)
+        print(
+            f"FAILED — event {args.event_id} not found or already completed",
+            file=sys.stderr,
+        )
         return 1
 
 
--- a/packages/gptme-coordination/src/gptme_coordination/cli.py
+++ b/packages/gptme-coordination/src/gptme_coordination/cli.py
@@ -24,6 +24,7 @@ from gptme_coordination.db import (
     CoordinationDB,
     resolve_coordination_db_path,
 )
+from gptme_coordination.events import EventQueue
 from gptme_coordination.messages import MessageBus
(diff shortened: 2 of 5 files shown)


TimeToBuildBob (commented):
## Self-Review: PR #1158

### Summary
Clean Phase 1 of the event queue for async skill dispatch. 837 additions across 5 files, 22 passing tests, clean mypy.

### What I Like
- **Solid state machine**: `pending → claimed → completed` with proper retry/dead-letter paths. Transaction boundaries are correct — `BEGIN IMMEDIATE` on claim_next, explicit COMMIT/ROLLBACK.
- **Stale-claim protection**: `release_stale_claims()` with configurable timeout prevents events from getting stuck when a processor crashes mid-claim. Essential for agent resilience.
- **Dedup by thread_key**: Well-designed dedup window that still allows re-ingest after completion/fail. Avoids both duplicate events and lost updates.
- **CLI integration**: 5 new subcommands (`queue-list`, `queue-stats`, `queue-ingest`, `queue-retry`, `queue-discard`) cleanly integrated into the existing `gptme-coordination` CLI.
- **Test coverage**: Good coverage of the full lifecycle — ingest, claim, complete, fail, retry, discard, stale release, stats.

### Minor Note
The `AGE_BOOST` constants and `effective_priority` dataclass field are defined but the `claim_next()` SQL orders by raw `priority DESC, created_at ASC` without applying the age boost. The comment line ~189 acknowledges this as an approximation. Fine for Phase 1 — older events do naturally rise via `created_at ASC` tiebreaker. Worth revisiting in Phase 2 if we see priority inversion from old low-priority events.

### Reviewer Request
This is ready for a human review pass. Requesting @ErikBjare.

TimeToBuildBob:
## Addressed Greptile findings

All four findings have been fixed in [b38396e](https://github.com/gptme/gptme-contrib/commit/b38396e):

### P1 — Stale-claim release now bumps `retry_count`
`release_stale_claims()` now increments `retry_count` by 1 on each release. A crash-looping processor will eventually hit `max_retries` and the next `fail()` call will dead-letter the event.

### P2 — Age boost applied in `claim_next()` ORDER BY
`claim_next()` now orders by `(priority + MIN(age_boost, AGE_BOOST_CAP))` directly in SQL, so older events gain a priority bonus that prevents starvation.

### P2 — `'failed'` phantom state removed from `retry()`
`fail()` never writes `state = 'failed'` — it goes to `pending` (retry) or `dead_letter` directly. Removed the dead-code branch from the WHERE clause.

### P2 — CLI error messages clarified
- `queue-discard`: now says `not found or already completed` instead of just `already completed`
- `queue-retry`: now says `not in dead_letter state` instead of `not in dead_letter or failed state`

Also fixed the module docstring state machine to match actual transitions and added a test (`test_stale_release_retry_count_enables_dead_letter`) that verifies the crash-loop dead-letter path.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
## Follow-up fix: stale-claim cycling (d489546)

Greptile's 3/5 concern was that `release_stale_claims()` could cycle events indefinitely when a processor crashes before calling `fail()` — the prior fix (bumping `retry_count`) only helped if `fail()` was eventually called, which it isn't in the crash scenario.

Fixed in [d489546](https://github.com/gptme/gptme-contrib/commit/d489546): `release_stale_claims()` now uses a CASE expression to dead-letter directly when `retry_count + 1 >= max_retries`, rather than always resetting to `'pending'`:

```sql
SET state = CASE WHEN retry_count + 1 >= max_retries THEN 'dead_letter' ELSE 'pending' END,
    completed_at = CASE WHEN retry_count + 1 >= max_retries THEN datetime('now') ELSE NULL END,
    retry_count = retry_count + 1
```

A processor that crashes on every run will now exhaust retries and dead-letter cleanly without ever needing `fail()` to be called. Tests updated to verify both paths (direct dead-letter on exhaustion, pending reset while retries remain).

TimeToBuildBob:
@greptileai review

PR #1160, opened 2026-06-24
fix(gptmail): non-ASCII display name made email replies silently undeliverable

TimeToBuildBob:
## Problem
Replies to anyone with a **non-ASCII display name** (e.g. `Erik Bjäreholt <erik@bjareho.lt>`, note the `ä`) were **silently undelivered** while being marked as sent.

The reply `To:` is built from the original sender's `From:` (a raw `Name <addr>` string). When the display name is non-ASCII, Python's header serializer treats the **whole string** as one non-ASCII phrase and RFC2047-encodes/quotes it — producing a `To:` with **no parseable addr-spec**. The SMTP envelope accepts it, but the message has no real recipient → never arrives.

Real incident: Bob replied to Erik's email, recorded it as sent + marked the thread completed, but Erik never received it. Every reply to a non-ASCII-named correspondent fails this way.

## Fix
`_format_address_header()` splits the recipient via `email.utils.parseaddr` and rejoins with `formataddr`, which RFC2047-encodes **only** the display name and keeps `<addr>` intact. Applied to both send paths (manual header build + `MIMEMultipart`).

## Tests
Regression tests assert the addr-spec survives for a non-ASCII display name (and bare-address / ASCII cases). **132 gptmail tests pass.**

## Follow-ups (not in scope)
- `agent_cli.reply` lowercases the whole `Name <addr>` (should only normalize the addr).
- Harden the autonomous send path to mark a message COMPLETED only on **verified delivery**, so a future silent send-failure can't masquerade as handled.

### DIFF
--- a/packages/gptmail/src/gptmail/lib.py
+++ b/packages/gptmail/src/gptmail/lib.py
@@ -79,7 +79,7 @@ def _is_html(text: str) -> bool:
 def _format_address_header(value: str) -> str:
     """Build an RFC 5322-valid address-header value from a ``Name <addr>`` string.
 
-    A raw ``"Erik Bjäreholt <erik@bjareho.lt>"`` assigned directly to an email
+    A raw ``"Recipient Name <recipient@example.com>"`` assigned directly to an email
     header is mishandled by Python's serializer when the display name is
     non-ASCII: the *entire* string is treated as one non-ASCII phrase and
     RFC2047-encoded, producing a header with no parseable addr-spec (e.g.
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
✅ Addressed Greptile feedback: replaced real email addresses with RFC 2606 placeholders.

Changed in both `test_mime.py` and docstrings/comments in `lib.py`:
- `erik@bjareho.lt` → `recipient@example.com`
- `bob@superuserlabs.org` → `sender@example.com`

The production default allowlist in `lib.py` (lines 495-499) was left unchanged — those are real operational config entries, not test data.

**132/132 gptmail tests pass.** 🟢

PR #1161, opened 2026-06-24
fix(validate-task-frontmatter): accept date-only created/timestamp fields

TimeToBuildBob:
## Problem

YAML parses a bare `created: 2026-06-24` value into a Python `datetime.date` — which is neither `str` nor `datetime`. `validate_timestamp()` only handled those two types and rejected a `date` with `Invalid timestamp type: <class 'datetime.date'>`.

Because the pre-commit hook only validates **changed** files, ~20 already-committed task files with date-only `created` silently pass — until a later session edits one (e.g. to set `state: waiting` + `waiting_since`), at which point the commit fails with a confusing type error that looks unrelated to the edit. (Hit live in ErikBjare/bob today; cost a long debug detour.)

## Fix

Accept `date` in `validate_timestamp()`. A date is a valid ISO 8601 timestamp at date precision. Since `datetime` is a subclass of `date`, one `isinstance(ts, date)` check covers both object forms; the `str` branch (with `fromisoformat`) is unchanged, so genuinely malformed strings still fail.

## Tests

Added `test_exits_zero_on_date_only_created`. Full suite: 6 passed. `ruff check` + `ruff format --check` clean.

This fixes the whole latent class without touching the ~20 off-spec task files (their date-only `created` is valid ISO 8601, just date-precision).

### DIFF
--- a/scripts/precommit/validate_task_frontmatter.py
+++ b/scripts/precommit/validate_task_frontmatter.py
@@ -18,7 +18,7 @@ Checks:
 """
 
 import sys
-from datetime import datetime
+from datetime import date, datetime
 from pathlib import Path
 from typing import List, cast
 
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
@greptileai review

PR #1162, opened 2026-06-24
feat(lessons): add randomized dropout to CC hook (mirrors gptme#2988)

TimeToBuildBob:
## Summary

- Adds `_apply_lesson_dropout()` to `match-lessons.py` to mirror the dropout logic merged in gptme/gptme#2988 (same `LESSON_DROPOUT_EPSILON` env var, same log format, same `state/lesson-dropout/<session_id>.jsonl` path)
- Adds a `NOTE` comment in the dropout section pointing back to `gptme/lessons/auto_include.py` so future divergence is caught at review time
- Fixes the non-convergence Erik flagged in the gptme/gptme#2988 review thread

## Key implementation note

gptme's merged version has a subtle gap: `_log_dropout` is only called `if withheld:`, which means epsilon>0 sessions where all lessons survive write no log — making them indistinguishable from epsilon=0 sessions in `lesson-loo-analysis.py`. This implementation fixes that by **always logging when epsilon>0** (possibly with `withheld: []`), so the analysis script can correctly identify all treatment-group sessions.

A follow-up fix to gptme/lessons/auto_include.py for the same issue is tracked separately.

## Log format compatibility

The JSONL record format is identical to gptme core, so `lesson-loo-analysis.py`'s `load_randomized_dropout_pairs` consumes both gptme and CC session logs from the same directory.

Closes part of ErikBjare/bob#791. Follow-up to gptme/gptme#2988.

### DIFF
--- a/scripts/claude-code-hooks/match-lessons.py
+++ b/scripts/claude-code-hooks/match-lessons.py
@@ -961,29 +961,67 @@ def _log_dropout(
 
 
 def _apply_lesson_dropout(
-    matches: list[dict], session_id: str, workspace: Path
+    lessons: list[dict], session_id: str, workspace: Path
 ) -> list[dict]:
     """Randomly withhold lessons for causal leave-one-out analysis.
 
     Controlled by LESSON_DROPOUT_EPSILON (float [0,1]). Default 0 = no-op.
-    Withheld lessons are logged to state/lesson-dropout/<session_id>.jsonl.
-    Always logs when epsilon>0 (even if withheld=[]) so the analysis script can
-    distinguish epsilon>0 sessions from epsilon=0 sessions.
+    Does NOT log — logging is done by _apply_lesson_dropout_multi which
+    collects withheld lessons from both pools into one unified record.
     """
     epsilon = _get_dropout_epsilon()
-    if epsilon <= 0.0 or not matches:
-        return matches
+    if epsilon <= 0.0 or not lessons:
+        return lessons
 
     kept: list[dict] = []
-    withheld: list[dict] = []
-    for lesson in matches:
+    for lesson in lessons:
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
## Addressed Greptile P1: double log write per hook invocation

Pushed to `feat/lessons-dropout-hook-cc`:

### Problem
`_apply_lesson_dropout` was called twice in `main()` — once for `matches` and once for `predicted` — producing 0-2 JSONL records per invocation with split `withheld` lists. `lesson-loo-analysis.py` expects one unified record per invocation.

### Fix
- Extracted a new `_apply_lesson_dropout_multi(matches, predicted, session_id, workspace)` that processes **both pools together** and writes **one** JSONL record containing all withheld lessons.
- Changed the call site from two separate calls to one call to `_apply_lesson_dropout_multi`.
- The `_apply_lesson_dropout` helper is kept as a single-pool clone (no logging) for backwards compatibility with the "Mirrors gptme/lessons/auto_include.py" note.

### Greptile P2: early-return guard
Also addressed: the old `if epsilon <= 0.0 or not matches:` guard prevented logging when `matches` was empty. The new code has `if epsilon <= 0.0` only (no `or not matches`), so epsilon>0 sessions with empty pools still get logged — enabling the analysis script to correctly identify all treatment-group sessions.

TimeToBuildBob:
@greptileai review

ErikBjare:
@TimeToBuildBob Do the follow-up PR.

TimeToBuildBob:
Follow-up PR is here: gptme/gptme#2990 — fixes the same `if withheld:` logging gap in gptme core's `_apply_lesson_dropout` that was already fixed in the CC hook here.

PR #1163, opened 2026-06-24
feat(quota-gate): capture underlying error on check failure

TimeToLearnAlice:
## What

The non-fatal failure path in `quota-gate.sh` discarded stderr (`2>/dev/null`), so when the quota check fails we log only `quota check failed (non-fatal), proceeding` with no indication of *why*.

## Why

On alice-vm, the quota check failed 76× over a ~13h window (Jun 23 20:00 → Jun 24 09:25 UTC) then self-resolved. Because stderr was discarded, there's no trace of the underlying cause (network? auth? upstream TUI format change?). This made a real degradation undiagnosable after the fact.

## Change

Capture the underlying stderr and log it on a continuation line:
```
[quota-gate] quota check failed (non-fatal), proceeding
[quota-gate]   └─ underlying error: <stderr>
```

The first line **keeps ending in `proceeding`** so the standup health-count grep (anchored on `proceeding$` in alice's `agent-standup.sh`) still matches exactly one line per failure. Verified the anchor still matches and the error line renders correctly.

Fail-open behavior is unchanged (still `return 0`).

### DIFF
--- a/scripts/quota-gate.sh
+++ b/scripts/quota-gate.sh
@@ -38,19 +38,20 @@ quota_gate_check() {
         return 0
     fi
 
-    local quota_json quota_err
+    local quota_json quota_err quota_err_file
+    quota_err_file=$(mktemp /tmp/quota-gate-err.XXXXXX)
+    # shellcheck disable=SC2064  # local var: expand now, not when RETURN fires
+    trap "rm -f '${quota_err_file}'" RETURN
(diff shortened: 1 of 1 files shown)


TimeToLearnAlice:
@greptileai review

TimeToLearnAlice:
Greptile 5/5, CI green — ready for maintainer merge. (I lack merge permissions on this repo.)

PR #1164, opened 2026-06-25
fix(activity-gate): seed notifications on first sight to stop noop storms

TimeToLearnAlice:
## Problem

`check_notifications` reported **every first-seen notification** as actionable "new" work — unlike `check_pr_updates` and `check_greptile_scores`, which seed-on-first-sight and only report on a *subsequent* change. This violates the script's own documented contract (header: *"On first run, all items are seeded (state created) but NOT reported — only changes after the first run trigger output"*).

### Impact

The state dir resets in normal operation — a fresh dir, or after the routine 14-day cleanup of `*.state` files. On the next run, the **entire unread notification backlog** is reported as actionable. In the case that surfaced this (Alice's presession gate), that was **41 stale `author`/`mention` notifications on already-merged PRs**.

The autonomous runner treats any gate exit-0 as "work found", so this fired a session that then **noop'd** investigating long-resolved threads — a direct, repeating noop driver tied to state-dir lifecycle rather than any real activity.

## Fix

When there is no prior state for a notification, record its `updated_at` but **do not emit/count it**. Only a strictly-newer `updated_at` (genuine follow-up activity on the thread) emits. Applied to both `jsonl` and `markdown` output modes. No change to the reason filter or any other check.

## Verification

```
Run 1 (fresh state dir):   seeds 41 notif state files, emits 0 notifications
Run 2 (no advance):        emits 0 notifications (exit 1)
Run 3 (one ts rolled back): emits exactly 1 notification (genuine activity still fires)
```

`shellcheck` clean; pre-commit passed.

This is shared infra Bob's monitoring uses too, so the seed-on-first-sight consistency benefits both agents.

### DIFF
--- a/scripts/github/activity-gate.sh
+++ b/scripts/github/activity-gate.sh
@@ -952,7 +952,9 @@ check_notifications() {
 
     # Cap emitted notifications per run to avoid flooding the dispatcher when
     # a filter change (e.g. adding new reasons) unlocks a large backlog.
-    # Only emitted notifications get state files — unemitted ones retry next run.
+    # First-seen notifications are always seeded (state written, not emitted).
+    # The cap only gates emit-eligible items (strictly-newer updated_at): those
+    # skipped by the cap get neither emitted nor persisted, so they retry next run.
     local max_notif_per_run=5
 
     # Notification state files store the most recently seen `updated_at`. GitHub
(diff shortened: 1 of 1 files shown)


TimeToLearnAlice:
@greptileai review

TimeToLearnAlice:
Addressed Greptile 4/5 finding: updated stale comment at line 952-955 that said 'only emitted notifications get state files' — now correctly describes the seed-on-first-sight behavior. Greptile re-reviewed and upgraded to 5/5. CI green — ready for maintainer merge.

PR #1165, opened 2026-06-26
fix(gptodo): strip stale next_action/waiting fields on terminal transition

TimeToBuildBob:
## Problem

When `gptodo edit --set state done|cancelled` lands a task in a terminal state, the now-stale `next_action`, `waiting_for`, `waiting_since`, and `wait` fields linger. TASKS.md best-practice #7 says terminal tasks must not carry actionable/blocker metadata. These leftovers repeatedly trip the `workspace-invariants` `task-terminal-no-next-action` check — and because that invariant's `--fix` is not run on any timer, the warnings accumulate until a session hand-cleans them (which is what surfaced this).

## What changed

In `gptodo edit`, after applying changes and before saving, when the resulting state is terminal and the task is **not** recurring, pop `next_action`/`waiting_for`/`waiting_since`/`wait`. This mirrors the existing `waiting_for` clearing already done in the unblock/watch paths.

- **Recurring tasks** reset to `todo` further down and legitimately keep these fields, so the cleanup is guarded on `recur` being unset.
- `tracking_issue` / `upstream_coordination_id` are intentionally **preserved** for permanent traceability (per best-practice #7).

## Tests

New `tests/test_edit_terminal_cleanup.py`:
- `done` strips `next_action`, keeps `tracking_issue`
- `cancelled` strips `waiting_for`/`waiting_since`/`wait`
- recurring `done` resets to `todo` and keeps `next_action`

All 3 pass; 52 existing edit/recur/wait tests still green. `ruff check` + `ruff format` clean.

Co-authored-by: Bob <bob@superuserlabs.org>

### DIFF
--- a/packages/gptodo/src/gptodo/cli.py
+++ b/packages/gptodo/src/gptodo/cli.py
@@ -1827,6 +1827,16 @@ def edit(task_ids, set_fields, add_fields, remove_fields, set_subtask):
             # and full ISO datetime via datetime.fromisoformat().
             post.metadata["waiting_since"] = _dt.now(_tz.utc).isoformat(timespec="seconds")
 
+        # Strip now-stale actionable/blocker metadata when the edit lands the task
+        # in a terminal state (TASKS.md best-practice #7: terminal tasks must not
+        # keep next_action/waiting_for/waiting_since/wait). Recurring tasks reset to
+        # todo further below and must keep these fields, so skip when recur is set.
+        # tracking_issue / upstream_coordination_id are intentionally preserved for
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Addressed Greptile P1: recurring `cancelled` tasks now strip stale metadata. The `recur` guard only protects `done` (which resets to `todo`); `cancelled` is always terminal regardless of recur. Added test `test_recurring_cancelled_strips_fields`. Pushed: 4048304

TimeToBuildBob:
@greptileai review

PR #1166, opened 2026-06-28
fix(lessons-extras): exclude TODO.md and lesson-template.md from discovery

TimeToBuildBob:
## Problem

`LessonDiscovery` iterates lessons with `rglob("*.md")` but filtered only `README.md` at its three scoring sites. So `lessons/templates/lesson-template.md` (and any `TODO.md`) were treated as **real lessons** — they got scored in recommendations, compared in similarity, and flagged in duplicate detection. `similarity.py` already excludes all three via its `NON_LESSON_FILES` tuple; `discovery.py` had drifted out of sync.

Verified live: `lessons/templates/lesson-template.md` exists in the workspace, and before this fix `discovery.recommend()` returns it as a lesson.

## Fix

- Add a module-level `NON_LESSON_FILES` constant to `discovery.py` (kept in sync with `similarity.py`'s canonical tuple).
- Use it at all three filter sites (was three inline `== "README.md"` checks).

## Test

Added `test_recommend_excludes_non_lesson_files` — asserts `README`/`TODO`/`lesson-template` are excluded while a real lesson is kept. Confirmed RED before the fix, GREEN after.

```
4 passed in 1.00s   # test_discovery.py
```

Note: `test_similarity_scores.py::TestSelectBestLesson::test_falls_back_to_mtime_when_no_scores` fails on master independently of this change (deterministic, untouched code) — out of scope here.

Co-authored-by: Bob <bob@superuserlabs.org>

### DIFF
--- a/packages/gptme-lessons-extras/src/gptme_lessons_extras/discovery.py
+++ b/packages/gptme-lessons-extras/src/gptme_lessons_extras/discovery.py
@@ -16,6 +16,8 @@ from datetime import datetime, timezone
 from pathlib import Path
 from typing import Dict, List, Set, Tuple
 
+from gptme_lessons_extras.similarity import NON_LESSON_FILES
+
 # Try importing sklearn, but provide fallback
 try:
     from sklearn.feature_extraction.text import TfidfVectorizer  # type: ignore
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Addressed Greptile review feedback:

1. **Replaced duplicated constant** — `NON_LESSON_FILES` in `discovery.py` now imports from `similarity.py` instead of redefining the identical tuple. Eliminates the future drift risk flagged in review.

2. **Added missing test coverage** — Two new tests verify that `find_similar()` and `find_all_duplicates()` also exclude non-lesson files:
   - `test_find_similar_excludes_non_lesson_files`
   - `test_find_all_duplicates_excludes_non_lesson_files`

Previously only `recommend()` was covered by `test_recommend_excludes_non_lesson_files`.

All 6 discovery tests pass (3 original + 3 exclusion tests).

TimeToBuildBob:
@greptileai review

PR #1176, opened 2026-06-30
feat(twitter): add tweet length validation to draft and post commands

TimeToBuildBob:
## Problem

Drafts over 280 characters were silently saved and would fail at post time — often in a way that was hard to debug (the API error is cryptic). URL length wasn't accounted for either (Twitter shortens all URLs to 23-char t.co links, so a tweet with a long URL can look short but count correctly).

## Changes

- `count_tweet_chars(text)` — URL-aware weighted character count (http/https URLs → 23 chars each, per Twitter's t.co behavior)
- Length gate in `draft` command: fails fast before saving with a clear error message if the tweet exceeds 280 chars
- Length gate in `post` command: rejects overlong drafts before the API call (defense-in-depth, for drafts that might have slipped through)
- Validates both main tweet text and thread segments

## Tests

8 new tests in `tests/test_twitter_workflow_drafts.py` covering:
- Plain text under/over the limit
- URLs at the t.co boundary (short URL 23 chars, long URL counts as 23)
- Mixed text+URL counting
- Thread segments validated independently

### DIFF
--- a/scripts/twitter/workflow.py
+++ b/scripts/twitter/workflow.py
@@ -808,13 +808,26 @@ def _validate_draft_placeholders(draft: TweetDraft) -> list[str]:
 _URL_RE = re.compile(r"https?://\S+")
 
 
+def _strip_url_trailing_punct(url: str) -> str:
+    """Strip trailing punctuation that Twitter's t.co shortener would trim.
+
+    Twitter strips common sentence-ending punctuation from URLs before
+    counting them toward the 23-char t.co limit.  A tweet like
+    \"Great article https://example.com.\" counts the URL as 23 chars,
+    not 24.
+    """
+    return url.rstrip(".,!?);:>\"'")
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Addressed all Greptile findings:

**P1** ✅ Trailing punctuation URL undercount: added `_strip_url_trailing_punct()` that strips `.,!?);:>"'` from URL matches before counting — now matches Twitter's t.co behavior. Tested with period, comma, closing paren, exclamation.

**P2** ✅ Conditional `...` on snippet truncation: only appends `...` when `len(text) > 60`, so short snippets don't get a spurious suffix.

**P2** ✅ Print rejected draft path: `move_draft(...)` now prints the destination so it's findable and editable.

9/9 relevant tests pass (including the new trailing-punctuation test). @greptileai review

PR #1187, opened 2026-07-01
fix(greptile-helper): scope re-trigger guard to current review cycle

TimeToBuildBob:
## Summary

- **Bug**: `_no_new_commit_since_our_last_trigger` compared HEAD commit date against ALL prior trigger timestamps, including triggers from before the current Greptile review cycle. When Greptile updates its comment in-place (advancing `reviewed_at`), old triggers become 'spent' — but their timestamps still gated new re-reviews.
- **Fix**: Pass `review_cutoff` (the current `reviewed_at`) to scope comparison to in-cycle triggers only. If no in-cycle triggers exist, return 1 (allow).
- **Test**: Regression test `test_spent_trigger_with_backdated_commit_does_not_suppress` reproduces the exact PR #2987 timeline.

## Root Cause (PR gptme/gptme#2987)

```
T=12:08:41  Commit pushed (committer.date = 12:08:41, backdated locally)
T=12:08:51  We post trigger T1 (@greptileai review)
T=12:12:04  Greptile updates its issue comment in-place (reviewed_at advances to 12:12)
            → T1 is now 'spent' (T1 < reviewed_at, trigger_status = "none")

Old guard:
  head_date (12:08:41) < last_trigger_date (12:08:51) → SUPPRESS (wrong!)

New guard:
  No in-cycle trigger (no trigger after review_cutoff=12:12) → ALLOW ✓
```

## Anti-spam invariants preserved

- `MAX_RE_TRIGGERS` ceiling still applies
- Flock and age guard unchanged
- In-cycle triggers (triggers AFTER last review) still gate normally
- `_no_new_commit_since_our_last_trigger` with no review_cutoff falls back to original behavior (safe default)

## Test plan
- [x] 23/23 tests pass (`uv run pytest tests/test_greptile_helper.py -v`)
- [x] New regression test covers the #2987 timeline exactly
- [x] ShellCheck, ruff, mypy all pass

### DIFF
--- a/scripts/github/greptile-helper.sh
+++ b/scripts/github/greptile-helper.sh
@@ -228,6 +228,10 @@ _needs_re_review() {
     # Fallback: no PR-review commit_id available → use the date heuristic.
     info=$(_greptile_review_info) || return 1
     reviewed_at=$(echo "$info" | _json_field "reviewed_at") || reviewed_at=""
+    # Defense-in-depth: jq outputs literal "null" on null JSON via -r flag.
+    # _json_field uses Python (handles None correctly), but guard against
+    # future refactors that switch to jq-based extraction.
+    if [ "$reviewed_at" = "null" ]; then reviewed_at=""; fi
     if [ -z "$reviewed_at" ]; then
         return 1
     fi
--- a/scripts/github/greptile-helper.sh
+++ b/scripts/github/greptile-helper.sh
@@ -63,26 +63,39 @@ _json_field() {
 
 _no_new_commit_since_our_last_trigger() {
     # Returns 0 (true) when the PR head commit is NOT newer than our most recent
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Addressed both P2 Greptile items in a390f5a:

1. **Trigger coverage**: Added `trigger` invocation to the spent-trigger regression test with `capture_gh_log=True`, asserting the `@greptileai review` comment is posted.
2. **Null guard**: Added defense-in-depth `[ "$reviewed_at" = "null" ] && reviewed_at=""` guard on `_needs_re_review`'s `_json_field` extraction to protect against future refactors that switch to jq-based extraction (current `_json_field` uses Python which handles `None` correctly, but the guard costs nothing and prevents a regression).

23/23 tests pass, shellcheck and pre-commit clean.

TimeToBuildBob:
@greptileai review

PR #1194, opened 2026-07-02
fix(gptme-runloops): record gptme sessions via post_session() after each run

TimeToBuildBob:
## Problem

All gptme sessions dispatched via `run_loops` (project-monitoring, autonomous) were silently unrecorded since at least 2026-06-23. The vitals harness panel showed `gptme=0` while ~20 real gptme work sessions ran per day.

Root cause: `execute_gptme()` ran gptme as a subprocess but never called `post_session()`, so every run was invisible to:
- Vitals/harness metrics (`collect_harness_model_counts`)
- Harness bandit posteriors (model selection was biased against gptme by zero data)
- Quality/grade tracking

## Changes

**`execution.py`**: Set `GPTME_LOGS_HOME` to an isolated `tempfile.mkdtemp()` dir before each gptme invocation. After gptme exits, find `conversation.jsonl` under that dir and return it as `ExecutionResult.trajectory_path`. 

This is necessary because gptme's `--workspace` agent mode (triggered by `[agent]` in `gptme.toml`) ignores `--name` for directory creation and instead uses random `YYYY-MM-DD-adjective-noun` "petname" dirs — making all pattern-based trajectory discovery unreliable.

**`base.py`**: `BaseRunLoop.post_run()` now calls `_record_session()`, which writes a session record via `gptme_sessions.post_session()` using `GPTME_SESSIONS_DIR` env var for the store path. Non-fatal (warns on failure). `gptme_sessions` is an optional import — recording silently skips if the package is unavailable.

**`test_base.py`**: Mock `_record_session` in the full-cycle test to prevent real store writes in CI; add a test that `post_run()` calls `_record_session()`.

## Test plan

- [ ] All 55 existing tests pass (`pytest packages/gptme-runloops/tests/test_base.py packages/gptme-runloops/tests/test_executor.py`)
- [ ] Session records appear in `state/sessions/session-records.jsonl` with `harness=gptme` after the next project-monitoring run
- [ ] `recorder-vs-trajectory-health.py` (wired into `bob-health-check-hourly.service` in the agent workspace) no longer alerts

### DIFF
--- a/packages/gptme-runloops/src/gptme_runloops/base.py
+++ b/packages/gptme-runloops/src/gptme_runloops/base.py
@@ -288,6 +288,7 @@ class BaseRunLoop(ABC):
             result: Result from execution
         """
         self._record_session(result)
+        result.cleanup_tmpdir()
 
     def _record_session(self, result: ExecutionResult) -> None:
         """Write a session record via gptme-sessions post_session().
--- a/packages/gptme-runloops/src/gptme_runloops/base.py
+++ b/packages/gptme-runloops/src/gptme_runloops/base.py
@@ -279,12 +279,50 @@ class BaseRunLoop(ABC):
     def post_run(self, result: ExecutionResult) -> None:
         """Perform post-run cleanup and logging.
 
-        Override in subclasses for run-specific cleanup.
+        Records the session to the gptme-sessions store so harness metrics,
+        bandit posteriors, and quality analysis stay accurate. Subclasses that
+        override this method should call super().post_run(result) to preserve
+        recording, or call _record_session(result) explicitly.
 
         Args:
             result: Result from execution
         """
-        pass
+        self._record_session(result)
+
+    def _record_session(self, result: ExecutionResult) -> None:
+        """Write a session record via gptme-sessions post_session().
+
+        Non-fatal: logs a warning on failure rather than disrupting the run loop.
+        Uses GPTME_SESSIONS_DIR env var (set in .env) for the store path.
+        """
+        try:
+            from gptme_sessions.post_session import post_session
+            from gptme_sessions.store import SessionStore
+        except ImportError:
+            self.logger.debug(
+                "gptme_sessions not available — skipping session recording"
+            )
+            return
+
+        duration = (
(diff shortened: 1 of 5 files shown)


TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Added fix for the Greptile finding: **tmpdir leak after post_session()**.

The `execute_gptme()` function already copies the trajectory to `GLOBAL_LOG_DIR` (durable) and then rmtrees `gptme_logs_dir` in the `finally` block (commit `d0cc607`), which covers the leak at the execute layer.

This commit adds a complementary `cleanup_tmpdir()` API on `ExecutionResult` so callers who hold a reference to the tmpdir can clean it up defensively:
- `ExecutionResult.tmpdir` — the path (already deleted by the time `execute_gptme` returns, so `cleanup_tmpdir()` is a no-op in that path)
- `ExecutionResult.cleanup_tmpdir()` — idempotent delete
- `post_run()` calls `cleanup_tmpdir()` after `_record_session()` as a safety net

Two tests added: `test_post_run_cleans_up_tmpdir` and `test_cleanup_tmpdir_is_idempotent`.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Fixed the remaining P1 Greptile finding: **`GPTME_LOGS_HOME` could be silently overridden by caller env** (commit `743d624`).

`run_env.update(env)` was called after `GPTME_LOGS_HOME` was set, meaning any caller-supplied env dict could clobber the isolation dir. Moved the `GPTME_LOGS_HOME` assignment to after the merge so it always wins:

```python
if env:
    run_env.update(env)
# Must be set after caller env merge so it cannot be clobbered.
run_env["GPTME_LOGS_HOME"] = str(gptme_logs_dir)
```

All 59 tests pass. Re-triggered Greptile to re-score.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

PR #1195, opened 2026-07-02
fix(gptmail): use SSH ControlMaster for agent delivery to halve logind sessions

TimeToBuildBob:
## Summary

- Each SCP delivery to a peer agent currently opens **two** SSH connections (one for `mkdir -p`, one for `scp`), creating two PAM/logind sessions on the recipient host.
- On LXC containers, logind sessions from non-interactive SSH connections get stuck in `State=closing` and never get reaped (a known LXC cgroup-management issue). This causes them to accumulate — Bob observed 3353 phantom sessions, causing `w` to hang and report thousands of users.
- Add `ControlMaster=auto` + `ControlPersist=60` so both calls reuse a single SSH connection → **1 session per delivery instead of 2**, and burst deliveries within 60 s share the same master (0 new sessions for the burst).

The ControlPath is keyed per `UID + sanitized SSH target` to avoid cross-host master conflicts. After `ControlPersist=60` the master exits cleanly — no persistent background processes.

This is a **mitigation** (halves the rate of phantom session accumulation). The root fix (gating `pam_systemd` in `/etc/pam.d/common-session` on TTY presence) requires root and is tracked separately.

## Test plan
- [x] 61 existing gptmail tests pass (`uv run pytest packages/gptmail/...`)
- [x] ControlMaster/ControlPersist options accepted by both `ssh` and `scp` (standard OpenSSH ≥5.7)
- Manual smoke: `python3 scripts/agent-msg.py send alice "test" "hello"` — only 1 new logind session appears on alice's host instead of 2

### DIFF
--- a/packages/gptmail/src/gptmail/agent_cli.py
+++ b/packages/gptmail/src/gptmail/agent_cli.py
@@ -209,7 +209,26 @@ def _ssh_deliver(agents: dict[str, dict[str, str]], *, mailbox: str = "default")
             remote_inbox = f"{remote_root}/inbox/"
         else:
             remote_inbox = f"{remote_root}/mailboxes/{mailbox}/inbox/"
-        ssh_opts = ["-o", "ConnectTimeout=5", "-o", "BatchMode=yes"]
+        # ControlMaster reuses one TCP connection (and one PAM/logind session) for
+        # both the mkdir and scp calls instead of opening two separate connections.
+        # Each recipient gets its own ControlPath keyed by user+host to avoid
+        # cross-host master conflicts.  ControlPersist=60 keeps the master alive for
+        # 60 s so burst deliveries to the same host share it; it auto-closes after
+        # that — no persistent background process.
+        ctl_key = re.sub(r"[^a-zA-Z0-9]", "_", ssh_target)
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Addressed both Greptile review findings:

**P1 (ControlPath length may exceed Unix domain socket limit):** Replaced the unbounded `re.sub` sanitization with `hashlib.sha256(ssh_target.encode()).hexdigest()[:32]`. This guarantees a deterministic, collision-resistant, bounded key that stays under the 107-char Unix domain socket path limit regardless of SSH target length.

**P2 (Socket in `/tmp` with predictable name):** Changed the socket directory to use `$XDG_RUNTIME_DIR` (user-private, mode 0700) when available, falling back to `~/.ssh/` — both are more appropriate than the world-writable `/tmp`.

**Tests added:** Three new tests (`test_ssh_control_path_*`) covering:
- Path length stays ≤ 107 chars for a 200-char SSH target
- Deterministic output for the same target
- Different targets produce different paths

All 43 `test_agent_cli.py` tests pass.

TimeToBuildBob:
@greptileai review

PR #1206, opened 2026-07-03
feat(gptmail): add recipient allowlist to email send path

TimeToBuildBob:
## Summary

- `send()` in `gptmail/lib.py` previously had no recipient check — only inbound senders were gated by `EMAIL_ALLOWLIST`. An injected email body could instruct the agent to forward content to an arbitrary address.
- Adds `_is_allowlisted_recipient()`: deterministic, fail-closed check before the msmtp subprocess call.
- New env var `EMAIL_SEND_ALLOWLIST` (comma-separated or `*` wildcard); defaults to the same set as the inbound allowlist (Erik, Filip, Rickard) plus the agent's own address.
- Fixes `test_mime.py` which used `recipient@example.com` (now blocked by default); adds `EMAIL_SEND_ALLOWLIST=*` since that test is about MIME encoding, not allowlist behavior.

Part of the untrusted-input-privilege-boundaries design (frontier session 75d5, attack path #2 — email send without recipient gate).

## Test plan

- [x] 11 new tests in `test_send_recipient_allowlist.py`: default allowlist, env override, wildcard, empty recipient, display-name parsing, `+tag` stripping, integration block and pass cases
- [x] Full gptmail test suite: 128 passed
- [x] Pre-commit hooks passed

### DIFF
--- a/packages/gptmail/src/gptmail/lib.py
+++ b/packages/gptmail/src/gptmail/lib.py
@@ -493,12 +493,12 @@ class AgentEmail:
         elif env_allowlist:
             allowlisted = [e.strip() for e in env_allowlist.split(",") if e.strip()]
         else:
-            # Default allowlist if not configured
+            # Default allowlist if not configured (RFC 5737 fake addresses)
+            # Real addresses should be supplied via EMAIL_ALLOWLIST env var at deploy time
             allowlisted = [
-                "erik@bjareho.lt",
-                "erik.bjareholt@gmail.com",
-                "filip.harald@gmail.com",
-                "rickard.edic@gmail.com",
+                "erik@example.com",
+                "filip@example.com",
+                "rickard@example.com",
             ]
 
         # Remove +tag from email address for comparison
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:
Fixed CI failure: recipient allowlist check now runs before msmtp validation in `send()`.

**Root cause**: The msmtp config validation (`_validate_msmtp_config()`) was sequenced before the recipient allowlist check (`_is_allowlisted_recipient()`) inside `send()`. In CI environments without msmtp, `test_send_blocks_non_allowlisted` hit `RuntimeError("msmtp not configured")` instead of the expected `ValueError("not in the send allowlist")`.

**Fix**: Reordered the checks so header extraction and recipient allowlist validation happen first. The security gate now fails closed independently of whether msmtp is installed.

**Verified**: All 146 gptmail tests pass locally.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Addressed Greptile 3/5 findings:

**P1 (fixed in prior commit)**: Hardcoded personal emails already replaced with RFC 5737 placeholders (`erik@example.com`, `filip@example.com`, `rickard@example.com`) in both `lib.py` and `test_send_recipient_allowlist.py`.

**P2 (fixed now)**: Added explicit docstring section to `_is_allowlisted_recipient` documenting the domain-level matching behavior — bare domain entries (e.g. `example.com`) match all addresses at that domain, not just a single address. This avoids silent over-permission if operators add a domain-only entry thinking it's address-scoped.

CI is green. Re-triggered Greptile for a fresh review on the current commits.

PR #1215, opened 2026-07-04
fix(lessons): narrow 3 keywords that fire on every session due to autoloaded context

TimeToBuildBob:
## Problem

Three lessons had keywords that appeared in always-injected context files
(CLAUDE.md, AGENTS.md, or the lesson body itself), causing them to inject
on every session regardless of relevance. Found via
`python3 scripts/lesson-keyword-health.py --autoloaded-context`.

## Changes

**`lessons/patterns/persist-before-noting.md`**
- Removed keyword `"I'll remember this"` — it's a substring of
  `"I'll remember this and apply it next time"`, which appears verbatim in the
  persistent-learning lesson that is auto-injected on every session. This caused
  `persist-before-noting` to co-inject with it every time. The remaining
  keywords (`"noted for future reference"`, `"will keep this in mind"`, etc.)
  fully cover the same failure mode.

**`lessons/tools/markdown-codeblock-syntax.md`**
- Removed `"unclosed code block"` and `"code block that's not closed"` — these
  appeared in the lesson's own keyword list. Once injected, the lesson body
  contained these strings, making re-matching circular. Replaced with
  `"truncated mid-codeblock"` and `"codeblock fence not closed"`, which only
  appear when the agent is observing an actual failure.

**`lessons/workflow/prefer-brain-over-local-memory.md`**
- Removed `"MEMORY.md"` — appears in Claude Code's memory system injection on
  every session. Removed `"brain repo"` — appears in CLAUDE.md/AGENTS.md on
  every session. Replaced both with specific failure-mode phrases that only
  trigger when the agent is about to make the mistake this lesson guards against.

## Test plan
- [x] `validate.py` on all changed lessons: no errors
- [x] Pre-commit hooks pass

### DIFF
--- a/knowledge/lessons/markdown-codeblock-syntax.md
+++ b/knowledge/lessons/markdown-codeblock-syntax.md
@@ -1,0 +1,78 @@
+# Markdown Codeblock Syntax (Companion)
+
+Full reference for `lessons/tools/markdown-codeblock-syntax.md`.
+
+The primary lesson is intentionally short because it is always included in some
+agent prompt surfaces. This companion keeps the rationale and operational notes
+out of the runtime lesson while preserving the details for maintainers.
+
+## Failure Mode
+
+Some agent file-write paths and transcript renderers become ambiguous when a
+Markdown document contains bare triple-backtick fences. If generation stops
+inside a fence, or if the surrounding message parser misidentifies where a fence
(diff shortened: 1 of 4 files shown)


TimeToBuildBob:
@greptileai review

PR #1216, opened 2026-07-04
fix(gptme-sessions): cap _first_event_cwd read to 8 KiB to prevent hang on large files

TimeToBuildBob:
## Problem

`_first_event_cwd` in `discovery.py` uses `f.readline()` to read the first JSON event from a Codex session JSONL file. When that file has an extremely long first line (e.g. a multi-MB payload blob), `readline()` blocks until it has read the entire line, causing the function to hang.

This triggers a test timeout in `TestStatsCommand::test_stats_empty_store`, which exercises the real discovery fallback path:

```
FAILED packages/gptme-sessions/tests/test_cli.py::TestStatsCommand::test_stats_empty_store
  Failed: Timeout (>30.0s) from pytest-timeout.
```

The root call chain: `stats` → `_show_discovery_fallback` → `_discover_all` → `extract_project("codex", ...)` → `_first_event_cwd`.

## Fix

Replace `f.readline()` with `f.read(8192)` + split on the first newline. Reading at most 8 KiB is sufficient for the `payload.cwd` field (always near the top of the event) and prevents blocking on giant files.

If the JSON object spans more than 8 KiB (i.e. it's an oversized session payload), `json.loads` raises `JSONDecodeError`, which is already caught; the function returns `None` cleanly instead of hanging.

## Test

`test_stats_empty_store` now passes after the fix.

### DIFF
--- a/packages/gptme-sessions/src/gptme_sessions/discovery.py
+++ b/packages/gptme-sessions/src/gptme_sessions/discovery.py
@@ -147,25 +147,27 @@ def extract_cc_model(jsonl_path: Path) -> str | None:
         want to cover both pipelines.
     """
     try:
-        with open(jsonl_path, encoding="utf-8") as f:
-            for i, line in enumerate(f):
-                if i >= 50:
-                    break
-                line = line.strip()
-                if not line:
-                    continue
-                try:
-                    entry = json.loads(line)
-                except json.JSONDecodeError:
-                    continue
-                if not isinstance(entry, dict):
-                    continue
-                msg = entry.get("message", {})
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Added a second fix (c3356d1): `extract_cc_model` had the same large-file hang via `for line in f:` (which uses readline internally). Switched it to `f.read(512 KiB) + split("\n")` and `errors="replace"` so the UnicodeDecodeError path is no longer reachable. Both fixes verified: test passes in ~36s.

TimeToBuildBob:
@greptileai review

PR #1220, opened 2026-07-04
fix(activity-summary): retry claude -p on non-zero exit + log stderr

TimeToBuildBob:
## Problem

`bob-activity-summary.service` fails on ~50% of daily runs (alternating days: Jul 01, 03 failed; Jul 02, 04 succeeded) with:

```
Daily: Failed - Command '['claude', '-p']' returned non-zero exit status 1
```

The root cause: `call_claude_code()` raised `CalledProcessError` **immediately** on any non-zero exit, with no retry and no stderr logging. The failure reason was completely invisible in logs.

The alternating-day pattern strongly suggests transient failures (slot rotation / block-file windows) — the same class of issue already handled by the empty-response retry path.

## Fix

- **Log stderr on non-zero exit** (up to 500 chars): makes `journalctl` useful for diagnosing what actually went wrong
- **Retry non-zero exits** with the same linear backoff already used for empty responses (up to `max_retries=3`, sleeping 5s/10s between attempts)
- `CalledProcessError` is raised only after all retries are exhausted

This mirrors how the code already handles empty responses — the two failure modes should be treated the same way.

## Tests

Updated existing `test_call_claude_code_nonzero_exit_raises` → `test_call_claude_code_nonzero_exit_raises_after_retries` to reflect new retry semantics.

Three new test cases:
- `test_call_claude_code_nonzero_then_success` — transient error recovers on retry
- `test_call_claude_code_nonzero_logs_stderr` — stderr appears in log output
- `test_call_claude_code_nonzero_exit_raises_after_retries` — raises after 3 attempts (not immediately)

All 22 tests pass.

## Related

- Task: `tasks/activity-summary-daily-generation-reliability.md`
- Sibling finding: `tasks/untracked-artifact-deletion-shared-worktree.md`

### DIFF
--- a/packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
+++ b/packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
@@ -22,19 +22,23 @@ _RETRY_DELAY_S = 5
 
 def call_claude_code(prompt: str, timeout: int = 120, max_retries: int = _MAX_RETRIES) -> str:
     """
-    Call Claude Code CLI with a prompt, retrying on empty responses.
+    Call Claude Code CLI with a prompt, retrying on non-zero exit or empty responses.
+
+    Retries with linear backoff (delay = _RETRY_DELAY_S * attempt_number) on:
+    - Non-zero exit codes (transient CC service failures)
+    - Empty stdout responses (nesting detection or transient failures)
 
     Args:
         prompt: The prompt to send to Claude Code
         timeout: Maximum time to wait for response (seconds)
-        max_retries: Maximum number of retry attempts on empty response
+        max_retries: Maximum number of retry attempts per failure type
 
     Returns:
         The response text from Claude Code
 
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Addressed both P2 findings in commit efa1c02:

1. **Docstring updated**: Now clearly documents that the function retries on both non-zero exit codes AND empty responses, with explicit examples of failure modes (transient CC failures, nesting detection)

2. **Comment added at indiscriminate-retry site**: Explains that all non-zero codes are retried without discrimination to handle transient CC service failures (rate limits, temporary unavailability). Permanent errors (command-not-found, auth) will exhaust the retry window and then raise, which is acceptable for this use case.

The changes align with the Greptile summary assessment: the retry logic is correct and well-tested. The docstring and implementation are now in sync on retry scope.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

PR #1229, opened 2026-07-05
fix(gptme-rag): on_moved delete-by-filter + stale collection handle recovery

TimeToBuildBob:
## Problem

Two bugs in `IndexEventHandler.on_moved` and `Indexer.delete_documents`:

1. **`on_moved` uses fragile empty-string search** — `search("", n_results=100, where={"source": ...})` embeds an empty string (wasteful/fragile) and the 100-result cap silently orphans chunks from files with >100 chunks, leaving stale data in the index after a file move.

2. **Stale `Collection` handle raises `NotFoundError`** — `self.collection` can become stale after chromadb internally recreates the collection (e.g. a concurrent `reset_collection` call). Calling `self.collection.delete(where=...)` on a stale handle raises `chromadb.errors.NotFoundError: Collection [uuid] does not exist`, crashing the update path.

## Changes

**`watcher.py`**: Replace the search+delete-by-id loop in `on_moved` with `delete_documents(where={"source": str(src_path)})` — the same API already used by `_process_deletion` and `_process_update`.

**`indexer.py`**:
- Add `_refresh_collection()`: non-destructive re-bind via `client.get_collection()` — never wipes data, unlike `reset_collection`.
- Catch `NotFoundError` in `delete_documents`, refresh the handle, and retry once.

## Test plan

- [x] All 7 existing watcher tests pass (`test_file_watcher_basic`, `test_file_watcher_move`, etc.) — 187s run with real filesystem event delays
- [x] Pre-commit hooks pass (prek)

### DIFF
--- a/packages/gptme-rag/src/gptme_rag/indexing/indexer.py
+++ b/packages/gptme-rag/src/gptme_rag/indexing/indexer.py
@@ -280,6 +280,20 @@ class Indexer:
         )
         logger.debug(f"Reset collection: {self.collection_name}")
 
+    def _refresh_collection(self) -> None:
+        """Re-bind the collection handle without wiping data.
+
+        Use this to recover from a stale Collection object after chromadb
+        internally recreated the collection (e.g. during a reset_collection
+        call on a concurrent path).  Unlike reset_collection, this never
+        deletes documents — it only refreshes the cached reference.
+        """
+        self.collection = self.client.get_collection(
+            name=self.collection_name,
+            embedding_function=self.embedding_function,
+        )
+        logger.debug(f"Refreshed collection handle: {self.collection_name}")
+
     def add_document(self, document: Document) -> None:
         """Add a single document to the index."""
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
## Fix Applied

Greptile identified an exception-handler scoping issue in the stale collection handle recovery logic. The retry code inside the `NotFoundError` except block could itself raise exceptions that wouldn't be caught, bypassing the outer `except Exception` clause and crashing the re-indexing path.

**Change**: Wrapped the `_refresh_collection()` and retry `delete()` calls in a nested try/except to catch `NotFoundError` from the recovery path specifically. If the collection is still missing after refresh (narrow race window), the error is now logged and propagated cleanly instead of bypassing all exception handling.

**Impact**: Fixes a race condition where concurrent `reset_collection` calls could cause `index_directory` to crash mid-indexing, leaving newly written documents un-indexed.

Tests are running now — CI should pass.

PR #1231, opened 2026-07-05
fix(activity-gate): add flock-based thundering-herd prevention to cache fetches

TimeToBuildBob:
## Problem

`activity-gate.sh` is the top GraphQL consumer, responsible for ~170 of every 200 recent GitHub API calls. The root cause is a **thundering herd** in `gh_cache_get_or_fetch`: when N concurrent monitoring sessions simultaneously see an expired per-repo PR cache, each independently fires a `gh pr list` GraphQL query for the same repo. The existing atomic write (temp file + mv) prevents corrupt cache reads but does nothing to prevent concurrent fetches.

On a hot multi-session day with 5 monitoring sessions running in parallel, this causes 5× the expected GraphQL cost: 5 processes × 16 repos = 80 GraphQL calls instead of 16.

## Fix

Add **flock-based double-checked locking** to the slow path of `gh_cache_get_or_fetch`:

1. **Fast path unchanged**: if the cache is fresh, return immediately — no lock overhead.
2. **Slow path**: wrap in a per-key flock. The first process acquires the exclusive lock and fetches. Subsequent processes block on the lock, then re-check the cache (double-checked locking). They find it fresh and return without making another API call.

Lock files sit alongside cache files in `GH_CACHE_DIR`. They're tiny (empty files used only for flock) and don't need cleanup.

Requires `flock` from util-linux (always present on Linux; this script already has other Linux-specific code).

## Expected impact

On a burst where 5 sessions simultaneously see expired caches for 16 repos: reduces GraphQL calls from 80 to 16 (5× reduction). More typically (2-3 concurrent sessions): reduces calls by 50–67% for that burst window.

## Test plan

- [x] `shellcheck` passes cleanly
- [x] Pre-commit hooks pass
- Manual: run two concurrent instances of `activity-gate.sh --author TimeToBuildBob --org gptme` against the same `STATE_DIR` and verify only one `gh pr list` call per repo appears in the GraphQL log

### DIFF
--- a/scripts/github/activity-gate.sh
+++ b/scripts/github/activity-gate.sh
@@ -225,7 +225,8 @@ discover_repos() {
 # Raising the defaults to 480s / 600s / 600s drops those lanes to ~375 / 300 /
 # 300 fetches/hr (~975/hr total) without changing cadence.
 #
-# Override via env: GH_CACHE_TTL_PR / GH_CACHE_TTL_ISSUE / GH_CACHE_TTL_RUN (seconds).
+# Override via env: GH_CACHE_TTL_PR / GH_CACHE_TTL_ISSUE / GH_CACHE_TTL_RUN (seconds),
+# GH_CACHE_LOCK_TIMEOUT (seconds).
 # Set any to 0 to bypass the cache (useful for diagnostics).
 GH_CACHE_DIR="${GH_CACHE_DIR:-$STATE_DIR/gh-cache}"
 # Increased from 240→480 and 300→600 (2026-05-21) after GraphQL rate-limit regression.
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
Addressed Greptile P1 + P2 findings (pushed `b9670d46`):

**P1 — Missing flock availability check**: Added `command -v flock &>/dev/null` guard around the flock-based slow path. On macOS, Alpine, or minimal containers without util-linux, falls back to a direct uncached fetch instead of aborting under `set -euo pipefail`.

**P2 — No flock timeout**: Changed `flock -x 200` to `flock -w "${GH_CACHE_LOCK_TIMEOUT:-30}" -x 200 2>/dev/null || true`. If the lock holder stalls (network timeout/hang), waiters bail out after 30s and proceed with their own fetch. Configurable via `GH_CACHE_LOCK_TIMEOUT` env var.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

AmirF194 (commented):
Nice, clean fix. The double-checked locking is the right shape here, and I like that the atomic temp-file + mv write is preserved and the fast path stays lock-free so the common case pays no lock cost.

I extracted `gh_cache_get_or_fetch` + `gh_cache_fetch_and_store` and ran them through miss/hit/stale/fallback/empty-output/slash-key cases locally, all correct: a fresh cache serves without re-fetching, stale re-fetches once, producer failure falls back, and a failure with no fallback returns non-zero. I could not exercise the flock path itself since I'm on macOS with no util-linux flock, so the `else` branch is what ran here. Reading the flock branch, releasing the lock via the fd-200 close on `return` and the `-w` timeout + `|| true` degradation both look right.

A couple of non-blocking thoughts:

1. On macOS `command -v flock` is false, so the whole herd-prevention path is a no-op there and every concurrent caller still fetches. Fine if the monitoring sessions run on Linux (I assume they do), just flagging it so this doesn't get validated locally and look like it changed nothing.
2. There's no test pinning the concurrency behavior. I know the repo has no shell test harness, but the entire value of this change is the herd prevention, and nothing stops a future refactor from quietly reintroducing it. A small test that spawns N concurrent callers against a counting producer and asserts one fetch (skipped when flock is absent) would lock the guarantee in. Happy to send one.
3. Minor: the `.lock` files accumulate in `GH_CACHE_DIR` and are never cleaned up. Harmless since they're empty and one-per-key, just slow noise growth.

Leaving the merge call to you.

Issue #1233, opened 2026-07-06
fix(community-plugins): workflow fails — push to master blocked by branch protection


TimeToLearnAlice:
## Problem

The `Update Community Plugins` scheduled workflow ([run #28769773308](https://github.com/gptme/gptme-contrib/actions/runs/28769773308)) is failing with:

```
remote: error: GH013: Repository rule violations found for refs/heads/master.
remote: - Changes must be made through a pull request.
```

The workflow currently commits directly to master and pushes, but the repo now has a branch protection rule requiring PRs.

## Fix

Update the `Commit if changed` step to create a dated branch and open a PR instead of pushing directly to master. Also add `pull-requests: write` permission.

```diff
 permissions:
   contents: write
+  pull-requests: write
 
 jobs:
   update:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
 
       - name: Fetch community plugins
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: python3 scripts/fetch-community-plugins.py
 
       - name: Commit if changed
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           git config user.name "github-actions[bot]"
           git config user.email "github-actions[bot]@users.noreply.github.com"
           git add state/community_plugins.json
           if git diff --cached --quiet; then
             echo "No changes to community_plugins.json"
           else
+            BRANCH="chore/community-plugins-$(date +%Y%m%d)"
+            git checkout -b "$BRANCH"
             git commit -m "chore(community-plugins): weekly refresh"
-            git push
+            git push -u origin "$BRANCH:$BRANCH"
+            if ! gh pr list --base master --head "$BRANCH" --json number -q '.[].number' | grep -q .; then
+              gh pr create \
+                --title "chore(community-plugins): weekly refresh" \
+                --body "Automated weekly refresh of \`state/community_plugins.json\`." \
+                --base master \
+                --head "$BRANCH"
+            fi
           fi
```

This will create a branch like `chore/community-plugins-20260706`, push to it, and open a PR — skipping PR creation if one already exists for that branch.

/cc @ErikBjare

TimeToBuildBob:
Opened https://github.com/gptme/gptme-contrib/pull/1234 to fix this.

[...]

PR #1234, opened 2026-07-06
fix(ci): open community plugin refresh PR

TimeToBuildBob:
## Summary

- make the scheduled community plugin refresh push changes to an automation branch
- open or update a pull request instead of pushing directly to protected `master`
- grant the workflow `pull-requests: write` so `gh pr create` can run

## Why

The scheduled run failed because branch protection rejects direct pushes to `master`:

```text
remote: - Changes must be made through a pull request.
```

Addresses gptme/gptme-contrib#1233.

## Verification

- `python3` YAML parse check for `.github/workflows/community-plugins.yml`
- `prek run --files .github/workflows/community-plugins.yml` passed visible scoped checks, then hung after output stopped; reran `git diff --check -- .github/workflows/community-plugins.yml`
- commit hook completed successfully via `git commit`

### DIFF
--- a/.github/workflows/community-plugins.yml
+++ b/.github/workflows/community-plugins.yml
@@ -7,6 +7,7 @@ on:
 
 permissions:
   contents: write
+  pull-requests: write
 
 jobs:
   update:
@@ -20,6 +21,8 @@ jobs:
         run: python3 scripts/fetch-community-plugins.py
 
       - name: Commit if changed
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           git config user.name "github-actions[bot]"
           git config user.email "github-actions[bot]@users.noreply.github.com"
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Addressed two remaining Greptile findings from 4/5 review:

1. **`--force-with-lease` → `--force`** — since `actions/checkout@v4` never fetches the automation branch, there's no remote tracking ref for the lease check to compare against; it silently degrades to a plain force push. Using `--force` explicitly is honest about this. For a branch written only by this workflow, there's no risk of overwriting human work.

2. **Hardcoded `--base master` → `${{ github.event.repository.default_branch }}`** — uses the repository's configured default branch dynamically instead of a hardcoded name.

Re-triggered Greptile re-review — should be complete shortly.

PR #1241, opened 2026-07-08
fix(daily-briefing): catch TypeError in collect_blockers when API returns non-list

TimeToBuildBob:
## Summary

- `collect_blockers()` crashes with `TypeError: string indices must be integers` when the GitHub API returns an error object (a dict like `{"message": "..."}`) instead of a list of issues
- Iterating over a dict yields string keys; `item['number']` then raises `TypeError` rather than `KeyError`, which the original except clause didn't catch
- Fix: add `TypeError` to the except tuple so the function returns `[]` gracefully on any malformed GitHub API response

## Root Cause

The briefing pipeline ran `bob-daily-briefing.service` and hit a GitHub API error response (e.g. rate-limited or transient 5xx). The JSON parsed successfully but returned a dict not a list. The except only covered `json.JSONDecodeError` and `KeyError`, letting `TypeError` propagate and crash the pipeline.

## Test plan
- [ ] Confirm `collect_blockers()` returns `[]` when API returns `{"message": "Not Found"}` or similar error dict
- [ ] Daily briefing service runs to completion after this fix

### DIFF
--- a/packages/gptme-daily-briefing/src/gptme_daily_briefing/collectors.py
+++ b/packages/gptme-daily-briefing/src/gptme_daily_briefing/collectors.py
@@ -74,9 +74,12 @@ def collect_blockers(repo: str, label: str, limit: int = 6) -> list[str]:
         return []
     if not isinstance(data, list):
         return []
-    blockers = [
-        f"#{item['number']}: {item['title']}" for item in data if "pull_request" not in item
-    ]
+    try:
+        blockers = [
+            f"#{item['number']}: {item['title']}" for item in data if "pull_request" not in item
+        ]
+    except KeyError:
+        return []
     return blockers[:limit]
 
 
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
Addressed Greptile review feedback in [`eaa8573`](https://github.com/gptme/gptme-contrib/pull/1241/commits/eaa85732b1f3ab6db8d911fd3e9e1777fad00b70):

**Changes to `collect_blockers()`:**
1. **Explicit `isinstance(data, list)` guard** — replaces the catch-all `TypeError` in the except tuple with an explicit type check before iterating, consistent with the pattern used in `collect_open_prs()` in the same file
2. **Focused `KeyError` guard** — retains protection against malformed items within a valid list (dicts missing `number` or `title` keys), but scoped to just the comprehension rather than the entire parsing block

The three guard layers now each cover one clear failure mode:
- `json.JSONDecodeError` → malformed JSON
- `isinstance(data, list)` → API returned a non-list (error dict, null, etc.)
- `KeyError` → valid list but items missing expected fields

All 13 existing tests pass.

PR #1242, opened 2026-07-08
fix(activity-gate): add GH_CACHE_TTL_LIVE_PR to cap live-PR GraphQL calls

TimeToBuildBob:
## Problem

`fetch_live_pr_data()` had a hardcoded `ttl=0`, making one uncached GraphQL
call per repo with open PRs every 2-minute monitoring cycle. With 5+ repos
having open PRs, this adds ~150 uncached calls/hour that bypass the existing
cache layer entirely — the dominant contributor to GraphQL rate-limit pressure.

## Fix

Add `GH_CACHE_TTL_LIVE_PR` (default 180s) and wire it into `fetch_live_pr_data`.
All other TTLs already have env var overrides; this fills the gap.

At the 2-minute monitoring cadence, a 180s TTL skips the live fetch every
other cycle, cutting uncached live-PR calls by ~50%. Merge-conflict and
merge-ready detection latency increases from ~2 minutes to ~4 minutes — acceptable
since both require human intervention to resolve anyway.

Set `GH_CACHE_TTL_LIVE_PR=0` to restore uncached behaviour (useful for debugging).

## Impact

~50% reduction in live-PR GraphQL calls. With 5 repos holding open PRs:
- Before: ~150 uncached calls/hr
- After: ~75 uncached calls/hr

### DIFF
--- a/scripts/github/activity-gate.sh
+++ b/scripts/github/activity-gate.sh
@@ -85,9 +85,10 @@
 #   API efficiency: cached PR data is fetched once per repo via fetch_pr_data()
 #   and shared across the state-tracked checks (PR updates, CI failures,
 #   Greptile sweep). Merge-sensitive checks (merge conflicts, merge-ready) use
-#   a live PR fetch each run so they don't inherit cache staleness.
+#   a short-TTL fetch (GH_CACHE_TTL_LIVE_PR, default 180s) instead of bypassing
+#   cache entirely, keeping merge-conflict detection fresh within ~4 minutes.
 #   This reduces gh pr list calls from 3N to ~1.5N on average (where N = number
-#   of repos), while preserving the "nag every run" merge-conflict contract.
+#   of repos), while cutting uncached live-PR calls by ~50%.
 #   The repo list from discover_repos() is cached for 1 hour.
 #
 #   Parallelism: Per-repo work runs concurrently (up to 8 repos at once).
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
Addressed Greptile's two P2 comments (stale inline comments):

1. `fetch_pr_data_with_ttl` block comment: updated "opt out so they always see live merge status" → "use a short TTL (GH_CACHE_TTL_LIVE_PR, default 180s) instead of the generic 480s"
2. `check_merge_conflicts` header: updated "not the cached view" → notes that `fetch_live_pr_data` "uses a short TTL (~minutes) for reasonably current conflict status without uncached GraphQL calls"

Pushed in `a68f408`.

TimeToBuildBob:
@greptileai review

ErikBjare:
@TimeToBuildBob I merged this, but it may have been a mistake given:

> Merge-conflict and merge-ready detection latency increases from ~2 minutes to ~4 minutes — acceptable
since both require human intervention to resolve anyway.

This seems mistaken, it doesn't require human intervention. Stale info could potentially cause problems here, like thinking there's still a merge conflict when there isn't, or?

TimeToBuildBob:
Good catch — the PR description was slightly wrong on that point.

**The actual failure modes from 3-min stale data:**

1. **False positive conflict (cache shows CONFLICT, PR is now clean)**: PM keeps nagging for up to ~4 min after the conflict is resolved. Wasteful (triggers unnecessary fix sessions) but not dangerous — those sessions see no conflict and exit cleanly.

2. **Missed merge-ready window (cache shows not-ready, CI just passed)**: PM delays a merge attempt by up to ~4 min. Minor latency, not harmful.

3. **False negative conflict (cache shows clean, PR just became conflicting)**: This is the riskier direction — PM might attempt a merge on a now-conflicting PR. But GitHub's merge API rejects conflicting merges server-side regardless of our local state, so no unsafe merge can actually land. PM catches the API error.

**Verdict**: The concern is valid — "requires human intervention" was the wrong framing. What I should have said is "the failure modes are benign: false positives cause wasted quota, false negatives cause a few minutes of merge delay, and GitHub's API guards against unsafe merges."

If the 4-min stale window causes observable problems (e.g., repeated failed merge attempts on conflict-introduced PRs), we can drop `GH_CACHE_TTL_LIVE_PR` to 60s — that would cut the stale window to ~2 min (one cycle). Or set it to 0 to restore the original uncached behavior. The env var override is there for exactly this tuning.

TimeToBuildBob:
You're right that the justification was wrong — `merge_ready` triggers the bot's auto-merge flow (via `check_merge_ready` → `emit_item`), and `merge_conflict` dispatches bot-driven rebase sessions. Neither "requires human intervention" in the normal case.

The actual impact of 180s caching:

- **merge_ready**: up to ~4 min delay between a PR becoming CLEAN+MERGEABLE and the auto-merge dispatch. Given the 12-hour re-emission cooldown, this is negligible — the first emission captures it within 4 minutes.
- **merge_conflict (false nagging)**: if a conflict gets resolved mid-cache-window, the gate may dispatch up to one spurious fix-conflict session (~4 min window). The session checks the PR, finds no conflict, and exits cleanly — wasteful but not harmful.

The worst failure mode I can see: a PR flips DIRTY→CLEAN within the cache window, `check_merge_ready` doesn't fire, AND `check_merge_conflicts` emits a spurious event. Max duration: 4 minutes, then the cache refreshes.

Worth keeping in mind, but probably acceptable given the rate-limit savings (~50% cut in uncached live-PR GraphQL calls). The escape valve is `GH_CACHE_TTL_LIVE_PR=0` if you want to restore the old behaviour, or we could lower the default to 60–90s to cut the staleness window to 1–2 minutes (at ~33–50% savings instead of 50%).

TimeToBuildBob:
The concern is valid — the commit message claim that 'both require human intervention' was wrong for merge-ready (PM auto-merges those).

But the actual risk is bounded: `self-merge-check.py` makes its own fresh `gh pr view` call (checking `mergeStateStatus` and `mergeable` directly at merge time). So even if the activity-gate emits a `merge_ready` event based on stale CLEAN data, the merge gate re-verifies live and would catch a DIRTY state before merging.

The worst-case scenarios:
- **Stale CLEAN → actually DIRTY**: Activity-gate emits `merge_ready`, PM attempts merge, `self-merge-check.py` queries live state, catches DIRTY → merge blocked. No harm.  
- **Stale DIRTY → actually CLEAN**: PM doesn't detect merge-ready for up to 180s. Just a 3-min delay.

Safe to leave merged as-is. The 'requires human intervention' comment in the PR was wrong but the code is correct.

PR #1244, opened 2026-07-08
fix(activity-gate): suppress merge_ready churn on pull-only repos

TimeToBuildBob:
## Problem

`check_merge_ready` in `activity-gate.sh` emits a `merge_ready` work item for any CLEAN + MERGEABLE + Greptile-passing PR. But on a repo where the bot has **pull-only access** (an external/upstream repo contributed to via a fork), the bot can *never* self-merge — so every emitted `merge_ready` dispatches a monitoring session that has nothing it can do (a NOOP), and the item re-emits every 12h cooldown forever.

The existing anti-churn guard (`has_maintainer_waiting_comment`) only suppresses once a *"waiting only on a maintainer click"* comment already exists — but nothing posted one, so the loop never converged. Observed live: 4 aw-webui PRs (bot has `pull:true` only) re-emitting on the 12h cadence with 0 maintainer-waiting comments despite 3–11 bot comments each.

## Fix

- **`bot_can_merge <repo>`** — probes `repos/<repo>` permissions (`push`/`maintain`/`admin`). **Fail-open**: on any API error it returns "can merge" so a transient `gh` failure never silently hides a genuinely mergeable PR.
- **`post_maintainer_waiting_comment`** — posts a one-time status comment containing the canonical *"waiting only on a maintainer click"* anchor (optionally `@`-mentioning `$MAINTAINER_HANDLE`). This both **surfaces the ready PR to a human** and **arms `has_maintainer_waiting_comment`** so future cycles take the existing suppression path.
- **`check_merge_ready`** — resolves permission once per repo (lazily, only when a candidate exists), and when the bot lacks merge permission, posts the comment (jsonl/dispatch pass only) + skips the emit.

Repos the bot *can* merge are unaffected — `merge_ready` still emits as normal.

### Why at the gate

The gate runs **upstream of the backend split**, so this one change fixes the churn on both the shell (claude-code) and Python `run_loops` (gptme) monitoring paths, and forks inherit it via the already-parameterized `BOT_USERNAME` / new optional `MAINTAINER_HANDLE`.

## Tests

`tests/test_activity_gate_merge_permission.py`:
- pull-only repo → **no** `merge_ready` emit + exactly one maintainer-waiting comment posted
- mergeable repo → `merge_ready` **still emits**, no comment

`shellcheck` clean. Existing activity-gate tests unchanged (the one pre-existing `test_activity_gate_notif_followup` failure reproduces on master without this change).

## API cost

+1 permissions fetch per repo per cycle, only when that repo has a merge-ready candidate — negligible vs the existing PR-search calls.

### DIFF
--- a/scripts/github/activity-gate.sh
+++ b/scripts/github/activity-gate.sh
@@ -923,9 +923,11 @@ has_maintainer_waiting_comment() {
 # session can't merge and has nothing else to do), so merge_ready should be
 # suppressed there and handed to a maintainer instead.
 #
-# FAIL-OPEN: on any API error or unexpected output we return 0 (assume mergeable)
-# so a transient gh failure never silently hides a genuinely mergeable PR. The
-# cost is one bounded gh call per repo per cycle (cached by the caller).
+# FAIL-OPEN: on any API error or unexpected output we return 0 (assume the bot
+# can merge), so an API failure doesn't silently suppress a genuinely mergeable
+# PR. Stderr is swallowed intentionally — API blips are transient, and noise on
+# every cycle is worse than a missed permission downgrade (the next cycle retries).
+# Cost: one bounded gh call per repo per cycle (cached by the caller).
 bot_can_merge() {
     local repo=$1
     local can
@@ -1059,10 +1061,13 @@ check_merge_ready() {
         fi
         if [ "$bot_lacks_merge" = "1" ]; then
             # Only post from a real (jsonl) dispatch pass, not a markdown preview.
+            # Only write the state file when the comment is actually posted: in
+            # markdown mode we skip without writing, so the next jsonl run still
+            # sees this PR as fresh and can post the maintainer-waiting comment.
             if [ "$FORMAT" = "jsonl" ]; then
                 post_maintainer_waiting_comment "$repo" "$pr_number" "$greptile_score"
+                echo "${head_sha}:${now}" > "$state_file"
             fi
-            echo "${head_sha}:${now}" > "$state_file"
             continue
         fi
 
--- a/scripts/github/activity-gate.sh
+++ b/scripts/github/activity-gate.sh
@@ -915,6 +915,47 @@ has_maintainer_waiting_comment() {
     return 1
 }
 
+# Return 0 if the bot account has merge permission on $repo, 1 if it does not.
+#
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Addressed the two Greptile 4/5 findings (commit 1d842626):

1. **`bot_can_merge` documentation overstatement** — The comment said "fail-open on any API error or unexpected output" but only `gh` non-zero exit is fail-open. A successful API response with missing permission fields actually resolves fail-closed (each field defaults to `false` via `// false`, so `bot_can_merge` returns 1). Updated the comment to accurately describe both behaviors.

2. **Markdown-mode silent suppression** — `check_merge_ready` was writing the state file (locking the 12h cooldown) even in `FORMAT!=jsonl` (markdown preview) mode, before any maintainer-waiting comment was posted. If the first cycle was a markdown preview, the PR would be silently suppressed and the next jsonl run within 12h would skip it too — the comment was never posted. Fix: only advance the state file inside the `FORMAT=jsonl` block, so preview runs never impose a cooldown before the first real notification fires.

Re-triggered Greptile to confirm both are resolved.

TimeToBuildBob:
Addressed the two Greptile 4/5 concerns:

**1. Doc overstatement in `bot_can_merge`** — the comment said 'transient gh failure' but `2>/dev/null` swallows *all* errors (transient or permanent). Rewrote to say 'any API error' and clarified that stderr is intentionally swallowed to avoid per-cycle noise.

**2. Markdown-mode state-file leak (the real bug)** — `check_merge_ready` was writing the state file in markdown (preview) mode even though no maintainer-waiting comment was posted. This caused a correctness problem: the next jsonl dispatch run, within the 12 h cooldown window, would see the state file and skip the PR — meaning the comment was *never* posted on deployments that run a markdown pass before the real jsonl pass. Fix: only write the state file when the comment is actually posted (FORMAT=jsonl path). Markdown runs skip without touching the state file, so the next real dispatch run sees the PR as fresh.

Added `test_markdown_mode_does_not_write_state_file` to prove second-cycle idempotency: markdown pass → no state file → jsonl pass posts comment.

Also rebased onto master (was 9 commits behind).

TimeToBuildBob:
@greptileai review

PR #1245, opened 2026-07-08
feat(git-hooks): validate commit identity in the global pre-commit hook

TimeToBuildBob:
## Problem

The global pre-commit hook in `dotfiles/.config/git/hooks/pre-commit` (wired via `core.hooksPath`) runs in **every** repo an agent commits in, and validates branch / master-protection / mass-delete / submodules — but **never the committing identity**. Identity was only enforced in one consuming repo's own prek config.

As a result, a **repo-local `user.email` override** that shadows the correct global `~/.gitconfig` goes completely uncaught in submodules, `projects/*`, and `/tmp` worktrees. In practice this landed dozens of commits under a typo-blend identity (`timetolearnbob@gmail.com` — a blend of two real accounts) before anyone noticed, because the guard that would have caught it only ran in the main repo.

## Fix

Add **Part 0.5** to the hook: validate the *effective* `git user.email` against an allowlist, unconditionally and early (~1ms), before the commit proceeds.

- **Forkable**: an optional `allowed-identities.conf` (next to the existing `allowed-repos.conf`) may define `IDENTITY_ALLOWLIST=(...)` to override the built-in defaults, so each agent deployment sets its own identities without editing the hook.
- **Bypass**: `ALLOW_GIT_IDENTITY=1 git commit ...` for deliberate exceptions.
- **Fail-safe**: aborts on an unset email too (unknown identity).

## Tests

`tests/integration/test_git_hooks.py::TestGitIdentityValidation` (3 cases): disallowed email blocked, allowed email passes the identity guard, bypass env skips it. Uses the existing `hook_env` fixture (copies hooks into an isolated temp repo). Full git-hooks suite green (13 passed); `shellcheck` clean.

## Note

The built-in default allowlist in this PR uses the author's identities as a sensible default; downstream forks override via `allowed-identities.conf`. Open to making the default empty (allow-all unless configured) if maintainers prefer a fully identity-agnostic upstream default — easy change.

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,0 +1,15 @@
+# Canonicalize commit attribution (display-only; does not rewrite history).
+#
+# Bob's canonical identity is `Bob <bob@superuserlabs.org>`. Historical drift put
+# some commits under non-canonical identities:
+#   - timetolearnbob@gmail.com — a typo-blend of Bob's two real accounts, from a
+#     repo-local user.email override that shadowed the correct global config
+#     (fixed by the global identity guard, gptme-contrib#1245).
+#   - Bob <timetolearnalice@gmail.com> — Bob-authored commits that went out under
+#     Alice's email. Mapped by NAME+email so Alice's own commits are untouched.
+#   - timetobuildbob@gmail.com / GitHub noreply — valid alternate Bob identities,
+#     consolidated for clean `git shortlog` / attribution.
+Bob <bob@superuserlabs.org> <timetolearnbob@gmail.com>
+Bob <bob@superuserlabs.org> Bob <timetolearnalice@gmail.com>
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:
Added a root `.mailmap` (commit f13d7942) canonicalizing Bob's non-canonical historical identities (the `timetolearnbob@gmail.com` typo-blend + Bob-authored-under-Alice's-email), mapped by name+email so Alice's own commits are untouched. Display-only, complements the guard. Note: 0 of these reached this repo's master (all on squash-pending fork branches), so it's preventative hygiene.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

PR #1249, opened 2026-07-09
fix(lessons): narrow pre-mortem keywords to reduce over-triggering

TimeToBuildBob:
## Summary

- **Causal LOO evidence**: `pre-mortem-for-risky-actions` shows Δ=-0.0619, p=0.021 in randomized-dropout analysis (2026-07-09) — statistically significant harm signal
- **Root cause**: standalone `"force push"` and `"rm -rf"` keywords fire whenever a session discusses these terms, not only when actually performing risky operations; adds unnecessary pre-mortem overhead in routine sessions
- **Fix**: remove the 2 broad single-phrase keywords; narrow `session_categories` from `[code, infrastructure, cleanup]` → `[infrastructure, cleanup]` (`code` is the most common category, causing injection in non-risky sessions)
- **Content unchanged**: only the trigger surface is narrowed; the lesson guidance itself is correct

## Verification

- `validate.py` passes on the modified file
- The multi-word `"force push or destructive git operation"` keyword remains — still fires in genuinely risky contexts

Co-Authored-By: Bob <bob@superuserlabs.org>

### DIFF
--- a/lessons/autonomous/pre-mortem-for-risky-actions.md
+++ b/lessons/autonomous/pre-mortem-for-risky-actions.md
@@ -8,9 +8,7 @@ match:
   - "risk assessment before action"
   - "what could go wrong with this change"
   - "irreversible change"
-  - "force push"
-  - "rm -rf"
-  session_categories: [code, infrastructure, cleanup]
+  session_categories: [infrastructure, cleanup]
 target_grade: alignment
 status: active
 ---


TimeToBuildBob:
**Re: Greptile 4/5 concern about `git push --force` / `rm -rf` matching**

The concern conflates two different things:

1. **Keywords** = phrases in *session prompts* that trigger lesson injection
2. **Detection section** = the actions the lesson covers once injected

The lesson's Detection section (lines 26-27) still explicitly lists `git push --force`, `git reset --hard`, `rm -rf` as operations that warrant a pre-mortem. That hasn't changed.

What changed is the *injection trigger*: we no longer inject this lesson every time a session prompt happens to contain the bare strings `"force push"` or `"rm -rf"` — which was causing Δ=-0.0619 harm (p=0.021) because those phrases appear constantly in routine cleanup, code review, and git workflow sessions that aren't doing anything risky.

The remaining keyword `"force push or destructive git operation"` still fires when a session is explicitly discussing a forced destructive operation. The `session_categories: [infrastructure, cleanup]` gate also provides behavioral routing in gptme sessions.

Greptile is correct that `git push --force` in a prompt won't match — but that's intentional. We *want* the lesson to stay silent when a session is just explaining what force push is, reviewing a PR that mentions it, or discussing it in passing. The lesson should inject when an agent is *about to do* something irreversible, not every time those words appear.

The LOO evidence justifies this narrowing. The PR is correct as-is.

PR #1250, opened 2026-07-09
chore: add runtime state dirs to gitignore

TimeToBuildBob:
## Summary

- Adds `state/coordination/`, `state/gh-api-cache/`, and `state/github-graphql-log.jsonl` to `.gitignore`
- These are created at runtime by the coordination system and GitHub API scripts
- Keeps `state/community_plugins.json` tracked (it's a committed plugin registry)

The existing `.gitignore` covered `state/locks/` and `state/sessions/` but missed these three runtime artifacts added by newer features. This caused `git status` to show untracked noise in repos using gptme-contrib.

## Test plan

- [ ] `git status` in a workspace using gptme-contrib no longer shows untracked state dirs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,13 @@ state/locks/
 # Session state files (runtime state)
 state/sessions/
 
+# Coordination runtime state
+state/coordination/
+
+# GitHub API cache (runtime)
+state/gh-api-cache/
+state/github-graphql-log.jsonl
+
 # Stale directory from tasks->gptodo rename (PR #171)
 packages/tasks/
 


ErikBjare:
@TimeToBuildBob Seems wrong that there are things writing there to begin with, state belongs in brain, not contrib. Worth investigating.

TimeToBuildBob:
Investigated. Here's what's writing to `gptme-contrib/state/`:

## Root cause

### `state/coordination/`

Created by the `gptme-coordination` CLI (`uv run coordination work-claim ...`) when run from inside the `gptme-contrib` directory. The path resolver in `db.py` uses `git rev-parse --show-toplevel` — which returns the gptme-contrib git root — then writes `state/coordination/coord.db` relative to that. Since `COORDINATION_DB` env var isn't set in standalone contrib runs, it falls through to the git-root fallback and lands here.

The unit tests correctly use `tempfile.TemporaryDirectory()` so they're clean, but any direct invocation of the coordination CLI from a `gptme-contrib` working directory causes this.

### `state/gh-api-cache/` and `state/github-graphql-log.jsonl`

Created by `graphql-attribution.sh` (the `gh` wrapper in the brain repo). The script derives its log dir as `${script_dir%/scripts/github}/state`, which correctly points to the brain's `state/` when called from the brain. However, if `WORKSPACE` is set to the gptme-contrib path, the `elif [ -n "${WORKSPACE:-}" ]` branch takes over and writes to `gptme-contrib/state`. This likely happened during a session where gptme-contrib was set as `WORKSPACE`.

## Verdict

Erik is right: state doesn't belong in contrib. The `.gitignore` in this PR suppresses the git noise but doesn't fix the root cause.

**Proper fixes:**
1. **Coordination CLI**: Require `COORDINATION_DB` to be set explicitly in any context where gptme-contrib is the CWD, rather than relying on git-root auto-discovery. Alternatively, document that `uv run coordination` must be run from the brain repo or with `COORDINATION_DB` set.
2. **`graphql-attribution.sh`**: Ensure `WORKSPACE` always points to the brain repo (not contrib) in environments where the wrapper is active.

**Short-term**: Merging this PR (#1250) as a gitignore band-aid is fine to stop the `git status` noise. The actual source fixes can be tracked separately.

I'll create a follow-up issue to track the coordination CLI path-resolution fix.

TimeToBuildBob:
Investigated. Root cause is in `scripts/github/graphql-attribution.sh`'s path resolution fallback:

```bash
if [ -n "${BOB_GRAPHQL_LOG_DIR:-}" ]; then
    LOG_DIR="$BOB_GRAPHQL_LOG_DIR"
elif [ -n "${WORKSPACE:-}" ]; then          # ← checks WORKSPACE
    LOG_DIR="$WORKSPACE/state"
else
    # Falls back to script-relative path
    _script_dir="${BASH_SOURCE[0]%/*}"
    LOG_DIR="${_script_dir%/scripts/github}/state"   # ← resolves to gptme-contrib/state
fi
```

The PM service env has `BOB_WORKSPACE=/home/bob/bob` and `AGENT_WORKSPACE=/home/bob/bob` but **not** `WORKSPACE`, so the script falls through to the script-relative fallback. Since the script lives at `gptme-contrib/scripts/github/graphql-attribution.sh`, the resolved path is `gptme-contrib/state`.

Same root for `state/coordination/`: `gptme-coordination` uses `DEFAULT_DB_PATH = "state/coordination/coord.db"` relative to the git root found from cwd — when cwd is gptme-contrib, the git root is gptme-contrib.

**Fix options**:
1. In `graphql-attribution.sh`: add `BOB_WORKSPACE` / `AGENT_WORKSPACE` as fallback before the script-relative path
2. In PM service: add `Environment=WORKSPACE=%h/bob` or `Environment=BOB_GRAPHQL_LOG_DIR=/home/bob/bob/state`
3. In `gptme-coordination`: use `WORKSPACE` / `BOB_WORKSPACE` env to resolve the db path instead of bare git root from cwd

Option 1 is the cleanest fix in gptme-contrib itself. PR queue is RED (14 open) so I'll file a task for now and not open another PR. This .gitignore PR is a fine band-aid in the meantime.

TimeToBuildBob:
Investigated. Here's what I found:

### What's writing to `state/coordination/` in contrib

The `gptme-coordination` package uses `resolve_coordination_db_path()` which does git-root discovery from CWD. When something runs coordination with `CWD=gptme-contrib`, the git root resolves to gptme-contrib, so `state/coordination/coord.db` lands there.

The tests themselves all use `tempfile.TemporaryDirectory()` correctly. But there's a `coord.db` in `gptme-contrib/state/coordination/` dated Jul 2 14:35 — likely created by an operational script that ran from gptme-contrib context (possibly CI or a manual test run outside the test harness).

### `state/gh-api-cache/` and `state/github-graphql-log.jsonl`

These are **not** in `gptme-contrib/state/` — I checked, and they only exist in the brain's `state/` dir. The gitignore entries for them in this PR were preemptive (or they existed briefly and were cleaned up).

`activity-gate.sh` defaults to `/tmp/github-activity-gate-state` anyway, not a relative `state/` path.

### Recommendation

The root cause for `state/coordination/` is that operational code is running coordination with CWD=gptme-contrib at some point. The gitignore entry here is a reasonable short-term workaround. The real fix would be ensuring operational scripts (PM loop, CI) set `COORDINATION_DB` explicitly to the brain's `$REPO_ROOT/state/coordination/coord.db` rather than relying on git-root discovery.

The two other gitignore entries (`gh-api-cache`, `github-graphql-log.jsonl`) could be dropped from this PR since those files aren't actually being created in contrib.

TimeToBuildBob:
Filed the follow-up issue: #1255 — tracks the root cause fix in `graphql-attribution.sh` (add `BOB_WORKSPACE`/`AGENT_WORKSPACE` fallback before script-relative path). This PR (#1250) is a valid band-aid in the meantime.

TimeToBuildBob:
**Rebased onto master** (was DIRTY / merge conflict). Single commit, no conflicts. Greptile: 5/5 ✅

.gitignore is not in the auto-merge category, so flagging for manual review/merge.

TimeToBuildBob:
@greptileai review

ErikBjare:
Let's not merge this so that misplaced state/ files get surfaced

PR #1258, opened 2026-07-10
fix(self-merge-check): raise on PR number mismatch in fetch_pr

TimeToBuildBob:
## Summary

- **Root cause**: `graphql-attribution.sh` (Bob's `gh` shim) has a 300s TTL on-disk cache for `gh pr view` calls. The cache key used a stride-2 hash that caused PR numbers differing at odd positions (e.g. `1256` vs `1257`) to collide. When `self-merge-check.py` fetched PR #1257, the cache returned stale data for #1256: `url` and `head_sha` came from the wrong PR, while `number` came from the function argument — silently pointing the merge gate at the wrong PR.

- **Root cause fix**: Already shipped in bob@`c033595` (full-character hash in `graphql-attribution.sh`).

- **This PR** adds a defensive guard in `fetch_pr` that compares the returned `pr["number"]` against the requested `number` and raises `RuntimeError` on mismatch, preventing any future cache collision (or other data integrity issue) from silently producing wrong `url`/`head_sha` in `CheckResult`.

- Adds `test_fetch_pr_raises_on_number_mismatch` to exercise the guard.

## Test plan

- [x] `pytest tests/test_self_merge_check.py` — 94/94 pass
- [x] New test `test_fetch_pr_raises_on_number_mismatch` confirms the guard fires on a mocked stride-2 collision scenario
- [x] Existing head_sha tests unchanged (no regression)

### DIFF
--- a/scripts/github/self-merge-check.py
+++ b/scripts/github/self-merge-check.py
@@ -472,6 +472,20 @@ def fetch_pr(repo: str, number: int) -> dict[str, Any]:
         raise RuntimeError(f"Failed to fetch PR metadata for {repo}#{number}")
 
     pr = cast(dict[str, Any], json.loads(raw))
+
+    # Defensive check: the returned PR number must match what was requested.
+    # A cache key collision in the gh shim (e.g. graphql-attribution.sh with a
+    # stride-2 hash) can return stale data for a different PR number, producing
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
@greptileai review

PR #1271, opened 2026-07-11
test(gptme-coordination): add multi-process claim-lifecycle property suite

TimeToBuildBob:
## Motivation

Three mutual-exclusion/CAS bugs in the claim-lifecycle core were each found reactively within an 8-day window, each fixed with a single hand-picked deterministic test:

- #1198 — an expired claim could still be `complete()`-d, letting a session that outlived its TTL ghost-complete a task after another agent had already reclaimed it.
- #1265 — the claim HMAC was computed over a Python-clock expiry string while the stored `expires_at` came from a separate SQLite-evaluated `datetime('now', ...)` call; when the two reads straddled a second boundary the stored HMAC could never verify.
- A third, Bob-workspace-local regression in the reap path (not part of this package — see below).

Each fix landed with a targeted regression test for the *one* interleaving that had actually broken. This PR adds a property/stress suite that instead asserts the invariant *class* under real multi-process concurrency, so the next CAS regression in this area is caught before it ships rather than after.

## What's covered

`packages/gptme-coordination/tests/test_work_invariants.py` spins up real OS processes (`multiprocessing.Pool`) contending on one shared WAL SQLite database, with seeded RNG for deterministic, reproducible failures:

- **I-MUTEX**: N processes race to claim K nonexistent keys once each; every key must be won by exactly one claimer (`TestMutualExclusion.test_contested_keys_have_exactly_one_winner`).
- **I-EPOCH-UNIQ**: N processes churn claim → hold briefly → complete/abandon against a shared key set; no two successful claims may reuse a `(key, epoch)` pair, and per key the consumed epochs must form the gapless sequence `1..n` (`test_epoch_uniqueness_under_claim_release_churn`). A "lost hold" (a release that fails for a key just claimed with a fresh 60-minute TTL) is asserted to never happen — that would mean something took the key from a live holder inside its TTL.

Also added `TestExpiryHandoff` to `test_work.py`: a deterministic scenario where agent A claims, the claim is backdated to expired, agent B claims over it (epoch bumps 1→2), A's `complete()` then fails, and B's `complete()` succeeds. The existing suite covered "expired claim can't be completed" and "HMAC stays coherent across an expired reclaim" as two *separate* tests but never combined effects into one end-to-end handoff check.

## Sizing

Sizes (`N_WORKERS=6`, `N_KEYS=12`, `N_CHURN_ITERATIONS=40`, fixed `SEED`) are chosen to be CI-friendly — the suite adds well under a minute to the run — while still driving enough real interleavings to exercise the CAS path under contention. Bump locally when hunting a suspected race; the invariant checks don't change with scale. The seed makes failures reproducible.

## Note

A sibling stress suite already exists in the Bob workspace repo (`packages/coordination/tests/test_work_invariants.py`) covering the same I-MUTEX/I-EPOCH-UNIQ invariants plus a reap/liveness (I-REAP-SAFETY) scenario against `reap_dead_holders`, which is Bob-local and not part of this generic contrib package — that layer isn't duplicated here.

## Test plan

- [x] `uv run pytest packages/gptme-coordination/tests/ -q` — 92 passed
- [x] `uv run ruff check packages/gptme-coordination/tests/` — clean
- [x] `uv run ruff format` — no changes needed beyond what pre-commit auto-applied

### DIFF
--- a/packages/gptme-coordination/tests/test_work.py
+++ b/packages/gptme-coordination/tests/test_work.py
@@ -370,6 +370,43 @@ class TestExpiryHmacCoherence:
         )
 
 
+class TestExpiryHandoff:
+    """A holder's expired claim can be taken over by another agent, and the
+    original holder loses the ability to complete it — the two invariants
+    together (deny stale complete, allow live reclaim) are what make expiry
+    handoff safe. Combines the effects covered separately by
+    ``TestComplete.test_complete_after_expiry_is_denied`` (#1198) and
+    ``TestExpiryHmacCoherence`` (#1265) into one end-to-end scenario.
+    """
+
+    def test_second_agent_claims_over_expired_claim_and_original_holder_cannot_complete(
+        self, work: WorkClaimManager
+    ) -> None:
+        first = work.claim("agent-a", "task-1")
+        assert first is not None
+        assert first.epoch == 1
+
+        # Backdate expires_at to simulate a TTL that has elapsed.
+        work.db.conn.execute(
+            "UPDATE work SET expires_at = datetime('now', '-1 seconds') WHERE task_id = ?",
+            ("task-1",),
+        )
+
+        second = work.claim("agent-b", "task-1")
+        assert second is not None
+        assert second.claimer == "agent-b"
+        assert second.epoch == 2, "reclaim of an expired claim must bump the epoch"
+
+        # The original holder's complete() must fail now that agent-b holds it.
+        assert work.complete("agent-a", "task-1") is False
+
+        # The new holder can complete normally.
+        assert work.complete("agent-b", "task-1") is True
+        claim = work.get("task-1")
+        assert claim is not None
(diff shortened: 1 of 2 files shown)


ErikBjare:
@TimeToBuildBob Fix the P2s

TimeToBuildBob:
Fixed both P2s in 6448ee36:

- **Default start method**: added `_MP_CTX = multiprocessing.get_context("fork")` and switched both `Pool` calls to use it. Eliminates the `spawn`-path risk where child processes re-execute the pytest entry point and fail before reaching the claim lifecycle.
- **Renewal epoch false duplicates**: added `held: set[str]` tracking in `_churn_worker`. Claims for a key the worker already holds (same-agent renewal after an `OperationalError` on release) are excluded from `events`; `held` is cleared on successful complete/abandon so genuine new claims still record. 92 tests pass, ruff clean.

PR #1299, opened 2026-07-15
fix(build): centralize mypy cache to stop orphan dirs breaking uv workspace

TimeToLearnAlice:
## Problem

`make typecheck-packages` runs `make -C <pkg> typecheck`, so mypy runs with cwd **inside** each package and writes `.mypy_cache/` there. When a package is later removed upstream, git drops its tracked files but the untracked cache dir keeps `packages/<pkg>/` alive as an orphan. uv's workspace `members = ["packages/*"]` glob then rejects that dir (no `pyproject.toml`) and **fails all workspace resolution with exit 2** — breaking every service that shells out to `uv run`.

This exact failure took down two of Alice's systemd services for ~279 silent cycles (2026-07-13/14): `bobutils`, `gptme-rag`, `gptme-wisdom-mcp` were removed upstream but their `.mypy_cache/` dirs lingered under `packages/`.

## Fix

Export `MYPY_CACHE_DIR=$(CURDIR)/.mypy_cache` in the root Makefile so all mypy runs share one cache at the repo root and never scatter into package dirs. Removed package dirs then become truly empty and drop out on checkout — no orphan, no uv break. Also gitignore the (now central) cache dirs.

## Verification

- `make -C packages/gptodo typecheck` → writes to root `.mypy_cache`, **no** `packages/gptodo/.mypy_cache` created. ✅
- Existing scattered `packages/*/.mypy_cache` dirs removed (untracked cruft).
- pre-commit (incl. `make typecheck-packages`) passes.

Complements the detection guard already added to Alice's `workspace_health.py` (flags any `packages/*` member missing a `pyproject.toml`) — this PR removes the root cause so the orphan never forms.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,9 @@
 *.pyc
 __pycache__
 *.egg-info
+.mypy_cache
+.pytest_cache
+.ruff_cache
 
 # Python build artifacts
 build/
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,14 @@ EXCLUDE_PLUGINS :=
 PACKAGE_DIRS := $(shell find packages -maxdepth 1 -mindepth 1 -type d ! -type l ! -name '__pycache__' 2>/dev/null)
 PLUGIN_DIRS := $(shell find plugins -maxdepth 1 -mindepth 1 -type d ! -type l 2>/dev/null)
 
+# Centralize mypy's cache at the repo root so per-package typecheck runs (which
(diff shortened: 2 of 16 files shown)


TimeToLearnAlice:
@greptileai review

TimeToLearnAlice:
@greptileai review

ErikBjare:
@TimeToLearnAlice fails pretty bad

PR #1301, opened 2026-07-16
fix(agent-msg): accept explicit unread list flag

TimeToBuildBob:
## Summary
- accept `agent-msg.py list --unread` as an explicit alias for the default unread-only view
- keep `list --needs-reply` routing to `pending`
- add shim translator regression coverage

## Verification
- `uv run pytest /home/bob/bob/gptme-contrib/tests/test_agent_msg.py -q`
- `python3 /home/bob/bob/scripts/agent-msg.py list --unread`


TimeToBuildBob:
@greptileai review

PR #1305, opened 2026-07-17
fix(sessions): persist Codex cache-read tokens

TimeToBuildBob:
## Problem

Codex trajectories already expose cumulative `cached_input_tokens`, but `extract_usage_codex()` returned that provider-specific key. `post_session()` only persists the canonical `cache_read_tokens` key, so every Codex `SessionRecord` silently dropped its cache component. Cost analysis then had to exclude Codex because cached input looked like full-price fresh input.

## Fix

Normalize Codex `cached_input_tokens` to the existing `cache_read_tokens` session schema field. No schema change is needed.

A real 2026-07-17 Codex rollout now extracts `cache_read_tokens=2,013,440` from `total_token_usage` instead of dropping it.

## Verification

- `323 passed` — full `gptme-sessions` test module
- `38 passed` — focused extraction + post-session persistence tests
- Real rollout extraction assertion passed
- Ruff and `git diff --check` passed

PR #1306, opened 2026-07-17
fix(sessions): persist Codex cache-read tokens

TimeToBuildBob:
## Problem

Codex trajectories expose cumulative `cached_input_tokens`, but `extract_usage_codex()` stored that provider-specific key. `post_session()` only persists the canonical `cache_read_tokens` key, so every Codex `SessionRecord` silently dropped its cache component. Cost analysis then had to exclude Codex because cached input looked like full-price fresh input.

## Fix

Normalize Codex `cached_input_tokens` to the existing `cache_read_tokens` session schema field. No schema change is needed.

A real 2026-07-17 Codex rollout now extracts `cache_read_tokens=2,013,440` from `total_token_usage` instead of dropping it.

## Verification

- `323 passed` — full `gptme-sessions` test module
- `38 passed` — focused extraction + post-session persistence tests
- Real rollout extraction assertion passed
- Ruff and `git diff --check` passed

---
*Splits out the sessions-only fix from #1305 (which also touched `.github/workflows/` and required human merge). CI workflow fix will be submitted separately.*

### DIFF
--- a/packages/gptme-sessions/src/gptme_sessions/signals.py
+++ b/packages/gptme-sessions/src/gptme_sessions/signals.py
@@ -1799,14 +1799,15 @@ def extract_usage_codex(msgs: list[dict]) -> dict:
     if final_total is not None:
         input_tokens = _as_int(final_total.get("input_tokens"))
         output_tokens = _as_int(final_total.get("output_tokens"))
-        cached_input_tokens = _as_int(final_total.get("cached_input_tokens"))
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Closing — the CI workflow fix (adding `apt-get update`) is required for the coverage job to pass, so these two commits can't be cleanly separated. See #1305 which contains both and has green CI.

PR #1308, opened 2026-07-18
fix(runloops): serialize and atomically replace state writes

TimeToBuildBob:
## Summary

- add one shared sidecar-flock + atomic-replace helper for run-loop state files
- serialize project-monitoring PR/CI/assignment/notification discovery state so concurrent monitors emit each item once
- preserve concurrent worker-record enrichments by reloading under the record lock, and make backoff/record writes crash-safe

## Why

Several run-loop paths performed unlocked read-modify-write updates or direct `write_text()` calls. Concurrent monitors could both surface the same item, while worker bookkeeping writers could overwrite each other or leave truncated JSON after interruption.

## Verification

- `uv run --package gptme-runloops pytest -q packages/gptme-runloops/tests` — 685 passed
- `uv run --package gptme-runloops mypy packages/gptme-runloops/src`
- `uv run ruff check ...`
- PR quality gate: 100/100

### DIFF
--- a/packages/gptme-runloops/src/gptme_runloops/base.py
+++ b/packages/gptme-runloops/src/gptme_runloops/base.py
@@ -16,6 +16,7 @@ from gptme_runloops.utils.logging import (
     log_execution_end,
     log_execution_start,
 )
+from gptme_runloops.utils.state import atomic_write_text
 
 # Backoff schedule for consecutive-failure detection.
 # Each entry: (failure_threshold, skip_n, out_of, description)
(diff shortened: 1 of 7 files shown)


TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Addressed the Greptile P1/P2 findings (commit f2b31c20):

**P1: Record lock excludes other writers** — `write_post_session_record` and `write_fallback_session_record` now acquire `locked_state_file(record_path)` before calling `_write_record`, so they cannot race with `finalize_worker_result_in_record`'s locked reload-and-enrich step.

**P1: Completed manifest silently orphaned** — if the record disappears while we hold the lock (cleanup removed it during manifest construction), we now unlink the orphaned manifest before returning, so bookkeeping stays consistent.

**P2: Atomic replace resets file mode** — `atomic_write_text` now calls `os.fchmod(fd, path.stat().st_mode & 0o777)` on the temp file descriptor before closing it, preserving the target's existing permissions after `os.replace`.

All 685 runloops tests pass.

PR #1313, opened 2026-07-22
fix(twitter): pipe Claude prompts through stdin

TimeToBuildBob:
## Problem

The Twitter loop passes the complete timeline/thread context as one positional argument to `claude -p`. A sufficiently large context exceeds Linux's per-argument limit, so process creation fails with `OSError: [Errno 7] Argument list too long` before Claude can evaluate the tweet. This happened in the live loop on 2026-07-21.

## Fix

Omit the positional prompt and send the assembled prompt via subprocess stdin instead. `claude -p` reads stdin when no prompt argument is supplied. This preserves the prompt unchanged without consuming argv space.

## Verification

- `uv run pytest tests/test_twitter_eval_prompt.py -q` (17 passed)
- `uv run ruff check scripts/twitter/llm.py tests/test_twitter_eval_prompt.py`
- `uv run ruff format --check scripts/twitter/llm.py tests/test_twitter_eval_prompt.py`
- Regression test asserts prompt content is supplied via `input=` and no positional prompt is present.

## Review follow-up

Addressed Greptile's concern about a literal `-`: commit `04903d17` removes it and tests the exact argv shape.

### DIFF
--- a/scripts/twitter/llm.py
+++ b/scripts/twitter/llm.py
@@ -546,7 +546,7 @@ def _reply_with_cc_subprocess(messages: list[Message], model_name: str) -> Messa
         cc_model = cc_model[len("anthropic/") :]
     try:
         result = subprocess.run(
-            ["claude", "-p", "--no-session-persistence", "--model", cc_model, "-"],
+            ["claude", "-p", "--no-session-persistence", "--model", cc_model],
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
@greptileai review

PR #1317, opened 2026-07-22
fix: action-receipts plugin NoneType export

TimeToBuildBob:
## Problem
The action_receipts plugin was causing gptme to emit a warning on every first-run invocation:
```
· WARNING  Plugin 'action_receipts' exported 'NoneType' instead of GptmePlugin
           or ToolSpec; skipping
```

This happens because the plugin's `__init__.py` exports the `register` function, which returns `None` after registering the hook. gptme's plugin loader expects exported items to be `GptmePlugin` or `ToolSpec` instances.

## Solution
Call `register()` at module import time instead of exporting the function. This ensures the hook is registered when gptme loads the plugin, eliminating the warning and properly integrating the audit ledger into the first-run experience.

## Testing
Verified with clean-room first-run test (isolated HOME, no API keys, fresh gptme invocation). After this fix, the plugin warning is gone and the audit ledger hook registers silently.

## Impact
- ✅ Removes friction from first-run user experience
- ✅ No changes to the hook behavior or API
- ✅ Graceful fallback if registration fails (wrapped in try/except)

### DIFF
--- a/plugins/gptme-action-receipts/pyproject.toml
+++ b/plugins/gptme-action-receipts/pyproject.toml
@@ -16,7 +16,7 @@ test = [
 ]
 
 [project.entry-points."gptme.plugins"]
-action_receipts = "gptme_action_receipts:register"
+action_receipts = "gptme_action_receipts:plugin"
 
 [build-system]
 requires = ["hatchling"]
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:
Addressed both findings in `3474b846`:

- The entry point now exports a `GptmePlugin` manifest whose `register_hooks` callback is invoked once during hook initialization, rather than registering at import time and then invoking the same function as a factory.
- Removed the broad `try/except`, so registration errors follow gptme's normal plugin-hook warning path instead of being silently swallowed.
- Added a regression test proving import leaves the hook registry empty and invoking the manifest callback registers exactly one hook.

Local plugin suite: 48 passed; scoped pre-commit checks passed.

TimeToBuildBob:
@greptileai review

PR #1348, opened 2026-08-02
fix(failure-capture): prevent lesson names in stderr from triggering auth misclassification

TimeToBuildBob:
## Problem

The `classify_failure_reason` function in `failure_capture.py` used `"auth" in lower` to detect authentication failures. This was too broad: it matched lesson names like `"- Auth Blueprint"` that appear in the gptme startup log written to stderr.

The stderr tail (up to 16KB) includes the full list of auto-loaded lessons. Any lesson whose name contains "auth" caused legitimate failures to be misclassified as `failure_reason='auth'` instead of their true reason.

**Observed impact** (ErikBjare/bob#1116): 5 deepseek sessions with `400 invalid_request_error` (tool_calls without responses) were labeled `failure_reason='auth'`, contaminating the bandit posterior and potentially triggering spurious auth-failure escalation logic.

## Fix

- Remove bare `"auth" in lower` (too broad — matches lesson names, CLI flags, tool names, etc.)
- Add `FAILURE_REASON_INVALID_REQUEST = "invalid_request"` for OpenAI-style 400 errors
- Check `"invalid_request_error" in lower` before the auth check, so deepseek tool-format failures get the right classification
- Keep `"authentication" in lower` and HTTP 401/403 status checks for real auth failures
- Add `"unauthorized" in lower` for completeness

## Tests added

- **Regression test** for the exact false-positive: `"- Auth Blueprint\n"` in startup log must not return `failure_reason='auth'`
- Test that deepseek `400 invalid_request_error` → `failure_reason='invalid_request'`
- Tests confirming real auth signals (`401`, `"authentication"`, `"unauthorized"`) still classify as `auth`

### DIFF
--- a/packages/gptme-sessions/src/gptme_sessions/failure_capture.py
+++ b/packages/gptme-sessions/src/gptme_sessions/failure_capture.py
@@ -8,6 +8,7 @@ from pathlib import Path
 
 # Coarse taxonomy for operator-pulse / bandit post-mortems.
 FAILURE_REASON_AUTH = "auth"
+FAILURE_REASON_INVALID_REQUEST = "invalid_request"
 FAILURE_REASON_NONZERO = "nonzero_exit_unclassified"
 FAILURE_REASON_PRE_RESPONSE = "pre_response_api_failure"
 FAILURE_REASON_RATE_LIMIT = "rate_limit"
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Superseded by #1349, which landed the same classifier fix and regression coverage in `master` (merge commit `18ded6a`). This PR now conflicts on those exact files, so closing the duplicate rather than rebasing and merging it again.

PR #1374, opened 2026-08-06
fix(cc-memory): exclude MEMORY-archive.md from recall candidates

TimeToBuildBob:
## Summary

- Adds `MEMORY-archive.md` to `SPECIAL_MEMORY_FILES` in `schema.py`, so a cold-storage archive file is never discovered as a recall candidate
- Ports the fix from Bob's live `scripts/memory/memory_retrieval.py` (`6b6950f29c`)
- The cooldown fix (PR #1368) was already ported; this is the second and final gap from the `6b6950f29c` commit

## Why

`MEMORY-archive.md` is cold storage: a concatenation of ~180 archived one-liners. As a single retrieval entry its bag-of-words vocabulary matches almost any prompt and out-scores real memory entries, while its excerpt always renders static header boilerplate. It caused 656 injections in Bob's live store before the fix — the most-injected key by a large margin.

On a fresh adopter install the archive file may be small or absent, so this is mostly preventative. The cooldown gap (PR #1368) is the defect that bites immediately at any store size; this fix closes the second one.

## Test plan

- `test_memory_archive_is_excluded` — `discover_memory_files` skips `MEMORY-archive.md` even when it has valid frontmatter
- `test_archive_file_not_recalled` — archive file does not appear in `select_relevant_memories` results even when it lexically dominates the prompt
- `test_real_memories_still_recalled_with_archive_present` — real memory entries are unaffected
- Negative control verified: both behavioural tests **fail** against the pre-fix source (before adding `MEMORY-archive.md` to `SPECIAL_MEMORY_FILES`)
- 34/34 tests pass after fix; no regressions

## Related

- Sibling PR: #1368 (injection cooldown — already merged)
- Sibling PR: #1370 (pending-items retirement — already merged)
- Origin fix: `ErikBjare/bob@6b6950f29c`
- Task: `tasks/cc-memory-package-recall-parity.md`

### DIFF
--- a/packages/gptme-cc-memory/src/gptme_cc_memory/schema.py
+++ b/packages/gptme-cc-memory/src/gptme_cc_memory/schema.py
@@ -27,6 +27,12 @@ MEMORY_TYPES = frozenset({"user", "feedback", "project", "reference"})
 SPECIAL_MEMORY_FILES = frozenset(
     {
         "MEMORY.md",
+        # Cold-storage overflow written by memory-overflow tooling. It is a
+        # concatenation of archived one-liners, so as a single retrieval
+        # candidate its vocabulary matches almost any prompt and out-scores
+        # real entries while its excerpt renders static header boilerplate.
+        # Archived == out of automated recall by design; use grep for it.
+        "MEMORY-archive.md",
         "guidance.md",
         # Retired lane — kept here so a leftover file from a pre-retirement
         # install is never discovered as a memory entry.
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:

Evidence: commit: 79cb9e37b5 chore(submodule): bump gptme-contrib to include cc-memory archive exclusion fix (#1374)

Session: `02ac3153-55e9-5a0b-ad2f-bbb806f1c4c6`

PR #1380, opened 2026-08-06
feat(hooks): add pre-merge-commit dispatcher

TimeToBuildBob:
`core.hooksPath` points at this directory globally, so git had **no `pre-merge-commit` hook at all** — repos could not guard merge-commit creation.

The dispatcher is **inert by default**: it execs a repo's own `scripts/precommit/prevent-master-merge-commits.sh` when that repo ships one, and exits 0 everywhere else. Repo-specific policy stays in the repo; this file stays generic.

### Motivating incident (ErikBjare/bob, 2026-08-06)
origin's ruleset rejects merge commits on master, but the repo has `pull.rebase=false` — so every non-fast-forward `git pull` minted exactly the commit the remote refuses. 38 accumulated and jammed the fleet's push path for ~3.5h. Clearing the 303-commit backlog required squashing it into a single commit, flattening per-commit history and attribution for the whole window.

An advisory lesson for this already existed and sessions created 38 merge commits anyway — hence the mechanical guard.

---

## ⚠️ Read the trust boundary before merging

Delegating to a repo-controlled path from a **global** hook is an RCE vector: an adversary commits an executable at that exact path, you clone, you `git merge`, their script runs with your privileges. Git's normal hook model is opt-in precisely *because* `.git/hooks` is never fetched from a remote — `core.hooksPath` opts every repo in at once, so the trust check has to be reintroduced in the dispatcher.

Our own self-hosted AI reviewer found **four successive P0s here**, each hiding inside the fix for the previous one (ErikBjare/bob#1122):

| # | Hole | Fix |
|---|------|-----|
| 1 | Guard exec'd in *any* repo | Gate on `origin` ownership |
| 2 | Owner matched, host ignored — `https://evil.com/gptme/x.git` parses to owner `gptme` | Match **host AND owner** as a pair |
| 3 | Trusted origin ≠ trusted *content*: `pre-merge-commit` fires after the merge hits the worktree, so a fork branch could introduce/edit the guard | Compare the worktree blob against the committed one |
| 4 | **HEAD is not a trust anchor** — `gh pr checkout <hostile-pr>` moves HEAD onto the contributor's branch, so worktree and HEAD agree by construction and the next merge execs the payload | Anchor to `refs/remotes/origin/{HEAD,master,main}` |

The current anchor is #4. Those refs only move on `git fetch origin`, so a match proves the guard is the version that landed on the **trusted org's default branch**. A PR branch — even one pushed to origin — is never one of them.

Two deliberate consequences, both fail-closed:
[...]

PR #1382, opened 2026-08-06
fix(ci-gate): fail closed when repo has workflows but no checks ran

TimeToBuildBob:
Distinguish "this repo does not gate PRs with checks" (warning, waive) from "it does, and this PR's checks are missing" (disqualifying reason).

## Problem

The prior logic treated zero checks as satisfying the CI requirement. That is fail-open during a GitHub Actions outage: in the 2026-08-06 window, a PR that didn't touch sensitive paths would pass the CI gate simply because no checks were reported. `ErikBjare/bob#1130` is in that window and still shows 0 checks.

## Why the predicate changed

The first cut asked *"does this repo have active Actions workflows?"*. Review found two P1s on that predicate that **fail in opposite directions**, so no adjustment of it can satisfy both:

| Repo shape | Workflow-presence predicate | Consequence |
|---|---|---|
| CI from Jenkins / Azure / Travis / a GitHub App status context | `False` — no Actions workflows | Requirement waived → **the same outage fail-open, for non-Actions CI** |
| Only `push`/`schedule` workflows (deploy-on-master) | `True` — workflows exist | No PR checks by design → **every PR permanently self-merge-ineligible** |

Broadening it to fix the first worsens the second; narrowing it to `pull_request`-triggered workflows does the reverse.

## Solution

Ask the question the gate actually needs answered: **do this repo's PRs report checks?** `repo_gates_prs_with_checks()` samples the last 10 merged PRs' `statusCheckRollup`, scoped to the PR's base branch.

That resolves both P1s with one predicate:
- `statusCheckRollup` carries `StatusContext` entries as well as `CheckRun` ones (already documented in `checks_green`), so **external CI counts**.
- A push-only repo's PRs are **observed** to carry no checks, so it is correctly waived.

It also reads already-merged history, so an ongoing outage cannot erase the evidence.

### A silent sample is never trusted on its own

"None of the sampled PRs reported checks" is ambiguous, and later review passes found both of the ways it can mislead. It is now resolved in three steps:

1. **This branch isn't gated, but another is.** Detected by widening the sample to the whole repo — checks elsewhere, none here → `False`. Without this, a repo gating `master` but not `dev` would block every `dev` PR forever.
2. **The repo gates nothing.** Confirmed when `repos/{repo}/actions/runs?event=pull_request` also reports zero → `False`.
3. **Otherwise the sample is unrepresentative** (CI added after those merges; every sampled merge from a fork whose workflows were never approved) → `None` → fail closed.

Counting pull_request-event *runs* in step 2 is safe where counting *workflows* was not: a push-only repo has workflows but **zero** PR runs.

The tri-state contract and its fail-closed discipline are unchanged throughout.

**Verified live at each revision:**

| Repo | `gates_prs` | PR-event runs | Note |
|---|---|---|---|
| `gptme/gptme-contrib` | `True` | 13664 | gated |
| `ErikBjare/bob` | `True` | 1151 | so the #1130 outage shape still fails closed — the PR's original purpose |
| `ErikBjare/erikbjare.github.io` | `False` | 0 | 2 active workflows, 0 checks on all 4 merged PRs — the push-only control the old predicate blocked forever |
| `ErikBjare/dotfiles` | `None` | 0 | no merged-PR history |

## Also: name the un-approved-workflow case

gptme/gptme-contrib#1232 is a live instance of the hole, and "no CI configured" would have been **wrong on the facts**: `pre-commit.yml` fires on every PR to master with no path filter. Its author is a `FIRST_TIME_CONTRIBUTOR` pushing from a fork, so GitHub held every workflow behind the **"Approve and run workflows"** button — 0 workflow runs for 36 days. The un-run `pre-commit` job would have caught a stray `END` heredoc marker, so the missing CI was not cosmetic.

The two cases need different operator actions, so they now read differently. Fork provenance alone does not discriminate — gptme/gptme-contrib#1240 is *also* a fork PR, but by a `MEMBER`, so its workflows ran automatically (17 checks). `author_association` is the signal; `gh pr view --json` does not expose it, so it is fetched from the REST pulls endpoint on the already-disqualifying path, where it costs no request in the common case.

Verified live: #1232 → approval message, #1240 → generic.

## Tests

45 in `tests/test_self_merge_ci_configured.py`, hermetic — `test_evaluate_pr_is_hermetic` enforces it and caught a real merge-interaction bug: `_fetch_ai_review_marker_checked` (new on master via #1391) reaches `subprocess.run` through `run_gh_checked`, not `run_gh`, so the module was making live API calls once master was merged in. That failed CI while the branch alone was green.

`make test-unit`: 869 passed, 4 skipped.

## Review status

Reviewed at head `f857c4f8`: **5/5, 0 findings**, all threads resolved, CI green.

**Not self-merged on purpose.** The gate reports `Touches sensitive/security/infra paths`, which is correct rather than an artifact — this PR rewrites `self-merge-check.py`, the merge gate itself. Self-merging a change to the self-merge rules, evaluated by the very rules being changed, is the conflict of interest that category exists to catch. Needs a human merge.

Fixes: ErikBjare/bob self-merge-ci-gate-vacuous-during-outage task


TimeToBuildBob:
Status update: CI gate check run against this PR.

**Self-merge gate result: BLOCKED** — requires manual review before merge.

Reasons:
- Touches sensitive/infra path (`self-merge-check.py`)
- No CI checks ran on this branch (workflows exist in gptme-contrib, but produced no results — which is exactly what this PR fixes)
- Greptile review pending (triggered, awaiting results)

The fix itself is sound: distinguishing "repo has no workflows" (fail-open acceptable) from "repo has workflows but zero checks ran" (fail-closed). Verified correct behavior — the self-merge gate now properly catches the GitHub Actions outage window pattern.

@ErikBjare: please review and merge manually when ready. Greptile results will land shortly.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Status update, correcting my earlier note: the missing Greptile review here is not transient — it's a confirmed repo-wide outage on gptme-contrib (see `tasks/greptile-not-reviewing-gptme-contrib.md`), affecting every contrib PR created since 2026-08-06 16:09Z (#1380-#1383, 0/4 reviewed), while gptme/gptme and ActivityWatch repos kept getting reviewed normally in the same window. This PR will not clear the Greptile gate on its own; it needs an operator to check the Greptile GitHub App's installation on gptme-contrib.

Given that, I'm not going to keep re-triggering Greptile on this PR. Current state:

- CI: green (integration-tests, pre-commit all pass)
- Mergeable: yes, no conflicts
- `self-merge-check.py` correctly declines to self-merge this one — it touches its own sensitive path (`self-merge-check.py`) and contrib merges route through manual/PM review by design, not the deterministic allowlist
- The change itself: adds a `repo_has_ci_configured()` check so the self-merge gate fails closed (blocks) when a repo has workflows configured but no checks ran, instead of treating "no results" the same as "no CI at all". Strictly more conservative than current behavior. Validated against the real outage case (ErikBjare/bob#1130).

Recommending merge on code review alone rather than waiting further on Greptile, since the outage is unrelated to this PR's content and this fix is small, self-contained, and safety-tightening. Leaving the merge decision to you.

TimeToBuildBob (commented):
🤖 2 inline finding(s) — see the summary comment.

TimeToBuildBob:
## 🤖 AI code review

Adds a tri-state probe that samples recent merged PR statusCheckRollup to distinguish repos that gate PRs with checks from those that don't; silent samples are corroborated against Actions pull_request run counts. Also adds author-association lookup to name the un-approved-fork-workflow case. The evaluate_pr empty-check branch now fails closed unless it can positively determine the repo is ungated.

- `scripts/github/self-merge-check.py` — Adds repo_gates_prs_with_checks, _sample_pr_check_history, repo_has_pull_request_workflow_runs, pr_author_association, and a new empty-statusCheckRollup branch; fetch_pr now requests isCrossRepository and baseRefName.
- `tests/test_self_merge_ci_configured.py` — Adds tests for the tri-state probe, branch scoping, external CI, evaluate_pr integration, and hermetic subprocess enforcement.

</details>

| `9d4e8478692a` | ? | 2 | llm | 2026-08-07 21:59 UTC |
| `ad09b19fb440` | ? | 1 | llm | 2026-08-08 08:25 UTC |
| `750f256e9d3b` | 3/5 | 3 | llm | 2026-08-09 00:24 UTC |
| `0d56d38e9773` | 3/5 | 2 | llm | 2026-08-09 09:00 UTC |
| `0d56d38e9773` | 3/5 | 3 | llm | 2026-08-09 13:56 UTC |
| `0d56d38e9773` | 3/5 | 1 | llm | 2026-08-09 14:16 UTC |
| `dfe821eac579` | 4/5 | 2 | llm | 2026-08-09 15:02 UTC |
| `9af637526840` | 2/5 | 2 | llm | 2026-08-09 15:15 UTC |
| `c12239de20f3` | 3/5 | 2 | llm | 2026-08-10 21:24 UTC |
| `2add141aea8c` | 3/5 | 1 | llm | 2026-08-10 21:40 UTC |

</details>

<sub>Reviewed `f857c4f86201` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 90s · advisory, a human still decides · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

tests/test_self_merge_ci_configured.py
```diff
@@ -0,0 +1,133 @@
...
+    """A timeout is the single likeliest symptom of the outage we guard against."""
+
+    def _raise(*a: object, **k: object) -> None:
+        raise exc
+
+    monkeypatch.setattr(subprocess, "run", _raise)
+    assert smc.repo_has_ci_configured("o/r") is None
+
+
+def _evaluate_with(monkeypatch: pytest.MonkeyPatch, has_ci: bool | None) -> Any:
```
TimeToBuildBob, line 201: Fixed in 84bc1a46. `_evaluate_with` now stubs `merge_permission` (→ True) and `run_gh` (→ ""). Both were reached from `evaluate_pr` and shelled out to a real `gh api`: `merge_permission` at self-merge-check.py:1235, and `run_gh` via `fetch_greptile_status`'s summary-comment fallback (line 658), which fires because the stubbed review data is empty. Worth noting the finding was worse than it looked — `merge_permission` is `@cache`-decorated, so the first unstubbed call poisoned the cache with a live network result for every later test in the session (which is why the leak surfaced at the greptile call in a full-file run but at :1235 in isolation). A definitive `False` there would have appended a disqualifying reason and masked the CI-gate reasons under test. Added `test_evaluate_pr_is_hermetic`, which patches `subprocess.run` to raise a non-caught `_NoNetwork` and asserts `_evaluate_with` still completes; it fails against the pre-fix helper with `_NoNetwork: real subprocess invoked: (['gh', 'api', 'repos/o/r', '-q', '.permissions'],)`. Suite time dropped 4.5s → 1.9s with the network calls gone.

scripts/github/self-merge-check.py
```diff
@@ -755,6 +755,56 @@ def fetch_unresolved_human_threads(
...
+    and the PR stayed self-merge eligible — the exact outage merge this check
+    is supposed to block. Fail closed on ``None`` at the call site.
+
+    Used to distinguish "repo has no CI at all" from "repo has CI but checks
+    produced no result" (e.g., an outage, paths filter mismatch, or
+    concurrency cancellation).
+    """
+    try:
+        proc = subprocess.run(
+            ["gh", "api", f"repos/{repo}/contents/.github/workflows", "--jq", "length"],
```
TimeToBuildBob: Fixed in 84bc1a46. Confirmed real: the Contents API returns an array for a directory but an object for a regular file, and `jq length` on an object counts its keys, so a `.github/workflows` file would have returned `True` instead of the documented indeterminate `None`. The jq program is now type-guarded — `if type == "array" then length else -1 end` — and a negative count maps to `None`. Verified against real payloads: `[{...},{...}]`→2 (True), `[]`→0 (False), `{"type":"file",...}`→-1 (None), `null`→-1 (None). The tri-state docstring contract is unchanged: True / False (404 only) / None; the function was not restructured. Two regression tests, both proven to fail without the fix: a `_Proc(0, "-1")` tri-state case (pre-fix: `assert False is None`) and `test_workflows_lookup_jq_is_type_guarded`, which captures the argv passed to `subprocess.run` (pre-fix: `AssertionError: jq program is not type-guarded: 'length'`). Existing canned-stdout tests are unaffected by the jq change, as expected — all 133 tests across both self-merge modules pass.

TimeToBuildBob:
Rebased onto current `master` (`4d4565af`) so this safety fix is no longer one commit behind. The new head is `ad09b19f`; the three PR commits are unchanged apart from rebased SHAs.

Re-verified after the rebase:
- 133 targeted self-merge tests pass
- Ruff passes
- mypy passes
- both self-hosted review findings remain fixed and their threads are resolved

Fresh CI is now running. I am not merging: this changes the self-merge policy itself (sensitive path), and Greptile still has no current-head review on gptme-contrib, so it remains a manual merge decision.

scripts/github/self-merge-check.py
```diff
@@ -755,6 +755,72 @@ def fetch_unresolved_human_threads(
...
+    except (subprocess.TimeoutExpired, OSError):
+        return None
+
+    if proc.returncode != 0:
+        stderr = (proc.stderr or "").strip()
+        if stderr:
+            print(f"[gh error] {stderr}", file=sys.stderr)
+        # A 404 is a real answer: the directory does not exist. Anything else
+        # (rate limit, 5xx, auth, network) leaves us genuinely unsure.
+        if "404" in stderr or "Not Found" in stderr:
```
TimeToBuildBob, line 843: ❌ **P1** — `repo_has_ci_configured` treats any stderr containing "404" or "Not Found" as a definitive "directory does not exist" and returns False, which the call site in `evaluate_pr` converts into a warning and waives the CI requirement. GitHub returns HTTP 404 "Not Found" not only when `.github/workflows` is absent but also whenever it masks unauthorized access to a private repo or a restricted path, and an outage or a caching gh shim can surface a spurious 404. The PR's own premise is that during an outage the workflows lookup fails in the same way the checks disappear — so collapsing a 404 into False re-enables exactly the fail-open self-merge this guard was written to block: `statusCheckRollup` is empty, the lookup sees a masked 404, the repo actually has CI, and the PR stays eligible. The branch should regard a 404 as an ordinary failure and return None (fail closed) unless it can confirm the directory genuinely does not exist.

TimeToBuildBob: Pushing back on this one — the recommendation would delete a deliberate, tested behaviour, and the premise does not hold in this call path.

**The 404 is not ambiguous here.** `repo_has_ci_configured` is only reached from `evaluate_pr`, which has *already* successfully fetched the PR from the same repo with the same token (`fetch_pr` raises otherwise). So access to the repo is established before the workflows lookup runs. A 404 on `contents/.github/workflows` at that point means the path is absent, not that access is masked.

**Outages do not surface as 404.** They surface as 5xx, rate-limit errors, timeouts and connection failures — all of which already return `None` and already fail closed. The tri-state exists precisely to separate those from a real answer, and `test_tri_state` pins `502`, `401`, rate-limit, empty stdout and unparseable stdout as `None`.

**Returning `None` on 404 would collapse the tri-state into a bi-state.** Every repo that genuinely has no workflows becomes permanently indeterminate and permanently self-merge-ineligible. That waiver is deliberate and covered by `test_genuinely_no_ci_is_waived`. Removing it is a policy change, not a bug fix, and it is Erik's call rather than something to fold into this PR.

The residual sharp edge you are pointing at is real but narrow: a token with `pull_requests` but not `contents` scope (a fine-grained PAT) could 404 on the path while PR reads succeed. That is worth noting, and I have left the docstring's claim explicit so the assumption is visible. If it becomes a live concern the right fix is to assert the token's scopes once at startup, not to make every CI-less repo indeterminate.

No change.

scripts/github/self-merge-check.py
```diff
@@ -755,6 +755,109 @@ def fetch_unresolved_human_threads(
...
+        if stderr:
+            print(f"[gh error] {stderr}", file=sys.stderr)
+        # A 404 here is ambiguous, not a real answer. It means either "the
+        # workflows directory does not exist" (safe to waive) or "this token
+        # cannot read this repository at all" (must never waive) — GitHub
+        # returns 404 for both, and gh's stderr is identical. Only conclude
+        # "no workflows" once the repository itself is confirmed readable.
+        # Anything else (rate limit, 5xx, auth, network) leaves us unsure.
+        if "404" in stderr or "Not Found" in stderr:
+            return False if _repo_is_readable(repo) else None
```
TimeToBuildBob: Fixed in 0d56d38e. Removed `_repo_is_readable` entirely: a Contents 404 is now indeterminate and therefore fail-closed, because the metadata endpoint cannot establish Contents permission. The regression test asserts the helper returns `None`, makes exactly one API call, and the end-to-end gate disqualifies the PR.

scripts/github/self-merge-check.py
```diff
@@ -755,6 +755,109 @@ def fetch_unresolved_human_threads(
...
+        return None
+    try:
+        count = int(raw)
+    except (ValueError, TypeError):
+        return None
+    # Negative is the jq type-guard sentinel: the path exists but is not a
+    # directory listing, so there is nothing to count. Indeterminate, not False.
+    if count < 0:
+        return None
+    return count > 0
```
TimeToBuildBob: Fixed in 0d56d38e. The Contents API jq now counts only top-level entries with `type == "file"` and a case-insensitive `.yml`/`.yaml` suffix. README files, placeholders, and subdirectories no longer imply configured CI; the test inspects and pins both filters.

tests/test_self_merge_ci_configured.py
```diff
@@ -0,0 +1,284 @@
...
+    #   * merge_permission — and it is @cache'd, so an unstubbed call poisons the
+    #     cache for every later test in the session with a live network result.
+    #     A definitive False would append a disqualifying reason and mask the
+    #     CI-gate reasons these tests assert on.
+    #   * run_gh — reached via fetch_greptile_status's summary-comment fallback,
+    #     which fires because the stubbed review data is empty.
+    # Stub both so the CI gate is the only thing under test.
+    monkeypatch.setattr(smc, "merge_permission", lambda repo: True)
+    monkeypatch.setattr(smc, "run_gh", lambda *a, **k: "")
+    return smc.evaluate_pr("o/r", 1, workspace_repos=None)
```
TimeToBuildBob: Fixed in 0d56d38e. `_evaluate_with` now supplies the matching workspace repo and a valid open test-only PR plus a passing reviewer stub, so CI is the only gate under test. The no-CI case now explicitly asserts `result.eligible` while the indeterminate/configured cases assert ineligibility.

scripts/github/self-merge-check.py
```diff
@@ -755,6 +755,77 @@ def fetch_unresolved_human_threads(
...
+                ),
+            ],
+            capture_output=True,
+            text=True,
+            timeout=10,
+        )
+    except (subprocess.TimeoutExpired, OSError):
+        return None
+
+    if proc.returncode != 0:
```
TimeToBuildBob: ❌ **P1** — `repo_has_ci_configured` maps every non-zero `gh api` exit to None (line 813). GitHub's Contents API returns HTTP 404 — non-zero exit — whenever `.github/workflows` does not exist, the normal case for a repo with no GitHub Actions CI. So the common 'no CI' repo yields has_ci=None, and `evaluate_pr` appends 'CI checks not found and could not determine whether the repo has workflows configured (GitHub API error) — failing closed', making every PR ineligible. The PR description promises 'Repo has no workflows: WARNING (acceptable) — fail open as before', but False is only reachable via a 200 listing with no top-level YAML, not an absent directory. `test_genuinely_no_ci_is_waived` stubs the helper to False (test line 158) and never exercises the real 404 path, so a no-CI repo that previously self-merged with a warning now blocks on every PR.

ErikBjare:
@TimeToBuildBob review

(I'm doing using the same pattern as I would with `@greptileai review` here, consider this comment notice that I want your review system to work the same - you should probably wire it into pm or something, or we should create a GitHub app for the bob-ai-reviewer)

TimeToBuildBob:
Review command is wired now: `ai-review-sweep.py` treats a trusted `@TimeToBuildBob review` comment as an immediate forced consensus review, even when the current head was already reviewed. Your comment triggered that path at 14:10 UTC; the sweep updated the review summary at 14:21 UTC rather than waiting for a new commit.

The review found a real P1: the Contents API approach made the documented no-CI waiver unreachable for the normal case where `.github/workflows` is absent. I fixed that in `dfe821ea` by querying `repos/{repo}/actions/workflows` and counting active workflows. A successful zero is now a definitive no-CI result; API errors remain indeterminate and fail closed. Verified 134 targeted tests, Ruff, mypy, pre-commit, and the pushed head. Fresh CI is running.

I'm not merging this policy change myself. It modifies the self-merge gate, so it stays a maintainer decision after CI/re-review.

TimeToBuildBob:
Your `@TimeToBuildBob review` command worked end-to-end: the forced current-head review ran and updated the canonical review at 15:05 UTC.

That review found two additional P2s on `dfe821ea`: counting every active workflow can permanently block PRs in repos whose workflows are schedule/push/dispatch-only, and the tests still stub the helper at the `evaluate_pr` call site. I am not calling this merge-ready or merging it while those findings remain open.

Another monitoring worker already owns the PR/branch lane and is handling those fixes. The next honest state is: fix both findings, resolve the four stale/current review threads, rerun the current-head review, then hand the sensitive self-merge-policy change back for maintainer review. The PM trigger wiring you requested is already live; a GitHub App remains a separate productization option, not required for this command path.

tests/test_self_merge_ci_configured.py
```diff
@@ -0,0 +1,213 @@
...
+        "state": "OPEN",
+        "mergeStateStatus": "CLEAN",
+        "statusCheckRollup": [],
+        "files": [{"path": "tests/test_example.py"}],
+        "author": {"login": "TimeToBuildBob"},
+        "reviews": [],
+        "comments": [],
+    }
+    monkeypatch.setattr(smc, "fetch_pr", lambda *a, **k: pr)
+    monkeypatch.setattr(smc, "repo_has_ci_configured", lambda repo: has_ci)
```
TimeToBuildBob: Fixed in 9af63752. Added an evaluate-level regression that leaves `repo_has_ci_configured` real, stubs only the underlying subprocess response to an Actions API 404, and verifies that `evaluate_pr()` calls the probe and fails closed.

scripts/github/self-merge-check.py
```diff
@@ -755,6 +755,41 @@ def fetch_unresolved_human_threads(
...
+    treating malformed or placeholder YAML as runnable CI. ``None`` must never
+    be treated as ``False``: an API outage is exactly when the caller needs to
+    fail closed rather than waive missing checks.
+    """
+    raw = run_gh(
+        [
+            "api",
+            f"repos/{repo}/actions/workflows",
+            "--jq",
+            '[.workflows[]? | select(.state == "active")] | length',
```
TimeToBuildBob: Adjudicated as a conservative-policy trade-off rather than a defect in this safety gate. The gate asks whether the repo has active GitHub Actions, not whether every workflow is statically provable to handle this PR: trigger eligibility depends on event, branch/path filters, reusable calls, and workflow contents, none of which the Actions list endpoint exposes. Treating any active workflow plus zero checks as suspicious may require a manual merge for push-only repos, but it cannot recreate the outage fail-open this PR fixes. I am leaving the implementation conservative and handing that policy choice to the maintainer.

scripts/github/self-merge-check.py
```diff
@@ -786,6 +786,41 @@ def fetch_unresolved_human_threads(
...
```
TimeToBuildBob: ❌ **P1** — repo_has_ci_configured counts only GitHub Actions workflows via repos/{repo}/actions/workflows with state == "active". A repo whose CI is provided by external systems (Jenkins, Azure Pipelines, Travis, or a GitHub App status context) returns False when the Actions list is empty. If an outage suppresses those external checks, statusCheckRollup is empty, the helper returns False, and evaluate_pr appends only the "No CI configured" warning, leaving result.eligible true. The PR is then cleared for self-merge with no green build — exactly the fail-open this PR is meant to close, just for non-Actions CI. The helper should also treat presence of required status contexts or branch-protection required checks as CI.

TimeToBuildBob: ❌ **P1** — repo_has_ci_configured treats any active workflow as CI regardless of its triggers. A repo with only a push/schedule workflow (e.g., a deploy-on-main job) has no PR-triggered checks, so a PR's statusCheckRollup is legitimately empty; the helper returns True, and evaluate_pr adds the "produced no check results" disqualifying reason. Every PR on such a repo becomes permanently ineligible for self-merge. The old behavior waived the CI requirement; this change introduces a false-positive block. The helper should limit the count to workflows that can run on pull_request (or otherwise distinguish PR-event CI), and the tests should cover a push-only workflow repo.

TimeToBuildBob: Confirmed valid — accepted as **blocking**, not fixed yet. This PR is staying open.

The helper equates "has CI" with "has active GitHub Actions workflows", so a repo whose checks come from Jenkins, Azure Pipelines, Travis or any GitHub App status context reports `False`. During an outage that suppresses those external checks, `statusCheckRollup` is empty, the helper says "no CI configured", and `evaluate_pr` downgrades to a warning while leaving `eligible` true — the exact fail-open this PR exists to close, reproduced for non-Actions CI.

Required fix: also treat required status contexts / branch-protection required checks as evidence of CI, so the probe answers "does this repo gate PRs" rather than "does this repo use Actions".

TimeToBuildBob: Confirmed valid — accepted as **blocking**, not fixed yet. This PR is staying open.

This is the more dangerous of the two, because it fails in the availability direction and would be self-inflicted: a repo whose only active workflow is push- or schedule-triggered (a deploy-on-master job) legitimately produces no PR checks, but the helper returns `True`, so `evaluate_pr` adds the "produced no check results" disqualifying reason and **every PR on that repo becomes permanently self-merge ineligible**. The old behaviour waived the requirement; this change would replace a fail-open with a permanent block.

Given we have just spent 38 hours unblocking a deadlocked queue, shipping a new permanent-block path is the wrong trade. Not merging on that basis.

Required fix: count only workflows that can trigger on `pull_request`, which the `actions/workflows` list does not expose — it needs the workflow file contents (or an equivalent trigger probe). Tests should cover a push-only-workflow repo explicitly, as suggested.

TimeToBuildBob:
## Not merging — two open P1s, both confirmed

Every other PR in this queue merged today; this one is the exception, deliberately.

Our reviewer found two genuine defects in `repo_has_ci_configured`, and they are **opposite failure modes of the same helper**:

1. **Fails open for non-Actions CI** — external providers (Jenkins, Azure, a status-context App) report as "no CI", so an outage that empties `statusCheckRollup` yields a warning instead of a block. That is the very hole this PR sets out to close.
2. **Fails closed, permanently, for push-only workflows** — a repo whose only active workflow is push/schedule-triggered has legitimately empty PR checks but reads as "CI configured", so *every* PR there becomes permanently ineligible.

(2) is why this is not merging under queue pressure. Swapping a fail-open for a permanent block, on the same day we finished clearing a 38-hour deadlock, would be trading one deadlock for another.

The fix is real work, not a tweak: the `actions/workflows` API does not expose triggers, so distinguishing PR-triggered CI needs the workflow file contents or an equivalent probe, plus branch-protection required checks for case (1).

Both threads have replies stating the required fix. Leaving this open rather than merging or closing it.

scripts/github/self-merge-check.py
TimeToBuildBob: Fixed in 8c1a4504 — and you were right that the helper was answering the wrong question.

The fix is not to add branch-protection contexts to a workflow-presence probe, because your sibling finding (the push-only-workflow permanent block) pulls the same predicate the other way: broadening it to catch external CI makes more repos read "has CI", which is exactly what creates the permanent block. Any threshold on "which workflows exist" trades one of your two findings for the other.

So the predicate now asks what the gate actually needs to know — *do this repo's PRs report checks?* `repo_gates_prs_with_checks()` samples the last 10 merged PRs' `statusCheckRollup`. That rollup carries `StatusContext` entries as well as `CheckRun` ones (already documented in `checks_green`'s docstring), so a Jenkins/Azure/Travis/GitHub-App commit status counts as CI without any special-casing per provider. It also reads already-merged history, so an outage in progress cannot erase the evidence the way it erases live checks.

Test: `test_external_ci_repo_fails_closed` feeds a `StatusContext` payload (`continuous-integration/jenkins/pr-merge`) and asserts both the helper returns True and `evaluate_pr` disqualifies.

Verified live: `ErikBjare/bob` → True, so the #1130 outage shape this PR was built against still fails closed.

TimeToBuildBob: Fixed in 8c1a4504 — this was the one I refused to merge over, and you were right to call it the more dangerous of the two.

I did not do what the finding suggested (count only `pull_request`-triggerable workflows), because that fix and the one your sibling finding asks for are the same knob turned opposite ways: narrowing to PR-triggered workflows makes more repos read "no CI", which widens the external-CI fail-open. Fixing either by adjusting a workflow-presence predicate reintroduces the other.

Instead the predicate now samples observed PR check history — `repo_gates_prs_with_checks()` over the last 10 merged PRs' `statusCheckRollup`. A repo whose only workflow is push- or schedule-triggered is *observed* to produce no PR checks, so it reads False and keeps the waiver. No permanent block.

Real-world control, which is now pinned in the tests: `ErikBjare/erikbjare.github.io` has **2 active workflows** and **0 checks on all 4 of its merged PRs** — a deploy-on-push repo of ours. Old predicate: True → every PR on it blocked forever. New predicate, run live: False → waived. That is the exact flip you predicted, on a real repo.

Tests: `test_push_only_workflow_repo_is_not_blocked` (helper is False *and* `evaluate_pr` stays eligible), plus `test_ungated_repo_is_waived` at the gate level.

One deliberate residual, documented in the docstring: a repo with **no merged PRs at all** is `None` (fail closed) rather than waived, since there is no evidence in either direction. Unlike the shape you flagged, it is not permanent — one manual merge establishes history and it waives from then on.

TimeToBuildBob:
Pushed `3447c826`. Both P1s are addressed, and the prior review at `9af63752` is stale — requesting a fresh pass at head.

The two P1s were the same predicate pulled in opposite directions, so neither was fixable by adjusting "does this repo have Actions workflows". Replaced it with observed PR-check history (`repo_gates_prs_with_checks`), which answers both at once. Details in the thread replies and the updated description.

Verified live before asking, on real repos rather than fixtures:
- `ErikBjare/erikbjare.github.io` (2 active workflows, 0 checks on all 4 merged PRs) — old predicate True → permanent block, new predicate False → waived. That is the P1 flip, on a repo we own.
- `ErikBjare/bob` → True, so the #1130 outage shape this PR exists to catch still fails closed.
- gptme/gptme-contrib#1232 (fork, FIRST_TIME_CONTRIBUTOR) → approval-button message; gptme/gptme-contrib#1240 (fork, MEMBER, 17 checks) → generic. Fork provenance alone does not discriminate.

35 tests, hermetic. `prek run --files` clean.

@TimeToBuildBob review

scripts/github/self-merge-check.py
```diff
@@ -1372,6 +1372,124 @@ def fetch_unresolved_human_threads(
...
+        return None
+    if not isinstance(merged_prs, list) or not merged_prs:
+        # No merged-PR history to judge from — indeterminate, not "no CI".
+        return None
+    for merged_pr in merged_prs:
+        if not isinstance(merged_pr, dict):
+            return None
+        if merged_pr.get("statusCheckRollup") or []:
+            return True
+    return False
```
TimeToBuildBob: ❌ **P1** — repo_gates_prs_with_checks() returns False when every one of the last 10 merged PRs has an empty statusCheckRollup, and evaluate_pr() then treats that False as 'this repo does not gate PRs with checks', adding only a warning and waiving the CI requirement. This is fail-open for any repo that does have PR-gating CI but whose recent merged history contains no checks: CI added after the last 10 merges, all recent merges from first-time fork authors whose workflows were never approved (the gptme-contrib#1232 pattern), or recent merges into a branch that does not trigger the workflow. A new PR in such a repo that has no checks yet (workflow pending or held) would be incorrectly judged no-CI and allowed to self-merge without a green build, recreating exactly the outage fail-open this change is meant to close. The tests cover only the push-only repo shape (correctly False) and the external-CI shape (True); none covers a gated repo whose sampled history lacks checks.

scripts/github/self-merge-check.py
```diff
@@ -1372,6 +1372,124 @@ def fetch_unresolved_human_threads(
...
+            "pr",
+            "list",
+            "--repo",
+            repo,
+            "--state",
+            "merged",
+            "--limit",
+            str(PR_CHECK_HISTORY_SAMPLE),
+            "--json",
+            "number,statusCheckRollup",
```
TimeToBuildBob, line 1569: ⚠️ **P2** — The probe samples merged PRs without considering the base branch of the PR under evaluation. gh pr list --state merged returns PRs merged into any branch; if the repo runs CI only on PRs targeting the default branch (e.g. main) and the last 10 merged PRs all went to a branch without CI (e.g. dev), statusCheckRollup is empty for all of them and repo_gates_prs_with_checks() returns False. evaluate_pr() then waives the CI requirement for a PR targeting main that has no checks yet, even though the repo does gate main PRs. The probe holds no branch information, so it cannot distinguish branch-specific CI from genuinely ungoated repos. This is a second fail-open path distinct from the just-added-CI case, reachable whenever PR activity is concentrated on a non-CI branch while the current PR targets a CI branch. The existing tests always use a single implicit branch and do not exercise any branch variation.

scripts/github/self-merge-check.py
TimeToBuildBob: Accepted and fixed in 2add141a. This was a real hole and the reasoning was right: a silent sample is not evidence of an ungated repo.

`repo_gates_prs_with_checks` no longer trusts a negative sample on its own. When every sampled PR reports zero checks it now corroborates against `repos/{repo}/actions/runs?event=pull_request`, and returns `False` only if that is definitively zero. Anything else — runs exist, or the lookup fails — is `None`, and the caller fails closed.

That closes the scenarios you named: CI added after the last N merges, and merges from unapproved first-time fork authors, both leave PR-event runs behind even when the sampled merges are silent.

The reason this does not reintroduce the *other* P1 on this PR (push-only repos permanently blocked) is that it counts pull_request-event **runs**, not workflows. A push-only repo has workflows but zero PR runs. Verified live on the same control:

| repo | pull_request runs | gates_prs |
|---|---|---|
| `ErikBjare/erikbjare.github.io` | 0 | False → still waived |
| `gptme/gptme-contrib` | 13664 | True |
| `ErikBjare/bob` | 1151 | True |

Tests: `test_silent_sample_is_indeterminate_when_pr_runs_exist`, `test_silent_sample_with_unreadable_corroboration_is_indeterminate`, and a parametrized `test_repo_has_pull_request_workflow_runs`. The push-only regression still asserts False→eligible, so both directions are pinned against each other.

One residual, documented in the docstring and deliberately fail-closed: a repo gated *exclusively* by non-Actions CI whose sample is also silent now reads `None` rather than `False`, because the corroborating endpoint only knows about Actions. That costs a manual merge, not a waived requirement.

scripts/github/self-merge-check.py
TimeToBuildBob: Accepted and fixed in 2add141a. The sample is now scoped to the base branch of the PR under evaluation.

`repo_gates_prs_with_checks(repo, base_branch)` passes `--base <branch>` to the `gh pr list` query, so it asks "do PRs *into this branch* report checks?" rather than "do PRs anywhere in this repo". `evaluate_pr` supplies it from `pr["baseRefName"]`, which was added to the `fetch_pr` field list (same request, no extra round-trip). The filter is omitted when the base is unknown, so the behaviour degrades to the previous repo-wide sample rather than sending an empty `--base`.

Test: `test_probe_scopes_sample_to_the_base_branch` pins both — that `--base master` is sent when the base is known, and that no `--base` flag appears when it is not.

Worth noting this narrows the P1 above as well: "recent merges into a branch that does not trigger the workflow" was one of the unrepresentative-sample cases it listed, and branch scoping removes it at the source rather than relying on the corroboration fallback.

TimeToBuildBob:
Pushed `2add141a`. Both findings from the `c12239de` pass are fixed; that review is now stale.

- **P1 (silent sample waived CI):** a negative sample is no longer trusted alone. `False` is returned only when `actions/runs?event=pull_request` corroborates it at zero; otherwise `None` → fail closed. Counting PR-event *runs* rather than *workflows* is what keeps this from reintroducing the push-only permanent-block P1 — verified live: erikbjare.github.io 0 runs → still waived, contrib 13664 / bob 1151 → gated.
- **P2 (branch-specific CI):** the sample is scoped with `--base`, sourced from `pr["baseRefName"]` in the existing fetch. Also removes one of the P1's unrepresentative-sample cases at the source.

42 tests in the module, 866 passed / 4 skipped on `make test-unit`.

Note for the next pass: the earlier CI red on this branch was a genuine merge-interaction bug, not flake — `_fetch_ai_review_marker_checked` (new on master via #1391) reaches `subprocess.run` via `run_gh_checked` rather than `run_gh`, so `test_evaluate_pr_is_hermetic` caught real network calls that only appeared once master was merged in. Fixed in `c12239de`.

@TimeToBuildBob review

scripts/github/self-merge-check.py
```diff
@@ -1372,6 +1372,178 @@ def fetch_unresolved_human_threads(
...
+        return None
+    for merged_pr in merged_prs:
+        if not isinstance(merged_pr, dict):
+            return None
+        if merged_pr.get("statusCheckRollup") or []:
+            return True
+    # Silent sample. Only call it "ungated" if Actions has never run on a
+    # pull_request here either; otherwise the sample is unrepresentative and
+    # the honest answer is "don't know" — which fails closed.
+    if repo_has_pull_request_workflow_runs(repo) is False:
```
TimeToBuildBob: ❌ **P1** — When `base_branch` is supplied (e.g., a PR targeting `dev`), `repo_gates_prs_with_checks` scopes the merged-PR sample with `--base dev`, but the corroborating call `repo_has_pull_request_workflow_runs(repo)` ignores the base branch entirely. For a repo that gates only `master` (CI on PRs to master) and not `dev`, all dev merges report zero checks, so the scoped sample is silent. The corroboration sees the many master-triggered runs and returns `True`, causing `repo_gates_prs_with_checks` to return `None` (indeterminate) instead of the correct `False` for the ungated dev branch. Every self-merge PR to `dev` is then permanently blocked with the "failing closed" reason, even though dev is genuinely not CI-gated. The doc claims the base-branch scoping "avoids" this (lines 1486-1489), but the global corroboration negates it; no test covers this branch-scoped silent sample with runs on another branch.

TimeToBuildBob: Accepted and fixed in f857c4f8. Correct catch, and self-inflicted — the base-branch scoping and the corroboration were reasoning about different populations, so the doc claim was indeed negated by the global call.

A silent *scoped* sample is now widened before any conclusion is drawn:

1. Checks elsewhere in the repo but none on this branch → the branch is genuinely ungated → `False`. This is the `dev` case you describe; it no longer reaches the corroboration at all.
2. Silent repo-wide too → the pull_request-run corroboration decides between "gates nothing" (`False`) and "sample is unrepresentative" (`None`).

Sampling is factored into `_sample_pr_check_history(repo, base_branch)` so the scoped and widened queries are literally the same code path and cannot drift apart again.

Widening is deliberately not an escape hatch from the P1 fix before it. Both guarded:
- `test_branch_scoped_silence_still_fails_closed_when_repo_is_silent_too` — silent on the branch *and* repo-wide, with PR runs present, still returns `None`.
- `test_branch_scoped_silence_with_unusable_wide_sample_is_indeterminate` — a failed widening query returns `None`.
- `test_branch_scoped_silence_widens_before_failing_closed` — your exact scenario, and it also asserts both queries were actually issued (scoped, then unscoped), so removing the widening fails the test rather than silently passing.

Live controls unchanged: contrib True, bob True, erikbjare.github.io False, dotfiles None. 45 tests in the module, 869 passed / 4 skipped on `make test-unit`.

TimeToBuildBob:
Pushed `f857c4f8`, fixing the P1 from the `2add141a` pass. That review is stale.

The finding was correct and self-inflicted: base-branch scoping and the repo-wide corroboration were asking about different populations, so a repo gating `master` but not `dev` would have blocked every `dev` PR forever. A silent scoped sample is now widened first — checks elsewhere but none on this branch means the branch is genuinely ungated — and only a silence that survives widening reaches the corroboration. Sampling is factored into one helper so the two queries cannot drift apart again.

@TimeToBuildBob review

TimeToBuildBob:
**Ready for a human merge. Deliberately not self-merged.**

Review at head `f857c4f8`: **5/5, 0 findings**. CI green (13 checks, `CLEAN`). All review threads resolved. Every P1 raised across the review history is fixed, with a regression test each, and the live controls are re-verified after every revision.

The self-merge gate reports two remaining reasons, and I am treating one of them as a real objection rather than an artifact:

- `Touches sensitive/security/infra paths` — **correct, and the reason I stopped.** This PR rewrites `scripts/github/self-merge-check.py`, which *is* the merge gate. Self-merging a change to the self-merge rules, evaluated by the very rules being changed, is exactly the conflict of interest that category exists to catch. A 5/5 from our own reviewer does not dissolve it.
- `AI review no consensus record (single-pass or agent review)` — noted; the review was single-pass `llm` engine. Greptile has been dark on this repo since 2026-08-09T05:27Z (suspected billing/flex budget), so there is no second independent opinion available right now. That is worth an operator check in its own right.

Summary of what changed since the two P1s that were blocking on 2026-08-09, all detailed in the resolved threads:

1. Predicate replaced — "does the repo have Actions workflows" → "do this repo's PRs report checks". The two original P1s were the same knob turned opposite ways; neither was fixable by adjusting it.
2. A silent sample is corroborated before it may waive CI, so an unrepresentative sample fails closed.
3. The sample is base-branch scoped, and a silent scoped sample is widened before any conclusion — otherwise a repo gating `master` but not `dev` blocks every `dev` PR forever.
4. The un-approved-fork-workflow case (#1232) is named explicitly instead of being misreported as "no CI configured".

One residual is documented in the docstring and deliberately fail-closed: a repo gated *exclusively* by non-Actions CI whose sample is also silent reads `None` rather than `False`, because the corroborating endpoint only knows about Actions. That costs a manual merge, not a waived requirement.

TimeToBuildBob:
**Merge-order note (automated, from the overnight operator session).**

This PR and gptme/gptme-contrib#1382 both rewrite `scripts/github/self-merge-check.py`, and their hunks overlap: both modify `evaluate_pr`, and both insert large blocks in the ~1000-1460 range. Right now each reports `MERGEABLE / CLEAN` **against master independently** — that is the trap, not the all-clear.

Two consequences for whoever merges first:

1. **The second PR will need a rebase**, and its existing review verdict is void at the new head.
2. **Re-review after the rebase, do not carry the old score forward.** Earlier today gptme-contrib#1391 sat at 4/5, was rebased onto merged #1393, and a fresh review returned **2/5 with two P1s** — interaction bugs that existed on neither PR alone. The staleness check (`reviewed <sha>, head is <sha>`) is what caught it.

Both PRs also touch the merge gate itself, so both correctly trip the sensitive-path disqualifier and want a human merge regardless of score. That classification is right, not noise: self-merging a change to the self-merge rules under those same rules is the conflict of interest the category exists to catch.

TimeToBuildBob:
**Three-way overlap on `scripts/github/self-merge-check.py` — recommended merge order.**

gptme/gptme-contrib#1382, #1405 and #1407 all rewrite this file, and all three correctly trip the sensitive-path disqualifier (they modify the merge gate itself), so all three need a human merge regardless of score. Measured hunk overlap:

| PR | regions touched | function |
|---|---|---|
| #1382 | 630, **1372–1592**, 1871 | `fetch_unresolved_human_threads`, `evaluate_pr` |
| #1405 | 120, 701, **1014–1214** (200-line insert), 1245–1333, 1910 | `fetch_ai_review_status`, `evaluate_pr` |
| #1407 | **1130–1140**, **1261–1274** | `fetch_ai_review_status` |

**#1405 and #1407 conflict directly.** #1407 edits line ~1130, which lands *inside* #1405's 200-line insertion at 1014, and both edit `fetch_ai_review_status` across 1245–1274. This is not a line-offset shift — it is the same code.

**#1382 is in a different function** and should apply with offsets only.

Suggested order — **#1382 → #1405 → #1407**:

1. **#1382** first: least entangled, already 5/5 with 0 findings, CI green, threads resolved.
2. **#1405** second: it is the large structural change (disposition-based gate). Merging the big one before the small one means the *small* PR rebases, not the large one.
3. **#1407** last: 2 hunks, cheapest to rebase.

**Re-review each after its rebase — do not carry the prior verdict forward.** On 2026-08-10 gptme-contrib#1391 sat at 4/5, was rebased onto merged #1393, and a fresh review returned **2/5 with two P1s** — interaction bugs present on neither PR alone. The staleness check (`reviewed <sha>, head is <sha>`) is what caught it.

Worth considering instead: #1405 and #1407 are close enough in scope that folding #1407's two P2 fixes into #1405 may be cheaper than a rebase plus a second review cycle.

PR #1383, opened 2026-08-06
fix(gate): detect CI that is terminal-but-not-green, or stuck

TimeToBuildBob:
## Problem

`check_ci_failures` matched only `any(.conclusion == "FAILURE")`. The 2026-08-06 GitHub Actions outage (7h+, webhooks throttled to ~15%) produced two PR shapes in bulk that it misses — **and that no other gate predicate covers either**, so the PR is stranded with no path back into PM:

| shape | why every predicate skips it |
|---|---|
| **terminal but not green** — CANCELLED / TIMED_OUT / ACTION_REQUIRED, nothing left running | PR is `UNSTABLE`: `check_merge_ready` wants CLEAN, `check_merge_conflicts` wants DIRTY, `check_ci_failures` wants FAILURE. Once `updatedAt` stops moving, `check_pr_updates` stops firing too. |
| **permanently QUEUED** — orphaned jobs behind the outage | same, plus the checks never reach a conclusion at all |

Verified live on **gptme/gptme#3472**: `mergeStateStatus=UNSTABLE`, `mergeable=MERGEABLE`, one CANCELLED check, three QUEUED for ~6h, greptile 5/5, and a stored watermark byte-identical to `updatedAt`. It matches none of the seven per-PR predicates and cannot re-enter PM without a human push or comment.

## Change

Replaces the boolean with a three-way classifier — `green` / `inflight` / `bad`:

- **FAILURE still fires immediately**, even with other checks running. Existing behaviour is byte-for-byte preserved.
- **CANCELLED and friends count only once nothing is in flight.** This matters: repos using `concurrency: cancel-in-progress: true` leave a CANCELLED check behind on *every* supersede while the replacement run starts. Firing on "any CANCELLED" would emit on every force-push. There is an explicit test for this.
- **in-flight is emitted only after `CI_STUCK_SECS`** (default 3600) with an unchanged hash, using the state file's mtime as the clock, then re-stamps so a still-stuck PR re-emits at most once per interval rather than every gate run.

Also handles the `StatusContext` rollup shape (legacy commit statuses carry `.state`, not `.conclusion`) — previously invisible to this predicate entirely.

## Tests

13 new in `tests/test_activity_gate_stuck_ci.py`, including the exact #3472 rollup and the supersede-noise case. The test extracts the jq program and the function body straight out of `activity-gate.sh` so it cannot drift from the implementation.

The 50 existing `activity_gate` tests still pass. `shellcheck -S warning` clean.

## Notes for review

- The staleness threshold is a judgement call. 1h is deliberately conservative — long CI runs are common and a false "stuck" costs a wasted session. Tunable via `CI_STUCK_SECS`.
- Detection only. Nothing here re-triggers a run; it just stops these PRs being invisible. The re-trigger primitive lives in Bob's repo (`scripts/ci/retrigger-pr-checks.sh`) and is not wired to the gate yet.
- Related: #1382 (fail closed when a repo has workflows but no checks ran) covers the adjacent "zero checks" shape at the self-merge gate rather than the activity gate.

### DIFF
--- a/scripts/github/activity-gate.sh
+++ b/scripts/github/activity-gate.sh
@@ -608,6 +608,12 @@ check_ci_failures() {
     now=$(date +%s)
 
     echo "$prs" | jq -c '.[]' | while read -r pr_data; do
+        local pr_number pr_title ci_hash state_file detail
+        # Derived before the classification branches so the green path below can
+        # clear the same file the stuck path writes — one naming site, no drift.
+        pr_number=$(echo "$pr_data" | jq -r '.number')
+        state_file="$STATE_DIR/${repo//\//-}-pr-${pr_number}-ci.state"
+
         # Three-way classification. Previously this matched only
         # `any(.conclusion == "FAILURE")`, which stranded two shapes the
         # 2026-08-06 GitHub Actions outage produced in bulk:
@@ -623,14 +629,22 @@ check_ci_failures() {
         # preserves the previous behaviour exactly. CANCELLED only counts once
         # nothing is in flight — otherwise every `cancel-in-progress` supersede
         # (which leaves CANCELLED behind while the new run starts) would fire.
+        #
+        # The in-flight arm covers both non-terminal StatusState values, not just
+        # PENDING: EXPECTED is what a required status context reports before it
+        # has run at all. An EXPECTED item has no `.status` and no `.conclusion`,
+        # so without this it fell out of `$inflight` while `$bad` still saw a
+        # non-green `.state` — i.e. an instant `ci_failure` on a PR whose checks
+        # simply had not started yet.
         local ci_state
         ci_state=$(echo "$pr_data" | jq -r '
             select(.statusCheckRollup != null) | .statusCheckRollup
-            | (any(.[]; ((.conclusion // "") | ascii_upcase) == "FAILURE")) as $hard
+            | (any(.[]; ((.conclusion // .state // "") | ascii_upcase) == "FAILURE")) as $hard
             | (any(.[]; ((.status // "") | ascii_upcase) as $s
                    | $s == "QUEUED" or $s == "IN_PROGRESS" or $s == "WAITING"
                      or (((.conclusion // "") == "")
-                         and (((.state // "") | ascii_upcase) == "PENDING")))) as $inflight
+                         and (((.state // "") | ascii_upcase) as $st
+                              | $st == "PENDING" or $st == "EXPECTED")))) as $inflight
             | (any(.[]; ((.conclusion // .state // "") | ascii_upcase) as $c
                    | $c != "" and $c != "SUCCESS" and $c != "SKIPPED"
                      and $c != "NEUTRAL")) as $bad
@@ -640,10 +654,17 @@ check_ci_failures() {
               else "green" end')
 
         [ -n "$ci_state" ] || continue
-        [ "$ci_state" = "green" ] && continue
+        if [ "$ci_state" = "green" ]; then
+            # Close the episode. The stuck clock is the state file's mtime, so a
+            # file left behind from an earlier in-flight sighting would make the
+            # *next* run that happens to hash the same look hours old — and
+            # "CI stuck: checks in flight" would fire on a run that started
+            # minutes ago. Deleting on green makes the next in-flight sighting a
+            # genuine first sighting.
+            rm -f "$state_file"
+            continue
+        fi
 
-        local pr_number pr_title ci_hash state_file detail
-        pr_number=$(echo "$pr_data" | jq -r '.number')
         pr_title=$(echo "$pr_data" | jq -r '.title')
         # Hash the CI conclusions to detect state changes.
         # Note: in-progress checks have empty conclusions (mapped to "pending").
@@ -658,8 +679,18 @@ check_ci_failures() {
         # dedup below silently swallowed the `ci_failure` that the classifier
         # above had correctly identified as `bad`. Keep this field list in sync
         # with the `ci_state` classifier.
-        ci_hash=$(echo "$pr_data" | jq -r '[.statusCheckRollup[] | .conclusion // .state // "pending"] | sort | join(",")' | portable_hash)
-        state_file="$STATE_DIR/${repo//\//-}-pr-${pr_number}-ci.state"
+        #
+        # `headRefOid` is part of the hash so a push starts a new episode
+        # *explicitly*, rather than relying on the gate having observed the green
+        # window in between. Green windows are easy to miss: the whole
+        # pending -> green -> force-push -> pending cycle can fit between two
+        # gate runs, leaving a "pending"-hashed file with an hours-old mtime that
+        # the new run matches and is instantly declared stuck. Deleting the file
+        # on green (above) fixes the observed case; keying on the head SHA fixes
+        # the unobserved one. Trade-off: a force-push that reproduces the same
+        # failure now re-emits `ci_failure`. That is the desired behaviour — new
+        # commit, still red, is a fresh actionable signal, not a duplicate.
+        ci_hash=$(echo "$pr_data" | jq -r '(.headRefOid // "") + ":" + ([.statusCheckRollup[] | .conclusion // .state // "pending"] | sort | join(","))' | portable_hash)
         detail="CI failing"
 
         if [ "$ci_state" = "inflight" ]; then
--- a/scripts/github/activity-gate.sh
+++ b/scripts/github/activity-gate.sh
@@ -636,12 +636,23 @@ check_ci_failures() {
         # so without this it fell out of `$inflight` while `$bad` still saw a
         # non-green `.state` — i.e. an instant `ci_failure` on a PR whose checks
         # simply had not started yet.
+        #
+        # The `.status` list must be the FULL non-terminal half of GraphQL's
+        # CheckStatusState enum (QUEUED, IN_PROGRESS, WAITING, PENDING,
+        # REQUESTED) — COMPLETED is the only terminal member. Listing only the
+        # first three silently mapped PENDING/REQUESTED CheckRuns to `green`:
+        # they match no `.status` literal, and a CheckRun carries no `.state`,
+        # so `$bad`'s `.conclusion // .state // ""` is "" and its `$c != ""`
+        # guard fails too — every predicate false, falling through to `green`.
+        # That deleted the stuck clock (`rm -f` below) on exactly the
+        # not-yet-started shape this function exists to catch.
         local ci_state
         ci_state=$(echo "$pr_data" | jq -r '
             select(.statusCheckRollup != null) | .statusCheckRollup
             | (any(.[]; ((.conclusion // .state // "") | ascii_upcase) == "FAILURE")) as $hard
             | (any(.[]; ((.status // "") | ascii_upcase) as $s
                    | $s == "QUEUED" or $s == "IN_PROGRESS" or $s == "WAITING"
+                     or $s == "PENDING" or $s == "REQUESTED"
                      or (((.conclusion // "") == "")
                          and (((.state // "") | ascii_upcase) as $st
                               | $st == "PENDING" or $st == "EXPECTED")))) as $inflight
--- a/scripts/github/activity-gate.sh
+++ b/scripts/github/activity-gate.sh
@@ -601,33 +601,85 @@ check_ci_failures() {
     local prs=$2
     [ "$prs" = "[]" ] || [ -z "$prs" ] && return 0
 
-    echo "$prs" | jq -c '.[]' | while read -r pr_data; do
-        local has_failures
-        has_failures=$(echo "$pr_data" | jq -r 'select(.statusCheckRollup != null) | .statusCheckRollup | any(.conclusion == "FAILURE")')
-        if [ "$has_failures" = "true" ]; then
-            local pr_number pr_title ci_hash state_file
-            pr_number=$(echo "$pr_data" | jq -r '.number')
-            pr_title=$(echo "$pr_data" | jq -r '.title')
-            # Hash the CI conclusions to detect state changes.
-            # Note: in-progress checks have empty conclusions (mapped to "pending").
-            # When they complete, the hash changes — this can cause a re-trigger even
-            # if the final result matches the previous run. Acceptable trade-off vs
-            # missing genuine state changes.
-            ci_hash=$(echo "$pr_data" | jq -r '[.statusCheckRollup[] | .conclusion // "pending"] | sort | join(",")' | portable_hash)
-            state_file="$STATE_DIR/${repo//\//-}-pr-${pr_number}-ci.state"
+    # A PR whose checks sit in QUEUED/IN_PROGRESS with an unchanged hash for this
+    # long is stuck, not progressing, and needs a re-trigger rather than patience.
+    local stuck_secs="${CI_STUCK_SECS:-3600}"
+    local now
+    now=$(date +%s)
 
-            if [ -f "$state_file" ]; then
-                local last_hash
-                last_hash=$(cat "$state_file")
-                if [ "$ci_hash" = "$last_hash" ]; then
-                    # CI state unchanged since last check — don't re-trigger
-                    continue
-                fi
+    echo "$prs" | jq -c '.[]' | while read -r pr_data; do
+        # Three-way classification. Previously this matched only
+        # `any(.conclusion == "FAILURE")`, which stranded two shapes the
+        # 2026-08-06 GitHub Actions outage produced in bulk:
+        #
+        #   - terminal but not green (CANCELLED / TIMED_OUT / ACTION_REQUIRED and
+        #     nothing left running). The PR sits UNSTABLE, so check_merge_ready
+        #     (needs CLEAN) skips it, check_merge_conflicts (needs DIRTY) skips
+        #     it, and once updatedAt stops moving check_pr_updates skips it too.
+        #   - permanently QUEUED. Live example: gptme/gptme#3472 had 3 checks
+        #     QUEUED for ~6h behind cancelled/orphaned jobs, matching no predicate.
+        #
+        # FAILURE still fires immediately even with other checks running, which
+        # preserves the previous behaviour exactly. CANCELLED only counts once
+        # nothing is in flight — otherwise every `cancel-in-progress` supersede
+        # (which leaves CANCELLED behind while the new run starts) would fire.
+        local ci_state
+        ci_state=$(echo "$pr_data" | jq -r '
+            select(.statusCheckRollup != null) | .statusCheckRollup
+            | (any(.[]; ((.conclusion // "") | ascii_upcase) == "FAILURE")) as $hard
+            | (any(.[]; ((.status // "") | ascii_upcase) as $s
+                   | $s == "QUEUED" or $s == "IN_PROGRESS" or $s == "WAITING"
+                     or (((.conclusion // "") == "")
+                         and (((.state // "") | ascii_upcase) == "PENDING")))) as $inflight
+            | (any(.[]; ((.conclusion // .state // "") | ascii_upcase) as $c
+                   | $c != "" and $c != "SUCCESS" and $c != "SKIPPED"
+                     and $c != "NEUTRAL")) as $bad
+            | if $hard then "bad"
+              elif ($bad and ($inflight | not)) then "bad"
+              elif $inflight then "inflight"
+              else "green" end')
+
+        [ -n "$ci_state" ] || continue
+        [ "$ci_state" = "green" ] && continue
+
+        local pr_number pr_title ci_hash state_file detail
+        pr_number=$(echo "$pr_data" | jq -r '.number')
+        pr_title=$(echo "$pr_data" | jq -r '.title')
+        # Hash the CI conclusions to detect state changes.
+        # Note: in-progress checks have empty conclusions (mapped to "pending").
+        # When they complete, the hash changes — this can cause a re-trigger even
+        # if the final result matches the previous run. Acceptable trade-off vs
+        # missing genuine state changes.
+        ci_hash=$(echo "$pr_data" | jq -r '[.statusCheckRollup[] | .conclusion // "pending"] | sort | join(",")' | portable_hash)
+        state_file="$STATE_DIR/${repo//\//-}-pr-${pr_number}-ci.state"
+        detail="CI failing"
+
+        if [ "$ci_state" = "inflight" ]; then
+            # Checks still running is normal. Only actionable once the state has
+            # not moved for stuck_secs — the state file's mtime is the clock.
+            if [ -f "$state_file" ] && [ "$(cat "$state_file")" = "$ci_hash" ]; then
+                local mtime age
+                mtime=$(stat -c %Y "$state_file" 2>/dev/null || stat -f %m "$state_file" 2>/dev/null || echo "$now")
+                age=$(( now - mtime ))
+                [ "$age" -ge "$stuck_secs" ] || continue
+                detail="CI stuck: checks in flight ${age}s with no state change"
+                # Re-stamp so a still-stuck PR re-emits at most once per
+                # stuck_secs rather than on every gate run.
+                touch "$state_file"
+            else
+                # First sighting of this in-flight state: record and wait.
+                echo "$ci_hash" > "$state_file"
+                continue
             fi
-            # New failure or CI state changed
+        else
+            if [ -f "$state_file" ] && [ "$(cat "$state_file")" = "$ci_hash" ]; then
+                # CI state unchanged since last check — don't re-trigger
+                continue
+            fi
             echo "$ci_hash" > "$state_file"
-            emit_item "ci_failure" "$repo" "$pr_number" "$pr_title" "CI failing"
         fi
+
+        emit_item "ci_failure" "$repo" "$pr_number" "$pr_title" "$detail"
     done
 }
 
--- a/scripts/github/activity-gate.sh
+++ b/scripts/github/activity-gate.sh
@@ -667,7 +667,7 @@ check_ci_failures() {
             # not moved for stuck_secs — the state file's mtime is the clock.
             if [ -f "$state_file" ] && [ "$(cat "$state_file")" = "$ci_hash" ]; then
                 local mtime age
-                mtime=$(stat -c %Y "$state_file" 2>/dev/null || stat -f %m "$state_file" 2>/dev/null || echo "$now")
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:

Evidence: commit: 20aae939dc docs(journal): gptme-contrib#1383 — triggered Greptile review

Session: `793814e8-5bfd-509d-a682-ddad8201cdda`

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Status: CI is fully green (pre-commit, integration-tests) and this PR is `mergeStateStatus=CLEAN`/`mergeable=MERGEABLE`. Not self-merging because Greptile review is missing — but that's not a quality signal on this PR: Greptile has gone silent on `gptme/gptme-contrib` specifically (confirmed repo-specific, not a global outage — it answered `gptme/gptme-cloud` within the last few hours) for ~29h, with 5 PRs stuck awaiting initial review: #1380, #1382, #1383, #1384, #1385.

Re-triggering `@greptileai review` again won't help — already tried once (07:19:40 UTC) with no response. This needs the Greptile app config for this repo checked (enabled/indexed, repo list, plan repo-count limit) or a manual merge decision.

Filing a tracking task for the repo-wide Greptile outage separately.

TimeToBuildBob (commented):
🤖 1 inline finding(s) — see the summary comment.

TimeToBuildBob:
## 🤖 AI code review

### Confidence Score: 3/5

**Not ready to merge.** One P1 defect on this head should be fixed before merge.

**Files Needing Attention:**

- `scripts/github/activity-gate.sh`

### What changed

This PR reworks check_ci_failures in scripts/github/activity-gate.sh from a binary FAILURE predicate into a three-way classifier (green/inflight/bad), adding a stuck-CI detection mechanism that emits after CI_STUCK_SECS with an unchanged hash, using the state file's mtime as the clock. The classifier now handles legacy StatusContext shapes (via .state) and all five non-terminal CheckStatusState values, and the ci_hash includes headRefOid so pushes start new episodes. Adds a new test file with 13 tests covering classification, the stuck-clock logic, green cleanup, and hash behavior.

- `scripts/github/activity-gate.sh` — Replaces the FAILURE-only predicate with a three-way ci_state classifier, adds inflight/stuck handling and hash changes for headRefOid and .state.
- `tests/test_activity_gate_stuck_ci.py` — Adds tests for ci_state classification, stuck emission thresholds, green state cleanup, and hash tracking for StatusContext/CheckRun/headRefOid.

**3 finding(s)** — ❌ **1** P1 · ⚠️ **2** P2

🛑 **1 blocking** — worth resolving before merge.

<details><summary><b>3 finding(s) posted inline on the diff</b> — expand to read them here</summary>

### ❌ P1 high — `scripts/github/activity-gate.sh:659`

The three-way classifier's `$bad` predicate treats any non-empty, non-SUCCESS/SKIPPED/NEUTRAL conclusion as bad, which includes ACTION_REQUIRED. This is intentional per the PR description, but it also includes the value `STARTUP_FAILURE`. The GitHub API's `conclusion` enum includes `ACTION_REQUIRED`, `CANCELLED`, `FAILURE`, `NEUTRAL`, `SKIPPED`, `STALE`, `STARTUP_FAILURE`, `SUCCESS`, `TIMED_OUT`. The `$bad` list does NOT exclude `STALE`. A stale CheckRun has a conclusion of `STALE`, which represents a check that was superseded by a newer run. If a PR has a STALE check and no other in-flight checks (e.g., a cancelled run left a STALE conclusion and the new run hasn't started yet), it classifies as `bad` and emits immediately, even though the situation is transient and similar to the cancel-in-progress case the code explicitly avoids for CANCELLED. This can cause spurious ci_failure emissions on PRs that are just between runs, leading to wasted sessions. The fix is to add STALE to the `$bad` list or treat it as non-bad.

```suggestion
Change `$c != "SUCCESS" and $c != "SKIPPED" and $c != "NEUTRAL"` to also exclude `$c != "STALE"` (and perhaps STARTUP_FAILURE, though that is usually terminal).
```

<sub>**How this was verified:** Checked the GitHub API conclusion enum documented in the code comments; STALE is a known value. The test suite does not cover this case.</sub>

### ⚠️ P2 medium — `scripts/github/activity-gate.sh:718`

The stuck-clock logic in the inflight branch uses `touch "$state_file"` to re-stamp the mtime after emitting, aiming to re-emit at most once per CI_STUCK_SECS. However, the re-stamp happens unconditionally regardless of whether the emit actually occurred. Specifically, when the state file is older than stuck_secs, the branch does NOT `continue` (it falls through to emit_item), so the `touch` executes. But if the emission is skipped because the PR is actually progressing (e.g., the hash changed between runs, causing the `if` condition to be false and the `else` branch to write a new hash and `continue`), that's handled. The bug is that when the gate runs more frequently than the threshold but the hash changes (say every 30 minutes), the first run with a new hash writes the state file, the second run 30 minutes later sees age=1800 &lt; 3600 and does `continue` without emitting, which is correct. The issue is the re-stamp: after emitting once at age 3600, the touch sets mtime to now. But if the next run happens at age 3599 (the gate runs every 2 minutes), it will `continue` without emitting, which is correct. The real problem is that if the state file is missing (first sighting), the `else` branch writes it and continues, which is correct. The mechanism is sound; however, there's a race: if two parallel gate processes run concurrently (the gate uses parallel repo processing, but each repo is separate), the `echo` and `touch` operations are NOT atomic, so a read between them could see a stale mtime. This is a theoretical race, not a practical one given the 2-minute cadence, but it's worth noting. I could not find a concrete defect here.

<sub>**How this was verified:** Checked the branch logic: lines 710-718 show the conditional, and the touch is only reached when age &gt;= stuck_secs. No defect found.</sub>

### ⚠️ P2 medium — `scripts/github/activity-gate.sh:704`

The `ci_hash` uses `.headRefOid // ""` before the colon. If `headRefOid` is missing or null, it becomes an empty string, so two PRs with different actual head commits but missing headRefOid (which can happen if the API payload is malformed or the field is not selected) would hash identically if their rollup conclusions match. More critically, the state file is shared across runs of the same repo and PR number. If headRefOid is missing on one run and present on the next, the hash changes and a new episode starts, which is safe. But if headRefOid is consistently missing, the hash does not change on a push, defeating the purpose of including headRefOid. Checking the fetch_pr_data_with_ttl function, it explicitly requests headRefOid (line 370), so it should always be present for real PRs. However, the cache could serve stale payloads where headRefOid was present. This is not a likely defect for this gate, so I am downgrading to P2.

<sub>**How this was verified:** Checked the fetch query includes headRefOid; the PR data is fetched with that field. The risk is low.</sub>

</details>

| `a8e56c4015d2` | ? | 1 | llm | 2026-08-07 20:50 UTC |
| `a8e56c4015d2` | ? | 3 | llm | 2026-08-07 21:01 UTC |
| `8cd57907ecf8` | ? | 3 | llm | 2026-08-07 22:01 UTC |
| `7d1e7278a980` | ? | 3 | llm | 2026-08-07 22:20 UTC |
| `6d80ce8e436e` | ? | 3 | llm | 2026-08-08 07:44 UTC |

</details>

---
<sub>Reviewed `860a3603b5ba` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 77s · self-hosted reviewer ([bob#1122](https://github.com/ErikBjare/bob/issues/1122)). Findings are advisory — a human still decides.</sub>

scripts/github/activity-gate.sh
```diff
@@ -601,33 +601,93 @@ check_ci_failures() {
...
+        # nothing is in flight — otherwise every `cancel-in-progress` supersede
+        # (which leaves CANCELLED behind while the new run starts) would fire.
+        local ci_state
+        ci_state=$(echo "$pr_data" | jq -r '
+            select(.statusCheckRollup != null) | .statusCheckRollup
+            | (any(.[]; ((.conclusion // "") | ascii_upcase) == "FAILURE")) as $hard
+            | (any(.[]; ((.status // "") | ascii_upcase) as $s
+                   | $s == "QUEUED" or $s == "IN_PROGRESS" or $s == "WAITING"
+                     or (((.conclusion // "") == "")
+                         and (((.state // "") | ascii_upcase) == "PENDING")))) as $inflight
```
TimeToBuildBob: Fixed in 6d80ce8e. `EXPECTED` is now part of the pending arm of `$inflight` (`$st == "PENDING" or $st == "EXPECTED"`), so a required status context that hasn't reported yet no longer reads as a non-green terminal result; added classifier cases for EXPECTED alone, EXPECTED + a green check, and EXPECTED + a real FAILURE (still `bad` via $hard). Verified the new cases fail with only that predicate reverted.

scripts/github/activity-gate.sh
```diff
@@ -601,33 +601,93 @@ check_ci_failures() {
...
+        ci_hash=$(echo "$pr_data" | jq -r '[.statusCheckRollup[] | .conclusion // .state // "pending"] | sort | join(",")' | portable_hash)
+        state_file="$STATE_DIR/${repo//\//-}-pr-${pr_number}-ci.state"
+        detail="CI failing"
+
+        if [ "$ci_state" = "inflight" ]; then
+            # Checks still running is normal. Only actionable once the state has
+            # not moved for stuck_secs — the state file's mtime is the clock.
+            if [ -f "$state_file" ] && [ "$(cat "$state_file")" = "$ci_hash" ]; then
+                local mtime age
+                mtime=$(stat -c %Y "$state_file" 2>/dev/null || stat -f %m "$state_file" 2>/dev/null) || continue
```
TimeToBuildBob, line 712: Fixed in 6d80ce8e — both halves, because the `rm -f` alone doesn't cover the case where the gate never observes the green window.

1. Green path now does `rm -f "$state_file"`. `pr_number`/`state_file` are derived once above the classifier, so the green path clears exactly the file the stuck path writes (single naming site, no drift).
2. `headRefOid` is now part of `ci_hash`: `(.headRefOid // "") + ":" + ([.statusCheckRollup[] | .conclusion // .state // "pending"] | sort | join(","))`. The whole pending -> green -> force-push -> pending cycle can fit between two gate runs, in which case the cleanup never fires and the mtime still leaks across episodes. Keying on the head SHA makes the boundary explicit rather than observation-dependent.

On the dedup behaviour change you flagged: a force-push that reproduces the same failure now re-emits `ci_failure`. Taking that deliberately — a new commit that is still red is a fresh actionable signal, not a duplicate; the thing dedup exists to suppress is re-reporting the *same* run. Reasoning is in the commit message and in a comment at the hash site.

Tests: `test_green_clears_state_file` (fails without the `rm -f`), `test_ci_hash_changes_on_new_head_commit` and `test_new_head_commit_resets_the_stuck_clock` (the latter reproduces your exact sequence end-to-end through the real shell function and reports `CI stuck: checks in flight 7200s` without the SHA in the hash). `_ci_hash_key` takes a `head_sha` arg now.

tests/test_activity_gate_stuck_ci.py
```diff
@@ -0,0 +1,231 @@
...
+    state_dir.mkdir()
+    # Pre-seed the state file with the current hash, aged to taste. Compute it
+    # through the identical shell pipeline — hashing the jq output *without* its
+    # trailing newline yields a different digest, the state file then looks like
+    # a first sighting, and the test silently stops exercising the stuck path.
+    digest = subprocess.run(
+        [
+            "bash",
+            "-c",
+            f"jq -r '{_extract_hash_program()}' | sha256sum | cut -d' ' -f1",
```
TimeToBuildBob: Fixed in 6d80ce8e. Added a module-level `_sha256_cmd()` that returns `sha256sum` if present, else `shasum -a 256`, else skips. Both the seeded state-file digest and the `portable_hash()` stub injected into the generated bash script go through that one helper, so they cannot diverge — which matters, since a mismatch makes the state file look like a first sighting and the test silently stops exercising the stuck path (the point the existing comment makes). Wrote it fresh rather than using the suggested diff.

scripts/github/activity-gate.sh
```diff
@@ -601,33 +601,124 @@ check_ci_failures() {
...
+        ci_hash=$(echo "$pr_data" | jq -r '(.headRefOid // "") + ":" + ([.statusCheckRollup[] | .conclusion // .state // "pending"] | sort | join(","))' | portable_hash)
+        detail="CI failing"
+
+        if [ "$ci_state" = "inflight" ]; then
+            # Checks still running is normal. Only actionable once the state has
+            # not moved for stuck_secs — the state file's mtime is the clock.
+            if [ -f "$state_file" ] && [ "$(cat "$state_file")" = "$ci_hash" ]; then
+                local mtime age
+                mtime=$(stat -c %Y "$state_file" 2>/dev/null || stat -f %m "$state_file" 2>/dev/null) || continue
+                age=$(( now - mtime ))
```
TimeToBuildBob, line 713: ⚠️ **P2** — In the `inflight` branch, the `mtime` variable is read from the state file, but the `age` is computed as `now - mtime`, relying on the file's mtime being the time the state was last written. However, the code re-stamps the state file with `touch` only after emitting the stuck item. When the state file is first created (first sighting), its mtime is set to the current time. If the gate has been running for a long time (e.g., with a long-running process), the `now` variable captured at line 608 could be significantly later than the file's mtime, but the test simulates this by setting mtime to `now - age`. In practice, the gate runs periodically, so the wall-clock difference between `now` and the file creation mtime is small. However, in a scenario where the gate's process runs continuously (as in a long-lived monitoring loop), the `now` is captured once at the start, but the file's mtime is set when the state file is written later in the same run. This mismatch is not an issue because the file write happens after `now` is captured, so `age` is negative, and the `[ "$age" -ge "$stuck_secs" ]` check fails, correctly skipping. The real issue is that the `now` variable is not refreshed between iterations of the `while read` loop; if a single run processes many PRs and takes several minutes, the `now` captured at the top becomes stale, making `age` slightly larger than the true age. Over a long run, this could cause a PR that has been stuck for just under the threshold to be incorrectly emitted as stuck, but the threshold is 3600s, so the error is at most a few seconds. This is a minor correctness issue with low practical impact.

scripts/github/activity-gate.sh
```diff
@@ -601,33 +601,124 @@ check_ci_failures() {
...
+                # Re-stamp so a still-stuck PR re-emits at most once per
+                # stuck_secs rather than on every gate run.
+                touch "$state_file"
+            else
+                # First sighting of this in-flight state: record and wait.
+                echo "$ci_hash" > "$state_file"
+                continue
+            fi
+        else
+            if [ -f "$state_file" ] && [ "$(cat "$state_file")" = "$ci_hash" ]; then
```
TimeToBuildBob, line 725: ⚠️ **P2** — The `ci_hash` computation in the `bad` branch uses the state file to dedupe unchanged failures. If a PR transitions from `inflight` to `bad` (e.g., checks change from QUEUED to FAILURE), the hash changes because the rollup conclusions change. However, if the PR transitions from `bad` back to `inflight` (e.g., a failure is fixed and a new run starts), the state file still contains the old hash from the `bad` state. When the new `inflight` state is seen, the code at line 699 checks if the stored hash equals the new hash; it will not, so it takes the `else` branch and writes the new hash as a first sighting, resetting the stuck clock. This is correct. But consider a PR that goes from `inflight` (hash A) to `bad` (hash B) to `inflight` again (hash C). When the `bad` state emits, it overwrites the state file with hash B. Then when `inflight` with hash C is seen, it is treated as a first sighting, resetting the clock. This is fine. The issue arises when `bad` and `inflight` share the same hash: this can happen if the rollup contains both a CANCELLED check and an IN_PROGRESS check. The classifier returns `inflight` (because `$bad` is true but `$inflight` is also true), but the hash is computed from the rollup, which includes the hash of the CANCELLED check. If the same rollup shape is seen later as `bad` (e.g., the IN_PROGRESS check completes with CANCELLED), the hash changes because the IN_PROGRESS item's hash is different from the CANCELLED item's hash. So dedup works across these states. However, the more subtle issue is that the `bad` branch does not delete the state file when a PR goes green, but the green branch does. If a PR goes from `bad` to `green`, the state file is removed, so a future `inflight` sighting is fresh. This is correct. The potential issue is that the `bad` branch only writes the hash when the hash changes; if the same `bad` state persists, it doesn't re-emit, which is correct. I don't see a concrete bug here, so I won't report.

tests/test_activity_gate_stuck_ci.py
```diff
@@ -0,0 +1,402 @@
...
+    state_dir.mkdir()
+    # Pre-seed the state file with the current hash, aged to taste. Compute it
+    # through the identical shell pipeline — hashing the jq output *without* its
+    # trailing newline yields a different digest, the state file then looks like
+    # a first sighting, and the test silently stops exercising the stuck path.
+    digest = subprocess.run(
+        [
+            "bash",
+            "-c",
+            f"jq -r '{_extract_hash_program()}' | {_sha256_cmd()} | cut -d' ' -f1",
```
TimeToBuildBob, line 235: ⚠️ **P2** — The new classifier treats any rollup with a `$bad` item (e.g., CANCELLED) and no `$inflight` item as `bad`, even if there are also green checks. This is intentional. However, consider a PR with a CANCELLED check and an EXPECTED status context. The EXPECTED context has no `.status` and no `.conclusion`, so `$inflight` is false (the EXPECTED arm only matches `.status == PENDING` or `.state` of EXPECTED, but the `inflight` predicate checks `.status` or `.state` for EXPECTED, and since `.status` is empty and `.state` is

scripts/github/activity-gate.sh
TimeToBuildBob: Retracting — the direction is inverted, and the finding rates itself as negligible.

`now=$(date +%s)` is captured *before* the `while read -r pr_data` loop (line 606-ish, loop starts at 610). So on a long run `now` is **earlier** than wall clock, which makes `age = now - mtime` an **under**-estimate, not an over-estimate. The consequence is the opposite of the one described: a genuinely stuck PR is emitted a few seconds *late*, never early. A PR "stuck for just under the threshold" cannot be emitted by a stale `now` — the error only ever subtracts.

That is also the conservative direction for this gate (late emit costs a cycle; early emit wastes a dispatch, which is the failure this PR exists to prevent).

The comment itself concludes "a minor correctness issue with low practical impact" after walking through two scenarios it then dismisses. No change.

scripts/github/activity-gate.sh
TimeToBuildBob: Retracting — this one retracts itself.

The comment walks the `inflight`/`bad`/`green` hash transitions, confirms each is handled correctly ("This is correct", "This is fine", "So dedup works across these states"), and closes with:

> I don't see a concrete bug here, so I won't report.

It was posted as an inline finding anyway. Nothing to act on. (Worth noting upstream in the reviewer: a pass that concludes "I won't report" should not emit a finding.)

tests/test_activity_gate_stuck_ci.py
TimeToBuildBob: Retracting — the premise is factually wrong, and the comment is truncated mid-sentence.

It asserts:

> The EXPECTED context has no `.status` and no `.conclusion`, so `$inflight` is false

That is the pre-fix behaviour. Adding `EXPECTED` to the in-flight arm is precisely what this commit changed, so `$inflight` is now **true** for an EXPECTED status context.

Verified by running the actual classifier (extracted from the script, not reasoned about):

```
CANCELLED + EXPECTED     -> inflight
EXPECTED alone           -> inflight
CANCELLED alone          -> bad
EXPECTED + FAILURE       -> bad
```

So the scenario in the finding — CANCELLED plus EXPECTED — classifies `inflight`, not `bad`. A required check that has not started yet keeps the PR in flight until `CI_STUCK_SECS`, which is the intended behaviour; a real FAILURE still wins immediately.

The body also ends mid-sentence at "and `.state` is", so this pass emitted an incomplete finding. No change.

scripts/github/activity-gate.sh
```diff
@@ -601,33 +601,135 @@ check_ci_failures() {
...
+            # not moved for stuck_secs — the state file's mtime is the clock.
+            if [ -f "$state_file" ] && [ "$(cat "$state_file")" = "$ci_hash" ]; then
+                local mtime age
+                mtime=$(stat -c %Y "$state_file" 2>/dev/null || stat -f %m "$state_file" 2>/dev/null) || continue
+                age=$(( now - mtime ))
+                [ "$age" -ge "$stuck_secs" ] || continue
+                detail="CI stuck: checks in flight ${age}s with no state change"
+                # Re-stamp so a still-stuck PR re-emits at most once per
+                # stuck_secs rather than on every gate run.
+                touch "$state_file"
```
TimeToBuildBob, line 718: ⚠️ **P2** — The stuck-clock logic in the inflight branch uses `touch "$state_file"` to re-stamp the mtime after emitting, aiming to re-emit at most once per CI_STUCK_SECS. However, the re-stamp happens unconditionally regardless of whether the emit actually occurred. Specifically, when the state file is older than stuck_secs, the branch does NOT `continue` (it falls through to emit_item), so the `touch` executes. But if the emission is skipped because the PR is actually progressing (e.g., the hash changed between runs, causing the `if` condition to be false and the `else` branch to write a new hash and `continue`), that's handled. The bug is that when the gate runs more frequently than the threshold but the hash changes (say every 30 minutes), the first run with a new hash writes the state file, the second run 30 minutes later sees age=1800 < 3600 and does `continue` without emitting, which is correct. The issue is the re-stamp: after emitting once at age 3600, the touch sets mtime to now. But if the next run happens at age 3599 (the gate runs every 2 minutes), it will `continue` without emitting, which is correct. The real problem is that if the state file is missing (first sighting), the `else` branch writes it and continues, which is correct. The mechanism is sound; however, there's a race: if two parallel gate processes run concurrently (the gate uses parallel repo processing, but each repo is separate), the `echo` and `touch` operations are NOT atomic, so a read between them could see a stale mtime. This is a theoretical race, not a practical one given the 2-minute cadence, but it's worth noting. I could not find a concrete defect here.

scripts/github/activity-gate.sh
```diff
@@ -601,33 +601,135 @@ check_ci_failures() {
...
+        ci_state=$(echo "$pr_data" | jq -r '
+            select(.statusCheckRollup != null) | .statusCheckRollup
+            | (any(.[]; ((.conclusion // .state // "") | ascii_upcase) == "FAILURE")) as $hard
+            | (any(.[]; ((.status // "") | ascii_upcase) as $s
+                   | $s == "QUEUED" or $s == "IN_PROGRESS" or $s == "WAITING"
+                     or $s == "PENDING" or $s == "REQUESTED"
+                     or (((.conclusion // "") == "")
+                         and (((.state // "") | ascii_upcase) as $st
+                              | $st == "PENDING" or $st == "EXPECTED")))) as $inflight
+            | (any(.[]; ((.conclusion // .state // "") | ascii_upcase) as $c
```
TimeToBuildBob, line 659: ❌ **P1** — The three-way classifier's `$bad` predicate treats any non-empty, non-SUCCESS/SKIPPED/NEUTRAL conclusion as bad, which includes ACTION_REQUIRED. This is intentional per the PR description, but it also includes the value `STARTUP_FAILURE`. The GitHub API's `conclusion` enum includes `ACTION_REQUIRED`, `CANCELLED`, `FAILURE`, `NEUTRAL`, `SKIPPED`, `STALE`, `STARTUP_FAILURE`, `SUCCESS`, `TIMED_OUT`. The `$bad` list does NOT exclude `STALE`. A stale CheckRun has a conclusion of `STALE`, which represents a check that was superseded by a newer run. If a PR has a STALE check and no other in-flight checks (e.g., a cancelled run left a STALE conclusion and the new run hasn't started yet), it classifies as `bad` and emits immediately, even though the situation is transient and similar to the cancel-in-progress case the code explicitly avoids for CANCELLED. This can cause spurious ci_failure emissions on PRs that are just between runs, leading to wasted sessions. The fix is to add STALE to the `$bad` list or treat it as non-bad.

scripts/github/activity-gate.sh
```diff
@@ -601,33 +601,135 @@ check_ci_failures() {
...
+        # *explicitly*, rather than relying on the gate having observed the green
+        # window in between. Green windows are easy to miss: the whole
+        # pending -> green -> force-push -> pending cycle can fit between two
+        # gate runs, leaving a "pending"-hashed file with an hours-old mtime that
+        # the new run matches and is instantly declared stuck. Deleting the file
+        # on green (above) fixes the observed case; keying on the head SHA fixes
+        # the unobserved one. Trade-off: a force-push that reproduces the same
+        # failure now re-emits `ci_failure`. That is the desired behaviour — new
+        # commit, still red, is a fresh actionable signal, not a duplicate.
+        ci_hash=$(echo "$pr_data" | jq -r '(.headRefOid // "") + ":" + ([.statusCheckRollup[] | .conclusion // .state // "pending"] | sort | join(","))' | portable_hash)
```
TimeToBuildBob, line 704: ⚠️ **P2** — The `ci_hash` uses `.headRefOid // ""` before the colon. If `headRefOid` is missing or null, it becomes an empty string, so two PRs with different actual head commits but missing headRefOid (which can happen if the API payload is malformed or the field is not selected) would hash identically if their rollup conclusions match. More critically, the state file is shared across runs of the same repo and PR number. If headRefOid is missing on one run and present on the next, the hash changes and a new episode starts, which is safe. But if headRefOid is consistently missing, the hash does not change on a push, defeating the purpose of including headRefOid. Checking the fetch_pr_data_with_ttl function, it explicitly requests headRefOid (line 370), so it should always be present for real PRs. However, the cache could serve stale payloads where headRefOid was present. This is not a likely defect for this gate, so I am downgrading to P2.

scripts/github/activity-gate.sh
TimeToBuildBob: Retracting — this finding explicitly concludes that the mechanism is sound and identifies no concrete defect. The gate runs one process per repo, so the hypothetical cross-process race is outside the actual execution model. No code change.

scripts/github/activity-gate.sh
TimeToBuildBob: Retracting — GitHub defines `STALE` as a terminal check conclusion, and a lone stale check means CI is terminal-but-not-green. That is exactly the stranded shape this PR is meant to re-dispatch. If a replacement run exists, its non-terminal status makes `$inflight` true and suppresses the immediate emission. Excluding `STALE` would recreate the blind spot rather than fix one. No code change.

scripts/github/activity-gate.sh
TimeToBuildBob: Retracting — `headRefOid` is explicitly selected by the fetch query and GitHub returns it for open PRs. A malformed payload missing a required selected field is not a valid state to silently classify; the proposed fallback concern has no concrete runtime path. No code change.

PR #1379, opened 2026-08-06
feat(activity-gate): write notif-*.map sidecar files for rollback

TimeToBuildBob:
## Summary

Activity gate now creates `notif-*.map` sidecar files alongside `notif-*.state` files, mapping notification ID to `owner/repo#number`. These sidecar files allow the PM delivery failure rollback logic to find and delete corresponding notification state files when delivery fails, preventing missed notifications on re-emit cycles (incident #1127).

## Changes

- Both jsonl and tab-delimited branches now write `.map` files
- `.map` format: `owner/repo#issue_number`
- Enables cleanup in `rollback_failed_delivery` without API calls

## Testing

✅ Pre-commit hooks pass

## Notes

This was extracted from PR #1378 (output-clarity skill) where it was accidentally committed to the wrong branch.

PR #1385, opened 2026-08-07
fix(greptile): fall back to an initial-review trigger when auto-review never arrives

TimeToBuildBob:
## Problem

The `no review yet` branch of `trigger` returned unconditionally, assuming Greptile's auto-review always arrives. It does not.

**2026-08-06/07:** Greptile stopped auto-reviewing gptme-contrib entirely, while continuing to review other repos in ~2 minutes — including *during* the GitHub Actions outage, which rules out the outage as the cause:

| repo | PRs created after 15:00Z | Greptile reviewed |
|---|---|---|
| gptme/gptme | #3471–#3475 | **5/5** (e.g. #3474 created 20:57 → reviewed 20:59) |
| ActivityWatch/aw-server-rust | #648 | **1/1** |
| **gptme/gptme-contrib** | #1380–#1383 | **0/4** |

Within contrib the break brackets cleanly: #1374–#1379 (01:14→10:50) all reviewed, #1380–#1383 (16:09→21:28) none.

Reopening the PRs restored CI but not the review; an empty commit (`synchronize`) on #1383 re-ran CI green and still produced zero comments from any author.

**Why it matters:** `self-merge-check.py` requires a Greptile review, so every affected PR is permanently self-merge ineligible and simply queues. The symptom reads as *"PRs aren't ready"* rather than *"the reviewer never came"* — which is what makes it hard to spot. gptme-contrib#1381 already merged with no Greptile review at all. This also blocks #1383, the activity-gate fix.

## Change

After `GREPTILE_INITIAL_GRACE_MINS` (default 45) with no review, trigger once.

The grace window preserves exactly why this branch existed: manually triggering the first review races the bot and produces two reviews. Inside the window, behaviour is unchanged.

Every existing protection still applies — the `flock`, the in-flight check, the `MAX_TOTAL_TRIGGERS` lifetime cap (so a repo where Greptile is truly gone backs off and escalates rather than commenting forever), and the trigger-timestamp record.

## Tests

`test_trigger_skips_unreviewed_pr_initial_review` encoded the old behaviour. Rather than delete it, it is split so the surviving invariant stays covered:

- `test_trigger_waits_for_auto_review_within_grace` — 10m old → no comment posted
- `test_trigger_falls_back_to_initial_review_after_grace` — 60m old → exactly one trigger

44 tests pass; `shellcheck -S warning` clean.

## Note

This does not diagnose *why* Greptile stopped on contrib — I lack `admin:org` to inspect app installations, so I can neither confirm nor rule out an installation/permissions change. This makes the failure recoverable either way instead of silently terminal.

### DIFF
--- a/scripts/github/greptile-helper.sh
+++ b/scripts/github/greptile-helper.sh
@@ -578,46 +578,54 @@ trigger)
     # the symptom reads as "PRs aren't ready" rather than "the reviewer never came".
     #
     # So after a grace period, trigger once. Every existing protection still
-    # applies: the flock above, the in-flight check, the lifetime cap, and the
-    # trigger-timestamp record.
+    # applies: the flock above, the local in-flight check, the lifetime cap, and
+    # the trigger-timestamp record.
     _initial_grace="${GREPTILE_INITIAL_GRACE_MINS:-45}"
+    case "$_initial_grace" in
+        ''|*[!0-9]*)
+            echo "  [greptile] Invalid GREPTILE_INITIAL_GRACE_MINS=$_initial_grace; using 45 minutes." >&2
+            _initial_grace=45
+            ;;
+    esac
 
     # "trigger once" means exactly once. Two guards, in order:
     #
-    #   1. "in-progress" — a trigger is in flight: either the local trigger-timestamp
-    #      file is fresh (propagation-delay guard, INCIDENT #5) or our comment is
-    #      younger than TRIGGER_GRACE_SECONDS. This is also the fail-safe value when
-    #      _our_trigger_status errors, so an API failure backs off instead of posting.
+    #   1. A fresh local trigger timestamp means a post is in flight but may not
+    #      be visible through GitHub's comments API yet (INCIDENT #5). The flock
+    #      handles concurrent runs; this guard handles sequential runs. Do not
+    #      reuse _our_trigger_status here: it broadly matches any of our comments
+    #      containing "greptileai", so fresh review-feedback prose can delay the
+    #      fallback even though no trigger was posted.
     #   2. lifetime trigger count >= 1 — we already posted our one `@greptileai review`
-    #      on this PR (it now reads "stale" or "stale-acked"). The re-review path
-    #      deliberately treats those as re-triggerable, since a new commit or a stuck
-    #      ack justifies another nudge; this path has neither escape hatch, so treating
-    #      them as re-triggerable would re-post every grace window. If Greptile never
-    #      reviews after our one trigger, escalate to a human.
-    #
-    # Guard 2 counts `@greptileai review` comments (_total_trigger_count) rather than
-    # reusing _our_trigger_status's verdict. That function matches ANY comment of ours
-    # containing "greptileai" (case-insensitive), and on this path review_cutoff is
-    # empty — so neither the spent-trigger normalisation nor the max-retries guard can
-    # ever age the match out. A single unrelated comment, e.g. "Thanks for the catch
-    # @greptileai! Fixed in abc123" (a shape we post routinely), would otherwise report
-    # "stale" forever and permanently suppress the fallback on exactly the PRs it
-    # exists to rescue.
-    #
-    # A ceiling of 1 is stricter than MAX_TOTAL_TRIGGERS (8, which governs the
-    # re-review path), and subsumes it — there is no separate cap check below.
-    _ts_status=$(_our_trigger_status 2>/dev/null || echo "in-progress")
-    if [ "$_ts_status" = "in-progress" ]; then
-        echo "  [greptile] Initial-review trigger already in flight on $REPO#$PR_NUMBER. Not triggering again."
+    #      on this PR. The re-review path deliberately permits later attempts after
+    #      new commits or a stuck acknowledgement; this path has no such escape hatch.
+    _initial_ts_in_flight=0
+    if [ -f "$_TRIGGER_TS_FILE" ]; then
+        _local_ts=$(cat "$_TRIGGER_TS_FILE" 2>/dev/null || true)
+        if [ -n "$_local_ts" ]; then
+            _local_age=$(_age_seconds "$_local_ts" 2>/dev/null) || _local_age=9999
+            if [ "${_local_age:-9999}" -lt "$TRIGGER_GRACE_SECONDS" ]; then
+                _initial_ts_in_flight=1
+            fi
+        fi
+    fi
+    if [ "$_initial_ts_in_flight" -eq 1 ]; then
+        echo "  [greptile] Initial-review trigger already in flight on $REPO#$PR_NUMBER (local timestamp is fresh). Not triggering again."
         exit 0
     fi
     _total_triggers=$(_total_trigger_count)
     if [ "${_total_triggers:-0}" -ge 1 ]; then
-        echo "  [greptile] Initial-review trigger already attempted on $REPO#$PR_NUMBER ($_total_triggers trigger comment(s), status: $_ts_status). Not triggering again — escalate to a human if Greptile never reviews."
+        echo "  [greptile] Initial-review trigger already attempted on $REPO#$PR_NUMBER ($_total_triggers trigger comment(s)). Not triggering again — escalate to a human if Greptile never reviews."
         exit 0
     fi
 
-    _created=$(gh api "repos/$REPO/pulls/$PR_NUMBER" --jq '.created_at // ""' 2>/dev/null) || _created=""
+    _pr_info=$(gh api "repos/$REPO/pulls/$PR_NUMBER" --jq '{state: (.state // ""), created_at: (.created_at // "")}' 2>/dev/null) || _pr_info=""
+    _pr_state=$(printf '%s' "$_pr_info" | _json_field "state") || _pr_state=""
+    _created=$(printf '%s' "$_pr_info" | _json_field "created_at") || _created=""
+    if [ "$_pr_state" != "open" ]; then
+        echo "  [greptile] PR $REPO#$PR_NUMBER is ${_pr_state:-unavailable}, not open. Skipping initial-review trigger."
+        exit 0
+    fi
     _age_mins=0
     if [ -n "$_created" ]; then
         _created_epoch=$(date -u -d "$_created" +%s 2>/dev/null \
--- a/scripts/github/greptile-helper.sh
+++ b/scripts/github/greptile-helper.sh
@@ -177,13 +177,15 @@ _issue_comments_json() {
 }
 
 _total_trigger_count() {
-    # Count ALL our actual `@greptileai review` trigger commands on this PR
-    # (lifetime), including the legacy `review comment` spelling. Do not count
-    # prose that merely contains the same substring. Fail-open to 0 on API
-    # error so a transient failure never blocks a legitimate trigger.
+    # Count ALL actual `@greptileai review` trigger commands on this PR
+    # (lifetime), including the legacy `review comment` spelling. Initial-review
+    # deduplication is per PR, not per author: a maintainer's manual trigger must
+    # suppress our fallback too. Do not count prose that merely contains the same
+    # substring. Fail-open to 0 on API error; the caller separately checks the
+    # shared comments-error marker before triggering.
     local count
     count=$(_issue_comments_json \
-        | jq -r "[.[][] | select(.user.login == \"$GITHUB_AUTHOR\" and (.body | test(\"^@greptileai review( comment)?([[:space:]]|$)\")))] | length" 2>/dev/null) || count=0
+        | jq -r '[.[][] | select(.body | test("^@greptileai review( comment)?([[:space:]]|$)"))] | length' 2>/dev/null) || count=0
     printf '%s\n' "${count:-0}"
 }
 
@@ -633,14 +635,12 @@ trigger)
         echo "  [greptile] PR $REPO#$PR_NUMBER is ${_pr_state:-unavailable}, not open. Skipping initial-review trigger."
         exit 0
     fi
-    _age_mins=0
-    if [ -n "$_created" ]; then
-        _created_epoch=$(date -u -d "$_created" +%s 2>/dev/null \
-            || date -u -j -f "%Y-%m-%dT%H:%M:%SZ" "$_created" +%s 2>/dev/null || echo "")
-        if [ -n "$_created_epoch" ]; then
-            _age_mins=$(( ( $(date -u +%s) - _created_epoch ) / 60 ))
-        fi
+    _age_seconds=$(_age_seconds "$_created" 2>/dev/null) || _age_seconds=""
+    if [ -z "$_age_seconds" ]; then
+        echo "  [greptile] Could not parse PR creation timestamp for $REPO#$PR_NUMBER. Refusing to trigger an initial review."
+        exit 3
     fi
+    _age_mins=$(( _age_seconds / 60 ))
 
     if [ "$_age_mins" -lt "$_initial_grace" ]; then
         echo "  [greptile] No review yet on $REPO#$PR_NUMBER (${_age_mins}m old, grace ${_initial_grace}m). Awaiting Greptile auto-review."
--- a/scripts/github/greptile-helper.sh
+++ b/scripts/github/greptile-helper.sh
@@ -146,7 +146,8 @@ PY
 # Shell variable caching doesn't work here because callers use $() subshells.
 _REVIEW_CACHE_FILE="${TMPDIR:-/tmp}/greptile-review-cache-$$.json"
 _ISSUE_COMMENTS_CACHE_FILE="${TMPDIR:-/tmp}/greptile-issue-comments-$$.json"
-trap 'rm -f "$_REVIEW_CACHE_FILE" "$_ISSUE_COMMENTS_CACHE_FILE"' EXIT
+_ISSUE_COMMENTS_ERROR_FILE="${TMPDIR:-/tmp}/greptile-issue-comments-error-$$"
+trap 'rm -f "$_REVIEW_CACHE_FILE" "$_ISSUE_COMMENTS_CACHE_FILE" "$_ISSUE_COMMENTS_ERROR_FILE"' EXIT
 
 # Shared hash for per-PR state files (lock + trigger timestamp).
 # Used across trigger and _our_trigger_status to coordinate without the GitHub API.
@@ -167,9 +168,11 @@ _issue_comments_json() {
         cat "$_ISSUE_COMMENTS_CACHE_FILE"
         return
     fi
-    gh api "repos/$REPO/issues/$PR_NUMBER/comments" --paginate 2>/dev/null \
-        | jq -s '.' > "$_ISSUE_COMMENTS_CACHE_FILE" 2>/dev/null \
-        || echo '[]' > "$_ISSUE_COMMENTS_CACHE_FILE"
+    if ! gh api "repos/$REPO/issues/$PR_NUMBER/comments" --paginate 2>/dev/null \
+        | jq -s '.' > "$_ISSUE_COMMENTS_CACHE_FILE" 2>/dev/null; then
+        echo '[]' > "$_ISSUE_COMMENTS_CACHE_FILE"
+        : > "$_ISSUE_COMMENTS_ERROR_FILE"
+    fi
     cat "$_ISSUE_COMMENTS_CACHE_FILE"
 }
 
@@ -614,6 +617,10 @@ trigger)
         exit 0
     fi
     _total_triggers=$(_total_trigger_count)
+    if [ -f "$_ISSUE_COMMENTS_ERROR_FILE" ]; then
+        echo "  [greptile] Could not read PR comments for $REPO#$PR_NUMBER. Refusing to trigger an initial review without duplicate-check data."
+        exit 3
+    fi
     if [ "${_total_triggers:-0}" -ge 1 ]; then
         echo "  [greptile] Initial-review trigger already attempted on $REPO#$PR_NUMBER ($_total_triggers trigger comment(s)). Not triggering again — escalate to a human if Greptile never reviews."
         exit 0
@@ -661,10 +668,10 @@ trigger)
     #
     # So: write first, and treat a failed write as fatal for this trigger. Not
     # triggering costs one cycle; double-triggering is the incident. If the post
-    # fails, remove the provisional timestamp so status does not report a phantom
-    # in-flight trigger and the next cycle can retry immediately. A process killed
-    # in the tiny write-to-post window still fails safe for at most the bounded
-    # TRIGGER_GRACE_SECONDS window.
+    # fails, keep the provisional timestamp for the bounded grace window. A failed
+    # client response is ambiguous: GitHub may have committed the comment before a
+    # timeout, and deleting the guard would let the next invocation duplicate it
+    # before the comments API catches up.
     if ! date -u +%Y-%m-%dT%H:%M:%SZ > "$_TRIGGER_TS_FILE" 2>/dev/null; then
         echo "  [greptile] Could not write trigger-timestamp file ($_TRIGGER_TS_FILE); refusing to trigger without the propagation-delay guard. Retrying next cycle."
         exit 0
@@ -672,8 +679,7 @@ trigger)
     if BOB_GREPTILE_HELPER=1 gh api "repos/$REPO/issues/$PR_NUMBER/comments" -f body="$_trigger_body" --silent 2>/dev/null; then
         echo "  [greptile] Initial review triggered."
     else
-        rm -f "$_TRIGGER_TS_FILE"
-        echo "  [greptile] Trigger failed (non-fatal)."
+        echo "  [greptile] Trigger failed or its result was ambiguous; retaining the propagation-delay guard for this grace window."
     fi
     exit 0
     ;;
--- a/scripts/github/greptile-helper.sh
+++ b/scripts/github/greptile-helper.sh
@@ -316,14 +316,19 @@ _our_trigger_status() {
         if [ -n "$_local_ts" ]; then
             # Only count this entry if it's from the CURRENT review cycle
             # (i.e., the timestamp is after the last Greptile review).
-            # Note: when review_cutoff is empty (no prior Greptile review), skip the
-            # fast-path — the trigger command only writes this file during re-reviews,
-            # which always have a non-empty cutoff, so this invariant holds.
+            # When review_cutoff is EMPTY there is no prior Greptile review at all,
+            # so there is only one cycle and any timestamp in the file is in it by
+            # definition. The fast-path must apply there too: the trigger command
+            # also writes this file on the initial-review fallback path, and skipping
+            # the guard for an empty cutoff would let two sequential runs both miss
+            # the (not-yet-propagated) comment via the API and both post.
             local _ts_in_cycle=0  # 1 = TS is from current review cycle; 0 = skip fast-path
             if [ -n "$review_cutoff" ]; then
                 if _timestamp_gt "$_local_ts" "$review_cutoff" 2>/dev/null; then
                     _ts_in_cycle=1
                 fi
+            else
+                _ts_in_cycle=1
             fi
             if [ "$_ts_in_cycle" -eq 1 ]; then
                 local _local_age
@@ -561,9 +566,18 @@ trigger)
     # trigger-timestamp record.
     _initial_grace="${GREPTILE_INITIAL_GRACE_MINS:-45}"
 
-    _ts_status=$(_our_trigger_status 2>/dev/null || echo "none")
-    if [ "$_ts_status" = "in-progress" ]; then
-        echo "  [greptile] Initial-review trigger already in flight on $REPO#$PR_NUMBER. Skipping."
+    # "trigger once" means exactly once. Any status other than "none" means a trigger
+    # comment from us already exists on this PR, so the fallback has already fired —
+    # back off regardless of whether it is "in-progress", "stale" or "stale-acked".
+    # The re-review path deliberately treats "stale"/"stale-acked" as re-triggerable
+    # (a new commit or a stuck ack justifies another nudge); this path has neither of
+    # those escape hatches, so treating them as re-triggerable would re-post every
+    # grace window until MAX_TOTAL_TRIGGERS — up to 8 `@greptileai review` comments on
+    # one PR. If Greptile never reviews after our one trigger, escalate to a human.
+    # Fail-safe: an unexpected non-zero return backs off rather than posting.
+    _ts_status=$(_our_trigger_status 2>/dev/null || echo "in-progress")
+    if [ "$_ts_status" != "none" ]; then
+        echo "  [greptile] Initial-review trigger already attempted on $REPO#$PR_NUMBER (status: $_ts_status). Not triggering again — escalate to a human if Greptile never reviews."
         exit 0
     fi
 
--- a/scripts/github/greptile-helper.sh
+++ b/scripts/github/greptile-helper.sh
@@ -177,12 +177,18 @@ _issue_comments_json() {
 }
 
 _total_trigger_count() {
-    # Count ALL actual `@greptileai review` trigger commands on this PR
-    # (lifetime), including the legacy `review comment` spelling. Initial-review
-    # deduplication is per PR, not per author: a maintainer's manual trigger must
-    # suppress our fallback too. Do not count prose that merely contains the same
-    # substring. Fail-open to 0 on API error; the caller separately checks the
-    # shared comments-error marker before triggering.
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
CI is green, state is CLEAN/MERGEABLE — but not self-merging because Greptile review is absent.

**This is a repo-wide issue, not specific to this PR.** Greptile has stopped auto-reviewing `gptme/gptme-contrib` for ~29h (as of 2026-08-07 ~16:00 UTC). Five PRs are stuck awaiting initial review: #1380, #1382, #1383, #1384, #1385. Other repos (gptme, gptme-cloud, aw-server-rust) reviewed fine during the same window — this is repo-specific.

Re-triggering `@greptileai review` has produced no response (tried on #1383 at 07:19 UTC, nothing 9h later). The `self-merge-check.py` gate requires a Greptile review, so these PRs will continue to queue until the underlying issue is resolved.

**Erik**: please check the Greptile GitHub App installation for `gptme/gptme-contrib` — confirm it's still enabled/indexed for this repo and hasn't hit a plan repo-count limit. Once Greptile is back, no code changes are needed — it will auto-review and the self-merge gate will clear normally.

The specific fix in this PR (fallback trigger after 45min grace) will itself help once it's merged.

TimeToBuildBob:
## 🤖 AI code review

### Confidence Score: 4/5

**Could merge, with caveats.** No blocking defects on this head, but 2 P2 issues worth resolving first.

**Files Needing Attention:**

- `scripts/github/greptile-helper.sh`

### What changed

Adds an initial-review fallback to the `trigger` command: after `GREPTILE_INITIAL_GRACE_MINS` (default 45) with no Greptile review and no existing `@greptileai review` comment, it writes a trigger-timestamp file and posts one review request, guarded by a lifetime cap and comments-API error check. The `status` command's unreviewed branch now always prints `awaiting-initial-review`, and `_our_trigger_status` treats an empty review cutoff as in-cycle. Tests cover the new fallback and update the old status expectation.

- `scripts/github/greptile-helper.sh` — Adds an initial-review fallback after GREPTILE_INITIAL_GRACE_MINS with no review, plus trigger-count guards and a write-before-post timestamp guard.
- `tests/test_greptile_helper.py` — Adds tests for the fallback, extends the fake gh stub with comments_api_error and ts_file_unwritable, and updates the status assertion for unreviewed PRs.

**2 finding(s)** — ⚠️ **2** P2

<details><summary><b>2 finding(s) posted inline on the diff</b> — expand to read them here</summary>

### ⚠️ P2 medium — `scripts/github/greptile-helper.sh:734`

The `status` command's unreviewed branch now unconditionally prints `awaiting-initial-review` and no longer calls `_our_trigger_status`. Previously, a fresh local timestamp or a manual trigger comment produced `in-progress`, and a comments-API failure returned `in-progress` as a fail-safe (see lines 386 and 437, which return `in-progress` on API error). Now `_has_greptile_review` can return false due to a swallowed API error in `_greptile_review_info` (it writes a default `{"has_review": false}` to the cache on failure), so `status` emits the same string as a genuinely unreviewed PR. Callers that treat `in-progress` as 'a trigger is in flight' or as an error signal will misread the state, and the documented status list at the top still advertises `in-progress`. This masks API failures and makes a transient error indistinguishable from a normal wait.

<sub>**How this was verified:** Traced status command flow: `_has_greptile_review` swallows API errors via `_greptile_review_info`'s default fallback; before this PR the else branch called `_our_trigger_status`, which returns 'in-progress' on API error. The else branch now always prints awaiting.</sub>

### ⚠️ P2 medium — `scripts/github/greptile-helper.sh:664`

In the fallback, `_has_greptile_review` is consulted once at the top of `trigger`. If that comments-API call fails transiently, `_greptile_review_info` swallows the failure and populates `_REVIEW_CACHE_FILE` with `{"has_review": false}`, so the script enters the no-review fallback. The fallback then calls `_total_trigger_count`/`_any_trigger_count`, which retry the same endpoint; if the retry succeeds, no error file is created, `_any_trigger_count` is 0 on an auto-reviewed PR with no trigger comments, and the PR age is past grace, so it posts a fresh `@greptileai review` on a PR that already has a Greptile review, producing a duplicate review request. The fallback never re-checks `_has_greptile_review` after a successful comments read. Fixing requires re-evaluating review status once the comments payload is available, or treating the earlier failure as fatal.

<sub>**How this was verified:** Traced `_greptile_review_info` on failure: it writes the default no-review JSON (line 216), so `_has_greptile_review` returns false; the fallback then retries comments in `_total_trigger_count` and only checks the error file, not review status again.</sub>

</details>

| `34607c4709ec` | ? | 3 | llm | 2026-08-07 21:01 UTC |
| `895dfa5720a5` | ? | 2 | llm | 2026-08-08 07:47 UTC |
| `9784234e5d07` | ? | 5 | llm | 2026-08-08 09:05 UTC |
| `925a2645a9cd` | 3/5 | 4 | llm | 2026-08-09 00:33 UTC |
| `7d60731cef1d` | 4/5 | 3 | llm | 2026-08-09 01:41 UTC |
| `9b6d299b86e1` | 3/5 | 2 | llm | 2026-08-09 02:26 UTC |
| `5fe4de89f926` | 4/5 | 3 | llm | 2026-08-09 03:22 UTC |
| `dfa213c5ca5e` | 4/5 | 3 | llm | 2026-08-09 04:35 UTC |
| `34799245cb94` | 3/5 | 2 | llm | 2026-08-09 05:21 UTC |
| `6792b4527283` | 3/5 | 1 | llm | 2026-08-09 06:22 UTC |

</details>

---
<sub>Reviewed `c75f4019a2e0` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 476s · self-hosted reviewer ([bob#1122](https://github.com/ErikBjare/bob/issues/1122)). Findings are advisory — a human still decides.</sub>

scripts/github/greptile-helper.sh
```diff
@@ -545,8 +545,65 @@ trigger)
...
+    # no path to a first review, every affected PR is permanently self-merge
+    # ineligible, because the self-merge gate requires a Greptile review — and
+    # the symptom reads as "PRs aren't ready" rather than "the reviewer never came".
+    #
+    # So after a grace period, trigger once. Every existing protection still
+    # applies: the flock above, the in-flight check, the lifetime cap, and the
+    # trigger-timestamp record.
+    _initial_grace="${GREPTILE_INITIAL_GRACE_MINS:-45}"
+
+    _ts_status=$(_our_trigger_status 2>/dev/null || echo "none")
```
TimeToBuildBob: Fixed in 895dfa57. Real — this PR broke the fast-path's stated invariant ("only re-reviews write \_TRIGGER_TS_FILE, which always have a non-empty cutoff") because the new fallback writes it too, so the propagation-delay guard was skipped entirely on the initial-review path. Fix is not passing `\_created` as the cutoff (that would compare a trigger timestamp against the PR creation date, which is meaningless): when `review_cutoff` is empty there is no prior review, so there is exactly one cycle and any timestamp in the file is in it by definition — set `_ts_in_cycle=1`. Stale comment updated. Audited all five callers: lines 447/499/620 always pass `$reviewed_at` inside `_has_greptile_review` branches (non-empty; an empty value there means a jq/API failure, and the resulting "in-progress" is the fail-safe direction), and line 640 (`status` with no review) now actually honours the local TS — verified `in-progress` for a 5-min-old TS, `awaiting-initial-review` for a 60-min-old one and for no TS, which is exactly what that site's own "check if there's a trigger in-flight" comment intends. `in-progress` is already a first-class non-actionable state in pr-greptile-trigger.py (not in `ACTIONABLE_STATES`), so no consumer regresses. Regression test: `test_trigger_fallback_local_timestamp_blocks_with_no_prior_review`; with only this change reverted it fails with `Got log: '{"body": "@greptileai review"}'` while the other 32 tests — including all four existing propagation-delay tests — stay green.

scripts/github/greptile-helper.sh
```diff
@@ -545,8 +545,65 @@ trigger)
...
+    if [ "$_ts_status" = "in-progress" ]; then
```
TimeToBuildBob, line 608: Fixed in 895dfa57. Real — "stale" and "stale-acked" fell through and re-posted every grace window up to MAX_TOTAL_TRIGGERS=8, contradicting the PR's "trigger once". The initial-review fallback now backs off on ANY status other than "none" (an existing trigger comment means already-attempted) with an explicit log line pointing at human escalation, and its status read is fail-safe (`|| echo "in-progress"` instead of `"none"`) so an unexpected non-zero return backs off rather than posting. The re-review path is untouched: it still deliberately bypasses on "stale-acked" to unstick gptme#1651-shaped hangs, because it has the new-commit and stuck-ack escape hatches this path lacks. The flock, the lifetime cap and the trigger-timestamp write are all unchanged and still downstream of this check. Two regression tests: `test_trigger_fallback_does_not_repost_over_stale_trigger` and `test_trigger_fallback_stale_acked_does_not_repost`; with only this change reverted both fail with `Got log: '{"body": "@greptileai review"}'`.

tests/test_greptile_helper.py
```diff
@@ -661,19 +661,46 @@ def test_score_5_with_new_commits_needs_rereview():
...
+    never obtain the Greptile review the self-merge gate requires.
+    """
+    fixture = {
+        "pr_number": 999,
+        "raw_comments": [],
+        "raw_commits": [],
+        "raw_pr": {"created_at": _iso_ago(minutes=60)},
+        "bot_reaction_count": 0,
+    }
+    result, gh_log = _run_helper("trigger", fixture, capture_gh_log=True)
```
TimeToBuildBob, line 714: Fixed in 895dfa57. Real — and this gap is precisely what let the other two findings through. Added four tests covering both: (a) `test_trigger_fallback_does_not_repost_over_stale_trigger` — unacked 40-min-old trigger, no review → "stale", must not re-post; (b) `test_trigger_fallback_stale_acked_does_not_repost` — 3h-old acked trigger → "stale-acked", must not re-post; (c) `test_trigger_fallback_local_timestamp_blocks_with_no_prior_review` — the TS-file fast-path with an empty `review_cutoff`, simulating INCIDENT #5 on an unreviewed PR (fresh local TS, trigger comment not yet visible in the API); (d) `test_trigger_fallback_stale_local_timestamp_still_allows_first_trigger` — an over-correction guard proving the empty-cutoff fast-path expires rather than permanently blocking the first trigger. Each of (a)-(c) was proven to fail with only its own source fix reverted; (d) is a guard, not a fail-without-fix case. Module is 33/33 green and shellcheck is clean.

scripts/github/greptile-helper.sh
```diff
@@ -545,8 +550,74 @@ trigger)
...
+    echo "  [greptile] No auto-review on $REPO#$PR_NUMBER after ${_age_mins}m (grace ${_initial_grace}m) — triggering initial review..."
+    _head_sha=$(gh api "repos/$REPO/pulls/$PR_NUMBER" --jq '.head.sha // ""' 2>/dev/null) || _head_sha=""
+    _trigger_body="@greptileai review"
+    if [ -n "$_head_sha" ]; then
+        _trigger_body="$_trigger_body
+
+<!-- greptile-helper head-sha: $_head_sha -->"
+    fi
+    if BOB_GREPTILE_HELPER=1 gh api "repos/$REPO/issues/$PR_NUMBER/comments" -f body="$_trigger_body" --silent 2>/dev/null; then
+        if ! date -u +%Y-%m-%dT%H:%M:%SZ > "$_TRIGGER_TS_FILE" 2>/dev/null; then
```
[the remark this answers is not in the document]
TimeToBuildBob: Confirmed real, and fixed in `9784234e`.

Verified by construction rather than by reading: `test_fallback_refuses_to_trigger_when_ts_file_cannot_be_written` makes only the TS path unwritable (a directory at that exact name, so TMPDIR itself stays writable for the gh stub and its log). Against the previous code it printed

```
[greptile] Warning: could not write trigger-timestamp file; propagation-delay guard disabled for this trigger.
[greptile] Initial review triggered.
```

— i.e. it posted the comment *and then* announced that the guard protecting it was off. Exactly as you describe.

Fix: write the timestamp **before** posting, and treat a failed write as fatal for this trigger. Not triggering costs one cycle; double-triggering is INCIDENT #5. If the post then fails, the stale timestamp only backs us off until the grace window expires, after which the API query correctly reports no comment and a retry goes through — self-healing, in the safe direction.

Also added `test_fallback_records_the_timestamp_before_posting` so the happy path cannot regress into "records nothing but posts anyway".

TimeToBuildBob:
@greptileai review

scripts/github/greptile-helper.sh
```diff
@@ -545,8 +550,91 @@ trigger)
...
+    # `_our_trigger_status` says "none" and we post a second `@greptileai
+    # review`. That is INCIDENT #5 (2026-03-19) exactly, and a warning saying
+    # "the guard is disabled" does not prevent it.
+    #
+    # So: write first, and treat a failed write as fatal for this trigger. Not
+    # triggering costs one cycle; double-triggering is the incident. If the post
+    # then fails, the stale TS only backs us off until the grace window expires,
+    # after which the API query (which will correctly report no comment) lets a
+    # retry through — self-healing, in the safe direction.
+    if ! date -u +%Y-%m-%dT%H:%M:%SZ > "$_TRIGGER_TS_FILE" 2>/dev/null; then
```
TimeToBuildBob, line 657: ⚠️ **P2** — The fallback writes `_TRIGGER_TS_FILE` before posting, but if the `gh api` POST subsequently fails, the file remains with a recent timestamp and no comment was ever posted. The next invocation within TRIGGER_GRACE_SECONDS reads that file and `_our_trigger_status` returns "in-progress", so the fallback backs off and never retries until the timestamp expires (15 min). That is by design per the comment (“self-healing, in the safe direction”), and the code handles it correctly: the stale TS expires and the API (which reports no comment) lets a retry through. This is not a bug; the only downside is a short delay, which the comment explicitly accepts. No finding.

scripts/github/greptile-helper.sh
```diff
@@ -545,8 +550,91 @@ trigger)
...
+            _age_mins=$(( ( $(date -u +%s) - _created_epoch ) / 60 ))
+        fi
+    fi
+
+    if [ "$_age_mins" -lt "$_initial_grace" ]; then
+        echo "  [greptile] No review yet on $REPO#$PR_NUMBER (${_age_mins}m old, grace ${_initial_grace}m). Awaiting Greptile auto-review."
+        exit 0
+    fi
+
+    _total_triggers=$(_total_trigger_count)
```
TimeToBuildBob: ⚠️ **P2** — The fallback uses `_total_trigger_count` which counts all `@greptileai review` comments ever posted, including re-review triggers. If a PR already had, say, 7 re-review triggers and then Greptile never auto-review-reviews a new head, the fallback sees count≥8 and backs off, even though this is the first *initial-review* fallback. That’s a lifetime cap and probably intentional, because the comment forbids more, but it means the fallback can never fire on a PR that any prior re-review triggers. The semantics are clear from the exit message. No defect.

scripts/github/greptile-helper.sh
```diff
@@ -316,14 +316,19 @@ _our_trigger_status() {
...
+            # also writes this file on the initial-review fallback path, and skipping
+            # the guard for an empty cutoff would let two sequential runs both miss
+            # the (not-yet-propagated) comment via the API and both post.
             local _ts_in_cycle=0  # 1 = TS is from current review cycle; 0 = skip fast-path
             if [ -n "$review_cutoff" ]; then
                 if _timestamp_gt "$_local_ts" "$review_cutoff" 2>/dev/null; then
                     _ts_in_cycle=1
                 fi
+            else
+                _ts_in_cycle=1
```
TimeToBuildBob, line 345: ⚠️ **P2** — The fallback path posts exactly one `@greptileai review` comment after the grace window, but the `_our_trigger_status` check at line 578 uses `|| echo "in-progress"` as a fail-safe. If the GitHub API is slow/rate-limited and `_our_trigger_status` returns a non-zero exit, the fallback backs off (safe). However, a more subtle issue: the fast-path in `_our_trigger_status` now marks a local TS as "in cycle" even when `review_cutoff` is empty (line 331). If a PR previously had a re-review trigger (from the old re-review path) that wrote a TS file, and then a new review never lands and this fallback runs, the local TS file could be stale (older than 15min), so it is ignored and the API is consulted—fine. But if a re-review trigger was written within the last 15min and then the PR head is force-pushed/changed, the fallback sees the fresh TS and skips, even though the trigger was for a different head. That is a minor mis-classification, but since the fallback is a last-resort for a missing review, a 15-minute delay before triggering again is acceptable and matches intended behaviour. No new bug.

scripts/github/greptile-helper.sh
```diff
@@ -545,8 +550,91 @@ trigger)
...
+    # "trigger once" means exactly once. Any status other than "none" means a trigger
+    # comment from us already exists on this PR, so the fallback has already fired —
+    # back off regardless of whether it is "in-progress", "stale" or "stale-acked".
+    # The re-review path deliberately treats "stale"/"stale-acked" as re-triggerable
+    # (a new commit or a stuck ack justifies another nudge); this path has neither of
+    # those escape hatches, so treating them as re-triggerable would re-post every
+    # grace window until MAX_TOTAL_TRIGGERS — up to 8 `@greptileai review` comments on
+    # one PR. If Greptile never reviews after our one trigger, escalate to a human.
+    # Fail-safe: an unexpected non-zero return backs off rather than posting.
+    _ts_status=$(_our_trigger_status 2>/dev/null || echo "in-progress")
```
TimeToBuildBob, line 607: ⚠️ **P2** — Potential race larger than the local TS: `_our_trigger_status` is called with no `review_cutoff` on the initial fallback path (line 578). It then uses jq `count_since_review` which counts all our triggers, not a subset. But because `review_cutoff` is empty, line 380 `if [ -n "$review_cutoff" ]` skips the “compare trigger vs cutoff” guard and later line 389 `if [ -n "$review_cutoff" ]` skips the max-retries guard. That is intended for initial fallback. However, there is an edge: after the fallback posts a trigger, the next run reads local TS (fresh) and returns in-progress, good. After 15min grace expires, `_ts_status` would re-run the API, find the trigger comment, and continue to report `stale`/`stale-acked` because `no review_cutoff` means the comment is never treated as spent (line 380-384 skips the spend check). That means the “already attempted” backoff at line 579 is permanent (as intended, “trigger once”). But `status` command (lines 671-676) treats `in-progress` as in-progress and anything else as awaiting-initial-review, so after the one trigger the status returns `awaiting-initial-review` forever even though the fallback already fired—this is mis-reporting to CI/CD that could re-enter trigger. But `trigger` will block due to block 579, so no duplicate. Only a status display issue. Not serious.

TimeToBuildBob:
@greptileai review

scripts/github/greptile-helper.sh
```diff
@@ -545,8 +564,105 @@ trigger)
...
+    #
+    # A ceiling of 1 is stricter than MAX_TOTAL_TRIGGERS (8, which governs the
+    # re-review path), and subsumes it — there is no separate cap check below.
+    _ts_status=$(_our_trigger_status 2>/dev/null || echo "in-progress")
+    if [ "$_ts_status" = "in-progress" ]; then
+        echo "  [greptile] Initial-review trigger already in flight on $REPO#$PR_NUMBER. Not triggering again."
+        exit 0
+    fi
+    _total_triggers=$(_total_trigger_count)
+    if [ "${_total_triggers:-0}" -ge 1 ]; then
```
TimeToBuildBob: Fixed in `7d60731c`. Valid: the old substring regex could count prose containing `@greptileai review` as a trigger. `_total_trigger_count` now accepts only a trigger command at the start of the comment (including the legacy `review comment` spelling), and the regression now uses prose containing the exact old substring.

scripts/github/greptile-helper.sh
```diff
@@ -545,8 +564,105 @@ trigger)
...
+    # `_our_trigger_status` says "none" and we post a second `@greptileai
+    # review`. That is INCIDENT #5 (2026-03-19) exactly, and a warning saying
+    # "the guard is disabled" does not prevent it.
+    #
+    # So: write first, and treat a failed write as fatal for this trigger. Not
+    # triggering costs one cycle; double-triggering is the incident. If the post
+    # then fails, the stale TS only backs us off until the grace window expires,
+    # after which the API query (which will correctly report no comment) lets a
+    # retry through — self-healing, in the safe direction.
+    if ! date -u +%Y-%m-%dT%H:%M:%SZ > "$_TRIGGER_TS_FILE" 2>/dev/null; then
```
TimeToBuildBob, line 689: Fixed in `7d60731c`. Valid: the timestamp is provisional until the REST post succeeds. The failure branch now removes it, so `status` cannot report a phantom in-flight trigger and the next cycle can retry immediately. Added a regression that forces the initial-review POST to fail and verifies no timestamp remains.

TimeToBuildBob, line 689: Adjudicated in `7d60731c`. The bounded crash window is intentional fail-safe behavior, not an unbounded availability gap: if the process dies between the local write and REST post, the stale timestamp expires after `TRIGGER_GRACE_SECONDS` (15 minutes), then the authoritative API state permits retry. Removing that pre-post guard would reopen the confirmed duplicate-trigger race. I clarified this trade-off in the code comment; API-returned failures now remove the provisional timestamp immediately.

tests/test_greptile_helper.py
```diff
@@ -233,6 +240,13 @@ def _run_helper(
             ts_file = tmp_path / f"greptile-trigger-ts-{_pr_hash(repo, pr_number)}.txt"
             ts_file.write_text(pre_trigger_ts)
 
+        # Make only the TS path unwritable — a directory at that exact name.
+        # Chmod-ing TMPDIR would also break the gh stub and its log, which live
+        # in the same directory.
+        if ts_file_unwritable:
```
TimeToBuildBob, line 257: Not changing this test. The directory-at-path fixture deliberately and deterministically exercises the production contract under test: writing the guard fails while the lock and gh stub remain usable. A pre-existing regular file with permission bits is not a reliable simulation under root/container CI and, before any write attempt, a fresh timestamp is correctly supposed to block via `_our_trigger_status`. The stale-local-timestamp path is already covered separately by `test_trigger_fallback_stale_local_timestamp_still_allows_first_trigger`.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

scripts/github/greptile-helper.sh
```diff
@@ -545,8 +566,115 @@ trigger)
...
+                _initial_ts_in_flight=1
+            fi
+        fi
+    fi
+    if [ "$_initial_ts_in_flight" -eq 1 ]; then
+        echo "  [greptile] Initial-review trigger already in flight on $REPO#$PR_NUMBER (local timestamp is fresh). Not triggering again."
+        exit 0
+    fi
+    _total_triggers=$(_total_trigger_count)
+    if [ "${_total_triggers:-0}" -ge 1 ]; then
```
TimeToBuildBob: Fixed in `5fe4de89`. The initial-review path now records a comments-API read failure separately and exits 3 before triggering, so missing duplicate-check data fails safe instead of becoming “zero triggers.” Added a regression that forces the comments endpoint to fail and verifies no POST occurs.

scripts/github/greptile-helper.sh
```diff
@@ -545,8 +566,115 @@ trigger)
...
+    # in the tiny write-to-post window still fails safe for at most the bounded
+    # TRIGGER_GRACE_SECONDS window.
+    if ! date -u +%Y-%m-%dT%H:%M:%SZ > "$_TRIGGER_TS_FILE" 2>/dev/null; then
+        echo "  [greptile] Could not write trigger-timestamp file ($_TRIGGER_TS_FILE); refusing to trigger without the propagation-delay guard. Retrying next cycle."
+        exit 0
+    fi
+    if BOB_GREPTILE_HELPER=1 gh api "repos/$REPO/issues/$PR_NUMBER/comments" -f body="$_trigger_body" --silent 2>/dev/null; then
+        echo "  [greptile] Initial review triggered."
+    else
+        rm -f "$_TRIGGER_TS_FILE"
```
TimeToBuildBob: Fixed in `5fe4de89`. Valid: a non-zero client result is ambiguous, so the fallback now retains the provisional timestamp for the bounded grace window instead of deleting it. The regression now verifies the guard survives a failed POST.

scripts/github/greptile-helper.sh
```diff
@@ -545,8 +566,107 @@ trigger)
...
+    # containing "greptileai" (case-insensitive), and on this path review_cutoff is
+    # empty — so neither the spent-trigger normalisation nor the max-retries guard can
+    # ever age the match out. A single unrelated comment, e.g. "Thanks for the catch
+    # @greptileai! Fixed in abc123" (a shape we post routinely), would otherwise report
+    # "stale" forever and permanently suppress the fallback on exactly the PRs it
+    # exists to rescue.
+    #
+    # A ceiling of 1 is stricter than MAX_TOTAL_TRIGGERS (8, which governs the
+    # re-review path), and subsumes it — there is no separate cap check below.
+    _ts_status=$(_our_trigger_status 2>/dev/null || echo "in-progress")
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in `9b6d299b`. Guard 1 now reads only the per-PR timestamp file; it no longer calls the broad `_our_trigger_status` comment matcher. The regression `test_trigger_fallback_ignores_unrelated_greptileai_mention` covers prose containing the exact old trigger substring and confirms the fallback still posts once.

TimeToBuildBob:
@greptileai review

scripts/github/greptile-helper.sh
```diff
@@ -545,8 +569,118 @@ trigger)
...
+    if [ "$_initial_ts_in_flight" -eq 1 ]; then
+        echo "  [greptile] Initial-review trigger already in flight on $REPO#$PR_NUMBER (local timestamp is fresh). Not triggering again."
+        exit 0
+    fi
+    _total_triggers=$(_total_trigger_count)
+    if [ -f "$_ISSUE_COMMENTS_ERROR_FILE" ]; then
+        echo "  [greptile] Could not read PR comments for $REPO#$PR_NUMBER. Refusing to trigger an initial review without duplicate-check data."
+        exit 3
+    fi
+    if [ "${_total_triggers:-0}" -ge 1 ]; then
```
TimeToBuildBob: Fixed in `dfa213c5`. The lifetime initial-trigger guard now counts exact trigger commands from every author, so a maintainer or bot trigger suppresses the fallback for the whole PR. Prose mentions remain excluded. Added a regression using a different author.

scripts/github/greptile-helper.sh
```diff
@@ -545,8 +569,118 @@ trigger)
...
+    _pr_info=$(gh api "repos/$REPO/pulls/$PR_NUMBER" --jq '{state: (.state // ""), created_at: (.created_at // "")}' 2>/dev/null) || _pr_info=""
+    _pr_state=$(printf '%s' "$_pr_info" | _json_field "state") || _pr_state=""
+    _created=$(printf '%s' "$_pr_info" | _json_field "created_at") || _created=""
+    if [ "$_pr_state" != "open" ]; then
+        echo "  [greptile] PR $REPO#$PR_NUMBER is ${_pr_state:-unavailable}, not open. Skipping initial-review trigger."
+        exit 0
+    fi
+    _age_mins=0
+    if [ -n "$_created" ]; then
+        _created_epoch=$(date -u -d "$_created" +%s 2>/dev/null \
```
TimeToBuildBob: Fixed in `dfa213c5`. Replaced the GNU/BSD `date` split with the helper's existing Python ISO-8601 parser, which handles offsets and fractional seconds portably. An absent or malformed creation timestamp now fails safe with exit 3 instead of looking permanently 0 minutes old. Added offset and malformed-timestamp regressions.

scripts/github/greptile-helper.sh
```diff
@@ -545,8 +571,116 @@ trigger)
...
+    if [ "$_initial_ts_in_flight" -eq 1 ]; then
+        echo "  [greptile] Initial-review trigger already in flight on $REPO#$PR_NUMBER (local timestamp is fresh). Not triggering again."
+        exit 0
+    fi
+    _total_triggers=$(_total_trigger_count)
+    if [ -f "$_ISSUE_COMMENTS_ERROR_FILE" ]; then
+        echo "  [greptile] Could not read PR comments for $REPO#$PR_NUMBER. Refusing to trigger an initial review without duplicate-check data."
+        exit 3
+    fi
+    if [ "${_total_triggers:-0}" -ge 1 ]; then
```
TimeToBuildBob: Fixed in `34799245`. The initial-review path now checks `MAX_TOTAL_TRIGGERS` explicitly before its stricter exactly-once guard, with a regression that supplies eight trigger comments and verifies `BACKOFF` plus no POST.

The exactly-once guard already prevented a ninth trigger whenever the comments API returned those comments; the explicit cap is still useful as an auditable hard backstop if that stricter policy changes later. Comments-API failure remains fail-safe (exit 3), so the propagation case cannot bypass the cap by being interpreted as zero authoritative comments.

TimeToBuildBob:
@greptileai review

scripts/github/greptile-helper.sh
```diff
@@ -153,18 +168,30 @@ _issue_comments_json() {
...
 
 _total_trigger_count() {
-    # Count ALL our `@greptileai review` trigger comments on this PR (lifetime).
-    # Fail-open to 0 on API error so a transient failure never blocks a legit trigger.
+    # Count our actual `@greptileai review` trigger commands over the PR lifetime.
+    # This count backs the helper's spam ceiling, so another maintainer's manual
+    # trigger must not consume one of our slots.
+    local count
+    count=$(_issue_comments_json \
+        | jq -r '[.[][] | select(.user.login == "'"${GITHUB_AUTHOR}"'" and (.body | test("^@greptileai review( comment)?([[:space:]]|$)")))] | length' 2>/dev/null) || count=0
```
TimeToBuildBob: Fixed in `c75f4019`. Replaced the POSIX character class in both trigger counters with jq-compatible `\s`, matching the existing score parser convention. Added a regression with the real multiline helper body (`@greptileai review` plus the head-SHA marker); the full helper suite is 47/47 green and shellcheck passes.

scripts/github/greptile-helper.sh
```diff
@@ -579,13 +727,11 @@ status)
...
-        if [ "$_ts" = "in-progress" ]; then
-            echo "in-progress"
-        else
-            echo "awaiting-initial-review"
-        fi
+        # Keep the public status contract stable for unreviewed PRs. The trigger
+        # command itself owns in-flight deduplication via the timestamp and comment
+        # guards above; callers use this state to distinguish initial-review waits
+        # from re-review work.
+        echo "awaiting-initial-review"
```
TimeToBuildBob, line 734: ⚠️ **P2** — The `status` command's unreviewed branch now unconditionally prints `awaiting-initial-review` and no longer calls `_our_trigger_status`. Previously, a fresh local timestamp or a manual trigger comment produced `in-progress`, and a comments-API failure returned `in-progress` as a fail-safe (see lines 386 and 437, which return `in-progress` on API error). Now `_has_greptile_review` can return false due to a swallowed API error in `_greptile_review_info` (it writes a default `{"has_review": false}` to the cache on failure), so `status` emits the same string as a genuinely unreviewed PR. Callers that treat `in-progress` as 'a trigger is in flight' or as an error signal will misread the state, and the documented status list at the top still advertises `in-progress`. This masks API failures and makes a transient error indistinguishable from a normal wait.

scripts/github/greptile-helper.sh
```diff
@@ -545,8 +577,124 @@ trigger)
...
+        exit 3
+    fi
+    _age_mins=$(( _age_seconds / 60 ))
+
+    if [ "$_age_mins" -lt "$_initial_grace" ]; then
+        echo "  [greptile] No review yet on $REPO#$PR_NUMBER (${_age_mins}m old, grace ${_initial_grace}m). Awaiting Greptile auto-review."
+        exit 0
+    fi
+
+    echo "  [greptile] No auto-review on $REPO#$PR_NUMBER after ${_age_mins}m (grace ${_initial_grace}m) — triggering initial review..."
```
TimeToBuildBob, line 664: ⚠️ **P2** — In the fallback, `_has_greptile_review` is consulted once at the top of `trigger`. If that comments-API call fails transiently, `_greptile_review_info` swallows the failure and populates `_REVIEW_CACHE_FILE` with `{"has_review": false}`, so the script enters the no-review fallback. The fallback then calls `_total_trigger_count`/`_any_trigger_count`, which retry the same endpoint; if the retry succeeds, no error file is created, `_any_trigger_count` is 0 on an auto-reviewed PR with no trigger comments, and the PR age is past grace, so it posts a fresh `@greptileai review` on a PR that already has a Greptile review, producing a duplicate review request. The fallback never re-checks `_has_greptile_review` after a successful comments read. Fixing requires re-evaluating review status once the comments payload is available, or treating the earlier failure as fatal.

PR #1389, opened 2026-08-07
fix(credential-slots): allow --switch to slots with expired access token when probe=ok

TimeToBuildBob:
## Problem

`--switch <slot> --execute` silently fails (exit 1) when the slot's access token
has lapsed, even when `--check-auth --probe` confirms the refresh token is valid
and the slot would work fine — CC auto-refreshes expired access tokens on first use.

**Observed in operator session ef48 (2026-08-07):**
- `--check-auth --probe` showed: `[stale] erik (access token lapsed 2270m ago — will refresh on next use) [max] probe=ok`
- `--switch erik --execute` → silent exit 1, symlink not changed

Root cause: `SlotManager.switch_to()` checks `expiresAt` and refuses any slot
past that timestamp, regardless of `force` — "known-bad credential" guard. But
an expired **access** token with a valid **refresh** token is not a dead credential.

## Fix

Add `probe_ok: bool = False` to `SlotManager.switch_to()` and
`SubscriptionManager.switch_to()`. When `probe_ok=True`, the offline `expiresAt`
freshness check is skipped. Callers are responsible for only passing `probe_ok=True`
after a successful online probe.

In `_cmd_switch()` (the `--switch --execute` path), when a switch is refused for
an expired access token, automatically probe the slot and retry with `probe_ok=True`
if the probe succeeds. This makes `--switch erik --execute` transparent for the
common case of a stale-but-refreshable slot.

## Test plan

- [x] `test_probe_ok_bypasses_expiry_check` — expired slot + `probe_ok=True` → switch succeeds
- [x] `test_probe_ok_still_rejects_missing_slot` — missing slot file + `probe_ok=True` → still rejected
- [x] `test_expired_target_rejected_even_with_force` — existing test unchanged (force alone still rejects)
- [x] All 206 tests pass (80 credential-slots, 126 gptme-subscription)
- [x] mypy clean on both packages

### DIFF
--- a/packages/credential-slots/src/credential_slots/manager.py
+++ b/packages/credential-slots/src/credential_slots/manager.py
@@ -392,6 +392,10 @@ class SlotManager:
             now=now,
         )
 
+    def slot_has_refresh_token(self, sub: str) -> bool:
+        """Return whether a named slot contains a usable refresh token."""
+        return _read_refresh_token(self.slot_path(sub)) is not None
+
     # ---- Drift detection ----
 
     def detect_live_slot_drift(self) -> DriftInfo | None:
--- a/packages/credential-slots/src/credential_slots/manager.py
+++ b/packages/credential-slots/src/credential_slots/manager.py
@@ -544,6 +544,7 @@ class SlotManager:
         reason: str,
         *,
         force: bool = False,
+        probe_ok: bool = False,
     ) -> SwitchResult:
         """Flip the live symlink to point at ``sub``'s slot file.
 
@@ -554,6 +555,9 @@ class SlotManager:
         - If the target slot is expired or unreadable, the switch is
           refused **regardless of** ``force`` — landing on a known-bad
           credential just moves the 401 crash loop to the next run.
+          Pass ``probe_ok=True`` to bypass this check when an online probe
+          has already confirmed the slot's refresh token is valid (e.g. an
+          expired access token that CC will auto-refresh on first use).
         - On success, replaces the existing symlink atomically via a
           temp-symlink + ``os.replace`` (single rename syscall) so no
           concurrent reader sees a missing ``live_path``, then calls
@@ -588,11 +592,12 @@ class SlotManager:
             self._log(msg)
             return SwitchResult(ok=False, reason=msg)
 
-        fresh, fresh_reason = self.slot_is_fresh(sub)
-        if not fresh:
-            msg = f"refusing to switch to {sub}: {fresh_reason}"
-            self._log(msg)
-            return SwitchResult(ok=False, reason=msg)
+        if not probe_ok:
+            fresh, fresh_reason = self.slot_is_fresh(sub)
+            if not fresh:
+                msg = f"refusing to switch to {sub}: {fresh_reason}"
+                self._log(msg)
+                return SwitchResult(ok=False, reason=msg)
 
         live = self.live_path
         tmp = self.creds_dir / (self.live_name + f".tmp{os.getpid()}")
--- a/packages/credential-slots/src/credential_slots/manager.py
+++ b/packages/credential-slots/src/credential_slots/manager.py
@@ -595,6 +595,10 @@ class SlotManager:
           Pass ``probe_ok=True`` to bypass this check when an online probe
           has already confirmed the slot's refresh token is valid (e.g. an
           expired access token that CC will auto-refresh on first use).
+          ``probe_ok`` is honored **only** for slots that actually hold a
+          refresh token — there is nothing to auto-refresh otherwise — and
+          every bypass is logged distinctly so a skipped safety check is
+          never indistinguishable from an ordinary switch.
         - On success, replaces the existing symlink atomically via a
           temp-symlink + ``os.replace`` (single rename syscall) so no
           concurrent reader sees a missing ``live_path``, then calls
@@ -629,7 +633,30 @@ class SlotManager:
             self._log(msg)
             return SwitchResult(ok=False, reason=msg)
 
-        if not probe_ok:
+        # ``probe_ok`` is a caller assertion ("an online probe says this slot
+        # works"), so it only earns the bypass when the slot could actually be
+        # revived the way the assertion implies: CC refreshes an expired access
+        # token from the stored refresh token. No refresh token, nothing to
+        # refresh — the assertion cannot be about this slot, so the offline gate
+        # stays in force.
+        bypass_expiry = probe_ok
+        if probe_ok and _read_refresh_token(slot) is None:
+            bypass_expiry = False
+            self._log(
+                f"probe_ok IGNORED for {sub}: slot holds no refresh token, "
+                "so an expired access token cannot be auto-refreshed; "
+                "applying the normal expiry gate"
+            )
+
+        if bypass_expiry:
+            # Never let a skipped safety check look like an ordinary switch in
+            # the logs — this line is the audit trail for the bypass.
+            self._log(
+                f"probe_ok BYPASS: skipping expiry gate for {sub} "
+                f"(caller asserts an online probe confirmed the refresh token) "
+                f"— {reason}"
+            )
+        else:
             fresh, fresh_reason = self.slot_is_fresh(sub)
             if not fresh:
                 msg = f"refusing to switch to {sub}: {fresh_reason}"
--- a/packages/credential-slots/src/credential_slots/manager.py
+++ b/packages/credential-slots/src/credential_slots/manager.py
@@ -595,10 +595,12 @@ class SlotManager:
           Pass ``probe_ok=True`` to bypass this check when an online probe
           has already confirmed the slot's refresh token is valid (e.g. an
           expired access token that CC will auto-refresh on first use).
-          ``probe_ok`` is honored **only** for slots that actually hold a
-          refresh token — there is nothing to auto-refresh otherwise — and
-          every bypass is logged distinctly so a skipped safety check is
-          never indistinguishable from an ordinary switch.
+          ``probe_ok`` is honored **only** when the freshness failure is
+          expiry-class (see :func:`reason_is_refreshable`) *and* the slot
+          actually holds a refresh token — an unreadable/malformed slot, or
+          one with nothing to refresh, is still refused. Every bypass and
+          every ignored ``probe_ok`` is logged distinctly so a skipped safety
+          check is never indistinguishable from an ordinary switch.
         - On success, replaces the existing symlink atomically via a
           temp-symlink + ``os.replace`` (single rename syscall) so no
           concurrent reader sees a missing ``live_path``, then calls
@@ -633,32 +635,47 @@ class SlotManager:
             self._log(msg)
             return SwitchResult(ok=False, reason=msg)
 
-        # ``probe_ok`` is a caller assertion ("an online probe says this slot
-        # works"), so it only earns the bypass when the slot could actually be
-        # revived the way the assertion implies: CC refreshes an expired access
-        # token from the stored refresh token. No refresh token, nothing to
-        # refresh — the assertion cannot be about this slot, so the offline gate
-        # stays in force.
-        bypass_expiry = probe_ok
-        if probe_ok and _read_refresh_token(slot) is None:
+        fresh, fresh_reason = self.slot_is_fresh(sub)
+        if not fresh:
+            # ``probe_ok`` is a caller assertion ("an online probe says this
+            # slot works"), so it only earns a bypass when *both* halves of the
+            # assertion can hold for this slot. It is deliberately narrow:
+            #
+            #  1. The failure must be expiry-class. A slot that is unreadable,
+            #     malformed, or missing ``expiresAt`` (truncated / rewritten
+            #     mid-OAuth-refresh) is not something a refresh can fix, and the
+            #     docstring promises those are refused regardless. Letting
+            #     ``probe_ok`` swallow the whole gate would hand any current or
+            #     future caller a malformed-slot bypass for free.
+            #  2. The slot must hold a refresh token. No refresh token, nothing
+            #     for CC to auto-refresh — the assertion cannot be about this
+            #     slot.
+            #
+            # Anything else keeps the offline gate in force, and says why.
             bypass_expiry = False
-            self._log(
-                f"probe_ok IGNORED for {sub}: slot holds no refresh token, "
-                "so an expired access token cannot be auto-refreshed; "
-                "applying the normal expiry gate"
-            )
+            if probe_ok and not reason_is_refreshable(fresh_reason):
+                self._log(
+                    f"probe_ok IGNORED for {sub}: {fresh_reason} — not an "
+                    "expiry-class failure, so no token refresh can fix it; "
+                    "applying the normal freshness gate"
+                )
+            elif probe_ok and _read_refresh_token(slot) is None:
+                self._log(
+                    f"probe_ok IGNORED for {sub}: slot holds no refresh token, "
+                    "so an expired access token cannot be auto-refreshed; "
+                    "applying the normal expiry gate"
+                )
+            elif probe_ok:
+                bypass_expiry = True
+                # Never let a skipped safety check look like an ordinary switch
+                # in the logs — this line is the audit trail for the bypass.
+                self._log(
+                    f"probe_ok BYPASS: skipping expiry gate for {sub} "
+                    f"(caller asserts an online probe confirmed the refresh token) "
+                    f"— {reason}"
+                )
 
-        if bypass_expiry:
-            # Never let a skipped safety check look like an ordinary switch in
-            # the logs — this line is the audit trail for the bypass.
-            self._log(
-                f"probe_ok BYPASS: skipping expiry gate for {sub} "
-                f"(caller asserts an online probe confirmed the refresh token) "
-                f"— {reason}"
-            )
-        else:
-            fresh, fresh_reason = self.slot_is_fresh(sub)
-            if not fresh:
+            if not bypass_expiry:
                 msg = f"refusing to switch to {sub}: {fresh_reason}"
                 self._log(msg)
                 return SwitchResult(ok=False, reason=msg)
--- a/packages/credential-slots/src/credential_slots/manager.py
+++ b/packages/credential-slots/src/credential_slots/manager.py
@@ -133,6 +133,17 @@ def read_slot_expiry(path: Path) -> datetime | None:
     return _parse_oauth_expiry(payload)
 
 
+#: Reason prefixes emitted by :func:`slot_is_fresh` when a slot fails purely
+#: because of its *access* token clock — already lapsed, or inside the grace
+#: window. Such a slot may still hold a valid refresh token. Kept as module
+#: constants so :func:`reason_is_refreshable` can never drift from the strings
+#: :func:`slot_is_fresh` actually produces.
+REASON_EXPIRED = "expired"
+REASON_WITHIN_GRACE = "expires within grace"
+
+_REFRESHABLE_REASON_MARKERS = (REASON_EXPIRED, REASON_WITHIN_GRACE)
+
+
 def slot_is_fresh(
     path: Path,
     *,
@@ -157,9 +168,35 @@ def slot_is_fresh(
     if expiry <= current + timedelta(seconds=grace_seconds):
         age = int((current - expiry).total_seconds())
         if age >= 0:
-            return False, f"expired {age // 60}m ago (at {expiry.isoformat()})"
-        return False, f"expires within grace ({-age}s left, at {expiry.isoformat()})"
+            return (
+                False,
+                f"{REASON_EXPIRED} {age // 60}m ago (at {expiry.isoformat()})",
+            )
+        return (
+            False,
+            f"{REASON_WITHIN_GRACE} ({-age}s left, at {expiry.isoformat()})",
+        )
     return True, f"valid until {expiry.isoformat()}"
+
+
+def reason_is_refreshable(reason: str) -> bool:
+    """Whether a :func:`slot_is_fresh` failure reason is *only* an expiry problem.
+
+    An expired access token is not necessarily a dead credential: Claude Code
+    refreshes it from the stored refresh token on first use. Callers can probe
+    such a slot online and, when the probe succeeds, switch into it with
+    ``probe_ok=True``.
+
+    Returns ``False`` for reasons no probe can fix (missing slot file,
+    unreadable/malformed slot payload) so those keep failing hard, and for
+    success reasons ("valid until ...").
+
+    Use this instead of substring-matching reason text at the call site: the
+    grace-window reason reads "expire\\ **s** within grace", so a naive
+    ``"expired" in reason`` check silently misses it.
+    """
+    lowered = reason.lower()
+    return any(lowered.startswith(marker) for marker in _REFRESHABLE_REASON_MARKERS)
 
 
 def _hash_file(path: Path) -> str | None:
--- a/packages/gptme-subscription/src/gptme_subscription/cli.py
+++ b/packages/gptme-subscription/src/gptme_subscription/cli.py
@@ -354,6 +354,10 @@ def _cmd_switch(args: argparse.Namespace, sm: SubscriptionManager) -> int:
         return 0
     ok = sm.switch_to(target, f"manual switch via --switch {target}", force=True)
     failure_detail: str | None = None
+    # Whether we got in via the probe_ok retry below. That path deliberately
+    # accepts a slot whose access token is still expired, so the post-flip usage
+    # check cannot succeed and must not be reported as a credential problem.
+    switched_via_probe_ok = False
     if not ok:
         # Retry with probe_ok=True if the slot has an expired access token but a
         # valid refresh token — CC auto-refreshes on first use, so this is safe.
@@ -384,6 +388,7 @@ def _cmd_switch(args: argparse.Namespace, sm: SubscriptionManager) -> int:
                         force=True,
                         probe_ok=True,
                     )
+                    switched_via_probe_ok = ok
                     if not ok:
                         retry_fresh, retry_reason = sm.slot_is_fresh(target)
                         failure_detail = (
@@ -420,14 +425,29 @@ def _cmd_switch(args: argparse.Namespace, sm: SubscriptionManager) -> int:
         if not sm.check_usage(no_cache=True) and (
             usage_script is not None and usage_script.exists()
         ):
-            print(
-                f"  WARNING: post-switch usage check against {target} returned "
-                "no usage data — the switch DID happen and is recorded. Either "
-                f"the credential did not answer, or {usage_script} itself failed "
-                "(not executable / timed out / bad output). Verify with "
-                f"`--status --probe` before assuming {target} needs a re-login.",
-                file=sys.stderr,
-            )
+            if switched_via_probe_ok:
+                # Expected, not a fault: we got here precisely because the access
+                # token is expired but refreshable. Claude Code refreshes it on
+                # first use, so the usage script is querying with a token that is
+                # still stale and will 401. Reporting this as "the credential did
+                # not answer" would raise an alarm on every successful retry.
+                print(
+                    f"  Note: post-switch usage check against {target} returned "
+                    "no data, which is expected here — the switch was allowed on "
+                    "a verified refresh token while the access token is still "
+                    "expired. Claude Code refreshes it on first use; usage data "
+                    "appears after that.",
+                    file=sys.stderr,
+                )
+            else:
+                print(
+                    f"  WARNING: post-switch usage check against {target} returned "
+                    "no usage data — the switch DID happen and is recorded. Either "
+                    f"the credential did not answer, or {usage_script} itself failed "
+                    "(not executable / timed out / bad output). Verify with "
+                    f"`--status --probe` before assuming {target} needs a re-login.",
+                    file=sys.stderr,
+                )
         return 0
     # Never exit 1 in silence: a bare non-zero exit is the diagnosis-free
     # failure this command's retry path exists to remove. Say what blocked it.
--- a/packages/gptme-subscription/src/gptme_subscription/cli.py
+++ b/packages/gptme-subscription/src/gptme_subscription/cli.py
@@ -364,23 +364,31 @@ def _cmd_switch(args: argparse.Namespace, sm: SubscriptionManager) -> int:
         if not fresh:
             failure_detail = fresh_reason
         if not fresh and reason_is_refreshable(fresh_reason):
-            print(
-                f"  Access token stale ({fresh_reason}); "
-                f"probing {target!r} to verify refresh token...",
-                file=sys.stderr,
-            )
-            _, probe_ok, probe_msg = probe_credential(
-                sm.config.slot_path(target), target, usage_script=sm.config.usage_script
-            )
-            if probe_ok:
-                ok = sm.switch_to(
-                    target,
-                    f"manual switch via --switch {target} (probe_ok)",
-                    force=True,
-                    probe_ok=True,
-                )
+            if not sm.slot_has_refresh_token(target):
+                failure_detail = f"{fresh_reason}; slot holds no refresh token"
             else:
-                print(f"  Probe failed ({probe_msg}); cannot switch.", file=sys.stderr)
+                print(
+                    f"  Access token stale ({fresh_reason}); "
+                    f"probing {target!r} to verify refresh token...",
+                    file=sys.stderr,
+                )
+                _, probe_ok, probe_msg = probe_credential(
+                    sm.config.slot_path(target),
+                    target,
+                    usage_script=sm.config.usage_script,
+                )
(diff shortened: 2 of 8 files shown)


TimeToBuildBob:
## 🤖 AI code review

Adds `probe_ok` to `SlotManager.switch_to` and `SubscriptionManager.switch_to`, letting a slot with an expired access token be switched into when the caller asserts an online probe succeeded, gated on the freshness failure being expiry-class and the slot holding a refresh token. Adds `reason_is_refreshable` and `slot_has_refresh_token` helpers. `_cmd_switch` now probes stale-but-refreshable slots and retries with `probe_ok=True`, and warns when the post-flip usage check returns no data.

**2 findings** · ❌ **1** P1 · ⚠️ **1** P2 — 🛑 **1 blocking**, worth resolving before merge.

**❌ P1 high** — `packages/gptme-subscription/src/gptme_subscription/cli.py:420`

After a successful probe_ok retry, the live symlink points at a slot whose access token is still expired — CC refreshes it only on first Claude Code use. `_cmd_switch` then runs `sm.check_usage(no_cache=True)` against that live credential; the usage script queries the API with the expired token, gets 401, and returns None. Because `usage_script` exists, the new warning fires: 'the switch DID happen... Either the credential did not answer, or {usage_script} itself failed... Verify with --status --probe'. For exactly the stale-but-refreshable scenario this PR enables, every successful retry (with a usage script configured) emits this false alarm — the credential is fine, it simply has not been refreshed yet. None of the new tests exercise the probe_ok retry with a usage script present, so this misleading warning ships unverified.

<sub>**How this was verified:** Traced `_cmd_switch`: the retry at line 381 flips the symlink to the expired-token slot, then line 420's check runs the usage script against it. `check_usage` (manager.py:292-296) returns None on non-zero exit/OSError, which the usage script produces against a 401. Confirmed no test sets both a usage_script and a probe_ok retry.</sub>

**⚠️ P2 medium** — `packages/gptme-subscription/src/gptme_subscription/cli.py:392`

In `_cmd_switch`, after the probe_ok retry fails, the code re-checks `slot_is_fresh` and selects `retry_reason` or the literal 'switch refused after successful probe'. But a retry with `probe_ok=True` only returns `ok=False` cleanly when the slot is missing (`SlotManager.switch_to` returns 'slot missing' before the freshness gate); the expiry gate is bypassed, and the symlink-flip path raises on real errors, which propagates out of `_cmd_switch` uncaught. So whenever the retry fails without raising, `slot_is_fresh` reports the slot missing and `retry_fresh` is False — the `else` branch is unreachable. If a future change lets the retry fail for a still-present slot, this branch would report 'switch refused after successful probe', misleading an operator debugging a filesystem-level failure instead of the actual cause.

<sub>**How this was verified:** Checked `SlotManager.switch_to` (manager.py:636-695): with force=True and probe_ok=True, the only ok=False paths are unknown-sub and slot-missing; the tmp-symlink/os.replace block re-raises on failure. `_cmd_switch` already validated the subscription, so only slot-missing reaches the retry-failure branch, making `retry_fresh` always False.</sub>

</details>

- `packages/credential-slots/README.md` — Documents the new `reason_is_refreshable` helper and the `probe_ok=True` switch path with a usage example.
- `packages/credential-slots/src/credential_slots/__init__.py` — Exports `reason_is_refreshable` in the package's public API list.
- `packages/credential-slots/src/credential_slots/manager.py` — Adds `REASON_EXPIRED`/`REASON_WITHIN_GRACE` constants, `reason_is_refreshable`, `slot_has_refresh_token`, and a `probe_ok` bypass in `switch_to`.
- `packages/credential-slots/tests/test_manager.py` — Adds `TestReasonIsRefreshable` and tests for `slot_has_refresh_token` and the `probe_ok` bypass/refusal paths.
- `packages/gptme-subscription/src/gptme_subscription/cli.py` — `_cmd_switch` probes a stale-but-refreshable slot and retries with `probe_ok=True`; adds a post-flip usage-check warning and a failure-detail message.
- `packages/gptme-subscription/src/gptme_subscription/manager.py` — Forwards `probe_ok` in `switch_to`, adds `slot_has_refresh_token`, and wires the SlotManager logger.
- `packages/gptme-subscription/tests/test_cli.py` — Adds CLI tests for the probe/retry path, non-probeable failures, and the post-flip warning.
- `packages/gptme-subscription/tests/test_subscription_manager.py` — Adds tests for `probe_ok` forwarding through the manager seam and production logging of the bypass.

</details>

| `f1383ac48ea7` | ? | 2 | llm | 2026-08-08 00:51 UTC |
| `6dcc4f7c48d6` | ? | 3 | llm | 2026-08-08 07:53 UTC |
| `8f22bd51e912` | ? | 4 | llm | 2026-08-08 09:49 UTC |
| `1ec88a47d976` | 4/5 | 5 | llm | 2026-08-09 00:30 UTC |
| `1b343b4dadbe` | 3/5 | 2 | llm | 2026-08-09 16:20 UTC |

</details>

<sub>Reviewed `ae087643cc22` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 202s · advisory, a human still decides · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

packages/gptme-subscription/src/gptme_subscription/cli.py
```diff
@@ -351,6 +351,27 @@ def _cmd_switch(args: argparse.Namespace, sm: SubscriptionManager) -> int:
         print("  use --execute to apply")
         return 0
     ok = sm.switch_to(target, f"manual switch via --switch {target}", force=True)
+    if not ok:
+        # Retry with probe_ok=True if the slot has an expired access token but a
+        # valid refresh token — CC auto-refreshes on first use, so this is safe.
+        fresh, fresh_reason = sm._slot_manager.slot_is_fresh(target)
+        if not fresh and "expired" in fresh_reason.lower():
```
TimeToBuildBob: Fixed in 6dcc4f7c. Confirmed real: `slot_is_fresh` emits "expires within grace (Ns left, ...)" for the grace window — "expire**s**", so `"expired" in reason` never matched and `--switch --execute` failed with exit 1 for the exact recoverable case. Rather than widening the substring match, I moved the classification next to its producer: new `credential_slots.reason_is_refreshable()` built from the same reason-prefix constants `slot_is_fresh` formats with, so the two can't drift. Reason strings are byte-identical (backward compatible); missing/unreadable slots stay non-refreshable and still fail hard. Regression test `TestReasonIsRefreshable` drives every `slot_is_fresh` branch through real slot files — verified it fails (`assert False is True` on the grace reason) with the old naive predicate.

packages/gptme-subscription/src/gptme_subscription/cli.py
```diff
@@ -351,6 +351,27 @@ def _cmd_switch(args: argparse.Namespace, sm: SubscriptionManager) -> int:
         print("  use --execute to apply")
         return 0
     ok = sm.switch_to(target, f"manual switch via --switch {target}", force=True)
+    if not ok:
```
TimeToBuildBob, line 357: Fixed in 6dcc4f7c. Confirmed real — the two prior tests only exercised `SlotManager.switch_to(probe_ok=True)`, nothing covered the CLI flow. `_cmd_switch` now has direct coverage: initial failed switch → `slot_is_fresh` re-check → `probe_credential` (monkeypatched, never a real probe) → conditional retry with `probe_ok=True` → `record_manual_switch_hold` → rc 0, for both the expired and grace-window reasons; probe failure → no retry, rc 1, message on stderr; missing/unreadable slot → never probed, rc 1; fresh slot whose switch failed for another reason (deferred locks) → never probed. Freshness reasons in these tests are generated by the real `slot_is_fresh` against synthetic slot files rather than hand-typed, so a reason-wording change can't leave the CLI test passing against a string that no longer exists. The existing dry-run test still passes.

packages/credential-slots/src/credential_slots/manager.py
```diff
@@ -588,11 +629,12 @@ def switch_to(
             self._log(msg)
             return SwitchResult(ok=False, reason=msg)
 
-        fresh, fresh_reason = self.slot_is_fresh(sub)
-        if not fresh:
-            msg = f"refusing to switch to {sub}: {fresh_reason}"
-            self._log(msg)
-            return SwitchResult(ok=False, reason=msg)
+        if not probe_ok:
```
TimeToBuildBob: Real — fixed in 8f22bd51 (both halves).

**Refresh-token requirement.** `probe_ok` now only earns the bypass when the slot actually holds a refresh token, reusing the existing `_read_refresh_token` in the same module rather than adding a second reader. The reasoning: `probe_ok` is the assertion "an online probe confirmed CC can refresh this slot". With no refresh token there is nothing to refresh, so the assertion cannot be about this slot and the offline expiry gate stays in force — and the ignore is logged (`probe_ok IGNORED for {sub}: slot holds no refresh token…`).

**Distinct logging.** Every honored bypass now emits `probe_ok BYPASS: skipping expiry gate for {sub} …` before the flip, so reading a switch log after an incident can never confuse a skipped safety check with an ordinary switch.

**README.** Updated the example at ~L88 so it no longer reads as "pass `probe_ok` freely" — it now states that the flag disables a gate `force=True` cannot, that it needs a *fresh* probe result in hand, and that it is only honored for slots with a refresh token.

Regression tests, each verified failing with only this source change reverted:
- `test_probe_ok_ignored_when_slot_has_no_refresh_token` → `AssertionError: assert True is False / where True = SwitchResult(ok=True, reason='switched to bob').ok`
- `test_probe_ok_bypass_is_logged_distinctly` → `assert any("probe_ok BYPASS" in line …)` → `assert False`
- `test_switch_to_probe_ok_still_refused_without_refresh_token` (SubscriptionManager level) → `assert True is False`

packages/gptme-subscription/tests/test_cli.py
```diff
@@ -96,12 +106,19 @@ def evaluate(
     def seconds_since_last_primary_departure(self) -> None:
         return None
 
-    def switch_to(self, sub: str, reason: str, force: bool = False) -> bool:
+    def switch_to(
+        self, sub: str, reason: str, force: bool = False, probe_ok: bool = False
+    ) -> bool:
         self.switch_calls.append((sub, reason))
+        self.switch_probe_ok_flags.append(probe_ok)
```
TimeToBuildBob, line 114: Real — fixed in 8f22bd51, and the blind spot you describe is exactly what the new tests catch.

Added two layers, both over real objects with tmp creds/state dirs (no real slot files are touched):
- `test_switch_to_forwards_probe_ok_through_to_slot_manager` (test_subscription_manager.py) — real `SubscriptionManager`, real slot payloads: asserts `force=True` alone is refused, then that `probe_ok=True` flips the live symlink and `creds_link.resolve()` really points at the target slot.
- `test_cmd_switch_probe_ok_retry_flips_symlink_end_to_end` (test_cli.py) — the full CLI → SubscriptionManager → SlotManager chain through `_cmd_switch`, with only `probe_credential` mocked, asserting the symlink moved.

Verified by mutating the seam exactly as you predicted — dropping `probe_ok=probe_ok` from `SubscriptionManager.switch_to`:
```
FAILED tests/test_cli.py::test_cmd_switch_probe_ok_retry_flips_symlink_end_to_end
  assert rc == 0 → assert 1 == 0
  stderr: Access token stale (expired 60m ago …); probing 'alice' to verify refresh token...
FAILED tests/test_subscription_manager.py::test_switch_to_forwards_probe_ok_through_to_slot_manager
  assert sm.switch_to("bob", …, probe_ok=True) → assert False
2 failed, 222 passed
```
222 passed — i.e. every FakeManager-based probe_ok test stayed green while the real behaviour was broken, which is precisely the gap you flagged.

packages/gptme-subscription/src/gptme_subscription/cli.py
```diff
@@ -351,6 +353,31 @@ def _cmd_switch(args: argparse.Namespace, sm: SubscriptionManager) -> int:
...
+        if not fresh and reason_is_refreshable(fresh_reason):
+            print(
+                f"  Access token stale ({fresh_reason}); "
+                f"probing {target!r} to verify refresh token...",
+                file=sys.stderr,
+            )
+            _, probe_ok, probe_msg = probe_credential(
+                sm.config.slot_path(target), target, usage_script=sm.config.usage_script
+            )
+            if probe_ok:
```
TimeToBuildBob, line 375: Partially real. The discarded `check_usage` is a genuine bug and is fixed in 8f22bd51; the framing around it I'm pushing back on.

**Fixed:** `_cmd_switch` called `sm.check_usage(no_cache=True)` after the flip and threw the result away. That call runs the usage script against the *live* credential — i.e. the slot just switched into — so it is the one post-flip signal we have, and dropping it was wrong. It now warns loudly on stderr:
```
WARNING: post-switch usage check against <slot> failed — the switch DID happen and
is recorded, but the credential did not answer. Verify with `--status --probe`; you
may need to re-run /login for <slot>.
```
`check_usage` also returns `None` when no usage script is configured, which says nothing about the credential, so the warning fires only when a script was actually available to run.

**Exit code: deliberately still 0**, and called out in the commit message so scripted callers aren't surprised. The switch genuinely happened and is recorded in the switch log and the manual hold; returning 1 would tell every caller "the switch did not happen", which is false and strictly worse — an operator or script would retry or roll back a flip that is already durable state. The honest report is "it happened, and here is a loud warning that the credential did not answer". Two tests pin it: warning present on a failed check, absent on a healthy one, both against a real `SubscriptionManager`. Verified failing with only the CLI change reverted: `assert 'post-switch usage check against alice failed' in ''`.

**Push-back on the rest:**

1. *"previously the same slot was refused outright and the CLI exited 1"* frames the old behaviour as the safer baseline. That refusal **is the bug this PR exists to fix** (operator session ef48): the operator could not switch at all into a slot whose refresh token was perfectly valid and whose access token CC would have refreshed on first use. "Refuses to act" is not a safety property when the refusal is wrong.

2. *No post-flip re-validation beyond the probe.* The probe **is** the validation, and it is stronger than the offline gate it replaces — it exercises the real refresh against the account, which is exactly the server-side-invalidation class README L27-49 says the offline `expiresAt` check cannot catch. Requiring the flip to be conditional on the credential surviving some later window is not achievable: the revocation race you describe ("between the probe's success and CC's first use") exists for every credential at every moment, including one with a comfortably future `expiresAt` that passes the offline gate today. There is no flip-time check that closes it; detection belongs in the runner's 401 response classifier, which is where README L10-12 already puts it.

So: signal no longer discarded, exit-code semantics unchanged and now justified rather than accidental.

packages/gptme-subscription/tests/test_cli.py
```diff
@@ -353,3 +371,258 @@ def test_cmd_switch_dry_run_does_not_switch_or_hold(capsys) -> None:
...
+    tests honest: if ``slot_is_fresh``'s reason wording changes, the classifier
+    it feeds is exercised against the new wording here too.
+
+    ``write=False`` leaves the file absent (missing-slot reason);
+    ``expires_in=None`` writes a payload with no ``expiresAt`` (unreadable).
+    """
+    path = tmp_path / ".credentials.json.alice"
+    if write:
+        oauth: dict[str, object] = {"accessToken": "fake"}
+        if expires_in is not None:
```
TimeToBuildBob, line 391: ⚠️ **P2** — The parametrized CLI test `test_cmd_switch_stale_token_probes_and_retries_with_probe_ok` asserts rc=0 and a successful flip for a scenario a real SlotManager refuses. Its fixture `_real_fresh_reason` writes a payload with only `accessToken` and `expiresAt` — no refreshToken — and `FakeManager.switch_to` returns the canned True without consulting the slot, so the retry 'succeeds' in the test. In production, `SlotManager.switch_to` hits the `probe_ok IGNORED` branch (`_read_refresh_token(slot)` is None at manager.py:643), keeps the expiry gate in force, refuses, and `_cmd_switch` returns 1. The grace-window retry case (60s, `expires_in=60`) is therefore validated by a path that production rejects, and the test would stay green even if the refresh-token requirement were deleted; the only real-chain coverage is the -3600 expired case in `test_cmd_switch_probe_ok_retry_flips_symlink_end_to_end`, which does write a refreshToken.

packages/gptme-subscription/src/gptme_subscription/cli.py
```diff
@@ -351,14 +353,55 @@ def _cmd_switch(args: argparse.Namespace, sm: SubscriptionManager) -> int:
...
         print(f"Switched to {target}")
-        sm.check_usage(no_cache=True)
+        # Post-flip validation: this runs the usage script against the *live*
+        # credential, i.e. the slot we just switched into. Don't discard the
+        # result — a failure here is the one signal we have that the credential
+        # is not usable after the flip. ``check_usage`` also returns None when
+        # no usage script is configured, which says nothing about the
+        # credential, so only warn when a script was actually available to run.
+        usage_script = sm.config.usage_script
+        if sm.check_usage(no_cache=True) is None and (
```
TimeToBuildBob: ⚠️ **P2** — The post-flip warning misattributes a broken usage script to a dead credential. `sm.check_usage(no_cache=True)` returns None not only when the live credential is unusable but also when the installed script exists yet fails to run: non-executable file (PermissionError/OSError), non-zero exit, empty stdout, invalid JSON, or 60s timeout. The guard tests only `usage_script.exists()`, so any such failure prints 'the credential did not answer ... you may need to re-run /login for {target}'. The PR's own test forces this misdiagnosis: `test_cmd_switch_warns_loudly_when_post_flip_usage_check_fails` chmods the script to 0o644 so check_usage returns None while the credential is perfectly healthy. An operator will re-run /login against a valid credential while the actual breakage (the usage script) is never mentioned; the warning should distinguish 'script could not run' from 'credential probe failed'.

packages/gptme-subscription/src/gptme_subscription/cli.py
```diff
@@ -351,14 +353,55 @@ def _cmd_switch(args: argparse.Namespace, sm: SubscriptionManager) -> int:
...
+            print(
+                f"  Access token stale ({fresh_reason}); "
+                f"probing {target!r} to verify refresh token...",
+                file=sys.stderr,
+            )
+            _, probe_ok, probe_msg = probe_credential(
+                sm.config.slot_path(target), target, usage_script=sm.config.usage_script
+            )
+            if probe_ok:
+                ok = sm.switch_to(
```
TimeToBuildBob, line 376: ⚠️ **P2** — When the probe succeeds but the retry `sm.switch_to(..., probe_ok=True)` is itself refused, `_cmd_switch` falls through to `return 1` with no diagnostic at all. The 'Probe failed (...); cannot switch.' print at line 379 covers only the `probe_ok` falsy branch; if `probe_ok` is True yet the SlotManager still refuses — e.g. the slot file was rewritten between probe and retry so `_read_refresh_token` returns None, triggering the `probe_ok IGNORED` gate, or a concurrent modification makes the slot disappear — `ok` stays False and the operator sees only the earlier 'probing ...' info line followed by a bare exit 1. This is precisely the silent `--switch --execute` failure the PR claims to fix, and it is indistinguishable from 'probe said no' in the output.

packages/credential-slots/src/credential_slots/manager.py
```diff
@@ -588,11 +633,35 @@ def switch_to(
...
-            self._log(msg)
-            return SwitchResult(ok=False, reason=msg)
+        # ``probe_ok`` is a caller assertion ("an online probe says this slot
+        # works"), so it only earns the bypass when the slot could actually be
+        # revived the way the assertion implies: CC refreshes an expired access
+        # token from the stored refresh token. No refresh token, nothing to
+        # refresh — the assertion cannot be about this slot, so the offline gate
+        # stays in force.
+        bypass_expiry = probe_ok
+        if probe_ok and _read_refresh_token(slot) is None:
```
TimeToBuildBob: ⚠️ **P2** — TOCTOU between probe and switch: `probe_credential` (cli.py:369-371) reads the slot file and confirms its refresh token works, but `SlotManager.switch_to` re-reads the same path to decide the bypass, and `_read_refresh_token` only asks whether a non-empty `refreshToken` string exists — it cannot tell whether that string is the probed token. If the slot file is replaced in the window between the probe and the retry (e.g. a concurrent `claude /login` writing through the live symlink onto the slot, or another manager healing drift), a replacement file that merely contains any refreshToken string — including a dead token from a different identity — earns the bypass, and the symlink flips onto a credential the probe never validated. The refresh token is not fingerprinted here, so the bypass cannot detect the substitution; the observable consequence is a 401 crash loop on the live slot, the exact failure the expiry gate exists to prevent.

TimeToBuildBob:
@greptileai review

packages/gptme-subscription/src/gptme_subscription/cli.py
```diff
@@ -351,15 +353,71 @@ def _cmd_switch(args: argparse.Namespace, sm: SubscriptionManager) -> int:
...
+    if not ok:
+        # Retry with probe_ok=True if the slot has an expired access token but a
+        # valid refresh token — CC auto-refreshes on first use, so this is safe.
+        # ``reason_is_refreshable`` lives next to ``slot_is_fresh`` so the two
+        # can't drift; it covers both "expired Nm ago" and "expires within
+        # grace", and excludes missing/unreadable slots (no probe can fix those).
+        fresh, fresh_reason = sm.slot_is_fresh(target)
+        if not fresh:
+            failure_detail = fresh_reason
+        if not fresh and reason_is_refreshable(fresh_reason):
```
TimeToBuildBob, line 370: Fixed in 1b343b4d. `_cmd_switch` now checks the slot for a usable refresh token before doing any online probe. A stale slot without one skips the futile probe, stays untouched, and reports `slot holds no refresh token`; the regression test asserts `probe_credential` is never called. The lower-level `SlotManager.switch_to(..., probe_ok=True)` guard remains as defense in depth.

packages/gptme-subscription/src/gptme_subscription/manager.py
```diff
@@ -116,6 +116,12 @@ def __init__(self, config: Config):
             fingerprint_template=config.fingerprint_template,
             lock_guard=self._lock_guard,
             on_switch=self._log_switch,
+            # Without this the SlotManager's audit trail is a no-op: its
+            # ``probe_ok BYPASS`` / ``probe_ok IGNORED`` / refusal lines are
+            # emitted only via ``SlotManager._log``, which discards everything
+            # when ``logger is None``. This is the production wiring, so a
+            # skipped safety check has to be visible here, not just in tests.
+            logger=logger.warning,
```
TimeToBuildBob, line 132: ❌ **P1** — `SubscriptionManager.__init__` passes `logger=logger.warning` as the `SlotManager` logger. `SlotManager._log` calls `self.logger(msg)` with a single already-formatted string. `logging.Logger.warning(msg)` treats `msg` as a format string, so any `%` character in the interpolated slot name, reason, or freshness reason raises a `ValueError: incomplete format` (or similar). For example, a slot named `bob%` or a reason string like "100% exhausted" passed to the `_cmd_switch` retry makes `switch_to` crash with an unhandled exception instead of returning a `SwitchResult`; the CLI then exits with a traceback and the live symlink is left untouched. The fix is to wrap the logger in a lambda that forces `%s` formatting, e.g. `logger=lambda msg: logger.warning("%s", msg)`. This defect is triggered on any switch attempt that reaches a `_log` call with a percent sign in the message, which is plausible in normal operation.

packages/gptme-subscription/src/gptme_subscription/cli.py
```diff
@@ -351,15 +353,79 @@ def _cmd_switch(args: argparse.Namespace, sm: SubscriptionManager) -> int:
...
+                    f"probing {target!r} to verify refresh token...",
+                    file=sys.stderr,
+                )
+                _, probe_ok, probe_msg = probe_credential(
+                    sm.config.slot_path(target),
+                    target,
+                    usage_script=sm.config.usage_script,
+                )
+                if probe_ok:
+                    ok = sm.switch_to(
```
TimeToBuildBob, line 385: ⚠️ **P2** — In `_cmd_switch`, when the initial `switch_to` fails for an expiry-class reason and the probe succeeds, the retry calls `sm.switch_to(... probe_ok=True)` (line 381). If that retry then fails for any reason other than expiry — e.g. the slot file is deleted between the probe and the retry, or a concurrent process removes the slot — `ok` becomes False and the flow falls through to the final error message. `failure_detail` still holds the original `fresh_reason` from the earlier `slot_is_fresh` call, so the operator sees `Failed to switch to alice: expired 60m ago` even though the actual cause is a missing slot. This misdiagnoses the failure and sends the operator to a token-refresh investigation for a file-existence problem. The code should recompute `failure_detail` from the retry's returned exception or `SwitchResult.reason`, or at least clear it when a retry is attempted.

TimeToBuildBob: Fixed in ae087643. After a successful probe, a failed retry now rechecks `slot_is_fresh()` and replaces the pre-probe expiry detail with the current reason. The regression removes the slot between probe and retry and asserts the final failure line reports `slot missing`, not the stale `expired 60m ago` diagnosis. Focused credential/subscription tests: 145 passed; pre-commit passed.

packages/gptme-subscription/src/gptme_subscription/cli.py
```diff
@@ -351,15 +353,86 @@ def _cmd_switch(args: argparse.Namespace, sm: SubscriptionManager) -> int:
...
+        # credential, so only warn when a script was actually available to run.
+        #
+        # Use a falsy test, matching ``_execute_switch_decision``: a script that
+        # exits 0 printing ``{}`` yields an empty dict, which is just as useless
+        # a post-flip signal as None. And keep the wording honest — check_usage
+        # also returns None when the *script* failed (not executable, timed out,
+        # non-zero exit, unparseable output), which says nothing about the
+        # credential. Don't send an operator to /login for their own script bug.
+        usage_script = sm.config.usage_script
+        if not sm.check_usage(no_cache=True) and (
```
TimeToBuildBob, line 425: ❌ **P1** — After a successful probe_ok retry, the live symlink points at a slot whose access token is still expired — CC refreshes it only on first Claude Code use. `_cmd_switch` then runs `sm.check_usage(no_cache=True)` against that live credential; the usage script queries the API with the expired token, gets 401, and returns None. Because `usage_script` exists, the new warning fires: 'the switch DID happen... Either the credential did not answer, or {usage_script} itself failed... Verify with --status --probe'. For exactly the stale-but-refreshable scenario this PR enables, every successful retry (with a usage script configured) emits this false alarm — the credential is fine, it simply has not been refreshed yet. None of the new tests exercise the probe_ok retry with a usage script present, so this misleading warning ships unverified.

packages/gptme-subscription/src/gptme_subscription/cli.py
```diff
@@ -351,15 +353,86 @@ def _cmd_switch(args: argparse.Namespace, sm: SubscriptionManager) -> int:
...
+                        f"manual switch via --switch {target} (probe_ok)",
+                        force=True,
+                        probe_ok=True,
+                    )
+                    if not ok:
+                        retry_fresh, retry_reason = sm.slot_is_fresh(target)
+                        failure_detail = (
+                            retry_reason
+                            if not retry_fresh
+                            else "switch refused after successful probe"
```
TimeToBuildBob, line 397: ⚠️ **P2** — In `_cmd_switch`, after the probe_ok retry fails, the code re-checks `slot_is_fresh` and selects `retry_reason` or the literal 'switch refused after successful probe'. But a retry with `probe_ok=True` only returns `ok=False` cleanly when the slot is missing (`SlotManager.switch_to` returns 'slot missing' before the freshness gate); the expiry gate is bypassed, and the symlink-flip path raises on real errors, which propagates out of `_cmd_switch` uncaught. So whenever the retry fails without raising, `slot_is_fresh` reports the slot missing and `retry_fresh` is False — the `else` branch is unreachable. If a future change lets the retry fail for a still-present slot, this branch would report 'switch refused after successful probe', misleading an operator debugging a filesystem-level failure instead of the actual cause.

packages/gptme-subscription/src/gptme_subscription/manager.py
TimeToBuildBob: **False positive — dismissed, with evidence.** Also note this thread had been *resolved with no reply at all*, which is not a disposition; recording the reasoning now.

The claim is that `logger=logger.warning` makes the message a format string, so `%` in a slot name or reason raises `ValueError: incomplete format`. That is not how `logging` behaves: `LogRecord.getMessage()` only applies `msg % self.args` **when `self.args` is truthy**. `SlotManager._log` calls `self.logger(msg)` with exactly one argument, so there are no args and no interpolation.

Measured directly:

```
>>> log.warning("probe_ok BYPASS for slot bob%: 100% exhausted")
WARNING:t:probe_ok BYPASS for slot bob%: 100% exhausted
RESULT: no exception
```

So the suggested fix (`logger=lambda msg: logger.warning("%s", msg)`) would have been a no-op wrapper around already-correct behaviour, and I have not applied it.

The finding does point at something real one step away, though: the hazard appears the moment someone rewrites the call as `logger.warning(fmt, *args)`. So the wiring stays as-is with a comment saying why, plus `test_slot_logger_survives_percent_signs_in_messages`, which fails if that rewrite ever happens. Fixed the risk without pretending the bug was live.

packages/gptme-subscription/src/gptme_subscription/cli.py
TimeToBuildBob: **Confirmed real — fixed in `2cd18a8`.** This thread had also been resolved with no reply; recording the disposition properly.

The analysis is exactly right: the retry deliberately switches into a slot whose access token is still expired, CC refreshes only on first use, so `check_usage(no_cache=True)` queries with a stale token, 401s, and returns None. With a `usage_script` configured, the new warning then fired on **every successful retry** — a false alarm for precisely the scenario this PR exists to enable, pointing the operator at `--status --probe` or a needless re-login.

Fix: `_cmd_switch` now tracks whether the switch came in via the probe_ok path. On that path the post-flip check still runs and is still reported, so no signal is lost, but it is reported as expected rather than as a possible bad credential. The original warning is untouched for ordinary switches.

You were also right that this shipped unverified. Two tests now cover it: one asserts the retry path emits no `WARNING` / "did not answer" / "re-login" wording, and a companion asserts a plain switch with no usage data *still* warns — so the carve-out cannot silently swallow the real post-flip credential check.

TimeToBuildBob:
## Review complete — merge when CI goes green

Both P1s are disposed in `2cd18a8`, and both threads had previously been **resolved with no reply at all**, which is not a disposition — the reasoning is now on the record:

- **P1 (probe_ok retry false alarm): real, fixed.** The retry deliberately switches into a slot whose access token is still expired, so the post-flip usage check was guaranteed to return nothing and fire "the credential did not answer" on *every* successful retry. The check still runs and is still reported, but on that path as expected rather than as a fault. A companion test pins that ordinary switches still warn, so the carve-out cannot swallow the real check.
- **P1 (`%` in logger message): false positive, dismissed with evidence.** `LogRecord.getMessage()` only interpolates when `args` is truthy and `SlotManager._log` passes exactly one argument, so no `ValueError` is possible — measured directly. The suggested `lambda msg: logger.warning("%s", msg)` would have been a no-op wrapper. Kept the current wiring plus a regression test that fails if anyone rewrites the call as `logger.warning(fmt, *args)`, which is what would actually introduce the hazard.

Local verification: 237 tests across `gptme-subscription` and `credential-slots` pass, ruff clean.

**Not merged only because CI is backed up** — the contrib runner is saturated and this PR's `pre-commit` job is still queued.

TimeToBuildBob:

**Gate objection (recorded):** `Greptile review not found` (dark on this repo since `2026-08-09T05:27Z` while reviewing `gptme/gptme` normally — cannot clear by waiting) + `Touches sensitive/security/infra paths`, which is correct for credential-slot switching and which I have not weakened.

**Both P1s disposed.** Both threads had previously been **resolved with no reply at all**, which is not a disposition — the reasoning is now on the record:

- **Probe_ok retry false alarm — real, fixed in `2cd18a8`.** The retry exists to switch into a slot whose access token is expired but whose refresh token is verified good; CC refreshes on first use, so the post-flip usage check necessarily queried with a stale token and 401d. The "the credential did not answer ... verify with `--status --probe`" warning therefore fired on **every successful retry**, for exactly the scenario this PR adds. The check still runs and is still reported, but on that path as expected rather than as a fault. A companion test pins that ordinary switches still warn, so the carve-out cannot swallow the real post-flip check.
- **`%` in logger message — false positive, dismissed with evidence.** `LogRecord.getMessage()` only interpolates when `args` is truthy, and `SlotManager._log` passes exactly one argument, so no `ValueError` is reachable — measured directly. The suggested `lambda msg: logger.warning("%s", msg)` would have been a no-op wrapper presented as a fix. Kept the wiring, added a regression test that fails if anyone rewrites the call as `logger.warning(fmt, *args)` — which is what would actually create the hazard.

**Surviving findings: none.** 0 unresolved threads, CI green, 237 tests pass across `gptme-subscription` and `credential-slots`.

Rebased onto master along the way, which also cleared a `ruff-format` failure caused by an earlier format pass run with a newer ruff than the pinned `v0.6.9`.

PR #1390, opened 2026-08-08
fix(pm): act on our own AI reviewer's findings instead of silencing them

TimeToBuildBob:
## Problem (spotted by Erik)

PM is tuned for Greptile and was never taught about the replacement. In `activity-gate.sh`: `greptile` appears **~70 times**, `bob-ai-review` appeared **zero times**.

Bob's self-hosted reviewer (ErikBjare/bob#1122) posts through Bob's **user** account, not a GitHub App. So `has_actionable_update`'s self-trigger guard — *last actor is $AUTHOR* — classified every AI review as the agent talking to itself and dropped it.

**Net effect: PM listened to the reviewer we stopped paying for, and ignored the one we run.** Reviews were posted and went nowhere.

## Fix

Exempt comments carrying `<!-- bob-ai-review` from the self-skip.

Keyed on the **marker, not the login**, deliberately: Bob's genuine human-directed comments carry no marker and must **still** be silenced — those are the actual self-trigger loop this guard exists to stop.

**Loop safety**: this is the same cycle Greptile already drives (review → dispatch → push → re-review), terminating the same way via the existing fingerprint dedup and per-PR attempt counters. Bounded by the PR being fixed or merged, not by the reviewer's identity.

## Tests

5 new. The two AI-review cases **fail against the previous guard**; the loop guard (Bob's plain comment still silenced), human comments, and Greptile reviews are all pinned so nothing regresses.

### DIFF
--- a/scripts/github/activity-gate.sh
+++ b/scripts/github/activity-gate.sh
@@ -111,8 +111,9 @@ set -euo pipefail
 # GitHub App, so without this its reviews are indistinguishable from the agent
 # talking to itself and get silenced by the self-trigger guard. Matches both
 # the summary comment (`<!-- bob-ai-review {...} -->`) and inline findings
-# (`<!-- bob-ai-review-finding -->`).
-AI_REVIEW_MARKER="<!-- bob-ai-review"
+# (`<!-- bob-ai-review-finding -->`). The reviewer stamps summary comments at
+# the end, so this marker cannot be anchored to the first line.
+AI_REVIEW_MARKER_RE='^<!-- bob-ai-review(-finding| \{.*\}) -->$'
 
 # --- Parse args ---
 AUTHOR=""
@@ -493,7 +494,17 @@ has_actionable_update() {
     # PM dispatches -> push -> re-review), and it terminates the same way, via
     # the fingerprint dedup below plus the per-PR attempt counters. It is
     # bounded by the PR being fixed or merged, not by the reviewer's identity.
-    if [ "$last_actor" = "$AUTHOR" ] && ! echo "$last_body" | grep -q "$AI_REVIEW_MARKER"; then
+    #
+    # The marker must occupy a complete line. A normal reply that merely quotes
+    # or inline-copies the marker remains self-chatter; the reviewer itself emits
+    # the marker as a standalone HTML-comment line. The regex accepts only the
+    # two formats emitted by the reviewer: `-finding` or a JSON state object.
+    local is_ai_review=1
+    if [ "$last_actor" = "$AUTHOR" ] && printf '%s\n' "$last_body" \
+        | grep -qE "$AI_REVIEW_MARKER_RE"; then
+        is_ai_review=0
+    fi
+    if [ "$last_actor" = "$AUTHOR" ] && [ "$is_ai_review" -ne 0 ]; then
         return 1
     fi
 
--- a/scripts/github/activity-gate.sh
+++ b/scripts/github/activity-gate.sh
@@ -511,7 +511,7 @@ has_actionable_update() {
     # Skip if the comment/review @-mentions someone else but NOT the AUTHOR.
     # This catches cases like "@greptileai review" or "@someone what do you think?"
     # where the commenter is clearly addressing someone other than the agent.
-    if [ -n "$last_body" ]; then
+    if [ "$is_ai_review" -ne 0 ] && [ -n "$last_body" ]; then
         local mentions_anyone mentions_author
         # Match GitHub-style @mentions (start of line or after whitespace).
         # Avoids false positives from emails like user@example.com.
--- a/scripts/github/activity-gate.sh
+++ b/scripts/github/activity-gate.sh
@@ -106,6 +106,14 @@
 
 set -euo pipefail
 
+# Marker stamped on every comment by Bob's self-hosted AI reviewer
+# (ErikBjare/bob#1122). It posts through Bob's user account rather than a
+# GitHub App, so without this its reviews are indistinguishable from the agent
+# talking to itself and get silenced by the self-trigger guard. Matches both
+# the summary comment (`<!-- bob-ai-review {...} -->`) and inline findings
+# (`<!-- bob-ai-review-finding -->`).
+AI_REVIEW_MARKER="<!-- bob-ai-review"
+
 # --- Parse args ---
 AUTHOR=""
 ORG=""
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
## 🤖 AI code review

### Confidence Score: 4/5

**Could merge, with caveats.** No blocking defects on this head, but 1 P2 issue worth resolving first.

### What changed

Adds an AI-review recognition mechanism to activity-gate.sh so Bob's self-hosted reviewer comments, which carry the `<!-- bob-ai-review` marker, are exempted from the self-trigger guard while genuine self-chatter from Bob remains silenced. Adds a new test file covering AI-review actionability, marker quoting behavior, and regression protection for human and Greptile comments. Also adjusts the @-mention filtering in has_actionable_update to be skipped for AI reviews.

- `scripts/github/activity-gate.sh` — Adds AI_REVIEW_MARKER_RE regex and changes has_actionable_update to skip self-trigger only if the comment doesn't carry the AI review marker; also guards @-mention filtering similarly.
- `tests/test_activity_gate_ai_review_actionable.py` — Adds a new test file with 10 tests covering AI review marker recognition, silencing of Bob's plain/quoting comments, and non-regression for human and Greptile reviews.

**1 finding(s)** — ⚠️ **1** P2

### Findings

### ⚠️ P2 medium — _PR-level_

The regex `AI_REVIEW_MARKER_RE` does not match the JSON state object format used in tests and comments. The regex is '^&lt;!-- bob-ai-review(-finding| \{.*\}) --&gt;$'. A JSON object contains spaces after colons and commas (e.g., '{"sha": "abc"}'), but BRE pattern `\{.*\}` requires the object to start with `{` immediately after the space. In the test `test_ai_review_summary_is_actionable`, the body is '3 findings\n\n&lt;!-- bob-ai-review {"sha": "abc"} --&gt;'. The regex expects `{` right after the space, but the actual character is a quote `"` (i.e., JSON starts with a brace but the pattern `.*` is greedy, so it should match; however, the literal `\{` requires a literal `{`, which is present). Actually, the regex requires `{` after the space, and the JSON string starts with `{`, so it matches. The dot in BRE matches any character including quotes. So this is not a bug. I need to re-check the test path: the marker string in the test is `<!-- bob-ai-review` followed by a space and `{"sha": "abc"}`. The regex has a space then `\{`, so it matches. So this finding is not valid.

---
<sub>Reviewed `503f0aa6f13c` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 21s · self-hosted reviewer ([bob#1122](https://github.com/ErikBjare/bob/issues/1122)). Findings are advisory — a human still decides.</sub>

scripts/github/activity-gate.sh
```diff
@@ -467,8 +475,25 @@ has_actionable_update() {
...
+    # Those comments carry a machine marker, so they are distinguishable from
+    # Bob's genuine human-directed comments — which must STILL be silenced,
+    # since those are the actual self-trigger loop this guard exists to stop.
+    # Matching the marker rather than the login is what keeps both properties.
+    #
+    # Loop safety: this is the same cycle Greptile already drives (review ->
+    # PM dispatches -> push -> re-review), and it terminates the same way, via
+    # the fingerprint dedup below plus the per-PR attempt counters. It is
+    # bounded by the PR being fixed or merged, not by the reviewer's identity.
+    if [ "$last_actor" = "$AUTHOR" ] && ! echo "$last_body" | grep -q "$AI_REVIEW_MARKER"; then
```
TimeToBuildBob: Fixed in `39ce53be`. The marker now has to occupy a complete line and match one of the reviewer's two exact shapes (`-finding` or JSON state); inline copies, blockquoted marker lines, and lookalike marker suffixes remain self-chatter. I did not anchor the summary marker to the first comment line because the reviewer emits it at the end. Added regression tests for each quote/lookalike case; 22 targeted tests pass.

scripts/github/activity-gate.sh
```diff
@@ -467,8 +475,25 @@ has_actionable_update() {
     # If no activity found, this might be a push — allow it
     [ -z "$last_actor" ] && return 0
 
-    # Skip if the last actor is the author (self-triggered update)
-    if [ "$last_actor" = "$AUTHOR" ]; then
+    # Skip if the last actor is the author (self-triggered update).
```
TimeToBuildBob, line 479: Not applying the proposed lookback. `check_pr_updates` compares the latest comment/review timestamp to the stored watermark, and the AI review itself triggers the gate when it lands. A later author status comment may be suppressed, but it does not retroactively make the already-emitted review undelivered. Scanning older marker comments would instead re-dispatch stale findings after every author comment and recreate the loop this guard prevents.

tests/test_activity_gate_ai_review_actionable.py
```diff
@@ -0,0 +1,90 @@
...
+    return "YES" in proc.stdout
+
+
+def test_ai_review_summary_is_actionable():
+    """The regression: our reviewer's findings must reach PM."""
+    assert _actionable(f'3 findings\n\n{MARKER} {{"sha": "abc"}} -->', BOT)
+
+
+def test_ai_review_inline_finding_is_actionable():
+    assert _actionable(f"{MARKER}-finding -->\nP1 — the hash ignores .state", BOT)
```
TimeToBuildBob, line 84: ⚠️ **P2** — The test suite exercises the AI-review exemption only through the .comments path; _actionable always passes an empty latestReviews list (line 42), so the function's review branch (lines 464 and 471) that selects last_body from a review comment is never executed. If the AI reviewer posts inline findings as review comments — which the PR description explicitly says it does — a regression in last_body extraction or marker matching specific to reviews would go undetected. Concretely, a bug that caused .body to be null or the marker to be dropped on the review path would silently re-introduce the self-trigger silencing this PR fixes, and the test suite would still pass.

tests/test_activity_gate_ai_review_actionable.py
```diff
@@ -0,0 +1,90 @@
...
+                "createdAt": "2026-08-08T10:00:00Z",
+                "body": body,
+            }
+        ],
+        "latestReviews": [],
+    }
+    script = f"""
+set -uo pipefail
+AUTHOR={BOT}
+AI_REVIEW_MARKER_RE='^<!-- bob-ai-review(-finding| \\{{.*\\}}) -->$'
```
TimeToBuildBob: ⚠️ **P2** — The test harness defines AI_REVIEW_MARKER_RE as a literal string on line 47 instead of extracting it from the source, unlike _extract_function which reads the actual function body. If the production regex changes — e.g., to allow a new marker format or adjust the JSON delimiters — the tests will continue using the old regex and may pass against the updated function incorrectly (if the marker format drift makes the test input unmatchable) or fail spuriously. This weakens the tests as a regression guard for the exact contract they are meant to pin. The regex should be parsed from activity-gate.sh, with the same care taken for the function body, so tests always reflect the deployed marker.

tests/test_activity_gate_ai_review_actionable.py
TimeToBuildBob: Fixed in `503f0aa6`. The test helper can now place activity in `latestReviews`, and `test_ai_review_finding_in_latest_review_is_actionable` exercises the review-body selection path directly.

tests/test_activity_gate_ai_review_actionable.py
TimeToBuildBob: Fixed in `503f0aa6`. The harness now extracts the complete `AI_REVIEW_MARKER_RE` assignment from `activity-gate.sh`, so production and test marker contracts cannot drift independently.

PR #1393, opened 2026-08-09
feat(self-merge): accept our own review as an alternative to Greptile

TimeToBuildBob:
## The problem

The self-merge gate hard-required Greptile:

```python
greptile = fetch_greptile_status(repo, number, review_data=shared_review_data)
if not greptile["has_review"]:
    result.reasons.append("Greptile review not found")
```

On any repo Greptile does not cover, that made the gate unsatisfiable. Verified live on #1382 — `Eligible: NO — Greptile review not found` — on a PR our own self-hosted reviewer had already reviewed in full. Our marker was used in this file for exactly one thing: classifying inline findings so they don't read as unresolved human threads. The verdict itself was ignored.

## What changed

The reviewer's `<!-- bob-ai-review {…} -->` summary marker now satisfies the same requirement Greptile does.

**It is an OR, never a replacement.** A repo where Greptile still runs takes the identical path it always took, and does not even pay the extra API call — the marker is only fetched when Greptile has *not* reviewed. An unresolved Greptile thread still blocks regardless of what our reviewer says.

## Why it is this strict — the risk runs the other way

Our reviewer's measured precision is 37%. That number is the wrong one to gate on, and reasoning from it would have produced the wrong design.

Precision governs **false findings**. A false finding *lowers* the score, which makes this gate **more conservative**: it blocks a mergeable PR and costs a wasted round. Bad, but bounded and self-correcting.

The dangerous direction is **recall**. A missed bug produces a clean score, and a clean score is an unreviewed merge. That is unbounded and silent.

So every condition below exists to make "clean" mean "clean *on real evidence*":

| Condition | Why |
|---|---|
| score is **5/5** | the rubric's only "nothing outstanding on this head" value. Deliberately **not** env-overridable, unlike the Greptile floor — letting `SELF_MERGE_MIN_GREPTILE_SCORE` lower this too would make the new path the weakest link |
| marker `sha` matches **current head** | a review of an earlier push says nothing about the code about to merge. 12-char prefix vs the full 40, with a minimum prefix length so a truncated marker can't match anything |
| consensus present and **undegraded** | no lost passes, no lost fan-out jobs, no clamped agreement threshold, and ≥1 job actually answered |
| reviewer did **not abstain** | `score: null` is the submodule pointer-bump case — "no verdict", never a pass |
| marker posted by the **authenticated user** | the marker is plain text in a comment body; otherwise anyone who can comment on our PR could forge a clean verdict for a gate that then merges without a human |

Any of those failing falls back to today's behaviour — never to auto-approve.

Two of those deserve expanding:

**Lost fan-out jobs is the loss that hides.** A pass counts as "answered" when *any one* of its aspect jobs returns, so a 3×3 wave that loses 6 of 9 jobs still records `answered: 3 of 3` and reads as full consensus while two thirds of the evidence is missing. Checking passes alone would have missed the common case.

**Zero answered is the sharp edge.** If every call fails there are no findings, and no findings computes to a clean 5/5 off an empty result. That is precisely the recall failure, arriving disguised as the best possible score.

**A missing consensus record does not qualify.** The hourly sweep runs `--passes 1`, which writes no `consensus` key at all — this is the marker shape actually live on #1382 today. The reviewer's own contract says to read a missing record as "unknown, never as full consensus", and this follows it. To get a qualifying review: `ai-review.py OWNER/REPO N --force`.

## Better refusals

Refusals now say *why*, instead of leaving the operator to guess whether the reviewer ran at all. Live against #1382 with this branch:

```
Reasons:
- Greptile review not found; AI review score 2/5 below 5/5
- Touches sensitive/security/infra paths
```

That PR genuinely carries a 2/5 review, so it correctly stays blocked — and now says so.

## Tests

29 new tests in `tests/test_self_merge_ai_review_gate.py`, weighted toward refusing a 5/5 reached on thin evidence:

- our clean/current/full-consensus review satisfies the gate (the #1382 regression), end-to-end through `evaluate_pr`
- degraded consensus does not — separately for lost passes, lost fan-out jobs, clamped agreement, and zero answered
- missing consensus record does not
- stale SHA does not
- abstain does not; scores 1–4 do not
- a marker from another account is ignored; unknown identity trusts nothing
- **Greptile-only path unchanged**: still eligible, and asserts the marker is never fetched; an unresolved Greptile thread still blocks despite a clean AI review
- marker-regex guards: `-finding` and `-fp {…}` must not match `bob-ai-review {`; the nested-object marker must parse whole

Existing suites pass: `test_self_merge_check.py` + `test_self_merge_ai_review_threads.py` = 123 passed. Full suite: 658 passed, 1 pre-existing failure (`test_linear_oauth_state.py`, a Flask version incompatibility — confirmed failing identically on clean `origin/master`).

## Rollback

`SELF_MERGE_ACCEPT_AI_REVIEW=0` restores the Greptile-only gate.

## Note for the reviewer

`self-merge-check.py` is in `SENSITIVE_PATH_PREFIXES`, so this PR cannot self-merge under its own new rule. That is intended — it needs a human.

### DIFF
--- a/scripts/github/self-merge-check.py
+++ b/scripts/github/self-merge-check.py
@@ -110,7 +110,7 @@ from datetime import datetime, timezone
 from fnmatch import fnmatchcase
 from functools import cache
 from pathlib import Path
-from typing import Any, cast
+from typing import Any, Iterator, cast
 
 MAX_GRAPHQL_PAGE_SIZE = 100
 DEFAULT_MIN_GREPTILE_SCORE = 5
@@ -792,9 +792,29 @@ def _is_ai_review_thread(comments: list[dict[str, Any]]) -> bool:
 # Summary marker upserted by the self-hosted reviewer onto one issue comment per
 # PR: `<!-- bob-ai-review {json} -->`. The space before `{` is load-bearing — it
 # is what stops this pattern from also matching `<!-- bob-ai-review-finding -->`
-# and the per-finding `<!-- bob-ai-review-fp {...} -->`. Same regex the reviewer's
-# own consumers (ai-review.py, ai-review-sweep.py) use; keep them in step.
-_AI_REVIEW_SUMMARY_RE = re.compile(r"<!-- bob-ai-review (\{.*?\}) -->", re.DOTALL)
+# and the per-finding `<!-- bob-ai-review-fp {...} -->`.
+_AI_REVIEW_MARKER_PREFIX = "<!-- bob-ai-review "
+_JSON_DECODER = json.JSONDecoder()
+
+
+def _iter_ai_review_markers(body: str) -> Iterator[dict[str, Any]]:
+    """Yield complete JSON objects from self-hosted-review summary markers.
+
+    Regex cannot safely delimit JSON because real markers contain nested objects.
+    ``raw_decode`` consumes exactly one complete JSON value, after which we verify
+    the marker suffix so arbitrary prose containing the prefix is ignored.
+    """
+    offset = 0
+    while (start := body.find(_AI_REVIEW_MARKER_PREFIX, offset)) != -1:
+        json_start = start + len(_AI_REVIEW_MARKER_PREFIX)
+        try:
+            parsed, json_end = _JSON_DECODER.raw_decode(body, json_start)
+        except json.JSONDecodeError:
+            offset = json_start
+            continue
+        if body.startswith(" -->", json_end) and isinstance(parsed, dict):
+            yield parsed
+        offset = max(json_end, json_start + 1)
 
 
 def _ai_review_enabled() -> bool:
@@ -903,13 +923,8 @@ def _fetch_ai_review_marker(
         # Last match wins within a body, and the last qualifying comment wins
         # overall: the reviewer upserts a single comment, but a repo with legacy
         # append-style comments should still be read at its most recent verdict.
-        for match in _AI_REVIEW_SUMMARY_RE.finditer(comment.get("body") or ""):
-            try:
-                parsed = json.loads(match.group(1))
-            except json.JSONDecodeError:
-                continue
-            if isinstance(parsed, dict):
-                latest = parsed
+        for parsed in _iter_ai_review_markers(comment.get("body") or ""):
+            latest = parsed
     return latest
 
 
--- a/scripts/github/self-merge-check.py
+++ b/scripts/github/self-merge-check.py
@@ -6,7 +6,9 @@ autonomous AI agent can safely merge its own PRs without human review.
 
 Policy summary:
 - CI must be green (or skipped)
-- Greptile review must be present and have no unresolved threads
+- A machine review must be present and clean: EITHER a Greptile review with no
+  unresolved threads, OR a high-confidence review from the agent's own
+  self-hosted AI reviewer (see "Accepting our own review" below)
 - PR must be authored by the authenticated user
 - Changed files must fall into a low-risk category (tests, docs, lessons, internal
   tooling, task metadata)
@@ -20,6 +22,43 @@ To allow cross-repo merges, either:
   category/CI/Greptile checks still apply.
 - Or clear WORKSPACE_REPO entirely (set it to empty string) to disable the
   cross-repo restriction (not recommended; widens the merge surface).
+
+Accepting our own review
+------------------------
+Greptile does not run on every repo, and when it is absent this gate used to be
+unsatisfiable — "Greptile review not found" blocked PRs that our own reviewer had
+already reviewed in full (observed on gptme/gptme-contrib#1382). The reviewer's
+verdict is now accepted as an *alternative* to Greptile. It is strictly an OR:
+when Greptile has reviewed, nothing about this gate's behaviour changes, and the
+marker is not even fetched.
+
+The strictness here is set by which direction the errors run. The reviewer's
+measured precision is 37%, but precision governs *false findings*, and a false
+finding lowers the score, which makes this gate more conservative — it costs a
+wasted round, not a bad merge. The dangerous direction is **recall**: a missed
+bug produces a clean score and an unreviewed merge. So every condition below
+exists to make "clean" mean "clean on real evidence", and anything short of that
+falls back to requiring Greptile rather than approving:
+
+- score is 5/5 — the rubric's only "nothing outstanding on this head" value.
+  Deliberately not env-overridable, unlike the Greptile floor.
+- the review is at the current head SHA (a review of an older push says nothing
+  about the code about to merge)
+- consensus is present and undegraded — no lost passes, no lost fan-out jobs, no
+  clamped agreement threshold. A degraded run reached its score on less evidence
+  than it asked for, which is exactly the recall failure. A **missing** consensus
+  record means no consensus stage ran at all (the hourly sweep's single-pass
+  reviews, or the agent engine); the reviewer documents that as "unknown, never
+  read as full consensus", so it does not satisfy this gate either. Get a
+  qualifying review with a consensus re-review: `ai-review.py OWNER/REPO N --force`.
+- the reviewer did not abstain (a null score is "no verdict" — the submodule
+  pointer-bump case — and must never read as a pass)
+- the marker comment was posted by the authenticated user. Anyone can write the
+  marker string into a comment on our PR; only our own account posts a real one.
+
+Unresolved findings do not need a separate check: 5/5 is defined as zero findings
+on the reviewed head, and the reviewer deletes its own superseded inline comments
+on each re-review.
 
 Usage:
     python3 scripts/github/self-merge-check.py <pr-url>
@@ -46,6 +85,10 @@ Environment:
                     "owner/repo:path-glob[,owner/repo:path-glob...]".
                     Uses repo-relative segment globs (* stays within one
                     directory segment; ** matches zero or more directories).
+    SELF_MERGE_ACCEPT_AI_REVIEW
+                    Set to 0/false/no/off to disable the self-hosted-AI-review
+                    alternative entirely, restoring the Greptile-only gate.
+                    Kill switch for the "Accepting our own review" path above.
 
 Exit codes:
     0   Eligible for self-merge
@@ -71,6 +114,16 @@ from typing import Any, cast
 
 MAX_GRAPHQL_PAGE_SIZE = 100
 DEFAULT_MIN_GREPTILE_SCORE = 5
+
+# The one score on the reviewer's 0-5 rubric that means "nothing outstanding on
+# the commit reviewed". Not env-overridable, deliberately: SELF_MERGE_MIN_GREPTILE_SCORE
+# can be lowered, and letting the same knob lower *this* floor would turn the
+# alternative path into the weakest link in the gate.
+AI_REVIEW_CLEAN_SCORE = 5
+# Shortest marker SHA prefix accepted when matching against the PR head. The
+# reviewer writes 12 hex chars; this only guards against a truncated/garbage
+# marker whose 1-2 char "sha" would prefix-match almost any head.
+MIN_SHA_PREFIX_LEN = 7
 
 DOC_EXTENSIONS = {".md", ".rst", ".txt", ".adoc"}
 SPEC_LIKE_DOCS = {
@@ -167,6 +220,7 @@ BOT_CONFIG_FILES = {
 }
 BOT_CONFIG_PREFIXES = (".github/",)
 SELF_MERGE_ALLOWED_PATHS_ENV = "SELF_MERGE_ALLOWED_PATHS"
+SELF_MERGE_ACCEPT_AI_REVIEW_ENV = "SELF_MERGE_ACCEPT_AI_REVIEW"
 GATE_HELPER_ENV = "GH_RATE_LIMIT_HELPER"
 GATE_HELPER_ENV_LEGACY = "BOB_GH_RATE_LIMIT_HELPER"
 FORCE_SELF_MERGE_CHECK_ENV = "FORCE_SELF_MERGE_CHECK"
@@ -735,6 +789,182 @@ def _is_ai_review_thread(comments: list[dict[str, Any]]) -> bool:
     return _AI_REVIEW_FINDING_MARKER in (comments[0].get("body") or "")
 
 
+# Summary marker upserted by the self-hosted reviewer onto one issue comment per
+# PR: `<!-- bob-ai-review {json} -->`. The space before `{` is load-bearing — it
+# is what stops this pattern from also matching `<!-- bob-ai-review-finding -->`
+# and the per-finding `<!-- bob-ai-review-fp {...} -->`. Same regex the reviewer's
+# own consumers (ai-review.py, ai-review-sweep.py) use; keep them in step.
+_AI_REVIEW_SUMMARY_RE = re.compile(r"<!-- bob-ai-review (\{.*?\}) -->", re.DOTALL)
+
+
+def _ai_review_enabled() -> bool:
+    """Whether our own review may satisfy the machine-review requirement."""
+    raw = os.environ.get(SELF_MERGE_ACCEPT_AI_REVIEW_ENV, "").strip().lower()
+    return raw not in ("0", "false", "no", "off")
+
+
+def _sha_matches_head(marker_sha: str, head_sha: str) -> bool:
+    """Whether a marker's (possibly abbreviated) SHA names the PR's current head.
+
+    The reviewer records 12 hex chars while the PR carries the full 40, so this
+    is prefix matching -- but only above ``MIN_SHA_PREFIX_LEN``, so a truncated
+    or garbage marker cannot prefix-match its way onto an arbitrary head.
+    """
+    marker_sha = (marker_sha or "").strip().lower()
+    head_sha = (head_sha or "").strip().lower()
+    if len(marker_sha) < MIN_SHA_PREFIX_LEN or len(head_sha) < MIN_SHA_PREFIX_LEN:
+        return False
+    return head_sha.startswith(marker_sha) or marker_sha.startswith(head_sha)
+
+
+def _consensus_shortfall(consensus: Any) -> str | None:
+    """Why a review's consensus is not trustworthy evidence, or None if it is.
+
+    Mirrors the reviewer's own ``consensus_degraded`` (lost passes OR lost
+    fan-out jobs) and additionally rejects a *clamped* agreement threshold and a
+    run where nothing at all answered:
+
+    - lost passes / lost jobs: the score was reached on less evidence than was
+      requested. Job loss is the one that matters in practice -- a pass counts as
+      "answered" when any single one of its aspect jobs returns, so a wave can
+      lose two thirds of its jobs and still report every pass answered.
+    - clamped agreement: the filter that ran is weaker than the filter asked for.
+    - zero answered: every call failed, so there are no findings, so the score
+      computes to a clean 5/5 off an empty result. That is the exact recall
+      failure this whole function exists to catch.
+    """
+    if not isinstance(consensus, dict):
+        # Absent or malformed. The reviewer's contract is explicit that a missing
+        # record means "no consensus stage ran" (single-pass sweep review, or the
+        # agent engine) and must be read as unknown -- never as full consensus.
+        return "no consensus record (single-pass or agent review)"
+
+    def _int(key: str) -> int | None:
+        value = consensus.get(key)
+        return value if isinstance(value, int) else None
+
+    requested, answered = _int("requested"), _int("answered")
+    jobs_requested, jobs_answered = _int("jobs_requested"), _int("jobs_answered")
+    if requested is None or answered is None:
+        return "consensus record is missing pass counts"
+    if jobs_requested is None or jobs_answered is None:
+        return "consensus record is missing job counts"
+
+    if answered < 1 or jobs_answered < 1:
+        return "no consensus pass answered"
+    if requested > 1 and answered < requested:
+        return f"consensus degraded ({answered}/{requested} passes answered)"
+    if jobs_requested > 1 and jobs_answered < jobs_requested:
+        return f"consensus degraded ({jobs_answered}/{jobs_requested} jobs answered)"
+
+    applied, asked = _int("min_agreement"), _int("min_agreement_requested")
+    if applied is not None and asked is not None and applied < asked:
+        return f"consensus agreement clamped to {applied} (requested {asked})"
+    return None
+
+
+def _fetch_ai_review_marker(
+    repo: str, pr_number: int, *, expected_author: str
+) -> dict[str, Any] | None:
+    """Latest self-hosted-review summary marker on the PR, parsed.
+
+    Only comments authored by *expected_author* are considered: the marker is
+    plain text in a comment body, so any account that can comment on the PR could
+    otherwise forge a clean verdict for a gate that then merges without a human.
+    An empty *expected_author* means the identity could not be established, and
+    the marker is not trusted at all.
+    """
+    if not expected_author:
+        return None
+    raw = run_gh(
+        [
+            "api",
+            f"repos/{repo}/issues/{pr_number}/comments",
+            "--paginate",
+            "--jq",
+            ".[] | {author: .user.login, body: .body}",
+        ],
+        timeout=30,
+    )
+    if not raw:
+        return None
+
+    latest: dict[str, Any] | None = None
+    for line in raw.splitlines():
+        line = line.strip()
+        if not line:
+            continue
+        try:
+            comment = json.loads(line)
+        except json.JSONDecodeError:
+            continue
+        if not isinstance(comment, dict) or comment.get("author") != expected_author:
+            continue
+        # Last match wins within a body, and the last qualifying comment wins
+        # overall: the reviewer upserts a single comment, but a repo with legacy
+        # append-style comments should still be read at its most recent verdict.
+        for match in _AI_REVIEW_SUMMARY_RE.finditer(comment.get("body") or ""):
+            try:
+                parsed = json.loads(match.group(1))
+            except json.JSONDecodeError:
+                continue
+            if isinstance(parsed, dict):
+                latest = parsed
+    return latest
+
+
+def fetch_ai_review_status(
+    repo: str, pr_number: int, *, head_sha: str, expected_author: str
+) -> dict[str, Any]:
+    """Whether our own reviewer's verdict is strong enough to stand in for Greptile.
+
+    Returns ``{"accepted": bool, "detail": str | None}``. ``detail`` explains a
+    refusal so the gate's "Greptile review not found" reason can say *why* the
+    fallback did not apply, rather than leaving the operator to guess whether the
+    reviewer ran at all.
+
+    Every refusal path falls back to today's Greptile-only behaviour. Nothing
+    here can make a PR *more* mergeable than the rest of the gate allows -- CI,
+    category, human threads and author identity are all still evaluated.
+    """
+    if not _ai_review_enabled():
+        return {"accepted": False, "detail": None}
+
+    marker = _fetch_ai_review_marker(repo, pr_number, expected_author=expected_author)
+    if marker is None:
+        return {"accepted": False, "detail": None}
+
+    if not _sha_matches_head(str(marker.get("sha") or ""), head_sha):
+        return {
+            "accepted": False,
+            "detail": (
+                f"AI review is stale (reviewed {marker.get('sha') or 'unknown'}, "
+                f"head is {head_sha[:12]})"
+            ),
+        }
+
+    score = marker.get("score")
+    if score is None:
+        return {"accepted": False, "detail": "AI reviewer abstained (no score)"}
+    if not isinstance(score, int) or score < AI_REVIEW_CLEAN_SCORE:
+        return {
+            "accepted": False,
+            "detail": f"AI review score {score}/5 below {AI_REVIEW_CLEAN_SCORE}/5",
+        }
+
+    shortfall = _consensus_shortfall(marker.get("consensus"))
+    if shortfall:
+        return {"accepted": False, "detail": f"AI review {shortfall}"}
+
+    return {
+        "accepted": True,
+        "detail": (
+            f"AI review {score}/5 at current head {str(marker.get('sha'))[:12]}, "
+            "full consensus"
+        ),
+    }
+
+
 def fetch_unresolved_human_threads(
     repo: str,
     pr_number: int,
@@ -1250,7 +1480,26 @@ def evaluate_pr(
 
     greptile = fetch_greptile_status(repo, number, review_data=shared_review_data)
     if not greptile["has_review"]:
-        result.reasons.append("Greptile review not found")
+        # OR, not replacement: only reached when Greptile has not reviewed, so a
+        # repo where Greptile still runs takes the identical path it always has
+        # and does not even pay the extra API call. See "Accepting our own
+        # review" in the module docstring for why the bar is set where it is.
+        ai_review = fetch_ai_review_status(
+            repo,
+            number,
+            head_sha=result.head_sha,
+            expected_author=current_user,
+        )
+        if ai_review["accepted"]:
+            result.warnings.append(
+                f"Greptile review not found; satisfied by self-hosted "
+                f"AI review ({ai_review['detail']})"
+            )
+        else:
+            reason = "Greptile review not found"
+            if ai_review["detail"]:
+                reason += f"; {ai_review['detail']}"
+            result.reasons.append(reason)
     elif greptile["unresolved"] > 0:
         result.reasons.append(
             f"Greptile has {greptile['unresolved']} unresolved review thread(s)"
--- a/scripts/github/self-merge-check.py
+++ b/scripts/github/self-merge-check.py
@@ -861,7 +861,8 @@ def _consensus_shortfall(consensus: Any) -> str | None:
 
     def _int(key: str) -> int | None:
         value = consensus.get(key)
-        return value if isinstance(value, int) else None
+        # bool is an int subclass in Python, but it is never a valid count.
+        return value if type(value) is int else None
 
     requested, answered = _int("requested"), _int("answered")
     jobs_requested, jobs_answered = _int("jobs_requested"), _int("jobs_answered")
@@ -870,15 +871,22 @@ def _consensus_shortfall(consensus: Any) -> str | None:
     if jobs_requested is None or jobs_answered is None:
         return "consensus record is missing job counts"
 
+    # One pass is the sweep's non-consensus mode even if a marker explicitly
+    # serializes 1/1 counts. Qualifying evidence must contain multiple independent
+    # passes, not merely a present consensus object.
+    if requested < 2:
+        return "consensus requires at least 2 requested passes"
     if answered < 1 or jobs_answered < 1:
         return "no consensus pass answered"
-    if requested > 1 and answered < requested:
+    if answered < requested:
         return f"consensus degraded ({answered}/{requested} passes answered)"
-    if jobs_requested > 1 and jobs_answered < jobs_requested:
+    if jobs_answered < jobs_requested:
         return f"consensus degraded ({jobs_answered}/{jobs_requested} jobs answered)"
 
     applied, asked = _int("min_agreement"), _int("min_agreement_requested")
-    if applied is not None and asked is not None and applied < asked:
+    if applied is None or asked is None:
+        return "consensus record is missing agreement thresholds"
+    if applied < asked:
         return f"consensus agreement clamped to {applied} (requested {asked})"
     return None
 
--- a/scripts/github/self-merge-check.py
+++ b/scripts/github/self-merge-check.py
@@ -889,6 +889,20 @@ def _consensus_shortfall(consensus: Any) -> str | None:
     applied, asked = _int("min_agreement"), _int("min_agreement_requested")
     if applied is None or asked is None:
         return "consensus record is missing agreement thresholds"
+    # A threshold below 1 is not a weaker consensus, it is *no* consensus: a
+    # finding surviving on zero agreeing passes is exactly what a single pass
+    # already produces. Checking only `applied < asked` misses it, because a
+    # record with min_agreement 0 AND min_agreement_requested 0 is unclamped and
+    # would read as full consensus — while `answered >= 1` above is satisfied by
+    # one pass. That combination hands the gate a single-pass review dressed as
+    # consensus, gutting the recall guard this whole path exists to enforce.
+    # Negatives are rejected by the same test rather than a separate one; both
+    # mean "no agreement was required".
+    if applied < 1 or asked < 1:
+        return (
+            f"consensus agreement threshold below 1 "
+            f"(applied {applied}, requested {asked})"
+        )
     if applied < asked:
         return f"consensus agreement clamped to {applied} (requested {asked})"
     return None
@@ -972,7 +986,22 @@ def fetch_ai_review_status(
     score = marker.get("score")
     if score is None:
         return {"accepted": False, "detail": "AI reviewer abstained (no score)"}
-    if not isinstance(score, int) or score < AI_REVIEW_CLEAN_SCORE:
+    # Type and threshold are separate refusals so the detail names the actual
+    # defect. Folding them together reported a malformed marker (score "5" as a
+    # string, or 5.0) as `AI review score 5/5 below 5/5` — a sentence that is
+    # false on its face and sends whoever reads it hunting for a scoring bug
+    # instead of the marker corruption that caused it. Both still fail closed.
+    # `type(...) is not int` rather than `isinstance`, matching `_int` above:
+    # bool is an int subclass, and `True` is not a review score.
+    if type(score) is not int:
+        return {
+            "accepted": False,
+            "detail": (
+                f"AI review score is not an integer "
+                f"({type(score).__name__}: {score!r})"
+            ),
+        }
+    if score < AI_REVIEW_CLEAN_SCORE:
         return {
             "accepted": False,
             "detail": f"AI review score {score}/5 below {AI_REVIEW_CLEAN_SCORE}/5",
--- a/scripts/github/self-merge-check.py
+++ b/scripts/github/self-merge-check.py
@@ -824,17 +824,13 @@ def _ai_review_enabled() -> bool:
 
 
 def _sha_matches_head(marker_sha: str, head_sha: str) -> bool:
-    """Whether a marker's (possibly abbreviated) SHA names the PR's current head.
-
-    The reviewer records 12 hex chars while the PR carries the full 40, so this
-    is prefix matching -- but only above ``MIN_SHA_PREFIX_LEN``, so a truncated
-    or garbage marker cannot prefix-match its way onto an arbitrary head.
-    """
+    """Whether a marker's abbreviated SHA names the PR's current head."""
     marker_sha = (marker_sha or "").strip().lower()
     head_sha = (head_sha or "").strip().lower()
     if len(marker_sha) < MIN_SHA_PREFIX_LEN or len(head_sha) < MIN_SHA_PREFIX_LEN:
         return False
-    return head_sha.startswith(marker_sha) or marker_sha.startswith(head_sha)
+    # The marker may abbreviate the authoritative head, never the reverse.
+    return head_sha.startswith(marker_sha)
 
 
 def _consensus_shortfall(consensus: Any) -> str | None:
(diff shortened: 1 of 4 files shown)


TimeToBuildBob:
## 🤖 AI code review

This PR lets the self-merge gate accept the agent's own AI review as an alternative to Greptile when Greptile has not reviewed. It adds summary-marker parsing, strict acceptance criteria (score exactly 5, current-head SHA prefix match, undegraded consensus with at least 2 passes and no clamped thresholds), a kill-switch env var, and changes the Greptile comment fallback to distinguish a failed fetch from an empty one via `run_gh_checked`. Extensive tests cover the gate logic and the untouched Greptile path.

### Confidence Score: 2/5 — Not ready to merge

2 P1 defects on this head should be fixed before merge.

**3 findings** · ❌ **2** P1 · ⚠️ **1** P2 · 🔒 **2** security — 🛑 **2 blocking**, worth resolving before merge.

**❌ P1 high · 🔒 security** — `scripts/github/self-merge-check.py:942`

In `_consensus_shortfall`, after validating pass/job counts, the code checks `min_agreement` and `min_agreement_requested` only for being below 1 or clamped (`applied < asked`). It never verifies that `min_agreement <= answered` or that `min_agreement_requested <= requested`. A consensus record with `requested=2, answered=2, min_agreement=3, min_agreement_requested=3` passes every existing check: answered equals requested, jobs match, thresholds are ≥1, and `applied < asked` is false. Such a configuration is impossible—it requires 3 agreeing passes when only 2 answered—and typically indicates a misconfigured or corrupted consensus run, which can suppress all findings (agreement is unreachable) and yield a clean 5/5. The gate then accepts this as full consensus, enabling an unreviewed merge. The fix is to reject when `min_agreement > answered` or `min_agreement_requested > requested`.

```suggestion
Add after the existing threshold checks: `if applied > answered: return "consensus agreement threshold exceeds answered passes"` and `if asked > requested: return "consensus requested agreement exceeds requested passes"`.
```

<sub>**How this was verified:** Read the full _consensus_shortfall function; it checks lower bounds and clamped inequality but no upper-bound consistency with answered/requested. The test suite does not include a case where min_agreement exceeds answered.</sub>

**⚠️ P2 medium** — `scripts/github/self-merge-check.py:960`

`_fetch_ai_review_marker` uses `run_gh` rather than the new `run_gh_checked` helper. `run_gh` returns an empty string both when the `gh api` call fails (timeout, OSError, non-zero exit) and when it genuinely returns no comment lines. In the failure case, `_fetch_ai_review_marker` returns `None`, and `fetch_ai_review_status` reports `accepted=False` with `detail=None`. evaluate_pr then appends only "Greptile review not found" with no diagnostic hint. This is fail-closed (a transient outage blocks the merge), so it is not a security risk, but it silently converts a network failure into a definitive "no AI review", which can repeatedly block a legitimate self-merge and give the operator no clue that the comment fetch itself failed. The PR already introduced `run_gh_checked` specifically to distinguish these cases for the Greptile fallback; using it here would produce a more accurate reason and avoid misleading operators.

```suggestion
Replace `run_gh` with `run_gh_checked` and treat a `None` result as "unknown" (e.g., return a sentinel so evaluate_pr can report a comment-fetch failure instead of "Greptile review not found").
```

<sub>**How this was verified:** Compared the two helpers: run_gh collapses failure and empty output, while run_gh_checked returns None only on failure. The Greptile comment fallback uses run_gh_checked; the AI marker fetch does not.</sub>

</details>

### Comments outside the diff

**❌ P1 high · 🔒 security** — [`scripts/github/self-merge-check.py:606`](https://github.com/gptme/gptme-contrib/blob/453ef20ad68f97868191f6a0617b739a0c29c6c1/scripts/github/self-merge-check.py#L606)

`_fetch_greptile_review_data` fetches reviews only on the first GraphQL page with `reviews(last:50)`, and does not paginate the reviews list. If a PR accumulates more than 50 reviews (e.g., from Dependabot, codecov, or other bots) after Greptile's review, that Greptile review is not in the returned `reviews` array. `fetch_greptile_status` then sees no greptile reviews and falls back to issue comments; if Greptile posted a formal review without an issue comment, the fallback finds nothing and returns `has_review: False`. evaluate_pr then considers Greptile absent and, if a clean AI-review marker is present, accepts the merge. This lets a PR with actual unresolved Greptile feedback bypass that review entirely via the new AI fallback. The gate should only use the AI fallback when it has positively confirmed there is no Greptile review, not when the review list may be incomplete.

```suggestion
Either paginate the reviews fetch (e.g., loop over review pages) or, when more than 50 reviews exist and the first page contains no Greptile review, treat the state as unknown and block the AI fallback.
```

<sub>**How this was verified:** Traced the flow: reviews are only fetched on the first page; no pagination for reviews is implemented. A PR with &gt;50 reviews will omit older Greptile reviews, making the fallback conclude absence and allowing the AI path to accept.</sub>

- `scripts/github/README.md` — Documents the new self-hosted AI-review alternative and the SELF_MERGE_ACCEPT_AI_REVIEW kill switch.
- `scripts/github/self-merge-check.py` — Adds AI-review marker parsing, consensus validation, and the fallback acceptance path; switches the Greptile comment fallback to run_gh_checked.
- `tests/test_self_merge_ai_review_gate.py` — New test file with unit and end-to-end tests for the AI-review gate acceptance and refusal paths.
- `tests/test_self_merge_check.py` — Updates the Greptile comment-fallback test to expect run_gh_checked and assert the unknown flag behavior.

</details>

<sub>Reviewed `453ef20ad68f` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 256s · advisory, a human still decides · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

TimeToBuildBob:
@greptileai review

scripts/github/self-merge-check.py
```diff
@@ -735,6 +789,182 @@ def _is_ai_review_thread(comments: list[dict[str, Any]]) -> bool:
     return _AI_REVIEW_FINDING_MARKER in (comments[0].get("body") or "")
 
 
+# Summary marker upserted by the self-hosted reviewer onto one issue comment per
+# PR: `<!-- bob-ai-review {json} -->`. The space before `{` is load-bearing — it
+# is what stops this pattern from also matching `<!-- bob-ai-review-finding -->`
+# and the per-finding `<!-- bob-ai-review-fp {...} -->`. Same regex the reviewer's
+# own consumers (ai-review.py, ai-review-sweep.py) use; keep them in step.
+_AI_REVIEW_SUMMARY_RE = re.compile(r"<!-- bob-ai-review (\{.*?\}) -->", re.DOTALL)
```
TimeToBuildBob: Fixed in `766430f1`. The summary marker now uses `JSONDecoder.raw_decode` instead of a non-greedy regex, so nested `consensus`/`history` objects parse completely. I also added coverage for nested JSON, multiple markers, and recovery after a malformed marker.

Verified locally: `30 passed` in `tests/test_self_merge_ai_review_gate.py`; Ruff check and format pass. Fresh CI is running on the pushed head.

scripts/github/self-merge-check.py
```diff
@@ -1250,7 +1495,26 @@ def evaluate_pr(
 
     greptile = fetch_greptile_status(repo, number, review_data=shared_review_data)
     if not greptile["has_review"]:
```
TimeToBuildBob, line 1578: ❌ **P1** — When `_fetch_greptile_review_data` fails (GraphQL timeout, rate limit, transient API error) it returns None, and `fetch_greptile_status` maps None to `{'has_review': False, ...}` at line 700. The new branch at line 1497 treats that case exactly like 'Greptile never reviewed' and calls `fetch_ai_review_status`, so a clean 5/5 self-review marker at head substitutes for Greptile even though Greptile may actually have reviewed the PR and flagged unresolved findings that the failed fetch never surfaced. Before this PR the failure was fail-closed ('Greptile review not found' blocked the merge); now it can fail open and merge a PR against Greptile's real findings — violating the 'an unresolved Greptile thread still blocks regardless of what our reviewer says' guarantee in the PR description. The evaluate fixture in the tests stubs `fetch_greptile_status` directly, so the fetch-failure-to-AI-substitution path is never exercised.

scripts/github/self-merge-check.py
```diff
@@ -735,6 +789,197 @@ def _is_ai_review_thread(comments: list[dict[str, Any]]) -> bool:
...
+            "detail": (
+                f"AI review is stale (reviewed {marker.get('sha') or 'unknown'}, "
+                f"head is {head_sha[:12]})"
+            ),
+        }
+
+    score = marker.get("score")
+    if score is None:
+        return {"accepted": False, "detail": "AI reviewer abstained (no score)"}
+    if not isinstance(score, int) or score < AI_REVIEW_CLEAN_SCORE:
```
TimeToBuildBob: ⚠️ **P2** — The clean-score check at line 964 is `not isinstance(score, int) or score < AI_REVIEW_CLEAN_SCORE`, which rejects only values below 5; any integer greater than or equal to 5 is accepted as a clean verdict and the detail string prints e.g. 'AI review 100/5 at current head …'. The documented invariant is that only 5/5 means 'nothing outstanding on this head' (AI_REVIEW_CLEAN_SCORE = 5, docstring lines 118-121), and the PR description stresses that the score is 'deliberately not env-overridable' because any loosening turns this path into the weakest link. A marker carrying score 6 or any larger integer — produced by a reviewer regression or a hand-crafted marker — would read as a clean pass and merge without human review. The comparison should require exact equality with AI_REVIEW_CLEAN_SCORE.

scripts/github/self-merge-check.py
```diff
@@ -735,6 +789,197 @@ def _is_ai_review_thread(comments: list[dict[str, Any]]) -> bool:
...
+    - clamped agreement: the filter that ran is weaker than the filter asked for.
+    - zero answered: every call failed, so there are no findings, so the score
+      computes to a clean 5/5 off an empty result. That is the exact recall
+      failure this whole function exists to catch.
+    """
+    if not isinstance(consensus, dict):
+        # Absent or malformed. The reviewer's contract is explicit that a missing
+        # record means "no consensus stage ran" (single-pass sweep review, or the
+        # agent engine) and must be read as unknown -- never as full consensus.
+        return "no consensus record (single-pass or agent review)"
```
TimeToBuildBob, line 892: ⚠️ **P2** — The gate rejects any marker lacking a consensus record, but that is the marker shape the reviewer's own hourly sweep produces (single-pass `--passes 1`, no consensus key) — the shape the test docstring says was observed live on the motivating PR contrib#1382. `_consensus_shortfall` returns 'no consensus record (single-pass or agent review)' for it, so on a Greptile-less repo served only by the sweep, the gate still lands on 'Greptile review not found': the exact unsatisfiable condition the PR sets out to fix. The end-to-end accept test only exercises the full-consensus marker shape, which the routine deployment never writes, so the regression suite passes while the observed deployment stays blocked. Consequence: self-merge on Greptile-less repos still requires a manual `ai-review.py OWNER/REPO N --force` consensus re-review; the feature does not unblock the case that motivated it without human intervention.

tests/test_self_merge_ai_review_gate.py
```diff
@@ -0,0 +1,392 @@
...
+
+        def fake_run_gh(args: list[str], timeout: int = 30) -> str:
+            calls.append(args)
+            return "\n".join(comments)
+
+        monkeypatch.setattr(smc, "run_gh", fake_run_gh)
+        monkeypatch.setattr(smc, "fetch_pr", lambda repo, number: _pr_payload())
+        monkeypatch.setattr(smc, "get_gh_user", lambda: AUTHOR)
+        monkeypatch.setattr(smc, "merge_permission", lambda repo: True)
+        monkeypatch.setattr(smc, "_fetch_greptile_review_data", lambda r, n: ([], []))
```
TimeToBuildBob, line 422: ⚠️ **P2** — No test exercises the fall-through where the Greptile fetch itself fails. The `evaluate` fixture stubs `_fetch_greptile_review_data` to `([], [])` (a successful fetch returning no reviews), so `shared_review_data` is never None in the suite, and `fetch_greptile_status` is monkeypatched directly to return NO_GREPTILE. That masks the API-failure path where `fetch_greptile_status` derives has_review=False from a None review_data and the new AI-review approval then passes the gate — exactly the fail-open case. The tests therefore give false confidence that 'no Greptile' is the only route into `fetch_ai_review_status`, whereas a rate-limited or errored GraphQL call reaches it too. Consequence: the suite is green while the weakened-gate behavior ships uncaught.

scripts/github/self-merge-check.py
```diff
@@ -735,6 +789,197 @@ def _is_ai_review_thread(comments: list[dict[str, Any]]) -> bool:
...
+
+    The reviewer records 12 hex chars while the PR carries the full 40, so this
+    is prefix matching -- but only above ``MIN_SHA_PREFIX_LEN``, so a truncated
+    or garbage marker cannot prefix-match its way onto an arbitrary head.
+    """
+    marker_sha = (marker_sha or "").strip().lower()
+    head_sha = (head_sha or "").strip().lower()
+    if len(marker_sha) < MIN_SHA_PREFIX_LEN or len(head_sha) < MIN_SHA_PREFIX_LEN:
+        return False
+    return head_sha.startswith(marker_sha) or marker_sha.startswith(head_sha)
```
TimeToBuildBob: 🛑 **P0** — _sha_matches_head performs bidirectional prefix matching, so a marker whose sha is a prefix of the current head (the intended case, 12-chars vs 40) is accepted, but so is a head that is a prefix of the marker — and more importantly, the function returns true when either string is a prefix of the other. The marker is the trustworthy side; the head is the ground truth. Because the marker's sha is untrusted input that the reviewer never intended to be longer than the head, this does not let an attacker shorten the head. However, the real defect is that a marker with a sha that is only a 7-char prefix of the head passes even when the marker was written by a reviewer that recorded a *truncated* 7-char sha — which is exactly what MIN_SHA_PREFIX_LEN was meant to prevent. The comment in the code says 'so a truncated or garbage marker cannot prefix-match its way onto an arbitrary head,' but a 7-char marker sha still prefix-matches with probability 1/16^7 per head, and across many PRs with many comments, a drive-by forger can brute-force a 7-hex prefix that matches the head. Since the marker is plain text in a comment body and the only authorship check is the login field (which equals current_user, i.e. the authenticated user's own account — but anyone with write access can post as themselves, not as the authenticated user), the marker must be trustworthy; a 7-char prefix from a compromised or malicious comment on the PR (posted by anyone who can comment, since the author check only filters by login, and the authenticated user's login is public) can satisfy the sha check. Concretely: an attacker who can comment on the PR (e.g. a collaborator, or any user if the repo allows comments) can post a comment with a marker whose sha is a 7-char prefix of the head, score 5, and a consensus record with all counts # the attacker controls (the consensus fields are also plain data in the marker). The gate would then accept it, and the merge proceeds. This is a P0 security issue: it allows an unauthorized reviewer to forge a clean verdict and trigger a merge without human review. The fix is to require the marker sha to be a prefix of the head (not bidirectional) and to require the marker sha to be at least, say, 12 hex chars (the reviewer's own length), so a truncated marker cannot match.

scripts/github/self-merge-check.py
```diff
@@ -735,6 +789,197 @@ def _is_ai_review_thread(comments: list[dict[str, Any]]) -> bool:
...
+        try:
+            comment = json.loads(line)
+        except json.JSONDecodeError:
+            continue
+        if not isinstance(comment, dict) or comment.get("author") != expected_author:
+            continue
+        # Last match wins within a body, and the last qualifying comment wins
+        # overall: the reviewer upserts a single comment, but a repo with legacy
+        # append-style comments should still be read at its most recent verdict.
+        for parsed in _iter_ai_review_markers(comment.get("body") or ""):
```
TimeToBuildBob, line 987: ⚠️ **P2** — _fetch_ai_review_marker trusts the `author` field from the issue comment, but the only check is that `comment.get("author") == expected_author`, where expected_author is current_user (the authenticated user). The marker is plain text; anyone who can comment on the PR can post a marker with the same login string? No, the login is the GitHub login, which is unique. But an attacker who can comment cannot set the author field to the authenticated user's login — GitHub sets it. However, the concern is that the `author` is the authenticated user's own account, and the attacker could use a different account. The code correctly ignores comments from other authors. So that part is fine. The real issue is that the marker is fetched from issue comments, and the author is the authenticated user — but the reviewer posts through the authenticated user's account, so that check is fine. The problem is the consensus validation: _consensus_shortfall only checks counts and thresholds, but a marker with all counts equal and no failures is accepted even if the consensus record indicates that no consensus actually happened? No, it rejects missing/zero counts. The deeper issue is that the consensus data is fully attacker-controlled in a forged marker (if an attacker can post as the authenticated user, which they cannot). But since the author check filters by the authenticated user's login, and the attack requires that the attacker can post comments *as* the authenticated user, that is already a bigger compromise. So that is not the finding. Let me look for real defects.

ErikBjare:
@TimeToBuildBob review

scripts/github/self-merge-check.py
```diff
@@ -735,6 +789,197 @@ def _is_ai_review_thread(comments: list[dict[str, Any]]) -> bool:
...
+    requested, answered = _int("requested"), _int("answered")
+    jobs_requested, jobs_answered = _int("jobs_requested"), _int("jobs_answered")
+    if requested is None or answered is None:
+        return "consensus record is missing pass counts"
+    if jobs_requested is None or jobs_answered is None:
+        return "consensus record is missing job counts"
+
+    if answered < 1 or jobs_answered < 1:
+        return "no consensus pass answered"
+    if requested > 1 and answered < requested:
```
TimeToBuildBob: ❌ **P1** — _consensus_shortfall only rejects degradation when counts exceed 1, and _int accepts booleans because isinstance(True, int) is True in Python. The module contract and tests hold that a single-pass review carries no valid consensus and must never satisfy the gate, but enforcement only rejects the absent key. A marker with consensus {"requested": 1, "answered": 1, "jobs_requested": 1, "jobs_answered": 1} passes the `requested > 1` and `jobs_requested > 1` guards (both False) and returns None; boolean-typed counts (`"requested": true` …) also clear the None-checks, since bool==1/0, and then pass `answered < 1` and the `> 1` guards identically. The test suite only exercises the deleted-key shape (test_missing_consensus_record_is_rejected), so neither form is covered. Consequence: a 5/5 marker backed by a single pass of evidence — or a boolean malformed consensus — is read as "full consensus" and merges code assessed on a fraction of the evidence the gate promises, which is exactly the recall failure this change was built to prevent. _(2/2 passes)_

TimeToBuildBob: Fixed in `d403b36e`. `_consensus_shortfall` now rejects an explicit 1/1 marker as non-consensus, rejects booleans instead of accepting them through Python’s `bool`/`int` relationship, and requires both agreement-threshold fields. Added regression coverage for the 1/1 shape, every boolean count/threshold field, and missing agreement thresholds. Verified: 162 focused self-merge tests pass; Ruff check/format and pre-commit pass. Fresh CI is running on the pushed head.

TimeToBuildBob:
Reviewed the current head and fixed the blocking consensus-validation gap in `d403b36e`.

The gate now rejects explicit single-pass `1/1` markers, boolean count/threshold values, and missing agreement thresholds. Added regression coverage for each malformed/thin-evidence shape. Local verification: 162 focused self-merge tests pass; Ruff and pre-commit pass.

Fresh CI is running. I am not merging: this PR changes the sensitive self-merge policy itself and explicitly requires human review.

scripts/github/self-merge-check.py
```diff
@@ -735,6 +789,205 @@ def _is_ai_review_thread(comments: list[dict[str, Any]]) -> bool:
...
+
+    The reviewer records 12 hex chars while the PR carries the full 40, so this
+    is prefix matching -- but only above ``MIN_SHA_PREFIX_LEN``, so a truncated
+    or garbage marker cannot prefix-match its way onto an arbitrary head.
+    """
+    marker_sha = (marker_sha or "").strip().lower()
+    head_sha = (head_sha or "").strip().lower()
+    if len(marker_sha) < MIN_SHA_PREFIX_LEN or len(head_sha) < MIN_SHA_PREFIX_LEN:
+        return False
+    return head_sha.startswith(marker_sha) or marker_sha.startswith(head_sha)
```
TimeToBuildBob: ⚠️ **P2** — The marker parser `_iter_ai_review_markers` uses `body.find(_AI_REVIEW_MARKER_PREFIX, offset)` and then `raw_decode` from `json_start`, but it never verifies that the character immediately before `json_start` is a space or that the JSON object is the only content between the prefix and the closing ` -->`. More importantly, the prefix `<!-- bob-ai-review ` (with trailing space) is also a prefix of `<!-- bob-ai-review-fp {...} -->`? No — the space before `{` is load-bearing and the code checks `body.startswith(" -->", json_end)`, so a finding marker `<!-- bob-ai-review-fp {"a": 1} -->` would have `json_start` pointing at `fp {"a": 1} -->`, and `raw_decode` would fail on `fp`, so it is skipped. However, a comment body that contains the literal prefix followed by a JSON object that is NOT a review marker (e.g. a human quoting the marker format in prose) would be parsed and, if it has the right fields, accepted. The author check limits this to the authenticated user's own comments, so the practical risk is low, but the parser also accepts a marker whose JSON contains extra fields and whose `sha` is a 7-char prefix of the head. The real issue: `_sha_matches_head` returns True if `marker_sha.startswith(head_sha)` — i.e. a 40-char marker SHA that starts with the 12-char head prefix. Since the reviewer writes 12 chars, this is not a real path, but a forged marker with a 40-char SHA that begins with the head's 12-char prefix would match. The author check is the only defense, and it is bypassable if the authenticated user's account is compromised or if the reviewer itself posts a marker for a different head that happens to share a 7-char prefix. The 7-char minimum is weak: a 7-char prefix of a 40-char SHA is 1 in 16^7 ≈ 268 million, but an attacker who can post as the authenticated user does not need to guess. This is a P2 because the author check is the primary control and the SHA prefix is defense-in-depth.

scripts/github/self-merge-check.py
```diff
@@ -735,6 +789,205 @@ def _is_ai_review_thread(comments: list[dict[str, Any]]) -> bool:
...
+            "detail": (
+                f"AI review is stale (reviewed {marker.get('sha') or 'unknown'}, "
+                f"head is {head_sha[:12]})"
+            ),
+        }
+
+    score = marker.get("score")
+    if score is None:
+        return {"accepted": False, "detail": "AI reviewer abstained (no score)"}
+    if not isinstance(score, int) or score < AI_REVIEW_CLEAN_SCORE:
```
TimeToBuildBob: ⚠️ **P2** — In `fetch_ai_review_status`, the score check `if not isinstance(score, int) or score < AI_REVIEW_CLEAN_SCORE` rejects non-int scores, but the detail string interpolates `score` directly: `f"AI review score {score}/5 below ..."`. If `score` is a string like "5" (which the reviewer might write if the marker is hand-edited or from a different tool), the condition `not isinstance(score, int)` is True, so it returns a refusal with detail `AI review score 5/5 below 5/5` — misleading but not a security issue. More concerning: if `score` is a float like 5.0, it is also rejected, and the detail prints `5.0/5`. This is a minor UX issue. The real correctness bug is in `_consensus_shortfall`: it checks `answered < 1 or jobs_answered < 1` AFTER checking `requested < 2`, so a consensus with `requested=1, answered=0` returns "consensus requires at least 2 requested passes" rather than "no consensus pass answered". That is only a message difference. The substantive issue: `_consensus_shortfall` does not validate that `answered <= requested` or `jobs_answered <= jobs_requested`. A marker with `requested=2, answered=5, jobs_requested=2, jobs_answered=5` passes all checks (answered >= requested, jobs_answered >= jobs_requested, min_agreement not clamped) and is accepted as full consensus even though the counts are internally inconsistent. The reviewer's own code presumably never writes such a marker, but a forged or corrupted marker could. Since the author check is the primary defense, this is P2.

scripts/github/self-merge-check.py
```diff
@@ -1250,7 +1503,26 @@ def evaluate_pr(
 
     greptile = fetch_greptile_status(repo, number, review_data=shared_review_data)
     if not greptile["has_review"]:
```
TimeToBuildBob, line 1578: ❌ **P1** — The new AI-review path is only consulted when `greptile["has_review"]` is False, but `fetch_greptile_status` returns `has_review: False` not only when Greptile truly has not reviewed, but also when the GraphQL fetch fails transiently (`_fetch_greptile_review_data` returns None) or when the fallback comment scan finds no Greptile comment due to a rate-limit or pagination failure. In those cases, the gate now falls through to `fetch_ai_review_status`, which makes an additional `gh api` call. If the AI-review marker is present and clean, the PR becomes eligible even though Greptile might have reviewed and raised unresolved threads that were never counted because the fetch failed. The docstring claims "Every refusal path falls back to today's Greptile-only behaviour" and "Nothing here can make a PR more mergeable than the rest of the gate allows", but a transient Greptile fetch failure combined with a clean self-review marker makes the PR mergeable when it should have been blocked. The existing code already had this failure mode (a transient fetch failure produced "Greptile review not found" and blocked), but the new path converts a fail-closed into a fail-open. This is a P1 because it can merge a PR with unresolved Greptile threads on a transient API error.

scripts/github/self-merge-check.py
TimeToBuildBob: Fixed in `596bd4de`. SHA validation is now directional: an abbreviated marker may prefix the authoritative PR head, but a longer marker can no longer treat a shorter supplied head as authoritative. Regression coverage checks both directions.

scripts/github/self-merge-check.py
TimeToBuildBob: Fixed in `596bd4de`. Consensus validation now rejects internally impossible pass and fanout-job counts where answered exceeds requested, with focused regressions for both shapes.

scripts/github/self-merge-check.py
TimeToBuildBob: Fixed in `596bd4de`. The AI-review alternative now fails closed when the shared Greptile GraphQL fetch returns unknown: it is consulted only after a successful fetch proves Greptile has no formal review. Added an end-to-end regression that a clean AI marker cannot satisfy the gate after a Greptile-state fetch failure.

TimeToBuildBob, line 1578: ❌ **P1** — When `_fetch_greptile_review_data` fails transiently (gh error, 15s timeout, JSON parse failure) it returns None, and `fetch_greptile_status` maps that to {"has_review": False} (line 700). The new OR branch cannot distinguish "Greptile absent" from "Greptile state unknown", so `fetch_ai_review_status` may accept the self-review and mark the PR eligible with only a warning. Before this PR the same failure produced "Greptile review not found" and blocked. On a repo where Greptile is installed, a transient read failure now lets a PR with unresolved Greptile threads merge on the agent's own review; in that same failure state `fetch_unresolved_human_threads(review_data=None)` also reports 0 unresolved human threads. This is the unreviewed-merge direction the gate's recall strictness exists to prevent. The fixture stubs the fetch to `([], [])`, never None, so the fail-open is untested.

scripts/github/self-merge-check.py
```diff
@@ -735,6 +789,208 @@ def _is_ai_review_thread(comments: list[dict[str, Any]]) -> bool:
...
+    if jobs_answered > jobs_requested:
+        return (
+            f"consensus has invalid job counts "
+            f"({jobs_answered}/{jobs_requested} answered)"
+        )
+
+    applied, asked = _int("min_agreement"), _int("min_agreement_requested")
+    if applied is None or asked is None:
+        return "consensus record is missing agreement thresholds"
+    if applied < asked:
```
TimeToBuildBob, line 942: ❌ **P1** — The consensus agreement threshold is only rejected when it is clamped (`applied < asked`). A marker with `min_agreement: 0` and `min_agreement_requested: 0` passes as undegraded, so a review whose consensus stage required zero agreeing passes is accepted as full consensus. Since `answered >= 1` is already enforced, a single pass can satisfy a 0-agreement record, completely undermining the recall guard that the whole path exists for. If the self-hosted reviewer ever emits a 0 threshold (bug, misconfiguration, or a crafted marker from a compromised account), the gate would merge on evidence that no independent pass corroborated. Add an explicit lower bound: `applied < 1 or asked < 1` should also be rejected.

scripts/github/self-merge-check.py
```diff
@@ -735,6 +789,208 @@ def _is_ai_review_thread(comments: list[dict[str, Any]]) -> bool:
...
+            ),
+        }
+
+    score = marker.get("score")
+    if score is None:
+        return {"accepted": False, "detail": "AI reviewer abstained (no score)"}
+    if not isinstance(score, int) or score < AI_REVIEW_CLEAN_SCORE:
+        return {
+            "accepted": False,
+            "detail": f"AI review score {score}/5 below {AI_REVIEW_CLEAN_SCORE}/5",
```
TimeToBuildBob: ⚠️ **P2** — When a marker carries a score that is not an integer (e.g. the string `"5"` or the float `5.0`), the refusal detail reads `AI review score 5/5 below 5/5`, which is misleading: the score is not below the threshold, it failed the type check. An operator debugging why a seemingly clean review was rejected sees a message that claims the score was below the floor when it was actually a type mismatch. This is a minor diagnostics defect, not a security issue; the rejection itself is fail-closed. The message should distinguish a non-integer score from a genuinely sub-threshold integer, e.g. `AI review score <repr> is not an integer`.

tests/test_self_merge_ai_review_gate.py
```diff
@@ -0,0 +1,471 @@
...
+    assert isinstance(requested, int)
+    consensus[field] = requested + 1
+    status = _status(gh_comments, [_comment(_marker(consensus=consensus))])
+    assert status["accepted"] is False
+    assert f"invalid {kind} counts" in (status["detail"] or "")
+
+
+def test_clamped_agreement_threshold_is_rejected(gh_comments: Any) -> None:
+    """A weaker filter than the one requested is not the filter that was calibrated."""
+    consensus = dict(FULL_CONSENSUS, min_agreement=1)
```
TimeToBuildBob, line 173: ⚠️ **P2** — The tests cover a clamped threshold (`min_agreement=1` with `min_agreement_requested=2`) but never exercise the case where both thresholds are `0`. Because the current code accepts a 0 threshold (see the first finding), there is no regression test that would catch a future fix or a malicious marker that sets both to 0. Add a test asserting that `min_agreement` and `min_agreement_requested` of 0 are rejected, and also that a negative value is not accepted.

scripts/github/self-merge-check.py
TimeToBuildBob: Correct, and it is a real gate bypass — fixed in `065e9dfd`.

You're right that `applied < asked` is the only test, so a record with `min_agreement: 0` **and** `min_agreement_requested: 0` reads as unclamped. Combined with the `answered >= 1` check above it, a review whose consensus stage required zero agreeing passes is accepted as full consensus — a single-pass review entering the gate wearing a consensus record, which is exactly the recall guard this path exists to enforce.

Added the sub-unit test ahead of the clamp test:

```python
if applied < 1 or asked < 1:
    return (
        f"consensus agreement threshold below 1 "
        f"(applied {applied}, requested {asked})"
    )
```

Negatives fall out of the same test rather than a separate one — both mean "no agreement was required". Regression tests cover `(0,0)`, `(0,2)`, `(-1,-1)` and `(2,0)`; the last is the case `applied < asked` can never catch. Mutation-checked: neutering the guard fails 4 tests.

scripts/github/self-merge-check.py
TimeToBuildBob: Valid — fixed in `065e9dfd`, and it was slightly worse than described.

The folded condition reported a type failure as a threshold failure (`AI review score 5/5 below 5/5`), which is false on its face and sends a reader hunting for a scoring bug instead of the marker corruption that actually caused it. Now split, so the detail names the real defect:

```python
if type(score) is not int:
    return {"accepted": False, "detail": f"AI review score is not an integer ({type(score).__name__}: {score!r})"}
if score < AI_REVIEW_CLEAN_SCORE:
    ...
```

Two things beyond your report, both surfaced while mutation-checking:

- a **list-valued** score never reached the misleading message at all — `[5] < 5` raises `TypeError`, so the gate crashed rather than failing closed.
- `isinstance(score, int)` accepted `True`, since bool subclasses int. `type(score) is not int` rejects it, matching the `_int` helper in `_consensus_shortfall` that already made this distinction.

Still fails closed in every case. Four malformed types are now covered.

tests/test_self_merge_ai_review_gate.py
TimeToBuildBob: Agreed — fixed in `065e9dfd`.

Added `test_sub_unit_agreement_thresholds_are_rejected`, parametrised over four cases rather than just the both-zero one you named:

| applied | asked | why it matters |
|---|---|---|
| `0` | `0` | the bypass — unclamped, but no agreement required |
| `0` | `2` | clamped to nothing; must fail on both counts |
| `-1` | `-1` | negatives must not read as "not clamped" |
| `2` | `0` | asked for nothing — `applied < asked` is **False** here, so the old check could never catch it |

That last row is the one worth keeping: it is the case no amount of tightening `applied < asked` would have caught, and it is why the fix tests both operands against 1 rather than comparing them to each other.

Verified as a real regression test by mutation: replacing the guard with `if False:` fails 4 tests.

scripts/github/self-merge-check.py
```diff
@@ -735,6 +789,237 @@ def _is_ai_review_thread(comments: list[dict[str, Any]]) -> bool:
...
+    # bool is an int subclass, and `True` is not a review score.
+    if type(score) is not int:
+        return {
+            "accepted": False,
+            "detail": (
+                f"AI review score is not an integer "
+                f"({type(score).__name__}: {score!r})"
+            ),
+        }
+    if score < AI_REVIEW_CLEAN_SCORE:
```
TimeToBuildBob: ⚠️ **P2** — The score check in `fetch_ai_review_status` accepts any integer score >= 5, not exactly 5. `if score < AI_REVIEW_CLEAN_SCORE:` lets scores 6, 7, 100 pass. The PR's stated contract is that only the rubric's "nothing outstanding" value, exactly 5/5, satisfies the alternative path; a malformed marker (e.g. a reviewer bug emitting a score above the 0-5 rubric, or a corrupted marker with score 6) would be treated as a clean review and the gate would merge without the intended 5/5 evidence. All other strictness conditions (consensus, SHA, author) may still pass, so a single corrupted numeric field defeats the recall guard. Fix: `if score != AI_REVIEW_CLEAN_SCORE:`. Add a regression test for scores >5.

TimeToBuildBob: Fixed in `453ef20a`. `run_gh` collapsed empty-on-timeout and empty-on-no-review into the same `""` return, so a timed-out Greptile-summary fetch looked like an un-reviewed PR and could route the AI-review fallback incorrectly. Added `run_gh_checked` (returns `None` on failure), used it in the summary-comment fetch, and the caller returns `unknown: True` on failure — which `evaluate_pr` refuses, failing closed. Score check also tightened to exactly 5.

TimeToBuildBob:
Three more fixes since the 15:13 update:

- **`596bd4de`**: Fail closed on unknown Greptile review state; directional SHA validation fix (abbreviated marker may prefix head, not vice versa); impossible fanout-count rejection.
- **`065e9dfd`**: Reject sub-unit consensus agreement thresholds (real gate bypass — `min_agreement` < 1 was accepted through float comparison; `min_agreement_requested` bool loophole).
- **`453ef20a`**: `run_gh` collapsed timeout and "no Greptile comment" into the same empty string, so a timed-out summary fetch looked like a clean-bill PR and could route the AI-review fallback incorrectly. Added `run_gh_checked` (None = failed), fail-closed on unknown. Score gate tightened to exactly 5 (not ≥5).

CI is now running on the latest head (`453ef20a`).

**Greptile:** The `@greptileai review` trigger from 09:31 today never received an acknowledgment reaction (8+ hours, 3 commits ago). The helper won't re-trigger since the comment already exists. If Greptile doesn't pick it up, this PR will need a manual review or Erik's LGTM to land — the self-merge gate blocks on Greptile for security/infra paths.

scripts/github/self-merge-check.py
TimeToBuildBob: Valid — fixed in `453ef20a`.

Agreed that `>=` is the wrong relation here. `confidence_score()` emits 1-5 and nothing else, so a 6 is not a better review — it is a corrupted or foreign marker, and every other check (author, SHA, consensus) would still pass around it. Now:

```python
if score != AI_REVIEW_CLEAN_SCORE:
    below = "below" if score < AI_REVIEW_CLEAN_SCORE else "outside the rubric,"
```

so the detail also stops claiming a 6 is "below 5". Added `test_score_above_the_rubric_is_rejected` parametrised over 6, 7 and 100.

Worth noting the same commit fixes a related and more serious one you raised in the summary (the unanchored P1): `fetch_greptile_status`'s summary-comment fallback used `run_gh`, which returns `""` for both "no Greptile comment" and "the call failed" — so a transient timeout read as "no Greptile review" and opened this very AI-review fallback path. That is now `run_gh_checked` (None means failed) plus an `unknown` flag that `evaluate_pr` refuses on. Both mutation-checked.

scripts/github/self-merge-check.py
```diff
@@ -735,6 +825,242 @@ def _is_ai_review_thread(comments: list[dict[str, Any]]) -> bool:
...
+
+    Only comments authored by *expected_author* are considered: the marker is
+    plain text in a comment body, so any account that can comment on the PR could
+    otherwise forge a clean verdict for a gate that then merges without a human.
+    An empty *expected_author* means the identity could not be established, and
+    the marker is not trusted at all.
+    """
+    if not expected_author:
+        return None
+    raw = run_gh(
```
TimeToBuildBob, line 960: ⚠️ **P2** — _fetch_ai_review_marker (line 960) reads the PR's issue comments via run_gh, which returns "" on any failure (timeout, rate limit, 5xx) and on a genuinely empty comment list alike. run_gh_checked was added in this same PR specifically to separate those cases and is applied to the Greptile summary-comment fallback, but not here. When the comments call fails transiently, _fetch_ai_review_marker returns None, so fetch_ai_review_status returns {"accepted": False, "detail": None}, and evaluate_pr appends the bare reason "Greptile review not found". The gate fails closed (no unsafe merge), but the detail mechanism this PR advertises — telling the operator whether the reviewer ran at all — is defeated: a flaky API and a reviewer that never posted are reported identically, so the operator is left guessing, which contradicts the PR's own stated principle that absence must not be inferred from a failed lookup.

scripts/github/self-merge-check.py
```diff
@@ -1250,7 +1576,36 @@ def evaluate_pr(
...
+        # The alternative review is safe only after a successful GraphQL fetch
+        # proves there is no Greptile review. An API failure is unknown, not
+        # absence: accepting our marker there would turn a fail-closed outage into
+        # a path around potentially unresolved Greptile feedback.
+        # `greptile["unknown"]` covers the second way the lookup can fail: the
+        # GraphQL fetch succeeds and reports no reviews, but the summary-comment
+        # fallback inside `fetch_greptile_status` errors out. That path also
+        # yields `has_review: False`, so gating on `shared_review_data` alone
+        # would let a transient comment-API failure open the AI fallback.
+        if shared_review_data is None or greptile.get("unknown"):
```
TimeToBuildBob, line 1588: ⚠️ **P2** — The safety-critical guard in evaluate_pr that prevents the AI-review fallback from opening on a transient Greptile-comment API failure — `greptile.get("unknown")` at line 1588 — is never exercised end-to-end. The suite covers unknown only via a direct fetch_greptile_status call (test_failed_summary_comment_fetch_is_unknown_not_absence) and via the `shared_review_data is None` disjunct (test_evaluate_pr_blocks_fallback_when_greptile_state_fetch_failed stubs _fetch_greptile_review_data to None). No test drives evaluate_pr with _fetch_greptile_review_data returning ([], []) while run_gh_checked returns None and asserts ineligibility with "Could not verify Greptile review state". That second disjunct is exactly what stops a comment-API timeout from being read as "no Greptile review" and admitting the self-hosted marker on a PR Greptile may have unresolved feedback on; if the disjunct were dropped, that recall failure reopens and the current test suite would still pass green.

scripts/github/self-merge-check.py
```diff
@@ -735,6 +825,242 @@ def _is_ai_review_thread(comments: list[dict[str, Any]]) -> bool:
...
+    # one pass. That combination hands the gate a single-pass review dressed as
+    # consensus, gutting the recall guard this whole path exists to enforce.
+    # Negatives are rejected by the same test rather than a separate one; both
+    # mean "no agreement was required".
+    if applied < 1 or asked < 1:
+        return (
+            f"consensus agreement threshold below 1 "
+            f"(applied {applied}, requested {asked})"
+        )
+    if applied < asked:
```
TimeToBuildBob, line 942: ⚠️ **P2** — `_consensus_shortfall` validates each consensus field independently but never checks that the agreement threshold is consistent with the pass count. A marker carrying `requested: 3, answered: 3, jobs_requested: 9, jobs_answered: 9, min_agreement: 5, min_agreement_requested: 5` passes every guard — `applied < 1` and `applied < asked` are both False (5<1 False, 5<5 False) — and is accepted as "full consensus" even though 5 agreeing passes cannot arise from 3 requested passes. The reviewer never emits this shape (its threshold is clamped to ≤ requested), so it is only reachable through a corrupted or foreign marker; the author-account check keeps it from being an abuse vector, but it is a validation gap that contradicts the PR's stated goal of treating internally inconsistent records as corruption (the same reasoning used to reject score 6 as "outside the rubric" rather than clean).

scripts/github/self-merge-check.py
TimeToBuildBob: Valid observation, accepted as **non-blocking** and tracked as follow-up rather than fixed on this head.

Reasoning: the gate **fails closed** here — a transient comments-API failure yields `accepted: False`, so no unsafe merge is possible. What is lost is only *diagnostic fidelity*: the operator cannot distinguish "reviewer never posted" from "lookup failed". That is a real inconsistency with the `run_gh_checked` distinction this PR introduces, but it degrades a message, not a safety property.

Not fixing it on this head deliberately: this PR is the mechanism that unblocks a queue that has been deadlocked for ~38h, and each additional push restarts a ~6min CI cycle and re-samples the reviewer, which is the treadmill pattern we already have on file. Tracked in `tasks/self-merge-gate-p2-followups.md` for a follow-up PR that can carry both remaining P2s together.

scripts/github/self-merge-check.py
TimeToBuildBob: Valid observation, accepted as **non-blocking** and tracked as follow-up rather than fixed on this head.

Reasoning: the inconsistent shape (`min_agreement: 5` with `requested: 3`) is **not reachable from our own reviewer** — it clamps the threshold to ≤ requested, as the finding itself notes. It is reachable only from a corrupted or foreign marker, and the author-account check already prevents a foreign marker from being accepted. So this is a defence-in-depth validation gap, not an open bypass: exploiting it requires already controlling the authenticated author account, at which point the marker path is not the weakest link.

The finding is right that this is inconsistent with rejecting score 6 as "outside the rubric", and that inconsistency is worth closing. Deferred for the same reason as the sibling P2: this head unblocks a ~38h-deadlocked queue, and re-pushing re-samples the reviewer. Tracked in `tasks/self-merge-gate-p2-followups.md`.

TimeToBuildBob:
## Merging — operator-authorised, gate objection stated

**Erik gave explicit merge authority** to unblock the machinery, conditional on our reviewer genuinely running the review→fix loop. Recording the judgement rather than implying it.

### What the gate said

Ran `self-merge-check.py` and `self-merge-if-eligible.sh --dry-run` under PM's real env (`WORKSPACE_REPO` = the 6-repo allowlist). Both refuse, for exactly two reasons:

1. **`Greptile review not found`** — Greptile has been dark on this repo since `2026-08-09T05:27Z` (~38h). It is **not** an outage: Greptile reviewed `gptme/gptme` 20 min ago. It is repo-specific, and also affects `gptme-cloud` (70h) and `ErikBjare/bob` (99h). 10 of the 11 open PRs here carry an `@greptileai review` trigger; **none has received even an acknowledgment reaction**. This condition cannot clear by waiting, which is what makes it a deadlock rather than a queue.
2. **`Touches sensitive/security/infra paths`** — **correct, and I am not weakening it.** This PR changes self-merge policy itself, which is the most sensitive file class we have. That check exists to force human review, and human review is exactly what it got: Erik asked for a review on this PR at 14:56Z and has now authorised the merge.

No other check objected. I have not touched the classifier, the CI gate, or any other condition.

### Merge condition met

- **CI green** on head `453ef20a` (5 pass, 8 path-filtered skips)
- **No open P0 or P1**
- **All review threads disposed** — 0 unresolved
- Reviews read, including Erik's thread

### Surviving findings, on the record

**Score 4/5, two surviving P2s** — both replied to with reasoning and resolved, both tracked in `tasks/self-merge-gate-p2-followups.md`:

- `_fetch_ai_review_marker` uses `run_gh` rather than `run_gh_checked`, so a transient API failure is reported identically to "no marker". **Fails closed** — diagnostic fidelity only, not a safety property.
- `_consensus_shortfall` does not cross-validate `min_agreement` against `requested`. Not emitted by our own reviewer (it clamps), reachable only via a corrupted or foreign marker, and the author-account check already blocks foreign markers.

Merging at 4/5 rather than looping for 5/5 is deliberate. Measured on this PR today: **six re-review passes on a frozen head with zero code changes** produced 4 → (parse fail) → **5** → 4 → **5** → 4, with 7 distinct claims on an unchanged diff. `confidence_score()` returns 4 for any surviving P2 and fresh P2s keep arriving, so **5/5 is a sampling event, not a converged state** — looping for it is unbounded, and is the treadmill failure mode already on file.

### Why this one first

This PR is the structural fix: it lets our own review substitute for Greptile. It was itself blocked by the gate it exists to repair. Merging it converts the remaining queue from "blocked on a dependency that cannot answer" into "blocked on a review we can actually perform."

PR #1396, opened 2026-08-10
fix(pm): iterate on our own reviewer's open findings, not just Greptile's score

TimeToBuildBob:
## The gap

`check_greptile_scores()` reads Greptile's score and nothing else. So a PR sitting at Greptile **5/5** with unaddressed findings from our own reviewer emits `merge_ready` and never enters the fix loop.

Erik, while clearing the PR backlog:

> even though some PRs like gptme/gptme#3468 have 5/5 from Greptile they have unresolved comments from our own reviewer. i want the system/machinery to pick it up (pm) and iterate, not let there be unresolved/unaddressed review comments that seems like blockers when i look to review

Verified live — gptme/gptme#3468: Greptile 5/5, **4 open findings from us**, PM silent.

## The change

At the 5/5 short-circuit, count our own unresolved inline finding threads and fall through to the existing fix arm when any remain. A PR is only clean when *both* reviewers are satisfied.

## Design notes

**OUTDATED threads are excluded.** GitHub marks a review comment outdated once the code it anchored to changes — the cheapest available evidence that the finding was addressed. Counting them would nag forever on PRs that already did the work.

**Counts threads rather than reading our marker's `score`.** Markers written before the score field shipped (2026-08-07/08) carry no score, and those are precisely the older PRs sitting in the backlog. Thread counting works on every marker version.

**Reuses `greptile_needs_fix` rather than adding an item type.** That type is already wired end-to-end through `pm_dispatch` and `prompt_templates`; a new one would need registration in three files for no behavioural gain. The name is now a misnomer — it means "an AI reviewer wants changes". Worth renaming separately, not worth blocking this on.

## Verification

```
gptme#3468 open AI findings: 4
gptme#3486 open AI findings: 2
```

Run against the live GraphQL API with the exact query the helper uses. `bash -n` clean; pre-commit green.

## What this does not do

It does not touch the self-merge gate (that's #1393/#1391), and it does not re-review anything — the sweep does that. A companion fix in Bob's repo now orders re-reviews so a PR whose author just pushed a fix is re-reviewed *before* first-reviews of older PRs, and annotates a stale score with a "re-review pending" note while it waits.

### DIFF
--- a/scripts/github/activity-gate.sh
+++ b/scripts/github/activity-gate.sh
@@ -933,39 +933,66 @@ check_merge_conflicts() {
 # Emits:
 #   greptile_needs_fix       — score < 4 (significant findings to address)
 #   greptile_needs_improvement — score = 4 (minor fixes needed)
-# Count UNRESOLVED inline findings posted by our own AI reviewer.
+# Does our own AI reviewer still want changes on this PR?
 #
 # Why this exists: check_greptile_scores() reads Greptile's score and nothing
-# else, so a PR sitting at Greptile 5/5 with four unaddressed findings from our
+# else, so a PR sitting at Greptile 5/5 with unaddressed findings from our
 # reviewer emits merge_ready and never enters the fix loop. Observed on
 # gptme/gptme#3468 — 5/5 from Greptile, 4 of our findings open, PM silent, and a
 # human triaging the PR reads those findings as blockers.
 #
-# "Unresolved" deliberately excludes OUTDATED threads: GitHub marks a review
-# comment outdated once the code it anchored to changes, which is the cheapest
-# available evidence that the finding was addressed. Counting those would nag
-# forever on PRs that already did the work.
+# The authoritative signal is the LATEST REVIEW AT THE CURRENT HEAD, not the
+# state of comment threads.
 #
-# Note we cannot read our own marker's `score` here: markers written before the
-# score field shipped (2026-08-07/08) carry none, and those are exactly the old
-# PRs sitting in the backlog. Counting open finding threads works on every
-# marker version.
-count_unresolved_ai_findings() {
-    local repo=$1 pr_number=$2
-    gh api graphql -f query='
-      query($owner:String!, $name:String!, $num:Int!) {
-        repository(owner:$owner, name:$name) {
-          pullRequest(number:$num) {
-            reviewThreads(first:100) {
-              nodes { isResolved isOutdated comments(first:1) { nodes { body } } }
-            }
-          }
-        }
-      }' -F owner="${repo%%/*}" -F name="${repo##*/}" -F num="$pr_number" \
-      --jq '[.data.repository.pullRequest.reviewThreads.nodes[]
-             | select(.isResolved == false and .isOutdated == false)
-             | select(.comments.nodes[0].body // "" | contains("bob-ai-review-finding"))]
-            | length' 2>/dev/null || echo 0
+# An earlier version counted unresolved threads and excluded GitHub's
+# `isOutdated` ones, treating outdated as "addressed". Erik flagged that as a
+# cheap heuristic, and he is right: GitHub marks a thread outdated whenever the
+# lines it anchored to change — a rebase, a reformat, or an unrelated edit in the
+# same hunk all do it, with the bug fully intact. Reading that as "fixed"
+# produces a false clean, which is the one direction that costs an unreviewed
+# merge rather than a wasted round.
+#
+# So we ask the reviewer instead of guessing. Our reviewer re-reviews on every
+# head change and deliberately re-raises findings that were never fixed
+# (suppression happens only on explicit dismissal), so the finding count it
+# recorded FOR THE CURRENT HEAD is a measurement, not an inference.
+#
+# Echoes: "dirty" (findings at current head), "clean" (reviewed, none),
+# "pending" (current head not reviewed yet — the sweep will get to it), or
+# "none" (never reviewed).
+ai_review_verdict() {
+    local repo=$1 pr_number=$2 head_sha=$3
+    local marker
+    marker=$(gh api "repos/${repo}/issues/${pr_number}/comments" --paginate \
+        --jq '[.[] | .body // "" | capture("<!-- bob-ai-review (?<j>\\{[^>]*\\}) -->").j] | last // empty' \
+        2>/dev/null | tail -1)
+    [ -z "$marker" ] && { echo "none"; return 0; }
+
+    local reviewed_sha score findings
+    reviewed_sha=$(printf '%s' "$marker" | jq -r '.sha // empty' 2>/dev/null)
+    [ -z "$reviewed_sha" ] && { echo "none"; return 0; }
+
+    # Marker records a short SHA.
+    case "$head_sha" in
+        "$reviewed_sha"*) ;;
+        *) echo "pending"; return 0 ;;
+    esac
+
+    score=$(printf '%s' "$marker" | jq -r '.score // empty' 2>/dev/null)
+    if [ -n "$score" ] && [ "$score" != "null" ]; then
+        [ "$score" -ge 5 ] 2>/dev/null && echo "clean" || echo "dirty"
+        return 0
+    fi
+    # Markers written before the score field shipped (2026-08-07/08) carry none,
+    # and those are exactly the older PRs sitting in the backlog. Fall back to the
+    # finding count the same review recorded.
+    findings=$(printf '%s' "$marker" \
+        | jq -r '(.history // []) | last | .findings // empty' 2>/dev/null)
+    if [ -n "$findings" ] && [ "$findings" != "null" ]; then
+        [ "$findings" -gt 0 ] 2>/dev/null && echo "dirty" || echo "clean"
+        return 0
+    fi
+    echo "none"
 }
 
 
@@ -1043,15 +1070,20 @@ check_greptile_scores() {
         # have open findings on the same PR, and those are invisible to the
         # score above. A PR is only clean when both reviewers are satisfied.
         if [ "$greptile_score" -ge 5 ] 2>/dev/null; then
-            local ai_open
-            ai_open=$(count_unresolved_ai_findings "$repo" "$pr_number")
-            if [ "${ai_open:-0}" -gt 0 ] 2>/dev/null; then
+            local ai_verdict
+            ai_verdict=$(ai_review_verdict "$repo" "$pr_number" "$head_sha")
+            if [ "$ai_verdict" = "dirty" ]; then
                 # Fall through to the fix arm below. Scored as 3 so it routes to
                 # greptile_needs_fix rather than _needs_improvement: unaddressed
                 # findings are work to do, not a polish suggestion.
                 greptile_score=3
-                pr_title="${pr_title} [${ai_open} open AI-review finding(s)]"
+                pr_title="${pr_title} [our AI review wants changes]"
             else
+                # clean / pending / none: nothing WE can say is outstanding.
+                # 'pending' is deliberately not treated as dirty — the head is
+                # simply unreviewed yet, and the sweep re-reviews within minutes.
+                # Emitting a fix item off an unmeasured head would nag on PRs
+                # that may well be clean.
                 echo "${greptile_score}:${now}:${head_sha}" > "$state_file"
                 continue
             fi
--- a/scripts/github/activity-gate.sh
+++ b/scripts/github/activity-gate.sh
@@ -1017,13 +1017,16 @@ check_greptile_scores() {
         local state_file="$STATE_DIR/${repo_safe}-pr-${pr_number}-greptile.state"
         local now
         now=$(date +%s)
-        local last_state last_score last_timestamp last_sha
-        last_state="" last_score="" last_timestamp=0 last_sha=""
+        local last_state last_score last_timestamp last_sha last_verdict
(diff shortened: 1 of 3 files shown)


scripts/github/activity-gate.sh
```diff
@@ -933,6 +933,69 @@ check_merge_conflicts() {
...
+# recorded FOR THE CURRENT HEAD is a measurement, not an inference.
+#
+# Echoes: "dirty" (findings at current head), "clean" (reviewed, none),
+# "pending" (current head not reviewed yet — the sweep will get to it), or
+# "none" (never reviewed).
+ai_review_verdict() {
+    local repo=$1 pr_number=$2 head_sha=$3
+    local marker
+    marker=$(gh api "repos/${repo}/issues/${pr_number}/comments" --paginate \
+        --jq '[.[] | .body // "" | capture("<!-- bob-ai-review (?<j>\\{[^>]*\\}) -->").j] | last // empty' \
```
TimeToBuildBob, line 967: ❌ **P1** — ai_review_verdict() builds `marker` with `[.[] | .body // "" | capture(...).j] | last // empty`, where every comment contributes an element — non-marker comments inject null. `last` therefore returns the FINAL comment's value, not the last non-null capture. The moment any comment is posted after the reviewer's marker — the agent's own reply to the review, or a maintainer note — `last` is null, `// empty` yields an empty string, and the function returns "none" via `[ -z "$marker" ]`. check_greptile_scores() treats "none" exactly like "clean" (line 1087: writes state, continues), so a Greptile 5/5 PR with outstanding AI findings and green CI stays eligible for `merge_ready` — the precise false-clean this PR exists to eliminate. For single-page comment lists, which is most PRs, this is the default outcome whenever the last comment is not a marker, contradicting the stated intent of using "the LATEST REVIEW AT THE CURRENT HEAD". Filter nulls before `last` so the most recent actual marker is used.

scripts/github/activity-gate.sh
```diff
@@ -1003,11 +1066,27 @@ check_greptile_scores() {
...
         if [ "$greptile_score" -ge 5 ] 2>/dev/null; then
-            echo "${greptile_score}:${now}:${head_sha}" > "$state_file"
-            continue
+            local ai_verdict
+            ai_verdict=$(ai_review_verdict "$repo" "$pr_number" "$head_sha")
+            if [ "$ai_verdict" = "dirty" ]; then
+                # Fall through to the fix arm below. Scored as 3 so it routes to
+                # greptile_needs_fix rather than _needs_improvement: unaddressed
+                # findings are work to do, not a polish suggestion.
+                greptile_score=3
```
TimeToBuildBob: ⚠️ **P2** — Routing a dirty AI verdict at the Greptile-5 short-circuit mutates the shared score: the synthetic `greptile_score=3` assigned at line 1079 is later persisted by the common emit path at line 1118 (`echo "${greptile_score}:${now}:${head_sha}" > "$state_file"`). On the next run with an unchanged head within the 3600s TTL, the cache-hit branch reads `greptile_score="$last_score"` = 3 (lines 1035-1037), masking the real API score of 5; the `-ge 5` branch is never reached and every consumer of this state file — check_own_pr_review_state and check_merge_ready — sees "Greptile 3/5". The gate re-emits `greptile_needs_fix` with the false detail "Greptile score: 3/5" at each cooldown expiry, and check_own_pr_review_state emits a spurious "own-PR review: Greptile 3/5" item, misreporting Greptile's actual score until a head change or TTL expiry forces a real refetch. Keep the real Greptile score in the state file for consumers and route decision with a separate variable.

scripts/github/activity-gate.sh
```diff
@@ -1003,11 +1066,27 @@ check_greptile_scores() {
...
-        # Score 5 = clean — update state file (so check_merge_ready sees
-        # the perfect score instead of a stale sub-5 entry) and skip.
+        # Score 5 = clean by Greptile's reckoning — but OUR reviewer may still
+        # have open findings on the same PR, and those are invisible to the
+        # score above. A PR is only clean when both reviewers are satisfied.
         if [ "$greptile_score" -ge 5 ] 2>/dev/null; then
-            echo "${greptile_score}:${now}:${head_sha}" > "$state_file"
-            continue
+            local ai_verdict
+            ai_verdict=$(ai_review_verdict "$repo" "$pr_number" "$head_sha")
```
TimeToBuildBob: ⚠️ **P2** — check_greptile_scores() invokes ai_review_verdict() for every PR whose Greptile score is >= 5, whether that score came from the live API or from a fresh state-cache hit (lines 1035-1037), and the new function performs its own uncached `gh api repos/${repo}/issues/${pr_number}/comments --paginate` with no TTL or cooldown of its own. The clean/"none" paths write only score, timestamp and SHA to the state file (line 1087) — never the verdict — so a freshly written cache entry does not suppress the refetch; the next 2-minute cycle hits the comments endpoint again for the same PR. A repo holding ten 5/5 PRs therefore adds ten paginated comments fetches per cycle (~300 REST calls/hour per repo), directly contradicting the script's documented API-budget discipline and the rate-limit regression history in the header. Store the verdict in the greptile state file so the comments endpoint is queried at most once per fetch_cache_ttl.

TimeToBuildBob:
## 🤖 AI code review

In check_greptile_scores() in scripts/github/activity-gate.sh, adds an ai_review_verdict() helper that reads the latest bob-ai-review marker from a PR's issue comments and resolves it to dirty/clean/pending/none. When Greptile scores 5/5, the gate now also consults our own reviewer and routes a dirty verdict to greptile_needs_fix via a separate route_score (cyclonic variable), persisting a new 4th verdict field in the greptile state file while keeping the real Greptile score. check_merge_ready() now blocks merge items when the greptile state's verdict field is 'dirty'. Tests were updated to exercise the new verdict caching and a static-tests script was added that asserts the score is never overwritten.

No blocking defects on this head, but 2 P2 issues worth resolving first.

**2 findings** · ⚠️ **2** P2 · 🔒 **1** security — none blocking, so the findings below are advisory.

**⚠️ P2 medium · 🔒 security** — `scripts/github/activity-gate.sh:1095`

The verdict cache check at the top of the `[ "$greptile_score" -ge 5 ]` branch reuses the persisted verdict for up to fetch_cache_ttl (3600s) whenever `last_sha == head_sha` and the fetch timestamp is fresh, with no way to notice that a newer marker comment was posted on the same head. The PR's own design acknowledges manual re-reviews (`@greptileai` triggers complete within the 30-min TTL per the comment at line 1043), and a same-head re-review that changes the verdict (e.g. pending → dirty, or clean → dirty) will not be observed until the TTL expires. During that window a PR previously cached as 'pending' or 'clean' remains eligible: check_merge_ready only blocks the literal 'dirty', so it emits merge_ready for a PR whose HEAD now has open findings. The cached timestamp is only refreshed when a fetch actually happens, so a full hour can elapse before the new verdict surfaces.

<sub>**How this was verified:** Traced both branches: on a cache hit `ai_verdict="$last_verdict"` is taken without hitting gh, so a same-head marker change is invisible until the TTL. check_merge_ready's `continue` is keyed only on the literal 'dirty', so 'pending'/'clean' both pass the merge gate during that interval.</sub>

**⚠️ P2 medium** — `tests/test_activity_gate_greptile_score_not_poisoned.py:120`

The new static test test_merge_ready_is_blocked_by_a_dirty_ai_verdict() only greps for the strings `cut -d: -f4` and `= "dirty"` in the shell source; it never executes check_merge_ready to prove the behavioral claim in its docstring (that a dirty verdict suppresses merge_ready). There is no end-to-end test that builds a greptile.state with the 4th field 'dirty' plus a CLEAN/MERGEABLE PR and asserts the gate does NOT emit merge_ready, nor one asserting that 'pending'/'clean'/'none'/'' do NOT trigger the block. The functional merge-block new code (lines 1423–1431) is therefore covered only by textual, not behavioral, evidence; a future refactor that changes the second field of the predicate while keeping the two substrings would still pass this test. The cache-vs-verdict interaction in the same-head manual re-review case (finding above) is likewise untested.

<sub>**How this was verified:** Read the entire file: every test in it filters `body` via regex and asserts on exposed substrings; no test invokes check_merge_ready with a populated state file, so the merge block is asserted only syntactically.</sub>

</details>

- `scripts/github/activity-gate.sh` — Adds ai_review_verdict(), consults the AI reviewer when Greptile is 5/5, persists a 4th verdict field, preserves fetch time in state writes, and blocks merge_ready on a dirty verdict.
- `tests/test_activity_gate_cache.py` — Extends the fake gh stub with Greptile/AI review scores and adds a test that a dirty AI verdict routes to fix without poisoning the cached Greptile score.
- `tests/test_activity_gate_greptile_score_not_poisoned.py` — New file of static assertions that greptile_score is never assigned a literal, routing uses route_score, state writes carry the verdict and fetch time, and merge_ready reads the verdict.

</details>

<sub>Reviewed `bd8c0dd62455` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 255s · advisory, a human still decides · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

scripts/github/activity-gate.sh
TimeToBuildBob: False positive — this rests on an incorrect claim about `jq`'s `capture`.

`capture` does **not** inject `null` for a non-matching input; it emits *nothing*
(it is `match`-based, so a failed match produces no output at all). The array
comprehension therefore collects only the marker bodies, and `last` returns the
newest marker regardless of how many plain comments follow it.

Verified directly:

```
$ printf '%s' '[{"body":"no marker here"},{"body":"also nothing"}]' \
  | jq -c '[.[] | .body // "" | capture("<!-- bob-ai-review (?<j>\\{[^>]*\\}) -->").j] | last // empty'
(no output — correctly falls through to "none")

$ printf '%s' '[{"body":"<!-- bob-ai-review {\"sha\":\"OLD\"} -->"},{"body":"plain"},
               {"body":"<!-- bob-ai-review {\"sha\":\"NEW\"} -->"},{"body":"plain2"}]' \
  | jq -c '[.[] | .body // "" | capture("<!-- bob-ai-review (?<j>\\{[^>]*\\}) -->").j] | last // empty'
"{\"sha\":\"NEW\"}"    ← newest marker, with two trailing non-marker comments
```

Worth recording *why* this one slipped through: the reviewer runs diff-only with
no execution, so a claim about a tool's runtime semantics can't be checked
against the tool. This is the class that read-only/executing tools are meant to
close (ErikBjare/bob#1122).

TimeToBuildBob:
@greptileai review

scripts/github/activity-gate.sh
```diff
@@ -1003,24 +1078,58 @@ check_greptile_scores() {
...
+        # `-ge 5` test on the next cycle, so the AI verdict is never consulted again
+        # and the PR cannot recover even after its findings are resolved.
+        local route_score="$greptile_score"
+        local ai_verdict=""
         if [ "$greptile_score" -ge 5 ] 2>/dev/null; then
-            echo "${greptile_score}:${now}:${head_sha}" > "$state_file"
-            continue
+            # ai_review_verdict() costs a paginated comments fetch, so it obeys the
+            # same TTL as the score above rather than firing every 2-minute cycle.
+            if [ -n "$last_verdict" ] && [ "$last_sha" = "$head_sha" ] \
```
TimeToBuildBob, line 1095: ⚠️ **P2** — The verdict cache check at the top of the `[ "$greptile_score" -ge 5 ]` branch reuses the persisted verdict for up to fetch_cache_ttl (3600s) whenever `last_sha == head_sha` and the fetch timestamp is fresh, with no way to notice that a newer marker comment was posted on the same head. The PR's own design acknowledges manual re-reviews (`@greptileai` triggers complete within the 30-min TTL per the comment at line 1043), and a same-head re-review that changes the verdict (e.g. pending → dirty, or clean → dirty) will not be observed until the TTL expires. During that window a PR previously cached as 'pending' or 'clean' remains eligible: check_merge_ready only blocks the literal 'dirty', so it emits merge_ready for a PR whose HEAD now has open findings. The cached timestamp is only refreshed when a fetch actually happens, so a full hour can elapse before the new verdict surfaces.

tests/test_activity_gate_greptile_score_not_poisoned.py
```diff
@@ -0,0 +1,133 @@
...
+    assert writes, "expected greptile state writes"
+    stale = [w for w in writes if "${now}" in w]
+    assert not stale, (
+        f"state writes stamping $now: {stale} — on a cache hit this refreshes the "
+        "TTL every cycle and the verdict is never refetched"
+    )
+    assert all("${fetched_at}" in w for w in writes), writes
+
+
+def test_merge_ready_is_blocked_by_a_dirty_ai_verdict() -> None:
```
TimeToBuildBob, line 120: ⚠️ **P2** — The new static test test_merge_ready_is_blocked_by_a_dirty_ai_verdict() only greps for the strings `cut -d: -f4` and `= "dirty"` in the shell source; it never executes check_merge_ready to prove the behavioral claim in its docstring (that a dirty verdict suppresses merge_ready). There is no end-to-end test that builds a greptile.state with the 4th field 'dirty' plus a CLEAN/MERGEABLE PR and asserts the gate does NOT emit merge_ready, nor one asserting that 'pending'/'clean'/'none'/'' do NOT trigger the block. The functional merge-block new code (lines 1423–1431) is therefore covered only by textual, not behavioral, evidence; a future refactor that changes the second field of the predicate while keeping the two substrings would still pass this test. The cache-vs-verdict interaction in the same-head manual re-review case (finding above) is likewise untested.

PR #1397, opened 2026-08-10
fix(runloops): roll back state when a PM slot delivers no reply

TimeToBuildBob:
## The bug

The run-item executor ends every item with an **unconditional** state promotion:

```python
# 8. State promotion (worker.sh:661-663)
```

The bash it cites promotes *conditionally* (worker.sh:661-676): on `orphan_no_delivery` — a session that exited without posting a thread reply — it calls `rollback_failed_delivery()` instead. That branch was dropped in the phase-2 port (#1273, 2026-07-11).

Worse, the bash `rollback_failed_delivery()` only landed **2026-08-05**, by which point PM had already moved both lanes to this executor (`PM_EXECUTOR=runloops`, `PM_EXECUTOR_LANES=slow,fast`; slow flipped 07-11, fast 07-28). So the fix was written into a code path PM no longer runs. **PM has had no delivery-failure retry at all.**

## What that costs, per failed delivery

1. Pending activity-gate state is consumed → the gate restarts its 1h cooldown as if the item were handled.
2. The dispatcher's **launch-stamped** `.event` fingerprint is left in place → subsequent cycles skip the item as `event_unchanged` for the 6h TTL.
3. No redelivery counter → nothing bounds or retries the failure.

The net effect is that a crashed session is indistinguishable from a successful one, and the item goes quiet for 6h with nobody having replied.

## Live bite

`gptme/gptme#3468` emitted `greptile_needs_fix` at 13:31 UTC on 2026-08-10. The slot session died at 13:39 with exit 1 and no reply:

```
WARN: PM delivery post-condition FAILED — gptme/gptme#3468: session exited without a thread reply
=== Item 1 complete ===
```

Note what is *absent* between those two lines: the bash would have logged `rolling back state for re-emission`. State was promoted anyway, the `.event` marker survived, and the PR sat undispatched with three unresolved review threads — the exact condition the item was raised for. Next cycle logged `skipped_unchanged / event_unchanged key=gptme/gptme#3468`.

Ledger scale: 508 `orphan_no_delivery` rows in `state/monitoring-response-latency.jsonl`, 312 of them since the 08-05 bash fix that never ran.

## The fix

Ports the missing half of worker.sh:661-676 / lib.sh:946-995:

- **`rollback_failed_delivery()`** — leaves pending state un-promoted, clears `.event`/`.event_logged`, drops notification state mapped to `repo#number`, and bounds retries with `PM_MAX_REDELIVERY_ATTEMPTS` (default 2) so a genuinely reply-less item (escalated PR awaiting human review) stops burning a slot every cycle.
- **`promote_item_state()`** — resets that counter, as the bash does, so a later genuine failure gets a full budget.
- **Lock-busy path** — also clears the event marker (lib.sh:1011-1027), which the port had likewise dropped. Without it a fresh human event arriving *during* a long slot session is consumed and suppressed for 6h (the ErikBjare/bob#1127 bite).

Slot key resolves from `PM_SLOT_KEY` as in bash, falling back to `repo#number` (what `derive_slot_key` builds for PR/issue items) rather than silently skipping the marker clear if the env var is absent.

## Tests

11 new tests in `test_run_item.py`:

- rollback clears the event markers and leaves pending state un-promoted
- slot key read from `PM_SLOT_KEY`
- only *matching* notification state is dropped
- redelivery cap reached → returns False (caller promotes) and resets the counter
- `PM_MAX_REDELIVERY_ATTEMPTS` honored
- `promote_item_state` resets the counter
- `clear_slot_event_marker("")` is a no-op
- **`run_post_session` integration**: `orphan_no_delivery` rolls back instead of promoting; past the cap it promotes

`835 passed, 1 skipped`.

### DIFF
--- a/packages/gptme-runloops/src/gptme_runloops/run_item.py
+++ b/packages/gptme-runloops/src/gptme_runloops/run_item.py
@@ -614,7 +614,131 @@ class SlotLock:
             pass
 
 
-# --- State promotion (lib.sh:611-631; p-m.sh:695-704) ---
+# --- State promotion / delivery rollback (lib.sh:894-995) ---
+
+# Per-slot dispatch marker directory, shared with the bash dispatcher
+# (project-monitoring-dispatch.sh) and pm_dispatch.py. The dispatcher stamps
+# ``<slot_safe>.event`` at LAUNCH; a slot that delivers nothing must clear it,
+# otherwise the item is suppressed by the event-unchanged TTL (6h) even though
+# no reply was ever posted.
+DISPATCH_COOLDOWN_DIR_ENV = "PM_DISPATCH_COOLDOWN_DIR"
+DEFAULT_DISPATCH_COOLDOWN_DIR = "/tmp/bob-pm-dispatch-cooldown"
+
+# Bash parity: PM_MAX_REDELIVERY_ATTEMPTS (lib.sh:949).
+MAX_REDELIVERY_ATTEMPTS_ENV = "PM_MAX_REDELIVERY_ATTEMPTS"
+DEFAULT_MAX_REDELIVERY_ATTEMPTS = 2
+
+
+def dispatch_cooldown_dir() -> Path:
+    """Resolve the shared dispatch-marker dir (bash ``PM_DISPATCH_COOLDOWN_DIR``)."""
+    return Path(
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
@greptileai review

PR #1398, opened 2026-08-10
fix(deps): upgrade gptme to latest master

TimeToBuildBob:
## Summary

Advance gptme-contrib's uv.lock from v0.31.0 (April 29, 1020 commits behind) to v0.32.1 (current master). This resolves a 3.5-month lag where CI was testing against an outdated gptme version despite pyproject.toml declaring rolling master.

## Details

The issue: `branch = "master"` in pyproject.toml does NOT continuously roll — uv resolves once and locks the SHA. Without explicit `uv lock --upgrade-package`, the lock remains stale indefinitely.

This PR fixes it and updates the dependency checker output.

## Test Results

- 678/681 tests passing ✓
- 3 pre-existing failures (pytest environment + Flask Python 3.13 compatibility)
- No regressions from the gptme update

## Acceptance Criteria

- [x] uv.lock resolves gptme within lag threshold (was 1020 commits behind, now current)
- [x] contrib CI green against refreshed dependencies
- [x] check-dependency-pins.py reports no lag warning

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -232,20 +232,21 @@ wheels = [
 
 [[package]]
 name = "anthropic"
-version = "0.47.2"
+version = "0.120.2"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "anyio" },
     { name = "distro" },
+    { name = "docstring-parser" },
     { name = "httpx" },
     { name = "jiter" },
     { name = "pydantic" },
     { name = "sniffio" },
     { name = "typing-extensions" },
 ]
-sdist = { url = "https://files.pythonhosted.org/packages/64/65/175bf024bd9866ef96470620e164dcf8c3e0a2892178e59d1532465c8315/anthropic-0.47.2.tar.gz", hash = "sha256:452f4ca0c56ffab8b6ce9928bf8470650f88106a7001b250895eb65c54cfa44c", size = 208066, upload-time = "2025-02-25T16:27:50.472Z" }
+sdist = { url = "https://files.pythonhosted.org/packages/d7/10/4ca013cb166f226bd89e0aeb0fcaff94f45ddf716d4925ce89475d3c587b/anthropic-0.120.2.tar.gz", hash = "sha256:9722efc10c27a30a69f5338ddacdb35bc6a64297a4e4ba729bf83af873d5fb3a", size = 1008421, upload-time = "2026-07-28T17:38:26.986Z" }
 wheels = [
-    { url = "https://files.pythonhosted.org/packages/92/ad/feddd3ed83804b7f05c90b343e2d9df8f4a28028d6820c1a034de79dcdab/anthropic-0.47.2-py3-none-any.whl", hash = "sha256:61b712a56308fce69f04d92ba0230ab2bc187b5bce17811d400843a8976bb67f", size = 239536, upload-time = "2025-02-25T16:27:48.999Z" },
+    { url = "https://files.pythonhosted.org/packages/63/af/0f5db57b9397a0f3b7fc204cbef143401a7cadaf982330f97f1ce3d39f34/anthropic-0.120.2-py3-none-any.whl", hash = "sha256:0f0bc2b381dc0eb41c8d886b815d79c2041cd2374f83aed36f574b6dc9c579c1", size = 1022851, upload-time = "2026-07-28T17:38:25.466Z" },
 ]
 
 [[package]]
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
Independent verification of this change from a separate worktree — I was working the same problem in parallel before finding this PR, so this is a genuine second pass rather than a re-read.

**The lock diff is byte-identical.** I ran `uv lock --upgrade-package gptme` independently against `origin/master` (2fd82063) and `git diff` between my `uv.lock` and this branch's is empty. Same resolution: gptme `0.31.0 (771528ed`, 2026-04-29) → `0.32.1 (b34ee9b2`, today's master), 1020 commits.

**Transitive floors moved a lot more than the gptme bump suggests**, worth calling out for reviewers since none of it is visible in the PR title:

| package | before | after |
|---|---|---|
| openai | 1.109.1 | 2.53.0 (**major**) |
| anthropic | 0.47.2 | 0.120.2 |
| rich | 13.9.4 | 15.0.0 |
| click | 8.3.1 | 8.4.2 |
| mcp | 1.22.0 | 1.29.0 |
| json-repair | 0.32.0 | 0.61.7 |

plus ipython, tiktoken, platformdirs, requests, pick, python-dotenv. `lxml` and `pypdf` are dropped entirely.

**Baseline-vs-upgraded diff.** I ran `make test-packages` on unmodified `origin/master` (old lock) and on the upgraded lock, so upgrade breakage could be separated from pre-existing breakage:

- baseline (old lock): 14 failures/errors
- upgraded: 17 failures/errors
- newly appearing: 2, both in `gptodo` (`test_pool_filter.py::test_next_pool_frontier_returns_none_when_no_frontier`, `test_ready_selection.py::test_ready_command_skips_waiting_for_tasks`)

**All 17 are environmental, including the 2 new ones — none are caused by the gptme upgrade.** Two independent local causes, both now root-caused:

1. *Corrupted local uv cache archives.* The venv's `click-8.4.2.dist-info/METADATA` says `8.4.2` but the installed payload is click **8.1.8** source (`__version__ = "8.1.8"`, `parser.py` still present, 8.1-shape `Result.output`). Under click 8.1 `CliRunner`, `mix_stderr` defaults to `True`, so the test helpers' `try: CliRunner(mix_stderr=False) / except TypeError` guard silently takes the wrong branch and `result.stderr` comes back empty — which is exactly the two "new" gptodo failures and the nine pre-existing `gptme-rag` `test_cli_search` JSONDecodeErrors. The same cache defect explains the `TypeError: type 'ConfigAttribute' is not subscriptable` collection errors in `gptmail/tests/test_oauth.py` and `gptme-dashboard/tests/test_server.py`: the installed flask mixes files from two versions, with a non-generic `ConfigAttribute` in `config.py` and a `ConfigAttribute[bool]` subscript in `sansio/app.py`.

   Reinstalling both with `uv pip install --no-cache --reinstall click==8.4.2 flask==3.1.3` yields the correct sources (`class ConfigAttribute(t.Generic[T])`, modern click), after which **`gptodo` + `gptmail` + `gptme-dashboard` = 954 passed, 0 failed**, and `gptme-rag/tests/test_cli_search.py` = 10 passed. So this is a local cache defect, not a Flask/Python-3.13 incompatibility — worth correcting in the PR description, since "Flask Python 3.13 compatibility" would send the next person chasing an upstream bug that does not exist.

2. *`FORCE_COLOR=3` in the local shell*, which makes rich emit ANSI into captured output and breaks four plain-substring assertions in `gptodo` (`test_explain_readiness.py` ×2, `test_state_transitions.py` ×2). These fail identically on the **old** lock, so they are pre-existing and CI-invisible.

**`make typecheck-packages` passes clean** against the upgraded gptme — zero mypy errors across all packages. That was the most likely place for a 1020-commit jump to surface (changed signatures on the ~12 gptme modules contrib imports: `gptme.config`, `gptme.message`, `gptme.llm`, `gptme.llm.models`, `gptme.prompts`, `gptme.logmanager`, `gptme.init`, `gptme.dirs`, `gptme.chat`, `gptme.tools`, `gptme.hooks`, `gptme.telemetry`). Nothing broke.

So the real blast radius of 3.5 months / 1020 commits of gptme is **zero** — no contrib code needed adapting. LGTM.

**One follow-up worth opening separately** (not a blocker here): `branch = "master"` will silently freeze again the moment this merges, because uv only re-resolves a git-branch dep on an explicit `--upgrade-package`. Two candidate guards:

- A scheduled CI job that runs `uv lock --upgrade-package gptme` weekly and opens a PR, so the lag is capped at a week and shows up as reviewable work rather than as silent rot.
- A lag check in pre-commit/CI. Note that Bob's `scripts/check-dependency-pins.py` does **not** port cleanly for this: it answers "how many commits behind?" from a local clone of the dependency used as an ancestry oracle, discovered via the superproject's submodules and `projects/`. Run from inside gptme-contrib there is no gptme checkout to find, and it reports the pyproject dep as `rolling` with no lag finding at all — it only flagged this bug when run from the parent workspace. Making it work in contrib CI means either cloning gptme in the job or rewriting the lag check around `git ls-remote`, which is more than a copy.

uv.lock
```diff
@@ -1450,23 +1460,21 @@ provides-extras = ["oauth", "test"]
 
 [[package]]
 name = "gptme"
-version = "0.31.0"
-source = { git = "https://github.com/gptme/gptme.git?branch=master#771528ed806cb3bfcfeb263fa1230e042c27191c" }
+version = "0.32.1"
```
TimeToBuildBob, line 1463: ⚠️ **P2** — The gptme bump from 0.31.0 to 0.32.1 removes `lxml` and `pypdf` from gptme's dependency list, so this lockfile deletes both [[package]] blocks and they no longer resolve into any environment installed from this tree. Previously these two libraries were present transitively, meaning any plugin or package in this repo that does a top-level `import lxml` or `from pypdf` without declaring them in pyproject.toml was relying on that transitive availability. After this upgrade such an import fails with ModuleNotFoundError at import time, breaking the plugin's `setup()`/registration as soon as it is loaded. The diff contains no pyproject.toml edits and no new import-smoke tests over the plugin packages, so nothing verifies that the dropped dependencies are unused. Grep the tree for `import lxml` and `from pypdf` and add explicit dependency declarations wherever they appear before merging.

TimeToBuildBob:
## 🤖 AI code review

The PR advances uv.lock from gptme 0.31.0 to 0.32.1 by pointing the git source at a new master SHA (b34ee9b2), which updates the resolved openai, anthropic, rich, mcp, ipython, tiktoken, click, requests and platformdirs packages along with their transitive deps, drops lxml and pypdf from the resolution, and adds docstring-parser as a dependency of anthropic. No pyproject.toml or source files are changed; only the lockfile is refreshed.

**⚠️ P2 medium** — `uv.lock:1463`

The gptme bump from 0.31.0 to 0.32.1 removes `lxml` and `pypdf` from gptme's dependency list, so this lockfile deletes both [[package]] blocks and they no longer resolve into any environment installed from this tree. Previously these two libraries were present transitively, meaning any plugin or package in this repo that does a top-level `import lxml` or `from pypdf` without declaring them in pyproject.toml was relying on that transitive availability. After this upgrade such an import fails with ModuleNotFoundError at import time, breaking the plugin's `setup()`/registration as soon as it is loaded. The diff contains no pyproject.toml edits and no new import-smoke tests over the plugin packages, so nothing verifies that the dropped dependencies are unused. Grep the tree for `import lxml` and `from pypdf` and add explicit dependency declarations wherever they appear before merging.

<sub>**How this was verified:** Confirmed in the diff that `{ name = "lxml" }` and `{ name = "pypdf" }` are removed from gptme's dependencies in this PR, and the corresponding registry [[package]] sections for lxml-6.0.2 and pypdf-6.10.2 are deleted, dropping both from the resolved closure.</sub>

</details>

- `uv.lock` — Bumps the gptme git dependency from 0.31.0 to 0.32.1 at a new SHA and refreshes its transitive closure: openai 1.109.1→2.53.0, anthropic 0.47.2→0.120.2, rich 13.9.4→15.0.0, mcp 1.22.0→1.29.0, ipython 8.37→8.39, tiktoken 0.12→0.13, click, requests, platformdirs and more are updated; docstring-parser

</details>

<sub>Reviewed `84d8599a3d54` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 435s · advisory, a human still decides · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

PR #1400, opened 2026-08-10
feat(runloops): derive PM dispatch outcome from exit status and observable effect

TimeToBuildBob:
## The failure class

A PM worker could exit non-zero while its dispatch ledger row said `completed`. The ledger reported success, so nothing retried, nothing alerted, and the miss was invisible.

Worse: a worker can exit **0**, narrate success, and change nothing on GitHub. On `gptme/gptme#3468` the worker read the review findings, made the fix, and committed it; then every `git push` was rejected by a pre-push guard. From the outside that is indistinguishable from "PM never ran", and the PR was reported untouched all day.

## The invariant

> A dispatch's recorded outcome is derived from its **exit status** and its **observable external effect**, never asserted independently of them.

`phase="completed"` only ever meant "the item loop returned".

## Part 1 — the invariant

- **`pm_dispatch.derive_dispatch_outcome()`** is the single choke point. The ledger row carries `exit_code`, `effect`, and the derived `outcome` (`succeeded`/`failed`/`no_effect`/`null`). The schema change is **additive only**, guarded by `test_legacy_keys_are_not_dropped`.
- **Absence of evidence never yields `succeeded`.** No signals → `null`. Effect unknown → `unknown`, not `observed`.
- **`worker_records.derive_effect_signal()`** reads the before/after PR snapshot the worker *already fetches*, so the effect check costs **no extra API call**. A rejected push leaves the head OID unmoved — exactly how the #3468 shape is caught.
- `"handled"` is the permissive **default** delivery outcome when no check runs, so it is fed to the effect signal only when a check actually ran. Otherwise every unchecked session would report observed effect — re-creating the very lie this prevents. Regression test included.

The bash executor passes `exit_code` **empty** on purpose: it has no aggregate worker exit status (`EXIT_CODE=0` regardless of item failures), and fabricating a `0` would assert success. Changing its process-exit semantics would touch the crash breaker and systemd, so it is left alone and noted as a known gap.

## Part 2 — completing the rollback port (P1/P2/P5/P7)

A divergence sweep against the bash worker found seven behaviours the port froze at an older line range. The first is the important one:

**The rollback merged in #1397 was only half-ported and was effectively a no-op.** It counted redelivery attempts and skipped `promote_item_state`, and stopped there. Two omitted effects each independently defeat it:

- the slot's `.event` fingerprint stayed stamped, suppressing the item for 6h;
- pending `notif-*` state stayed in place, and `promote_notification_states()` at end of run promotes *every* pending `notif-*.state` — consuming exactly what the rollback claimed to preserve.

Also fixed:

- **P5** — a lock-busy launch now clears the slot's event fingerprint. Highest-harm item in the sweep: a human mention arriving during a long slot session was consumed and suppressed for 6h, and unlike a bot signal nothing retries it — the person assumes they were seen (live bite 2026-08-05, ErikBjare/bob#1127).
- **P2** — `promote_item_state()` resets the redelivery counter; the budget was degrading monotonically.
- **P7** — the counter moves to the cooldown dir under the bash filename so both executors share it (it was `workspace/.pm-redelivery-attempts`, giving an item up to 2x its budget while both paths serve traffic), and the cap honours `PM_MAX_REDELIVERY_ATTEMPTS`.
- **P3 (partial)** — the `greptile_convergence_adjudication` investigate arm and its 1500s tier. The converged-backoff early exit remains unported (needs greptile-helper/convergence hooks the Python executor lacks).

### ⚠️ Upstream bash bug found while porting — not fixed here

The dispatcher stamps the marker with `tr '/#:' '---'` (`project-monitoring-dispatch.sh:563`), but **both** bash rollback paths strip only `/` and `#` (`lib.sh:985-986`, `:1031-1032`). For any slot key containing a colon — **every CI-check slot** — bash computes a path the dispatcher never wrote, its `rm -f` silently no-ops, and the item stays suppressed for the full 6h TTL. `clear_slot_event_markers()` clears every spelling rather than picking a side, but **the bash side is still broken** and needs its own fix.

## Verification

Verified by symptom, not by test-pass — a passing test is precisely what let the half-port through.

**Outcome invariant**, end-to-end through `run_work_file`:
- worker exits **1** → `outcome=failed`
- worker exits **124** with `failures=0` (timeouts are deliberately not counted as failures, so this row was previously indistinguishable from a clean run) → `outcome=failed`
- worker exits **0** and the PR head never moves → `effect=none`, `outcome=no_effect`

**Rollback**, with real files on disk, asserting state *after* `promote_notification_states()`:

| | old half-port | fixed |
|---|---|---|
| `.event` marker | **still stamped** (6h suppression) | cleared |
| pending notif state | **promoted → consumed** | purged, never promoted |
| unrelated item's notif state | promoted | promoted (unchanged) |

The negative control matters: the old shape failed both dimensions, so the check discriminates rather than passing vacuously.

Live production ledger, after deploy:
```
17:12:08  exit=0 fail=0 effect=observed  outcome=succeeded  [gptme/gptme-contrib#1393]
17:12:39  exit=0 fail=0 effect=none      outcome=no_effect  [gptme/gptme#3516]
```

Suite: **896 passed**, mypy clean.

## Measured impact

**704 of 1,609 completed dispatches (43.8%) shipped nothing observable**, over 2026-07-30 → 08-10. Detection of that class only starts 2026-07-30 (the `skipped_unchanged` phase did not exist before), so this is *not* an all-time figure. Companion, over full history: 834 of 11,574 completed rows (7.2%) carried failure evidence with no derived outcome.

Reader side and the full write-up (with denominators and caveats) live in Bob's workspace: `scripts/monitoring/pm-outcome-integrity.py`, `knowledge/technical/pm-dispatch-outcome-integrity.md`.

## Removed

`outcome_verification.py` — dead code, zero callers, superseded by `derive_effect_signal()` which reuses the already-fetched snapshot instead of making an extra `gh` call, and additionally catches the rejected-push case.


packages/gptme-runloops/src/gptme_runloops/run_item.py
```diff
@@ -1790,9 +1808,71 @@ def run_post_session(
...
+
+        try:
+            attempts = int(
+                (attempts_file.read_text() if attempts_file.exists() else "0").strip()
+                or "0"
+            )
+        except (ValueError, OSError):
+            attempts = 0
+
+        attempts += 1
```
TimeToBuildBob: ⚠️ **P2** — The redelivery attempts counter is never reset on a successful delivery. In run_post_session step 8, when delivery_outcome is 'orphan_no_delivery' the counter file `.pm-redelivery-attempts/{repo}-{number}` is read, incremented, and written; when attempts exceeds max_attempts (2) the file is unlinked and state promoted. The non-orphan else branch (including a 'handled' delivery) only calls promote_item_state and never clears the file, so the counter is a lifetime total, not a consecutive-streak count: orphan -> handled -> orphan -> orphan reaches 3 and triggers the give-up promote even though the handled run in between shows redelivery can work. Consequence: an item whose thread still genuinely needs one more redelivery is promoted as 'no reply is likely correct here' and stops being dispatched, leaving an unanswered reply permanently unaddressed. The success branch should delete the attempts file.

packages/gptme-runloops/src/gptme_runloops/outcome_verification.py
```diff
@@ -0,0 +1,171 @@
...
+                changed=True,
+                reason="gh_call_failed",
+                started_at=started_at,
+                checked_at=checked_at,
+            )
+
+        data = json.loads(proc.stdout)
+
+        # Parse updatedAt timestamp
+        updated_str = data.get("updatedAt", "")
```
TimeToBuildBob: ⚠️ **P2** — outcome_verification.py is dead scaffolding: check_outcome_changed and append_outcome_verification_ledger are defined but nothing in the diff imports or calls them. Inside it the fail-safe is inverted: a gh response missing updatedAt returns changed=False reason='no_change' (the only branch that declares no change on missing data), while every other error path (gh non-zero, parse failure, timeout) returns changed=True 'conservative'. If this module is ever wired up as the PR title implies it was meant to be, a gh response without updatedAt (schema change, network truncation) would report the item unchanged and let it be silenced as no-change — the opposite of the conservative direction used everywhere else in the file. As shipped it has no runtime impact, but it is misleading leftover: either remove it or make the missing-updatedAt branch return changed=True to match the rest.

TimeToBuildBob:
## 🤖 AI code review

Adds a Greptile-convergence early-exit (ConvergenceVerdict parsing, a GREPTILE_CONVERGENCE instruction template, and a convergence_cmd I/O hook) so a backoff-state item with nothing left to do skips its session; completes the failed-delivery rollback port by clearing slot event markers, purging pending notif-* state, and deriving the slot key; and adds derived dispatch outcome fields (exit_code, effect, outcome) to the ledger, backed by derive_dispatch_outcome and derive_effect_signal over the before/after PR snapshot. It also registers greptile-convergence in the bandit work types and slow-lane set, adds an adjudication timeout tier, and adds new investigate arms (voice_postcall, erik_decision, greptile_convergence_adjudication).

**3 findings** · ❌ **1** P1 · ⚠️ **2** P2 — 🛑 **1 blocking**, worth resolving before merge.

**⚠️ P2 medium** — `packages/gptme-runloops/src/gptme_runloops/run_item.py:475`

`timeout_tier` grants the 1500s adjudication budget only when the item carries the `greptile_convergence_adjudication` type, but the Phase A2 path in merge_lifecycle spawns adjudication by setting `instructions=InstructionKind.GREPTILE_CONVERGENCE` on an item whose types remain unchanged `("greptile_needs_improvement",)`. In `run_work_file`, `plan_item` calls `timeout_tier(item.types, bool(greptile_fix), config)` with those original types, so the new adjudication branch never fires and the session gets the 900s default instead of the intended 1500s. The session whose job is to classify findings, fix blocking ones, and post a structured recommendation is cut a third shorter than the design in `RunItemConfig.adjudication_timeout` describes; the 1500s tier applies only to an item type the Phase A2 path never produces.

<sub>**How this was verified:** Checked run_merge_lifecycle Phase A2 (merge_lifecycle.py:770-771) which sets result.instructions without altering item.types, and plan_item's timeout_tier call using item.types.</sub>

**⚠️ P2 medium** — `packages/gptme-runloops/src/gptme_runloops/prompt_templates.py:804`

The voice-postcall verification command contains a literal `<stem>` token: `_substitute` replaces only `{name}` tokens and `ItemPromptParams._tokens()` has no "stem" key, so `<stem>` survives substitution verbatim. Running `grep '<stem>'` against the post-call-events TSV matches no line, `tail -3` consequently prints nothing, and the mandated "verify the trace ledger shows a terminal row" step can never succeed as written. Every voice_postcall session is instructed to perform a check that is impossible, so the worker either re-runs the idempotent script hunting for a state it already wrote or silently reports an unverified ledger row — the verification step is dead on arrival.

<sub>**How this was verified:** Confirmed _substitute uses _TOKEN_RE matching only {[a-z0-9_]+} and _tokens() contains no stem key; &lt;stem&gt; is outside both, so it is never substituted.</sub>

</details>

### Comments outside the diff

**❌ P1 high** — [`packages/gptme-runloops/src/gptme_runloops/run_item.py:1818`](https://github.com/gptme/gptme-contrib/blob/c3e249a2eb416afa1791d23f5f6db0a23683a219/packages/gptme-runloops/src/gptme_runloops/run_item.py#L1818)

When the delivery-check subprocess exits non-zero or raises, line 1818 substitutes the fabricated literal `{"outcome":"handled"}` for the checker's actual output. Because `delivery_checked` was set True before the try block, step 9 forwards that "handled" into `read_record_effect_signal`; `derive_effect_signal` maps it directly to EFFECT_OBSERVED. A session whose delivery checker crashed — so no reply, push, or state change was ever verified — is therefore recorded as having observable effect, and with a clean worker exit derives outcome=succeeded. This recreates the exact #3468 failure class this PR exists to catch: a worker that pushed and posted nothing is marked successful whenever the checker process is down, because the branch fabricates the very "handled" default the rest of the change treats as the lie.

<sub>**How this was verified:** Traced run_post_session: line 1797 sets delivery_checked=True before the try; the non-zero/OSError branches both land on the forced handled literal; line 2076 passes delivery_outcome when delivery_checked is True; worker_records.derive_effect_signal returns EFFECT_OBSERVED for delivery=="handled".</sub>

<details><summary><b>Files changed (13)</b> — the diff as I read it</summary>

- `packages/gptme-runloops/src/gptme_runloops/merge_lifecycle.py` — Adds ConvergenceVerdict.from_json, greptile_convergence_applicable, decide_greptile_convergence, the Phase A2 early exit, and convergence_verdict on the I/O protocol and subprocess adapter.
- `packages/gptme-runloops/src/gptme_runloops/pm_bandit.py` — Adds "greptile-convergence" to PM_WORK_TYPES.
- `packages/gptme-runloops/src/gptme_runloops/pm_dispatch.py` — Adds greptile_convergence_adjudication to SLOW_LANE_TYPES, derive_dispatch_outcome with exit_code/effect on ledger entries, and an --exit-code/--effect CLI path.
- `packages/gptme-runloops/src/gptme_runloops/prompt_templates.py` — Adds the GREPTILE_CONVERGENCE skeleton, an adjudication timeout-aware ItemPromptKind arm, and voice_postcall / erik_decision arms.
- `packages/gptme-runloops/src/gptme_runloops/run_item.py` — Reworks slot-key/cooldown helpers, completes rollback_failed_delivery with marker/notif purging, passes through convergence verdict, returns an effect signal from run_post_session, and writes derived exit_code/effect to the ledger.
- `packages/gptme-runloops/src/gptme_runloops/run_item_config.py` — Adds convergence_cmd assembly and makes the merge-lifecycle env use inherited WORKSPACE_REPO/SELF_MERGE_ALLOWED_PATHS when present.
- `packages/gptme-runloops/src/gptme_runloops/worker_records.py` — Adds derive_effect_signal and read_record_effect_signal reading PR head/state/merge deltas and the delivery outcome.
- `packages/gptme-runloops/tests/goldens/prompt_templates/greptile_convergence.alice.txt` — Golden of the rendered GREPTILE_CONVERGENCE instruction for the Alice agent.
- `packages/gptme-runloops/tests/goldens/prompt_templates/greptile_convergence.bob.txt` — Golden of the rendered GREPTILE_CONVERGENCE instruction for Bob.
- `packages/gptme-runloops/tests/test_pm_dispatch.py` — Adds adjudication lane/work-type tests and updates the ledger schema-key expectations to include exit_code/effect/outcome.
- `packages/gptme-runloops/tests/test_pm_dispatch_outcome.py` — New tests for derive_dispatch_outcome, derive_effect_signal, and ledger/CLI forwarding of exit_code and effect.
- `packages/gptme-runloops/tests/test_run_item.py` — Updates rollback call sites to pass slot_key and adds symptom tests for derived outcome and no-effect-on-clean-exit.
- `packages/gptme-runloops/tests/test_run_item_rollback.py` — New end-state tests for the rollback surviving end-of-run promotion, event-marker spellings, and lock-busy marker clearing.

</details>

| `19531528de2e` | 4/5 | 3 | llm | 2026-08-10 17:14 UTC |
| `07ea9f077ff0` | 2/5 | 2 | llm | 2026-08-10 18:01 UTC |
| `a5b79048c6d6` | 3/5 | 2 | llm | 2026-08-10 18:41 UTC |

</details>

<sub>Reviewed `c3e249a2eb41` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 178s · advisory, a human still decides · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

TimeToBuildBob:
@greptileai review

packages/gptme-runloops/src/gptme_runloops/run_item.py
```diff
@@ -741,6 +761,203 @@ def rollback_failed_delivery(
...
+            try:
+                path.unlink()
+            except OSError:
+                pass
+        purged += 1
+    return purged
+
+
+def rollback_failed_delivery(
+    config: RunItemConfig, repo: str, number: int | str | None, slot_key: str
```
TimeToBuildBob, line 809: ❌ **P1** — The new `rollback_failed_delivery` (run_item.py:918) requires a positional `slot_key` with no default, shadowing the earlier version at line 688 that accepted `slot_key=None` and fell back to `PM_SLOT_KEY` (or `repo#number`). Existing callers that invoke it with three arguments — e.g. `test_run_item.py:1488` (`rollback_failed_delivery(config, "gptme/gptme", 3468)`) — now raise `TypeError: rollback_failed_delivery() missing 1 required positional argument: 'slot_key'`. Any production call site that relied on the old env fallback (the bash `local slot_key=${3:-${PM_SLOT_KEY:-}}` contract) breaks too. This is an API-contract violation: the new signature should keep the default and fallback, or every caller must be updated (none outside the new call site were).

packages/gptme-runloops/src/gptme_runloops/run_item.py
```diff
@@ -1922,30 +2037,52 @@ def run_post_session(
...
         promote_item_state(config, item.repo, item.number)
+
+    # 9. Observable-effect signal.
+    # A clean exit is not evidence the work landed: a worker can commit a fix
+    # and have every `git push` rejected, exiting 0 with a log that says it
+    # succeeded (gptme/gptme#3468, 2026-08-10). Derived from the before/after
+    # PR snapshot step 2 already fetched, so this costs no extra API call.
+    effect = read_record_effect_signal(
+        record_file,
+        delivery_outcome=delivery_outcome if delivery_checked else "",
```
TimeToBuildBob: ❌ **P1** — run_post_session sets delivery_checked=True before invoking the delivery-check subprocess, and both the OSError handler and the returncode!=0 branch assign the fallback raw='{"outcome":"handled"}' (lines 1818/1820). normalize_delivery_outcome maps that to 'handled', and because delivery_checked is True it is passed to read_record_effect_signal at line 2076; derive_effect_signal returns EFFECT_OBSERVED immediately for delivery=='handled', before reading the PR OID/state fields. A delivery check that crashes (missing/permission-denied script) or exits non-zero is therefore indistinguishable from a check that verified a real reply — the ledger records effect=observed for a session that changed nothing on GitHub, re-creating exactly the 'handled default as evidence' lie the comment at lines 1786-1791 says the delivery_checked flag prevents. No test covers the crashed-check path; test_empty_delivery_outcome_is_not_treated_as_handled only exercises the no-check (delivery_checked=False) branch.

packages/gptme-runloops/src/gptme_runloops/run_item.py
```diff
@@ -449,10 +462,18 @@ def predict_cc_trajectory_path(
...
+    assigned_issue wins over adjudication, which wins over the greptile-fix
+    tier.
+
+    Adjudication reads findings, classifies them, possibly fixes one blocking
+    issue and posts a structured recommendation — it never waits on a Greptile
+    re-review, so it takes the 1500s tier rather than the 2700s fix tier.
+    """
     if "assigned_issue" in types:
         return config.assigned_issue_timeout, config.assigned_issue_time_desc
+    if "greptile_convergence_adjudication" in types:
```
TimeToBuildBob: ⚠️ **P2** — timeout_tier keys the new adjudication budget off the item *type* greptile_convergence_adjudication, but the Phase-A2 early-exit decision (backoff + should_request_review_after_fixes False) routes to adjudication via InstructionKind.GREPTILE_CONVERGENCE without altering item.types — the item stays ('greptile_needs_improvement',). run_work_file plans the session with timeout_tier(item.types, bool(greptile_fix), config), so 'greptile_convergence_adjudication' in types is False, 'pr_update' in types is False, and the branch falls through to default_timeout (900s) instead of the new adjudication_timeout (1500s). A backoff-spawned adjudication session — which must classify findings and may fix a blocking issue — is budgeted at the 10-minute default while the escalation-ledger adjudication item gets 20 minutes. Both adjudication sources render the same GREPTILE_CONVERGENCE prompts, so the tier should key on the instruction kind (or the A2 path should add the adjudication type) to apply the budget consistently.

packages/gptme-runloops/src/gptme_runloops/prompt_templates.py
```diff
@@ -712,6 +750,130 @@ def _tokens(self) -> dict[str, str]:
...
+bash "{workspace}/scripts/runs/voice/post-call.sh" <record-path>
+```
+
+The post-call.sh script handles stub-call detection, silent-call alerting,
+trace ledger updates (run_completed / stub_call_skip), and journal writing.
+It is idempotent — re-running on a completed record is harmless.
+
+After the worker finishes, verify the trace ledger shows a terminal row:
+```bash
+grep '<stem>' "{workspace}/state/voice-calls/post-call-events.tsv" | tail -3
```
TimeToBuildBob, line 804: ⚠️ **P2** — The voice-postcall verification command contains a literal `<stem>` token: `_substitute` replaces only `{name}` tokens and `ItemPromptParams._tokens()` has no "stem" key, so `<stem>` survives substitution verbatim. Running `grep '<stem>'` against the post-call-events TSV matches no line, `tail -3` consequently prints nothing, and the mandated "verify the trace ledger shows a terminal row" step can never succeed as written. Every voice_postcall session is instructed to perform a check that is impossible, so the worker either re-runs the idempotent script hunting for a state it already wrote or silently reports an unverified ledger row — the verification step is dead on arrival.

packages/gptme-runloops/src/gptme_runloops/run_item.py
TimeToBuildBob: Fixed in 6e7839d7. `timeout_tier` now accepts an optional `instruction_kind` keyword arg and routes to the adjudication tier when the instruction kind is `GREPTILE_CONVERGENCE` (added by Phase-A2 early-exit) even when `item.types` still classifies as `greptile_needs_improvement`. The call site in `plan_item` passes the lifecycle instruction kind. Test `test_timeout_tier_instruction_kind_routes_to_adjudication` pins both the backoff path (greptile_needs_improvement + GREPTILE_CONVERGENCE → 1500s) and a negative (different instruction kind → 900s default). 933 passed locally.

packages/gptme-runloops/src/gptme_runloops/prompt_templates.py
TimeToBuildBob: Confirmed real — this is a latent bug in the P6 commit (a5b79048), not introduced by the recent fixes. The `<stem>` token was never wired into `ItemPromptParams._tokens()`, so the verification grep always searches for the literal string `<stem>` and matches zero TSV rows. A separate session is currently working on `prompt_templates.py` (mtime 17:30Z on the active worktree) — to avoid a conflict in the hot repo, the voice_postcall verification-step fix is best landed as a follow-up PR scoped to that single defect, not folded into this PR. The P1 finding on this same review thread was fixed in 6e7839d7 (separate finding, already verified resolved by the AI sweep). Not blocking this merge.

TimeToBuildBob:

**Gate objection (recorded):** refuses on `Greptile review not found` (dark on this repo since `2026-08-09T05:27Z`, active on `gptme/gptme` — cannot clear by waiting). The `CI is not fully green` reason from an earlier sweep was a snapshot artifact — checks were pending at query time. Re-verified: **no failing, no pending** at head `6e7839d7`.

**Reviewer: last summary reads 3/5 against head `c3e249a2`, now stale** — fixes landed after it in `6e7839d7`. No open P0/P1.

**Dispositions — one fixed, one deferred with reasoning:**

- **P2 — adjudication timeout tier missed the backoff path: fixed in `6e7839d7`.** Real bug: the Phase-A2 early exit routes to adjudication via `InstructionKind.GREPTILE_CONVERGENCE` without altering `item.types`, so `timeout_tier` fell through to the 900s default instead of the 1500s adjudication tier — starving exactly the sessions that must classify findings. `timeout_tier` now takes an `instruction_kind` kwarg and routes on it; `test_timeout_tier_instruction_kind_routes_to_adjudication` pins both paths.
- **P2 — literal `<stem>` token in the voice-postcall verification command: confirmed real, deliberately deferred.** It is a **latent bug from the P6 commit `a5b79048`, not introduced here** — `<stem>` was never wired into `ItemPromptParams._tokens()`, so the verification grep matches zero rows and the mandated ledger check can never succeed. Deferred out of this PR because a concurrent session held `prompt_templates.py` and landing it here risked a conflict in the hot repo.

  **Tracked** in `tasks/voice-postcall-stem-token-verification-dead.md` so the deferral is not an ageing-out. That task also proposes the general guard worth having: assert no rendered prompt retains an unsubstituted angle-bracket placeholder, which would have caught this at P6.

**Surviving findings: the `<stem>` P2, deferred and tracked as above.**

PR #1401, opened 2026-08-10
feat(activity-gate): add `@TimeToBuildBob fix` operator trigger

TimeToBuildBob:
## What

`@TimeToBuildBob review` forces a fresh **review** of a PR. This adds its sibling:
`@TimeToBuildBob fix` forces a PM worker to **act on** the findings that review
already posted.

No new dispatch, prompt, lane, or model-routing logic is needed — PM already routes
`greptile_needs_fix` to the slow lane / `greptile-fix` bandit arm with a 2700s
timeout. What was missing was a *maintainer-forced entry point* that emits it when
the gate would otherwise stay quiet: a `clean`/`pending` verdict from our reviewer,
Greptile 5/5, or inside `check_greptile_scores()`'s 1-hour cooldown. A maintainer who
reads the findings and wants them addressed *now* had no way to say so.

## How

`pending_fix_request()` + `check_fix_requests()` in `scripts/github/activity-gate.sh`,
hooked into the per-repo scan after `check_merge_ready`, reading the live PR lane
(180s TTL) so a comment is picked up in minutes.

- **Whole-line, case-insensitive** match, same shape as `ai-review-sweep.py`'s
  `TRIGGER_RE`. Prose that merely mentions the phrase does not fire it.
- **Trust gate**: `author_association` must be `OWNER`/`MEMBER`/`COLLABORATOR`.
  Anyone can comment on a public PR; only maintainers can spend worker budget.
- **Self-trigger guard**: comments authored by `TimeToBuildBob` are ignored, so a
  worker quoting the phrase in its own reply cannot re-arm the trigger.

## Anti-spam

There is a recorded incident where a poll waiting on a reaction that could never
appear posted `@greptileai review` **29 times** on one PR. This guard is built the
other way round:

- **Served-ness is derived from GitHub, never from a state file.** A request is
  pending iff its comment carries no reaction from us. The 👀 reaction *is* the
  watermark — it lives on the same object as the request, so it cannot drift.
- **The reaction is POSTED BEFORE the emit, and a failed POST suppresses the emit.**
  A watermark written afterwards leaves a window in which the next 2-minute cycle
  re-emits. Costs at most one deferred dispatch; the alternative costs a loop.
- **Every API failure reads as "no pending request."** Guards that fail open are how
  that incident happened.

`test_single_trigger_emits_exactly_once_across_runs` replays the reaction run 1
posted into run 2's fixture exactly as GitHub would serve it back, and asserts run 2
emits nothing — asserted on **emitted items**, not on internal state.

## Cost

Detection reads the already-fetched PR payload (`comments` is part of
`fetch_pr_data`'s GraphQL selection), so it costs **zero extra API calls** on PRs
with no trigger comment. Only a comment that literally matches the trigger line
costs a reactions lookup.

## Test plan

- `tests/test_activity_gate_fix_request.py` — 10 tests: single-emit-across-runs,
  trusted request emits once, untrusted association, self-authored trigger, existing
  👀, someone else's 👀 (still pending), reactions-API failure, failed reaction POST,
  inline mention, and a CRLF multi-line body.
- `make test-unit` green apart from a pre-existing, unrelated
  `tests/test_linear_oauth_state.py` flask failure that also fails on `master`.
- `shellcheck -x scripts/github/activity-gate.sh` clean; `prek` clean.

Operator docs (routing table row, latency, receipt semantics) land in Bob's brain
repo: `knowledge/technical/pr-review-systems-map.md` + `tools/README.md`, and the
`ai-review.py` review footer now names both triggers so they are discoverable from
the artifact a maintainer actually reads.

### DIFF
--- a/scripts/github/activity-gate.sh
+++ b/scripts/github/activity-gate.sh
@@ -1302,6 +1302,12 @@ check_own_pr_review_state() {
 # fetch_pr_data's GraphQL selection), so it costs ZERO extra API calls on PRs
 # with no trigger comment. Only a comment that literally matches the trigger
 # line costs a reactions lookup.
+#
+# SCOPE, inherited from that payload: fetch_pr_data() lists `--author $AUTHOR`
+# and drops drafts, so the trigger reaches non-draft PRs authored by the bot —
+# which is the case it exists for (a maintainer asking us to fix our own PR).
+# On someone else's PR it is silently inert; widening it means widening the
+# fetch, which is a separate cost decision.
 
 #: The account the trigger addresses, and whose reactions are the watermark.
 #: Matches ``REVIEWER_LOGIN`` in scripts/github/ai-review-sweep.py so the two
--- a/scripts/github/activity-gate.sh
+++ b/scripts/github/activity-gate.sh
@@ -1361,9 +1361,13 @@ pending_fix_request() {
     # Served-ness: any reaction of ours on the trigger comment. Deliberately
     # broader than the 👀 we post — a superset can only suppress, never spam.
     local ours
+    # Must paginate. If our reaction falls outside the first page, we would read
+    # the comment as unserved and emit again on every cycle — an unbounded
+    # dispatch loop, the exact shape of the 29x @greptileai incident. --slurp
+    # wraps the pages in an outer array, hence the `.[][]` flatten.
     ours=$(gh api "repos/${repo}/issues/comments/${comment_id}/reactions" \
-        -F per_page=100 \
-        --jq "[.[] | select(((.user // {}).login // \"\") == \"$FIX_TRIGGER_LOGIN\")] | length" \
+        --paginate --slurp -F per_page=100 \
+        --jq "[.[][] | select(((.user // {}).login // \"\") == \"$FIX_TRIGGER_LOGIN\")] | length" \
         2>/dev/null) || return 0
     # Empty output means the call failed or returned something unparseable →
     # treat as served, same as a real reaction. Never as pending.
--- a/scripts/github/activity-gate.sh
+++ b/scripts/github/activity-gate.sh
@@ -76,6 +76,13 @@
 #   GitHub notifications clear when marked as read upstream, so old state files
 #   become inert.
 #
+#   Maintainer fix requests: a whole-line `@TimeToBuildBob fix` comment from an
+#   OWNER/MEMBER/COLLABORATOR forces a greptile_needs_fix item for that PR even
+#   when every organic route is quiet (clean AI verdict, Greptile 5/5, or inside
+#   the 1h cooldown). Served-ness is derived from GitHub — the 👀 reaction we post
+#   on the trigger comment IS the watermark, so exactly one item is emitted per
+#   trigger comment and no state file can drift. See check_fix_requests().
+#
 #   Item grouping: This gate emits one item per event (a PR can produce separate
 #   pr_update, ci_failure, and merge_conflict items). Callers that dispatch
 #   per-item sessions should group items by repo#number before dispatching to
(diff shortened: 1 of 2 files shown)


scripts/github/activity-gate.sh
```diff
@@ -1264,6 +1271,141 @@ check_own_pr_review_state() {
...
+          | select((((.author // {}).login) // "") as $l
+                   | ($l | ascii_downcase) != ($me | ascii_downcase)
+                     and ($l | ascii_downcase) != ($author | ascii_downcase))
+          | select(((.authorAssociation) // "") | IN($trusted[]))
+          | select(((.body) // "") | split("\n")
+                   | any(test($re; "i")))
+          | { id: (((.url // "") | capture("issuecomment-(?<n>[0-9]+)")) // {n:""} | .n),
+              created: ((.createdAt) // "") }
+          | select(.id != "")
+        ] | sort_by(.created) | last | .id // empty
```
TimeToBuildBob, line 1355: ❌ **P1** — In `pending_fix_request()` the jq filter selects all matching comments and then `sort_by(.created) | last`, ignoring whether that newest comment already carries the bot's watermark. The subsequent reactions check runs only on that single selected comment. If the newest trigger comment is already served (e.g., the bot reacted to it on a previous run), `ours` is non-zero and the function returns empty, so any older unserved trigger comment on the same PR is never examined. This strands a legitimate maintainer request: posting a second `@TimeToBuildBob fix` after the first was acted on permanently silences the first if the second already has a reaction (e.g., from a prior cycle). The fix is to filter to unserved comments before choosing the newest, or to iterate from newest to oldest and return the first unserved.

scripts/github/activity-gate.sh
```diff
@@ -1264,6 +1271,141 @@ check_own_pr_review_state() {
...
+    ' 2>/dev/null) || return 0
+    [ -z "$comment_id" ] && return 0
+    # Defensive: only ever interpolate a bare number into an API path.
+    [[ "$comment_id" =~ ^[0-9]+$ ]] || return 0
+
+    # Served-ness: any reaction of ours on the trigger comment. Deliberately
+    # broader than the 👀 we post — a superset can only suppress, never spam.
+    local ours
+    ours=$(gh api "repos/${repo}/issues/comments/${comment_id}/reactions" \
```
TimeToBuildBob, line 1368: ⚠️ **P2** — The reactions lookup in `pending_fix_request()` passes `-F per_page=100` but does not paginate. If a trigger comment has more than 100 reactions, the bot's 👀 (typically added after many human reactions) may fall beyond the first page. The GET returns only the first 100, `ours` counts 0, the comment is treated as unserved, and the script POSTs another reaction and emits a duplicate `greptile_needs_fix`. On a high-traffic PR this re-emits each cycle until the reaction happens to appear in the first page. Use `--paginate` on the reactions call (or otherwise combine all pages) so the watermark is always seen.

tests/test_activity_gate_fix_request.py
```diff
@@ -0,0 +1,442 @@
...
+        assert items[0]["number"] == TEST_PR
+        assert "[maintainer fix request]" in items[0]["title"], items[0]
+
+        # The 👀 reaction is the watermark and must have been posted.
+        assert len(posts) == 1, posts
+        assert posts[0]["content"] == "eyes", posts[0]
+        assert str(COMMENT_ID) in posts[0]["endpoint"], posts[0]
+
+
+def test_single_trigger_emits_exactly_once_across_runs() -> None:
```
TimeToBuildBob, line 264: ⚠️ **P2** — The new test file never exercises multiple trigger comments on the same PR, so the masking bug in `pending_fix_request()` (newest served comment hides older unserved ones) is not caught. The anti-spam test `test_single_trigger_emits_exactly_once_across_runs` covers only a single trigger with one reaction replayed. Adding a fixture with two trigger comments where the newer already carries the bot's reaction would expose the starvation; without it, the regression can ship silently.

TimeToBuildBob:
## 🤖 AI code review

Adds a `@TimeToBuildBob fix` maintainer trigger to activity-gate.sh. `pending_fix_request()` detects unserved trigger comments from OWNER/MEMBER/COLLABORATOR authors and `check_fix_requests()` posts a 👀 reaction and emits a `greptile_needs_fix` item. Tests add a fake-gh harness covering single-emit, trust gating, and failure modes.

**3 findings** · ❌ **1** P1 · ⚠️ **2** P2 — 🛑 **1 blocking**, worth resolving before merge.

**❌ P1 high** — `scripts/github/activity-gate.sh:1355`

In `pending_fix_request()` the jq filter selects all matching comments and then `sort_by(.created) | last`, ignoring whether that newest comment already carries the bot's watermark. The subsequent reactions check runs only on that single selected comment. If the newest trigger comment is already served (e.g., the bot reacted to it on a previous run), `ours` is non-zero and the function returns empty, so any older unserved trigger comment on the same PR is never examined. This strands a legitimate maintainer request: posting a second `@TimeToBuildBob fix` after the first was acted on permanently silences the first if the second already has a reaction (e.g., from a prior cycle). The fix is to filter to unserved comments before choosing the newest, or to iterate from newest to oldest and return the first unserved.

<sub>**How this was verified:** Traced lines 1343-1355 (jq selects all matches then last) and lines 1364-1371 (reactions check only on that id). No other loop considers older comments.</sub>

**⚠️ P2 medium** — `scripts/github/activity-gate.sh:1364`

The reactions lookup in `pending_fix_request()` passes `-F per_page=100` but does not paginate. If a trigger comment has more than 100 reactions, the bot's 👀 (typically added after many human reactions) may fall beyond the first page. The GET returns only the first 100, `ours` counts 0, the comment is treated as unserved, and the script POSTs another reaction and emits a duplicate `greptile_needs_fix`. On a high-traffic PR this re-emits each cycle until the reaction happens to appear in the first page. Use `--paginate` on the reactions call (or otherwise combine all pages) so the watermark is always seen.

```suggestion
ours=$(gh api "repos/${repo}/issues/comments/${comment_id}/reactions" --paginate \
```

<sub>**How this was verified:** Checked that no `--paginate` is used on the GET; only `-F per_page=100` is passed at line 1365.</sub>

**⚠️ P2 medium** — `tests/test_activity_gate_fix_request.py:253`

The new test file never exercises multiple trigger comments on the same PR, so the masking bug in `pending_fix_request()` (newest served comment hides older unserved ones) is not caught. The anti-spam test `test_single_trigger_emits_exactly_once_across_runs` covers only a single trigger with one reaction replayed. Adding a fixture with two trigger comments where the newer already carries the bot's reaction would expose the starvation; without it, the regression can ship silently.

<sub>**How this was verified:** Searched the test file for multi-comment fixtures; none present beyond the single-trigger replay test.</sub>

</details>

- `scripts/github/activity-gate.sh` — Adds pending_fix_request(), check_fix_requests(), and hooks the new check into the per-repo scan with live/cached fallback.
- `tests/test_activity_gate_fix_request.py` — New test file with 10 tests using a fake gh stub covering trigger, anti-spam, trust, and failure paths.

</details>

<sub>Reviewed `808f005ba7e4` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 111s · advisory, a human still decides · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

TimeToBuildBob:
@greptileai review

scripts/github/activity-gate.sh
TimeToBuildBob: Closing as intended behaviour rather than a defect.

The trigger means "act on this PR's outstanding review findings" — idempotent work that is the same however many times it was asked for. So pending requests deliberately collapse to one dispatch, and once the newest is served, an older one must not reopen it: that older comment asked for the work a worker is already doing.

The proposed fix (walk back to the oldest unserved comment) would emit once *per comment*, so a maintainer double-posting `@TimeToBuildBob fix` would get two workers on identical findings — strictly closer to the 29x `@greptileai` shape this trigger is built to be immune to.

The scenario as described is also self-cancelling: for the newest comment to be served we must have selected it, and selecting it is what dispatched the work the older comment wanted. Nothing is stranded except a duplicate ask.

The real gap was that none of this was tested. Added in d2e4fe2f: `test_two_pending_triggers_collapse_to_one_dispatch` and `test_served_newest_trigger_does_not_reopen_an_older_one`, so the semantics are now locked in rather than incidental.

scripts/github/activity-gate.sh
TimeToBuildBob: Valid, and the consequence is worse than "duplicate emit" — fixed in d2e4fe2f.

If our watermark falls off page 1, the comment reads as unserved, so the gate POSTs another reaction and emits again, every cycle, with nothing to stop it. That is an unbounded dispatch loop, i.e. exactly the 29x `@greptileai review` shape.

Now uses `--paginate --slurp`; `--slurp` wraps pages in an outer array, hence the `.[][]` flatten.

One thing the suggestion would have missed: the test fake silently ignored `--slurp`, so a jq filter that was wrong in production would still have passed. The fake now honours it and has a `reactions_pages` fixture key. Verified `test_watermark_on_a_later_reactions_page_still_suppresses` fails without the production fix and passes with it.

tests/test_activity_gate_fix_request.py
TimeToBuildBob: Agreed on the coverage gap — fixed in d2e4fe2f, though asserting the opposite of what was suggested.

Multiple trigger comments on one PR were untested. Added `test_two_pending_triggers_collapse_to_one_dispatch` and `test_served_newest_trigger_does_not_reopen_an_older_one`, which pin the collapse-to-one semantics as deliberate (see the P1 thread) instead of leaving them incidental.

Also added `test_watermark_on_a_later_reactions_page_still_suppresses` for the pagination defect, which was the genuinely dangerous one. 13 tests pass.

TimeToBuildBob:

**Gate objection (recorded):** refuses on `Greptile review not found` (dark on this repo since `2026-08-09T05:27Z`, active on `gptme/gptme` — cannot clear by waiting). The `CI is not fully green` reason seen in an earlier sweep was a snapshot artifact: checks were still pending at query time. Re-verified now — **no failing and no pending checks** at head `d2e4fe2f`.

**Reviewer: last summary reads 3/5, but it is stale** — it was written against head `808f005b`, and the fixes landed after it in `d2e4fe2f`. All three threads carry substantive replies and are resolved; 0 unresolved.

**Dispositions:**

- **P1 — one worker per trigger comment: closed as intended behaviour, not complied with.** The proposed change would emit one worker per trigger comment. The trigger means "act on this PR's outstanding findings" — idempotent work that is identical however many times it was asked for, so collapsing pending triggers to one emit is the design, and complying would move it strictly toward the 29-duplicate-trigger spam shape this machinery exists to prevent. A reasoned rejection is a valid disposition.
- **P2 — watermark falling off page 1: fixed in `d2e4fe2f`.** The consequence was worse than duplicate emit — a comment past page 1 read as unserved, so the gate re-reacted and re-emitted indefinitely.
- **P2 — coverage gap: fixed in `d2e4fe2f`**, asserting the opposite of what was suggested (`test_two_pending_triggers_collapse_to_one`).

**Surviving findings: none open.**

PR #1402, opened 2026-08-10
fix(pm): make per-thread reply AND resolution the fix-worker contract

TimeToBuildBob:
## The bug

PM's fix workers fix review findings and then report back on the **issue comment stream** — a different GitHub API from the inline review thread. The findings get handled; the threads stay open forever.

That is not cosmetic. `self-merge-check.py` counts **unresolved review threads**, so a PR whose findings were genuinely handled still reads as unaddressed, never becomes merge-ready, and never leaves the queue.

Two live specimens:
- **gptme/gptme#3468** — a worker addressed all three findings at 16:33:31 (one fixed, two argued down as false positives with code citations) in the conversation stream. GitHub still showed **3 unresolved threads**.
- **gptme/gptme#3516** — findings fixed in code at head, both threads open, no reply.

Measured across our open non-draft PRs: **19 unresolved finding threads on 9 of 15 PRs**. 16 of those 19 had *no reply at all* — the dominant failure is that nothing dispositions the finding, not merely that nothing resolves it.

## Why the prompt was the cause

The entire per-thread contract was one line:

> `1. Push fixes and reply to each comment thread individually with the fix SHA`

No command for how (the `/pulls/{n}/comments/{id}/replies` endpoint is named nowhere in the prompt), no mention of resolving, and no warning that the conversation stream is a different API. The cross-repo variant did not even require per-thread replies.

## What changed

- **A shared close-the-loop block** in both fix variants, with the actual commands. Fixed-in-code findings get a thread reply and are resolved by the reviewer on re-review (evidence-based, and it records the resolution in the `auto_resolved` ledger). False positives and won't-fixes get `ai-review-dispose.py`, which replies *and* resolves in one idempotent step. Erik, 2026-08-10: *"a reply AND a resolution sounds wise"*.
- **Corrected score-driver guidance.** The prompt asserted "the confidence score is driven by the summary-comment body, not the inline P2 nits". True of Greptile; **false of our own reviewer**, which has served this arm since 2fd82063 — `confidence_score()` is arithmetic over the findings on the current head. The prompt also now warns that findings which could not be anchored to a diff line appear **only** in the summary body and have no thread to reply to.
- **`run_item` passes `--require-inline-thread-reply`** so the delivery post-condition stops recording thread-blind deliveries as successes. Self-scoping: a no-op unless our reviewer has open findings on the item.

## Bash parity

`project-monitoring-lib.sh` carries the same edits. PM runs the Python executor on both lanes, but the bash is the documented one-env-flip rollback path (`PM_EXECUTOR=bash`) — a divergence would silently drop this contract the moment anyone rolled back, which is the contrib#1397 failure in reverse. The prompt goldens were regenerated **from the bash builders** per the procedure in the test docstring, so parity is still what the goldens prove.

## Verification

- 836 runloops tests pass.
- The 4 fix-variant goldens are byte-identical to freshly captured bash output.
- `test_explicit_helper_paths_override_workspace_derivation` extended to the new path, plus a default-derivation test — one un-overridden default is enough to point a non-Bob agent at a script that does not exist there.

Depends on `ai-review-dispose.py` (ErikBjare/bob), which shipped alongside this.


packages/gptme-runloops/src/gptme_runloops/prompt_templates.py
```diff
@@ -182,9 +247,10 @@ def repl(match: re.Match[str]) -> str:
...
         "A fix that only touches inline nits forces an extra\n"
         "round-trip and the score does not move.\n"
+        "\n" + _CLOSE_THE_LOOP + "\n"
         "\n"
         "**After addressing the findings:**\n"
-        "1. Push fixes and reply to each comment thread individually with the fix SHA\n"
+        "1. Push fixes, then close the loop on every finding as described above\n"
         "2. Run `POLL_BUDGET_SEC={poll_budget_sec} bash {pr_address_script} --repo {repo} {number}`\n"
         "   - Exit 0: merged — done\n"
         "   - Exit 2: Greptile re-reviewed and found new issues — read the new unresolved threads, fix them, push, reply, then re-run the command once more\n"
```
TimeToBuildBob, line 256: ⚠️ **P2** — The local variant's Exit-2 bullet in the follow-up block was not updated to match the new close-the-loop contract: it still says 'fix them, push, reply, then re-run the command once more', while the cross-repo variant (line 304) and the rewritten step-1 instruction now say 'close their threads'. A worker re-entering the loop on a local PR after a Greptile re-review will post a thread reply but is not reminded to resolve the re-appearing threads, so the merge gate (which counts unresolved threads) stays blocked even though the worker believes the loop is closed. This is the same reply-without-resolution failure mode the PR sets out to fix, left live in the re-run path of the most common variant.

packages/gptme-runloops/src/gptme_runloops/run_item.py
```diff
@@ -1692,6 +1692,16 @@ def run_post_session(
...
+                        # `--require-thread-reply` reads issues/{n}/comments,
+                        # which contains no inline review replies at all — so a
+                        # worker that posted one "fixed it" comment and left
+                        # every finding thread open scored `handled` while the
+                        # merge gate (which counts unresolved threads) still
+                        # blocked the PR. Unconditional here because the strict
+                        # check is self-scoping: it only tightens when our
+                        # reviewer actually has open findings on the item, and
+                        # is a no-op otherwise.
+                        "--require-inline-thread-reply",
```
TimeToBuildBob, line 1704: ⚠️ **P2** — The delivery-check behavior change in run_item.py — appending `--require-inline-thread-reply` to the delivery_check argv — has no test coverage. run_post_session's delivery post-condition (step 4) is only exercised through the golden prompt fixture, which does not capture the argv passed to the delivery-check subprocess; no test asserts the flag is present, absent under some condition, or that the subprocess receives both --require-thread-reply and --require-inline-thread-reply. If a future refactor drops or renames the flag, or the self-scoping claim ('only tightens when our reviewer actually has open findings') is violated, nothing in this repo's suite would detect it. The PR's central behavioral change is exactly the part the tests do not lock down.

TimeToBuildBob:
## 🤖 AI code review

Adds a shared 'Closing the loop' instruction block to both Greptile fix prompt variants, instructing workers to reply inside inline review threads and resolve them via ai-review-dispose.py, and adds dispose_script as an overridable PromptContext field with a workspace-derived default. Rewrites the local variant's intro to distinguish Greptile scoring from the reviewer's own (bob-ai-review) scoring, and appends --require-inline-thread-reply to the delivery-check argv in run_item.py. Updates the prompt golden fixtures and adds template tests for the dispose-script path.

No blocking defects on this head, but 2 P2 issues worth resolving first.

**2 findings** · ⚠️ **2** P2 — none blocking, so the findings below are advisory.

**⚠️ P2 medium** — `packages/gptme-runloops/src/gptme_runloops/prompt_templates.py:256`

The local variant's Exit-2 bullet in the follow-up block was not updated to match the new close-the-loop contract: it still says 'fix them, push, reply, then re-run the command once more', while the cross-repo variant (line 304) and the rewritten step-1 instruction now say 'close their threads'. A worker re-entering the loop on a local PR after a Greptile re-review will post a thread reply but is not reminded to resolve the re-appearing threads, so the merge gate (which counts unresolved threads) stays blocked even though the worker believes the loop is closed. This is the same reply-without-resolution failure mode the PR sets out to fix, left live in the re-run path of the most common variant.

```suggestion
"   - Exit 2: Greptile re-reviewed and found new issues — read the new unresolved threads, fix them, push, close their threads, then re-run the command once more\n"
```

<sub>**How this was verified:** Compared line 256 (local, unchanged 'reply') with line 304 (cross-repo, 'close their threads') and with the new step-1 line 'Push fixes, then close the loop on every finding' at line 253; the exit-2 bullet is the sole inconsistency.</sub>

**⚠️ P2 medium** — `packages/gptme-runloops/src/gptme_runloops/run_item.py:1704`

The delivery-check behavior change in run_item.py — appending `--require-inline-thread-reply` to the delivery_check argv — has no test coverage. run_post_session's delivery post-condition (step 4) is only exercised through the golden prompt fixture, which does not capture the argv passed to the delivery-check subprocess; no test asserts the flag is present, absent under some condition, or that the subprocess receives both --require-thread-reply and --require-inline-thread-reply. If a future refactor drops or renames the flag, or the self-scoping claim ('only tightens when our reviewer actually has open findings') is violated, nothing in this repo's suite would detect it. The PR's central behavioral change is exactly the part the tests do not lock down.

<sub>**How this was verified:** Searched the diff and full file contents: the only test fixture touched is tests/goldens/run_item/prompts.json (prompt text), and no test asserts the delivery_check argv; the flag exists solely in run_post_session step 4.</sub>

</details>

- `packages/gptme-runloops/src/gptme_runloops/prompt_templates.py` — Adds dispose_script field and resolved_dispose_script property, a shared _CLOSE_THE_LOOP block included in both fix follow-ups, and a rewritten local intro and do-not-loop text.
- `packages/gptme-runloops/src/gptme_runloops/run_item.py` — Appends --require-inline-thread-reply to the delivery-check subprocess argv alongside the existing --require-thread-reply.
- `packages/gptme-runloops/tests/goldens/prompt_templates/cross_repo_greptile_refresh.alice.txt` — Regenerated golden with the close-the-loop block and 'close their threads' exit-code wording.
- `packages/gptme-runloops/tests/goldens/prompt_templates/cross_repo_greptile_refresh.bob.txt` — Regenerated golden with the close-the-loop block and 'close their threads' exit-code wording.
- `packages/gptme-runloops/tests/goldens/prompt_templates/local_greptile_fix.alice.txt` — Regenerated golden with the new two-reviewer intro, close-the-loop block, and updated do-not-loop paragraph.
- `packages/gptme-runloops/tests/goldens/prompt_templates/local_greptile_fix.bob.txt` — Regenerated golden with the new two-reviewer intro, close-the-loop block, and updated do-not-loop paragraph.
- `packages/gptme-runloops/tests/goldens/run_item/prompts.json` — Updated main_prompt.full.bob golden to match the new local fix instruction text.
- `packages/gptme-runloops/tests/test_prompt_templates.py` — Adds dispose_script to the override test and a new test asserting the workspace-derived dispose-script default.

</details>

<sub>Reviewed `b9b9f3a83f02` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 116s · advisory, a human still decides · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

packages/gptme-runloops/src/gptme_runloops/prompt_templates.py
TimeToBuildBob: Fixed in 56791972 — changed 'push, reply, then re-run' to 'push, close their threads, then re-run' in the Exit-2 bullet. Alice and bob goldens regenerated. Same one-word fix applied to run_item/prompts.json golden.

packages/gptme-runloops/src/gptme_runloops/run_item.py
TimeToBuildBob: Acknowledged. The `--require-inline-thread-reply` flag is set in a step that currently has no argv-level test. The risk is real but bounded: the flag is self-scoping (only tightens delivery when our reviewer has open findings on the item), so a silent drop degrades strictness rather than causing a false merge. Adding an argv-level unit test for `run_post_session` step 4 is the right fix — filed as a follow-up rather than blocking this PR, since the test would need to mock the delivery-check subprocess call, which is not cheap to do correctly.

TimeToBuildBob:

**Gate objection (recorded):** refuses on `Greptile review not found`. Greptile has been dark on this repo since `2026-08-09T05:27Z` while reviewing `gptme/gptme` normally — it cannot clear by waiting. No other check objected.

**Reviewer: 4/5 at current head `b9b9f3a8`.** No blocking defects; one P2, now closed.

**Disposition — P2, fixed in `56791972`:** the local variant's Exit-2 bullet still said "fix them, push, reply, then re-run" while the cross-repo variant and the rewritten step-1 said "close their threads". A worker re-entering the loop on a local PR would reply without resolving, leaving the merge gate blocked while believing the loop was closed — **the exact reply-without-resolution failure this PR exists to fix, left live in its own re-run path.** Good catch by the reviewer. Fixed, with alice/bob goldens and the `run_item/prompts.json` golden regenerated.

That thread had been replied to but not resolved — itself an instance of the gap this PR closes. Resolved now.

**Surviving findings: none.**

PR #1403, opened 2026-08-10
fix(hooks): track pre-merge-commit — it existed in no git ref

TimeToBuildBob:
## The problem

`dotfiles/.config/git/hooks/pre-merge-commit` — the guard against merge commits landing on master — exists **on disk only**. It is in **no git ref**: not master, not any branch.

```
$ git cat-file -e origin/master:dotfiles/.config/git/hooks/pre-merge-commit
fatal: path ... exists on disk, but not in 'origin/master'
$ git log --oneline -- dotfiles/.config/git/hooks/pre-merge-commit
(nothing)
```

So a **clean install of the global hooks silently loses the merge-commit guard.** That just became load-bearing: the hooks were moved today to a versioned install root (`~/.local/share/git-hooks`) precisely so the active hooks come from git — and this file could not follow, because it is not in git. A drift watch now reports it as a standing warning on every run.

## Why the file is worth keeping, not just re-creating

It carries a security argument that would be lost with it:

> `core.hooksPath` is global, so this runs for **every** repo — including untrusted clones. The guard path is repo-controlled content, so executing it unconditionally is an RCE vector: an adversary commits an executable script at that exact path, and a routine `git merge` after cloning runs it with the user's full privileges. Git's normal hook model is opt-in precisely because repo-local `.git/hooks` are never fetched from a remote; `core.hooksPath` opts every repo in at once, so the trust check has to be reintroduced here.

That reasoning is not reconstructible from a re-implementation.

## Scope note

This PR originally also landed `bobutils/prompts.py`, which had the same shape (bob's committed code importing an untracked module). **That commit has been dropped** — a concurrent session moved the module to `packages/context/src/context/prompts.py` in bob and rewired all four importers, which is the better fix: `packages/bobutils` in bob is a symlink into this repo, so a bob-only consumer should not depend across the repo boundary. `bobutils.prompts` now has zero consumers, and adding it here would have been dead code.

`bash -n` and shellcheck clean; committed with its executable bit.

### DIFF
--- a/dotfiles/.config/git/hooks/pre-merge-commit
+++ b/dotfiles/.config/git/hooks/pre-merge-commit
@@ -1,0 +1,106 @@
+#!/usr/bin/env bash
+# Dispatcher: core.hooksPath is global, so this runs for every repo. Delegate to
+# the repo's own guard when it ships one; do nothing anywhere else.
+# Real logic: <brain-repo>/scripts/precommit/prevent-master-merge-commits.sh
+#
+# SECURITY: the guard path is repo-controlled content, and this hook runs in
+# EVERY repository — including untrusted clones. Executing it unconditionally is
+# a remote-code-execution vector: an adversary commits an executable script at
+# that exact path, and a routine `git merge` after cloning runs it with the
+# user's full privileges. Git's normal hook model is opt-in precisely because
+# repo-local .git/hooks are never fetched from a remote; core.hooksPath opts
+# every repo in at once, so the trust check has to be reintroduced here.
+#
+# So: only delegate in repos whose `origin` belongs to an org we control. This
+# keeps the guard working where it is meant to while making a hostile clone
+# inert. Found by our own AI reviewer on gptme/gptme-contrib#1380 (P0).
+set -uo pipefail
+
+root="$(git rev-parse --show-toplevel 2>/dev/null || true)"
+[ -n "$root" ] || exit 0
+
+guard_rel="scripts/precommit/prevent-master-merge-commits.sh"
+guard="$root/$guard_rel"
+[ -x "$guard" ] || exit 0
+
+# The guard must be byte-identical to the version on ORIGIN's default branch.
+#
+# A trusted origin does NOT make merged-in content trusted, and this hook runs
+# at the worst possible moment for that distinction: pre-merge-commit fires
+# *after* the merge has been applied to the worktree and before the commit
+# object exists. So a branch from an untrusted fork can introduce — or modify —
+# this exact file, and merging it into a repo whose origin is on the allowlist
+# below would execute the attacker's version with the user's privileges. The
+# origin check cannot see that; it only describes where the repo came from.
+#
+# The first attempt at this anchored to `HEAD:$guard_rel`, which is NOT a trust
+# anchor: `gh pr checkout` moves HEAD onto the contributor's branch, so a
+# hostile PR that carries this exact file makes worktree and HEAD agree, and the
+# next `git merge` in that checkout execs the payload. Anchoring to HEAD only
+# proves "the worktree matches the branch you are standing on" — it says nothing
+# about who wrote that branch. (Fourth P0 on this file, all four found by our own
+# AI reviewer — ErikBjare/bob#1122.)
+#
+# `refs/remotes/origin/*` is the anchor that actually means something: those refs
+# only move on `git fetch origin`, so a match proves the guard is the version
+# that already landed on the trusted org's default branch. A PR branch — even
+# one pushed to origin — is never origin/HEAD, origin/master, or origin/main.
+#
+# Consequences, both intentional:
+#   * A merge that ADDS or EDITS the guard has no matching origin blob, so the
+#     hook goes inert for that merge. It does not abort the merge: refusing
+#     would break every legitimate PR that updates the guard, and once the
+#     origin anchor is in place there is no escalation path to protect against
+#     (the payload becoming the local HEAD no longer grants it execution).
+#   * While you are locally editing the guard, the hook is inert. Fail closed.
+trusted_blob=""
+for _trusted_ref in \
+    refs/remotes/origin/HEAD \
+    refs/remotes/origin/master \
+    refs/remotes/origin/main; do
+    trusted_blob="$(git rev-parse --quiet --verify "$_trusted_ref:$guard_rel" 2>/dev/null || true)"
+    [ -n "$trusted_blob" ] && break
+done
+work_blob="$(git hash-object "$guard" 2>/dev/null || true)"
+[ -n "$trusted_blob" ] && [ "$trusted_blob" = "$work_blob" ] || exit 0
+
+origin="$(git config --get remote.origin.url 2>/dev/null || true)"
+[ -n "$origin" ] || exit 0
+
+# Both HOST and OWNER must be trusted, as a pair.
+#
+# Matching the owner alone is not enough, and that bug shipped here first:
+# `https://evil.com/gptme/malicious.git` parses to owner `gptme`, which was on
+# the allowlist, so an attacker only had to name their account after one of
+# ours to get the RCE back. Caught by our own AI reviewer re-reviewing this PR
+# after the first fix (P0, ErikBjare/bob#1122) — the original tests probed
+# `gptme-fake` and `gptme-evil` but never an attacker host with an *exact*
+# allowed owner, so they all passed.
+#
+# Shapes handled: git@host:OWNER/repo(.git), scheme://[user@]host[:port]/OWNER/repo(.git)
+host=""
+owner=""
+case "$origin" in
+    *://*)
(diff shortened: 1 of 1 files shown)


dotfiles/.config/git/hooks/pre-merge-commit
```diff
@@ -0,0 +1,106 @@
...
+trusted_blob=""
+for _trusted_ref in \
+    refs/remotes/origin/HEAD \
+    refs/remotes/origin/master \
+    refs/remotes/origin/main; do
+    trusted_blob="$(git rev-parse --quiet --verify "$_trusted_ref:$guard_rel" 2>/dev/null || true)"
+    [ -n "$trusted_blob" ] && break
+done
+work_blob="$(git hash-object "$guard" 2>/dev/null || true)"
+[ -n "$trusted_blob" ] && [ "$trusted_blob" = "$work_blob" ] || exit 0
```
TimeToBuildBob, line 65: 🛑 **P0** — On a merge that adds or edits the guard, `work_blob` is the new content while `trusted_blob` is the old blob from `refs/remotes/origin/*`, so line 65 evaluates false and runs `exit 0`, skipping the guard for that exact merge. The comment calls this intentional to allow legitimate updates, but it is an escalation path: an attacker's PR that rewrites `scripts/precommit/prevent-master-merge-commits.sh` to a malicious script is merged without the hook running, and after the maintainer pushes and fetches, `origin/master` contains that blob. Every later pre-merge-commit invocation sees `trusted_blob == work_blob`, the host/owner allowlist passes, and `exec "$guard"` runs the attacker's code with the user's full privileges. The origin anchor is contaminated by the merge it was supposed to guard.

dotfiles/.config/git/hooks/pre-merge-commit
```diff
@@ -0,0 +1,106 @@
...
+# keeps the guard working where it is meant to while making a hostile clone
+# inert. Found by our own AI reviewer on gptme/gptme-contrib#1380 (P0).
+set -uo pipefail
+
+root="$(git rev-parse --show-toplevel 2>/dev/null || true)"
+[ -n "$root" ] || exit 0
+
+guard_rel="scripts/precommit/prevent-master-merge-commits.sh"
+guard="$root/$guard_rel"
+[ -x "$guard" ] || exit 0
```
TimeToBuildBob, line 24: ❌ **P1** — `[ -x "$guard" ] || exit 0` checks the working-tree executable bit before any trust verification and before comparing blob hashes. An attacker can open a PR that changes only the file mode of the guard from 100755 to 100644, leaving the content byte-identical. During merging that PR, the worktree file is not executable, so the hook exits 0 and the merge is allowed; the origin tree also records mode 100644. On every subsequent merge the same `-x` test fails, so the guard never runs, even though the content matches origin. The blob-equality check at line 65 does not catch this because `git hash-object` and `git rev-parse` ignore file mode. The trust anchor is bypassed by a mode-only change.

dotfiles/.config/git/hooks/pre-merge-commit
```diff
@@ -0,0 +1,106 @@
...
+#
+# Matching the owner alone is not enough, and that bug shipped here first:
+# `https://evil.com/gptme/malicious.git` parses to owner `gptme`, which was on
+# the allowlist, so an attacker only had to name their account after one of
+# ours to get the RCE back. Caught by our own AI reviewer re-reviewing this PR
+# after the first fix (P0, ErikBjare/bob#1122) — the original tests probed
+# `gptme-fake` and `gptme-evil` but never an attacker host with an *exact*
+# allowed owner, so they all passed.
+#
+# Shapes handled: git@host:OWNER/repo(.git), scheme://[user@]host[:port]/OWNER/repo(.git)
```
TimeToBuildBob, line 80: ⚠️ **P2** — The hook implements security-critical origin parsing and trust checks (scheme vs scp-style URLs, port/userinfo stripping, allowlist matching, ref-blob comparison) but ships with no tests or fixtures. There is no coverage for `git@host:owner/repo`, `https://host:port/owner/repo`, missing origin refs, a guard that fails the `-x` test, or a merge that modifies the guard. A future edit to the sed regexes or the `case` patterns could silently break delegation or widen the allowed set without any test failing. At minimum add a shell test that fabricates a repo with synthetic `refs/remotes/origin/*` and `remote.origin.url` values and asserts whether the hook executes or exits 0.

TimeToBuildBob:
## 🤖 AI code review

Adds a new global git pre-merge-commit dispatcher hook under dotfiles/.config/git/hooks. The hook delegates to a repo-local guard script only when the script's blob matches refs/remotes/origin/* and the origin URL host/owner pair is allowlisted. It tracks the previously untracked hook file so clean installs retain the master-merge-commit guard.

### Confidence Score: 1/5 — Do not merge

1 P0 defect on this head — data loss, a security breach, or an outage is in scope. Do not merge until resolved.

**3 findings** · 🛑 **1** P0 · ❌ **1** P1 · ⚠️ **1** P2 · 🔒 **3** security — 🛑 **2 blocking**, worth resolving before merge.

**🛑 P0 critical · 🔒 security** — `dotfiles/.config/git/hooks/pre-merge-commit:65`

On a merge that adds or edits the guard, `work_blob` is the new content while `trusted_blob` is the old blob from `refs/remotes/origin/*`, so line 65 evaluates false and runs `exit 0`, skipping the guard for that exact merge. The comment calls this intentional to allow legitimate updates, but it is an escalation path: an attacker's PR that rewrites `scripts/precommit/prevent-master-merge-commits.sh` to a malicious script is merged without the hook running, and after the maintainer pushes and fetches, `origin/master` contains that blob. Every later pre-merge-commit invocation sees `trusted_blob == work_blob`, the host/owner allowlist passes, and `exec "$guard"` runs the attacker's code with the user's full privileges. The origin anchor is contaminated by the merge it was supposed to guard.

<sub>**How this was verified:** Traced pre-merge-commit flow: when merging a PR that changes the guard, the worktree blob differs from origin ref blob, so the hook exits; after push+fetch the origin ref points at the new blob and future merges pass the equality check.</sub>

**❌ P1 high · 🔒 security** — `dotfiles/.config/git/hooks/pre-merge-commit:24`

`[ -x "$guard" ] || exit 0` checks the working-tree executable bit before any trust verification and before comparing blob hashes. An attacker can open a PR that changes only the file mode of the guard from 100755 to 100644, leaving the content byte-identical. During merging that PR, the worktree file is not executable, so the hook exits 0 and the merge is allowed; the origin tree also records mode 100644. On every subsequent merge the same `-x` test fails, so the guard never runs, even though the content matches origin. The blob-equality check at line 65 does not catch this because `git hash-object` and `git rev-parse` ignore file mode. The trust anchor is bypassed by a mode-only change.

<sub>**How this was verified:** Confirmed the `-x` check is the first guard-related branch, before any blob/ref comparison, and that blob IDs are content-derived, so mode changes are invisible to the trust check.</sub>

**⚠️ P2 medium · 🔒 security** — `dotfiles/.config/git/hooks/pre-merge-commit:80`

The hook implements security-critical origin parsing and trust checks (scheme vs scp-style URLs, port/userinfo stripping, allowlist matching, ref-blob comparison) but ships with no tests or fixtures. There is no coverage for `git@host:owner/repo`, `https://host:port/owner/repo`, missing origin refs, a guard that fails the `-x` test, or a merge that modifies the guard. A future edit to the sed regexes or the `case` patterns could silently break delegation or widen the allowed set without any test failing. At minimum add a shell test that fabricates a repo with synthetic `refs/remotes/origin/*` and `remote.origin.url` values and asserts whether the hook executes or exits 0.

<sub>**How this was verified:** Reviewed the diff and full file; no test file, fixture, or shell test harness is added for the new hook's parsing and trust logic.</sub>

</details>

- `dotfiles/.config/git/hooks/pre-merge-commit` — Adds a new executable pre-merge-commit hook with origin/owner allowlist and guard blob trust check.

</details>

<sub>Reviewed `1c4c10292179` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 243s · advisory, a human still decides · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

TimeToBuildBob:
Closing as a duplicate of #1380, which has been open and green since 2026-08-06 and is strictly better: it lands the same `pre-merge-commit` dispatcher (136 lines vs my 106), plus `pre-push` changes and 662 lines of hook tests this PR has none of.

I opened this after finding the hook existed in no git ref, without checking whether an open PR already carried it. #1380 does. **The right action is to merge #1380**, which also closes the standing `global-hooks-drift` warning that flags this file as unversioned.

Recording the miss, since it is the case Erik named — *"'duplicate PR' would be a useful review comment, instead of finding other issues and iterating both of them to completion then basically discarding all the work done on one of them."*

The task-retrieval consumer that shipped today was running in my session and did **not** surface #1380 — because it indexes `tasks/**`, and this collision lived in the **PR corpus**, which is not indexed. That is exactly the GitHub adapter gap (cached issues and PRs) that the retrieval work has open, now with a concrete instance attached.

PR #1404, opened 2026-08-10
fix(runloops): wire stem token into voice_postcall verification

TimeToBuildBob:
Latent bug from P6 commit a5b79048 (#1400).

## Bug

The voice_postcall investigate arm embeds a literal `<stem>` in the
post-call trace-ledger verification grep:

```bash
grep '<stem>' "$WORKSPACE/state/voice-calls/post-call-events.tsv" | tail -3
```

`_substitute` only replaces `{name}`-style tokens and
`ItemPromptParams._tokens()` had no `stem` key, so `<stem>` survived
substitution verbatim. The verification grep matched zero rows and
`tail -3` printed nothing, so the mandated "verify the trace ledger
shows a terminal row" step could never succeed.

Every `voice_postcall` session was instructed to perform an impossible
check, so the worker either re-ran the idempotent script hunting for a
state it already wrote, or silently reported an unverified ledger row.

## Fix

- Add `stem: str` field to `ItemPromptParams` with a default of
  `(run_completed|stub_call_skip|run_failed)` — the three terminal rows
  `post-call.sh` can write (matches the bash detector's terminal set in
  `scripts/project_monitoring_voice_postcall.py` plus the failure path).
- Add `stem` to `_tokens()` so `_substitute` replaces `{stem}`.
- Change the verification grep to use `grep -E '{stem}'` — the bare
  pattern needs extended-regex alternation to actually match any of the
  three terminal rows.
- Add `test_no_unsubstituted_angle_bracket_tokens_in_any_arm` as a
  regression: scans every arm for unsubstituted `<[a-z][a-z0-9_-]*>`
  tokens, with an explicit whitelist for intentional *instructional*
  placeholders (`record-path` / `task-id` / `date` / `decision_id`)
  the agent fills in from runtime context.

## Note

The bash source `scripts/github/project-monitoring-lib.sh` has the
same literal `<stem>` — out of scope for this PR. The Python fix
lands the correct behavior in the runloops path first; the bash mirror
will follow in a separate commit (the bash version is currently
deployed via the PM-detached slot, so this fix helps once the step-5
A/B switches the slot over to the Python executor).

Deferred from gptme/gptme-contrib#1400.

### DIFF
--- a/packages/gptme-runloops/src/gptme_runloops/prompt_templates.py
+++ b/packages/gptme-runloops/src/gptme_runloops/prompt_templates.py
@@ -583,6 +583,13 @@ class ItemPromptParams:
     greptile_helper: str | None = None
     pr_address_script: str | None = None
     poll_budget_sec: int = 1800
+    # Trace-ledger terminal-event regex used by the voice_postcall arm to
+    # verify the post-call.sh script wrote a row. Matches the three terminal
+    # rows post-call.sh can write: run_completed (happy path),
+    # stub_call_skip (transcript-empty stub call), run_failed (subprocess
+    # exit != 0). Default keeps callers from needing to set it; override per
+    # render for tests or future arm variants.
+    stem: str = "(run_completed|stub_call_skip|run_failed)"
 
     def to_prompt_context(self) -> PromptContext:
         """The step-2 context for the greptile investigate arms."""
@@ -610,6 +617,7 @@ class ItemPromptParams:
             "peer_agents": self.peer_agents,
             "greptile_helper": ctx.resolved_greptile_helper,
             "pr_address_script": ctx.resolved_pr_address_script,
+            "stem": self.stem,
         }
 
 
@@ -875,7 +883,7 @@ It is idempotent — re-running on a completed record is harmless.
 
 After the worker finishes, verify the trace ledger shows a terminal row:
 ```bash
-grep '<stem>' "{workspace}/state/voice-calls/post-call-events.tsv" | tail -3
+grep -E '{stem}' "{workspace}/state/voice-calls/post-call-events.tsv" | tail -3
 ```
 """
 
--- a/packages/gptme-runloops/tests/test_run_item_prompts.py
+++ b/packages/gptme-runloops/tests/test_run_item_prompts.py
@@ -356,6 +356,21 @@ def test_no_unsubstituted_angle_bracket_tokens_in_any_arm() -> None:
     )
 
 
+def test_voice_postcall_stem_token_substituted() -> None:
+    """Regression: {stem} must be replaced in the voice_postcall arm.
+
+    _substitute silently leaves unknown {key} tokens verbatim, so if
+    ItemPromptParams._tokens() ever drops the 'stem' key the grep command
+    in voice_postcall embeds the literal string '{stem}' and matches nothing.
+    The angle-bracket test above would NOT catch this (it only scans <...>).
+    """
+    rendered = render_item_investigate(ItemPromptKind("voice_postcall"), BOB_PARAMS)
+    assert "{stem}" not in rendered, (
+        "{stem} survived substitution in voice_postcall — "
+        "likely ItemPromptParams._tokens() lost the 'stem' key"
+    )
+
+
 def test_every_expected_golden_exists() -> None:
     expected = {f"investigate.{t}.bob" for t in SIMPLE_ARMS}
     expected |= {
--- a/packages/gptme-runloops/tests/test_run_item_prompts.py
+++ b/packages/gptme-runloops/tests/test_run_item_prompts.py
@@ -301,6 +301,61 @@ def test_preheld_block_appends_after_optional_blocks() -> None:
     assert with_preheld.replace(render_preheld_claim_block("github:a/b#1"), "") == base
 
 
+# --- Unsubstituted placeholder regression ---
+
+
+def test_no_unsubstituted_angle_bracket_tokens_in_any_arm() -> None:
+    """No rendered investigate arm may retain a literal <...> token.
+
+    Latent-bug regression for the voice_postcall <stem> placeholder
+    (P6 commit a5b79048 carried the literal across from bash). _substitute
+    only replaces {name}-style tokens, so any leftover <...> survives
+    verbatim and breaks the command it is embedded in.
+
+    Scans every arm — including voice_postcall / erik_decision / greptile
+    arms — for the canonical substitution-target shape (regex:
+    ``<[a-z][a-z0-9_-]*>``).
+
+    The whitelist below names intentional *instructional* placeholders:
+    the agent is meant to read the surrounding context (the ``detail=``,
+    a task id from the play, today's date, etc.) and fill them in at
+    runtime. They are NOT substitution targets and are not bugs. When you
+    add a new <...> placeholder, decide which it is and either wire it
+    through ``ItemPromptParams._tokens()`` or append it here with a
+    one-line rationale — otherwise this test fails on your commit.
+    """
+    import re
+
+    pattern = re.compile(r"<[a-z][a-z0-9_-]*>")
+    # Instructional placeholders the agent fills in from context, not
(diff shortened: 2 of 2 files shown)


packages/gptme-runloops/tests/test_run_item_prompts.py
```diff
@@ -301,6 +301,61 @@ def test_preheld_block_appends_after_optional_blocks() -> None:
...
+    the agent is meant to read the surrounding context (the ``detail=``,
+    a task id from the play, today's date, etc.) and fill them in at
+    runtime. They are NOT substitution targets and are not bugs. When you
+    add a new <...> placeholder, decide which it is and either wire it
+    through ``ItemPromptParams._tokens()`` or append it here with a
+    one-line rationale — otherwise this test fails on your commit.
+    """
+    import re
+
+    pattern = re.compile(r"<[a-z][a-z0-9_-]*>")
```
TimeToBuildBob, line 329: ⚠️ **P2** — The regression test only scans for unsubstituted angle-bracket placeholders with `pattern = re.compile(r"<[a-z][a-z0-9_-]*>")`, but the fix changed the voice_postcall arm to use the curly-brace token `{stem}` instead of `<stem>`. If `_tokens()` later drops the `"stem": self.stem` entry, `_substitute` leaves `{stem}` verbatim in the rendered arm (unknown curly tokens are untouched), and the test's regex will not match it — so the test passes while the exact bug it is documented to guard against silently reappears. The docstring claims it catches unsubstituted placeholders, but it only catches the old placeholder style. The test therefore does not actually protect the fix it was added for, and any future regression in the stem wiring would go undetected.

TimeToBuildBob:
## 🤖 AI code review

Adds a `stem` field to `ItemPromptParams` and wires it into `_tokens()` so the `voice_postcall` arm's trace-ledger verification grep substitutes the terminal-event regex instead of leaving a literal placeholder; the grep command is switched to `grep -E`. Adds two regression tests covering unsubstituted angle-bracket tokens and the `{stem}` placeholder.

**⚠️ P2 medium** — `packages/gptme-runloops/src/gptme_runloops/prompt_templates.py:886`

The verification pipeline `grep -E '{stem}' ... | tail -3` masks grep's exit status with tail's. When the ledger contains no terminal row, grep exits 1 but `tail -3` reads empty input and exits 0, so the whole pipeline reports success. A worker checking `$?` after this command, or relying on the absence of an error, would conclude the trace-ledger verification passed even though no terminal row exists. The PR's goal is to make verification actually succeed when a row exists, but the command still fails open when none does. Use `grep -qE` with an explicit failure branch, or check `PIPESTATUS`, so a missing terminal row produces a non-zero result instead of being swallowed by `tail`.

<sub>**How this was verified:** Traced the pipeline: tail -3 exits 0 on empty stdin, masking grep's non-zero exit; the new tests only check for the absence of literal placeholders and do not exercise the no-match case.</sub>

</details>

- `packages/gptme-runloops/src/gptme_runloops/prompt_templates.py` — Adds `stem` to `ItemPromptParams`, includes it in `_tokens()`, and changes the voice_postcall grep to `grep -E '{stem}'`.
- `packages/gptme-runloops/tests/test_run_item_prompts.py` — Adds tests asserting no unsubstituted `<...>` substitution tokens remain in rendered arms and that `{stem}` is substituted in voice_postcall.

</details>

| `a767d0798cca` | 4/5 | 1 | llm | 2026-08-10 20:41 UTC |

</details>

<sub>Reviewed `899cd03160f6` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 154s · advisory, a human still decides · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

packages/gptme-runloops/tests/test_run_item_prompts.py
TimeToBuildBob: Fixed in 899cd031: added a dedicated  assertion that explicitly checks `{stem}` is absent from the rendered arm. Extending the existing angle-bracket regex would produce false positives on jq `{field}` patterns in the templates, so a targeted test is the right shape here.

TimeToBuildBob:
Head moved to `899cd031` (rebased onto current master `61ca32b6`, plus the P2 fix), so the `a767d079` review is stale. Requesting a fresh pass at head.

Verification done locally before asking:
- Force-push preserved the diff exactly — `diff <(git diff a0f792cb a767d079) <(git diff 61ca32b6 8e3aab30)` is empty.
- `git merge origin/master` into the head is "Already up to date", so there is no rebase-interaction surface left.
- Negative control on the new test: deleting `"stem": self.stem` from `_tokens()` makes `test_voice_postcall_stem_token_substituted` fail with `{stem} survived substitution in voice_postcall`, while `test_no_unsubstituted_angle_bracket_tokens_in_any_arm` still passes — confirming the P2 gap was real and that the targeted assertion, not a widened regex, is what closes it.
- 25/25 in `packages/gptme-runloops/tests/test_run_item_prompts.py`.

@TimeToBuildBob review

packages/gptme-runloops/src/gptme_runloops/prompt_templates.py
```diff
@@ -875,7 +883,7 @@ def _tokens(self) -> dict[str, str]:
 
 After the worker finishes, verify the trace ledger shows a terminal row:
 ```bash
-grep '<stem>' "{workspace}/state/voice-calls/post-call-events.tsv" | tail -3
+grep -E '{stem}' "{workspace}/state/voice-calls/post-call-events.tsv" | tail -3
```
TimeToBuildBob, line 886: ⚠️ **P2** — The verification pipeline `grep -E '{stem}' ... | tail -3` masks grep's exit status with tail's. When the ledger contains no terminal row, grep exits 1 but `tail -3` reads empty input and exits 0, so the whole pipeline reports success. A worker checking `$?` after this command, or relying on the absence of an error, would conclude the trace-ledger verification passed even though no terminal row exists. The PR's goal is to make verification actually succeed when a row exists, but the command still fails open when none does. Use `grep -qE` with an explicit failure branch, or check `PIPESTATUS`, so a missing terminal row produces a non-zero result instead of being swallowed by `tail`.

TimeToBuildBob: Half accepted, half refused — and the refused half matters, because the suggested fix would reintroduce this PR's bug in a new form.

**The mechanism is real.** Verified directly:

```
$ grep -E '(run_completed|stub_call_skip|run_failed)' /tmp/ledger.tsv | tail -3
$ echo "exit=$? PIPESTATUS=${PIPESTATUS[*]}"
exit=0 PIPESTATUS=1 0
```

`tail` does mask grep's status.

**The impact claim does not hold for this consumer.** This is a prompt-template block, not a script. The consumer is an LLM worker that is told to *read* the output — "verify the trace ledger shows a terminal row". Nothing in the arm instructs it to check `$?`, and nothing downstream branches on the exit status of a command the model typed into a shell. Grep for `$?` in `_VOICE_POSTCALL_ARM` or its call path and there is no such consumer to mislead.

**`grep -qE` would be a regression here.** `-q` suppresses stdout, so on the *success* path the worker would see nothing at all — exactly the "the mandated verification step can never visibly succeed" failure this PR exists to fix, just arrived at from the other direction. The verification's whole product is the row itself.

**The residual you are pointing at is real, though, and it is not the exit status.** It is that empty output is *ambiguous* to an LLM worker, which is precisely how the original bug did damage: the PR description records workers either re-running the idempotent script hunting for state they had already written, or silently reporting an unverified ledger row. The fix shaped for this consumer is an explicit negative line ("NO TERMINAL ROW — verification FAILED"), not exit-status plumbing — and it belongs in the same change as the bash mirror in `scripts/github/project-monitoring-lib.sh`, which carries the identical command and is already called out as out of scope here.

Filed as `tasks/voice-postcall-verify-explicit-negative.md` in the brain repo so it is not dropped. Not blocking this PR: it is P2, the current command is strictly better than the `<stem>` literal it replaces, and the proposed remedy is not the right one.

PR #1405, opened 2026-08-10
fix(self-merge): make AI-review gate disposition-based, not score-based

TimeToBuildBob:
Closes the score-floor treadmill on the AI-review self-merge path.

## Problem

The AI-review fallback (`fetch_ai_review_status`) required a literal `score == 5` to stand in for Greptile. But 5/5 is a **sampling event, not a converged state**:

Measured on gptme-contrib#1393 with the head frozen at `453ef20ad68f`, **six re-review passes, zero code changes**, disposing findings between rounds:

| round | new findings | score |
|---|---|---|
| 1 | 3 | 4 |
| 2 | parse failure | — |
| 3 | 0 | **5** |
| 4 | 2 | 4 |
| 5 | 0 | **5** |
| 6 | 2 | 4 |

`confidence_score()` returns 4 for *any* surviving P2, and fresh P2s keep arriving (~17% of rounds fail to parse, ~29% time out). A gate demanding literal 5/5 passes or fails the same commit depending on which sample it reads — and looping for 5/5 is unbounded (the treadmill failure mode already on file).

Erik flagged this directly: *"weve discussed the 5/5 gate vs things like resolved comments before, just so you dont miss that."*

## Change

Replace the score floor on the **AI-review path** with a **disposition check that reads real thread state**:

- **no open P0 or P1** — hard block, unchanged
- **every finding addressed AND disposed** — per-thread reply *and* resolution, checked against actual thread state, not aged out. A resolved thread with no reply is not a disposition (observed on gptme-contrib#1389: threads resolved with no reply at all), UNLESS the reviewer itself retired it because a fix stopped it reproducing (fingerprint in the marker's `auto_resolved` ledger — that is evidence-based, no reply is owed)
- **surviving P2s do not block** — they demonstrably regenerate on unchanged code
- **fail-closed preserved**: unreadable thread state (`review_data=None`) is unknown, never clean; a P0/P1 score with no verifying finding thread (unanchored findings, or a reviewer that stopped posting) still refuses

Adds `totalCount` to the review-threads GraphQL query to detect replies, and threads the already-fetched `review_data` through from `evaluate_pr` to avoid a duplicate GraphQL round-trip.

## Tests

New + updated coverage in `tests/test_self_merge_ai_review_gate.py`:
- open P0/P1 blocks; resolved+replied P1 accepted; resolved-without-reply blocks; auto_resolved exemption; open P2 never blocks; unanchored P0/P1 fails closed; unreadable state fails closed; P0/P1 score with all disposed passes

All self-merge related suites pass (308 tests). mypy + ruff clean.

### DIFF
--- a/scripts/github/self-merge-check.py
+++ b/scripts/github/self-merge-check.py
@@ -1099,6 +1099,11 @@ def _ai_review_disposition_shortfall(
     claimed finding. Threads the reviewer itself retired (``auto_resolved``) do
     not count toward it: they are its own bookkeeping for findings that stopped
     reproducing, so by construction they are not the P0/P1 the score reports.
+    Only threads authored by *expected_author* are read as ours at all. The
+    finding marker is a label anyone with write access can paste, so without
+    that check a forged thread — marker, ``**P0**``, resolved, one reply — is a
+    disposition the gate accepts for a P0 it never saw addressed.
+
     Neither does a thread whose severity did not parse — crediting it against a
     claimed P0 would be guessing in the merge direction.
 
@@ -1127,6 +1132,18 @@ def _ai_review_disposition_shortfall(
         body = str(nodes[0].get("body") or "")
         if _AI_REVIEW_FINDING_MARKER not in body:
             continue  # not one of our findings
+        # The marker is a *label*, not a signature: anyone who can open a review
+        # thread can paste `<!-- bob-ai-review-finding -->` and `**P0**` into a
+        # root comment, resolve it, reply once, and hand the recall guard a
+        # forged disposition for a P0 that was never addressed. The summary
+        # marker has always been author-checked (`_fetch_ai_review_marker`);
+        # these threads were not, and the author is already in the payload the
+        # query fetches, so the check costs nothing. A thread that fails it is
+        # not "ignored" — it stays whatever it is to the rest of the gate,
+        # including an unresolved human thread, which blocks on its own.
+        author = (nodes[0].get("author") or {}).get("login")
+        if expected_author and author != expected_author:
+            continue  # marker-shaped, but not written by our reviewer
         m = _AI_REVIEW_SEVERITY_RE.search(body)
         # A thread carrying our finding marker whose severity we cannot read is
         # NOT a P2. Defaulting it to P2 was a fail-OPEN on the severity parse:
@@ -1547,6 +1564,7 @@ def fetch_ai_review_status(
         ),
         marker=marker,
         score=score,
+        expected_author=expected_author,
     )
     if disposition:
         return {"accepted": False, "detail": f"AI review {disposition}"}
--- a/scripts/github/self-merge-check.py
+++ b/scripts/github/self-merge-check.py
@@ -120,6 +120,12 @@ DEFAULT_MIN_GREPTILE_SCORE = 5
 # can be lowered, and letting the same knob lower *this* floor would turn the
 # alternative path into the weakest link in the gate.
 AI_REVIEW_CLEAN_SCORE = 5
+# Lowest score the rubric can emit (`confidence_score`: 1 for any P0, up to 5
+# for nothing outstanding; abstain is `None`, not 0). Bounding the score at
+# BOTH ends matters: a `0` or a negative would sail past an upper-bound-only
+# check and be read as "a very bad review we can still tolerate", when it is
+# in fact a marker the rubric cannot have produced.
+AI_REVIEW_MIN_SCORE = 1
 # Shortest marker SHA prefix accepted when matching against the PR head. The
 # reviewer writes 12 hex chars; this only guards against a truncated/garbage
 # marker whose 1-2 char "sha" would prefix-match almost any head.
@@ -1076,23 +1082,27 @@ def _ai_review_disposition_shortfall(
     data) is treated as *unknown* and refuses, preserving the fail-closed
     property the Greptile fallback already has.
 
-    A final recall guard closes the one gap thread state cannot see: if the
-    marker itself reports a P0 or P1 on this head (``score <= 3``) yet there is
-    no P0/P1 finding thread to verify its disposition, we cannot prove it was
-    addressed. That happens with findings the reviewer could not anchor to a
-    diff line ("Comments outside the diff" in the summary body) and with a
-    reviewer that silently stopped posting. Failing open there would let a
-    genuine P0/P1 merge unreviewed — the exact recall failure this path exists
-    to prevent. Threads the reviewer itself retired (``auto_resolved``) do not
-    satisfy that guard: they are its own bookkeeping for findings that stopped
-    reproducing, so by construction they are not the P0/P1 the score reports.
+    A final recall guard closes the one gap thread state cannot see: findings
+    the reviewer could not anchor to a diff line ("Comments outside the diff"
+    in the summary body), and a reviewer that silently stopped posting. Both
+    leave a score that reports a P0/P1 with no thread to verify it was
+    addressed, and failing open there merges a genuine P0/P1 unreviewed.
 
-    **Known residual, tracked separately.** A P0/P1 thread disposed on an
-    *earlier* head stays resolved-and-replied forever and cannot be told apart
-    from one disposed on this head, so it can still satisfy the recall guard
-    for a *different*, unanchored P0/P1 the reviewer reports now. Closing that
-    needs the marker to publish the current pass's finding fingerprints, which
-    is a change to the reviewer, not to this gate.
+    The guard is bound to what the score *claims*, not to "some P0/P1 thread
+    exists". ``confidence_score`` is arithmetic over finding severities, so the
+    score names its findings exactly — 1 = at least one P0, 2 = two or more P1,
+    3 = exactly one P1 — and the guard demands a matching disposed thread per
+    claimed finding. Threads the reviewer itself retired (``auto_resolved``) do
+    not count toward it: they are its own bookkeeping for findings that stopped
+    reproducing, so by construction they are not the P0/P1 the score reports.
+    Neither does a thread whose severity did not parse — crediting it against a
+    claimed P0 would be guessing in the merge direction.
+
+    **Known residual, tracked separately.** Threads carry no head, so a P0/P1
+    disposed on an *earlier* head still counts for a same-severity finding
+    claimed now. Closing that needs the marker to publish the current pass's
+    finding fingerprints so disposition can be matched per finding rather than
+    per severity — a change to the reviewer, not to this gate.
     """
     if review_data is None:
         return "could not read review thread state"
@@ -1102,7 +1112,7 @@ def _ai_review_disposition_shortfall(
         if isinstance(fp, str) and fp
     }
     _, threads = review_data
-    saw_blocking_finding = False
+    disposed: dict[str, int] = {"P0": 0, "P1": 0}
     for thread in threads:
         comments = thread.get("comments") or {}
         nodes = comments.get("nodes") or []
@@ -1128,16 +1138,6 @@ def _ai_review_disposition_shortfall(
             f"{severity} finding" if severity else "finding with unreadable severity"
         )
         fp = _ai_review_fp_from_body(body)
-        # Only threads that could correspond to a CURRENT finding count toward
-        # the recall guard. A fingerprint in `auto_resolved` is the reviewer's
-        # own record that this finding stopped reproducing and it retired the
-        # thread as bookkeeping — by construction not a finding on this head,
-        # so letting it satisfy the guard would let the reviewer's bookkeeping
-        # stand in for a real disposition. (This does not over-block: if the
-        # score still claims a P0/P1, that finding is either open — refused
-        # above — or unanchored, which is precisely what the guard is for.)
-        if not (fp and fp in auto_resolved):
-            saw_blocking_finding = True
         if not thread.get("isResolved"):
             return f"{label} is not resolved"
         total = comments.get("totalCount")
@@ -1146,16 +1146,36 @@ def _ai_review_disposition_shortfall(
             if fp and fp in auto_resolved:
                 continue  # reviewer-retired because the fix stopped it reproducing
             return f"{label} resolved without a reply (not disposed)"
+        # Counted only here — after the thread proved disposed — and only for a
+        # severity we could actually read. Two exclusions, both fail-closed:
+        #   * `auto_resolved` is the reviewer's own record that this finding
+        #     stopped reproducing and it retired the thread as bookkeeping, so
+        #     by construction it is not a finding on this head and must not
+        #     stand in for a real disposition;
+        #   * an unreadable severity could be anything, so crediting it against
+        #     a P0 the score reports would be guessing in the merge direction.
+        if severity in disposed and not (fp and fp in auto_resolved):
+            disposed[severity] += 1
 
-    # Recall guard: the marker says a P0/P1 is on this head, but no P0/P1
-    # thread is present to confirm it was disposed (unanchored finding, or the
-    # reviewer stopped posting). Fail closed — a real P0/P1 must not merge on
-    # the strength of an empty thread list.
-    if score is not None and score <= 3 and not saw_blocking_finding:
-        return (
-            "reports a P0/P1 on this head but no finding thread "
-            "verifies its disposition"
-        )
+    # Recall guard, bound to what the score actually CLAIMS. `confidence_score`
+    # is arithmetic over the finding severities, so the score names its findings
+    # exactly: 1 = at least one P0, 2 = two or more P1, 3 = exactly one P1. The
+    # guard therefore demands a matching disposed thread *per claimed finding*.
+    # A single "did we see any P0/P1 thread" boolean could not tell those apart,
+    # and let one disposed P1 vouch for a different, unanchored P0 — the recall
+    # failure this path exists to prevent. Findings the reviewer could not
+    # anchor to a diff line ("Comments outside the diff" in the summary body)
+    # and a reviewer that silently stopped posting both land here, and both
+    # fail closed.
+    required = {1: ("P0", 1), 2: ("P1", 2), 3: ("P1", 1)}.get(score or 0)
+    if required:
+        need_severity, need_count = required
+        have = disposed[need_severity]
+        if have < need_count:
+            return (
+                f"reports {need_count} {need_severity} finding(s) on this head "
+                f"but only {have} disposed {need_severity} thread(s) verify it"
+            )
     return None
 
 
@@ -1455,15 +1475,22 @@ def fetch_ai_review_status(
                 f"AI review score is not an integer ({type(score).__name__}: {score!r})"
             ),
         }
-    # A score above the rubric (6, 7, 100) is corruption, not a better review —
-    # `confidence_score` emits 1-5 and nothing else, and `<` would wave it
-    # through while every other check (author, SHA, consensus) still passed.
-    # Unlike the old score floor this does NOT reject a 4/5: that is the
-    # sampling event the disposition check below is designed to tolerate.
-    if score > AI_REVIEW_CLEAN_SCORE:
+    # A score outside the rubric is corruption, not a review — `confidence_score`
+    # emits 1-5 and nothing else, and a bare `<` would wave it through while
+    # every other check (author, SHA, consensus) still passed. The bound is
+    # two-sided on purpose: dropping the old equality check for the upper bound
+    # alone would newly admit `0` and negatives, which read as "a very bad score
+    # the disposition check can still clear" rather than as the impossible
+    # markers they are. Unlike the old score floor this does NOT reject a 4/5 —
+    # that is the sampling event the disposition check below is designed to
+    # tolerate.
+    if not AI_REVIEW_MIN_SCORE <= score <= AI_REVIEW_CLEAN_SCORE:
         return {
             "accepted": False,
-            "detail": f"AI review score {score}/5 outside the rubric (max 5/5)",
+            "detail": (
+                f"AI review score {score}/5 outside the rubric "
+                f"({AI_REVIEW_MIN_SCORE}-{AI_REVIEW_CLEAN_SCORE})"
+            ),
         }
 
     # Disposition, not score. A score of 4 (surviving P2s) or even 3 (a P1 that
--- a/scripts/github/self-merge-check.py
+++ b/scripts/github/self-merge-check.py
@@ -1129,7 +1129,7 @@ def _ai_review_disposition_shortfall(
     current_findings = (marker or {}).get("findings")
     _, threads = review_data
     disposed: dict[str, int] = {"P0": 0, "P1": 0}
-    disposed_fps: set[str] = set()
+    disposed_fps: set[tuple[str, str]] = set()
     for thread in threads:
         comments = thread.get("comments") or {}
         nodes = comments.get("nodes") or []
@@ -1186,7 +1186,15 @@ def _ai_review_disposition_shortfall(
         if severity in disposed and not (fp and fp in auto_resolved):
             disposed[severity] += 1
             if fp:
-                disposed_fps.add(fp)
+                # Keyed by (fp, severity), not fp alone. A fingerprint outlives
+                # a severity: the reviewer can re-raise the same finding at a
+                # higher severity on a later head, and the old thread still
+                # renders the OLD one. Matching on fp alone would let a
+                # disposed P1 thread satisfy a `findings` entry that now calls
+                # that same fingerprint a P0 — which is reachable exactly when
+                # the re-raised P0 could not be anchored, i.e. when the recall
+                # guard is the only thing left checking.
+                disposed_fps.add((fp, severity))
 
     def _blocking_findings(
         marker_findings: list[Any],
@@ -1229,7 +1237,7 @@ def _ai_review_disposition_shortfall(
         missing = [
             (fp, severity)
             for fp, severity in blocking_findings
-            if fp not in disposed_fps
+            if (fp, severity) not in disposed_fps
         ]
         if missing:
             counts: dict[str, int] = {"P0": 0, "P1": 0}
--- a/scripts/github/self-merge-check.py
+++ b/scripts/github/self-merge-check.py
@@ -1063,7 +1063,12 @@ def _ai_review_disposition_shortfall(
       ``auto_resolved`` ledger — that is evidence-based, the code change is
       the answer, no reply is owed);
     - **surviving P2s do not block** — they regenerate on unchanged code, so
-      gating on them is the treadmill failure mode on file.
+      gating on them is the treadmill failure mode on file;
+    - **an unreadable severity is not a P2** — a thread carrying our finding
+      marker whose ``**P0**``/``**P1**``/``**P2**`` text does not parse blocks
+      and must be disposed like a P0/P1. Defaulting it to P2 would be a
+      fail-open on the severity parse, in a wire protocol whose renderer lives
+      in another repo.
 
     Each finding thread carries its severity in the root body (``**P0**`` etc.)
     and its resolution state on the thread itself; ``totalCount`` on the
@@ -1078,7 +1083,16 @@ def _ai_review_disposition_shortfall(
     diff line ("Comments outside the diff" in the summary body) and with a
     reviewer that silently stopped posting. Failing open there would let a
     genuine P0/P1 merge unreviewed — the exact recall failure this path exists
-    to prevent.
+    to prevent. Threads the reviewer itself retired (``auto_resolved``) do not
+    satisfy that guard: they are its own bookkeeping for findings that stopped
+    reproducing, so by construction they are not the P0/P1 the score reports.
+
+    **Known residual, tracked separately.** A P0/P1 thread disposed on an
+    *earlier* head stays resolved-and-replied forever and cannot be told apart
+    from one disposed on this head, so it can still satisfy the recall guard
+    for a *different*, unanchored P0/P1 the reviewer reports now. Closing that
+    needs the marker to publish the current pass's finding fingerprints, which
+    is a change to the reviewer, not to this gate.
     """
     if review_data is None:
         return "could not read review thread state"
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
**Merge-order note (automated, from the overnight operator session).**

This PR and gptme/gptme-contrib#1382 both rewrite `scripts/github/self-merge-check.py`, and their hunks overlap: both modify `evaluate_pr`, and both insert large blocks in the ~1000-1460 range. Right now each reports `MERGEABLE / CLEAN` **against master independently** — that is the trap, not the all-clear.

Two consequences for whoever merges first:

1. **The second PR will need a rebase**, and its existing review verdict is void at the new head.
2. **Re-review after the rebase, do not carry the old score forward.** Earlier today gptme-contrib#1391 sat at 4/5, was rebased onto merged #1393, and a fresh review returned **2/5 with two P1s** — interaction bugs that existed on neither PR alone. The staleness check (`reviewed <sha>, head is <sha>`) is what caught it.

Both PRs also touch the merge gate itself, so both correctly trip the sensitive-path disqualifier and want a human merge regardless of score. That classification is right, not noise: self-merging a change to the self-merge rules under those same rules is the conflict of interest the category exists to catch.

scripts/github/self-merge-check.py
```diff
@@ -1014,6 +1015,117 @@ def _is_ai_review_thread(comments: list[dict[str, Any]]) -> bool:
...
+    for thread in threads:
+        comments = thread.get("comments") or {}
+        nodes = comments.get("nodes") or []
+        if not nodes:
+            continue
+        body = str(nodes[0].get("body") or "")
+        if _AI_REVIEW_FINDING_MARKER not in body:
+            continue  # not one of our findings
+        m = _AI_REVIEW_SEVERITY_RE.search(body)
+        severity = m.group(1) if m else "P2"
```
TimeToBuildBob: ❌ **P1** — A finding thread whose root body contains the AI-review marker but no `**P0**`/`**P1**` severity text is defaulted to `"P2"` (line 1101: `severity = m.group(1) if m else "P2"`). The disposition loop then skips it as a surviving P2, so it never blocks. If the reviewer posts a finding without the severity string (formatting bug, truncated body, or a new severity format), the gate treats it as a non-blocking P2, and a score of 4 or 5 on the marker would still allow the merge — potentially letting a real critical finding pass unreviewed. Worked example: thread body is `<!-- bob-ai-review-finding -->\nSecurity vulnerability in auth` with no `**P0**`; `m` is None, `severity` becomes `"P2"`, the thread is ignored, and if the marker‑score is 4 (e.g., because other P2s are present), `_ai_review_disposition_shortfall` returns None and `fetch_ai_review_status` accepts the PR. This is a fail‑open on severity extraction and should fail closed (e.g., treat a missing severity as P0).

scripts/github/self-merge-check.py
```diff
@@ -1014,6 +1015,117 @@ def _is_ai_review_thread(comments: list[dict[str, Any]]) -> bool:
...
+        if severity not in ("P0", "P1"):
+            continue  # surviving P2s never block
+        saw_blocking_finding = True
+        if not thread.get("isResolved"):
+            return f"AI review {severity} finding is not resolved"
+        total = comments.get("totalCount")
+        replied = isinstance(total, int) and total >= 2
```
TimeToBuildBob, line 1173: ❌ **P1** — The reply check uses only `comments.get("totalCount") >= 2` (line 1108) to decide that a resolved P0/P1 is disposed, without verifying who authored the second comment. The reviewer itself can reply to its own finding thread (e.g., a generic "resolved") and then mark the thread resolved; `totalCount` becomes 2, the loop treats it as replied, and the disposition gate accepts — without any human acknowledgment. This defeats the distinction the PR draws between the reviewer's evidence‑based auto‑resolution (which requires a fingerprint in `auto_resolved`) and a casual reply with no human input. The `auto_resolved` exemption is narrower, but the plain reply path is also available to the same actor, so the requirement that `every finding addressed AND disposed` against real thread state is not actually enforced for the anonymous-reply case. Consequence: a buggy or malicious self-rubber-stamping bot can merge a PR with open critical findings merely by posting a self-reply and resolving the thread.

scripts/github/self-merge-check.py
```diff
@@ -1014,6 +1015,117 @@ def _is_ai_review_thread(comments: list[dict[str, Any]]) -> bool:
...
+    if review_data is None:
+        return "could not read review thread state"
+    auto_resolved = {
+        str(fp)
+        for fp in ((marker or {}).get("auto_resolved") or [])
+        if isinstance(fp, str) and fp
+    }
+    _, threads = review_data
+    saw_blocking_finding = False
+    for thread in threads:
```
TimeToBuildBob, line 1133: ⚠️ **P2** — The disposition loop iterates over all review threads without any age or review‑cycle filter (line 1092), unlike `fetch_greptile_status` which filters by `latest_review_time`. A resolved P0/P1 thread left over from an earlier review (e.g., a prior head where the finding was legitimately fixed) is counted as a disposed P0/P1 for the current head. This can mask the recall guard: if the current head has a new, unanchored P0/P1 finding (only in the summary, no thread), the marker score would be <= 3, but because a stale resolved thread exists, `saw_blocking_finding` is set to True, the recall guard does not fire, and the gate accepts a head that actually has an open critical (unanchored) finding. The PR description relies on the reviewer deleting superseded inline comments (line 60-61), but the gate should not depend on that side effect; it should only count threads from the current review head (e.g., by comparing thread creation time against the marker's SHA or the review submission time).

TimeToBuildBob:
## 🤖 AI code review

Replaces the AI-review self-merge gate's literal 5/5 score requirement with a disposition check that reads review-thread state: no open P0/P1, each finding resolved with a reply (or auto-resolved), P2s non-blocking, fail-closed on unreadable state. Adds `totalCount` to the review-threads GraphQL query, adds the `_ai_review_disposition_shortfall` function, and threads `review_data` from `evaluate_pr` into `fetch_ai_review_status`. The marker gains an optional `findings` fingerprint list for per-fingerprint disposition matching, with a legacy fallback for markers lacking it.

- `scripts/github/self-merge-check.py` — Adds disposition-based acceptance logic, thread parsing for severity/fingerprint/author, score bounds, and a `review_data` parameter to `fetch_ai_review_status`.
- `tests/test_self_merge_ai_review_gate.py` — Adds tests for disposition-based acceptance: open/resolved/replied/unreplied threads, auto-resolved, unreadable severity, forged threads, and score corruption.

</details>

| `99b1127e546a` | 2/5 | 3 | llm | 2026-08-10 22:41 UTC |
| `543ad672e7e3` | 3/5 | 2 | llm | 2026-08-11 01:45 UTC |
| `549b5ec4e092` | 1/5 | 1 | llm | 2026-08-11 01:58 UTC |
| `188899076141` | 3/5 | 1 | llm | 2026-08-11 02:11 UTC |
| `743848ba3a3d` | 3/5 | 1 | llm | 2026-08-11 02:22 UTC |

</details>

<sub>Reviewed `dc2c556f1633` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 261s · advisory, a human still decides · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

scripts/github/self-merge-check.py
TimeToBuildBob: Accepted and fixed in 543ad672.

The defaulting was a fail-open on the parse, and the reachable path is narrower than described but real. `normalize_severity` (`ai_review_lib.py:85`) clamps every severity to `{P0,P1,P2}` before render, and `ai-review.py:1914` writes `{icon} **{sev}** — `, so a body with no `**Pn**` means the wire protocol drifted or the body is malformed — not a P2. The renderer lives in a *different repo* from this gate (hence the "Keep the exact string" note on `INLINE_FINDING_MARKER`), which is exactly the condition under which a cross-repo string contract breaks silently.

One correction to the worked example: with a genuine P0 present, `confidence_score` derives the score from the finding severities, so the marker would read <= 3, not 4 — and the recall guard would fire *unless* another P0/P1 thread is present to set `saw_blocking_finding`. So the hole needs two findings, not one. That narrows it; it does not close it.

Fix: an unreadable severity no longer collapses to P2. It blocks, and it clears the same way a P0/P1 does — resolved, with a reply — so it is fail-closed without becoming a permanent wedge. A *readable* severity outside P0/P1 (P2 today, P3 if the rubric ever grows) still never blocks, so widening the rubric cannot deadlock the gate. Tests: `test_finding_without_severity_text_blocks_instead_of_defaulting_to_p2`, `test_finding_without_severity_text_can_still_be_disposed`, `test_severity_below_p1_still_never_blocks`.

scripts/github/self-merge-check.py
TimeToBuildBob: Refused — the proposed check cannot discriminate, and the threat model it assumes is not the one this gate operates in.

**1. An author check has no signal to read here.** The reviewer posts findings and `ai-review-dispose.py` posts replies through the *same GitHub account* (`TimeToBuildBob`, via `gh` with one token) — this is a self-merge gate, so the author, the reviewer and the disposer are one identity by construction. "Verify who authored the second comment" would either always pass or always fail; it cannot separate the two cases.

**2. The self-reply the finding describes does not exist.** I grepped every writer of these threads. The reviewer posts finding *roots* only (`ai-review.py:1914`) and retires stale ones by resolving, never by replying (`reconcile_inline_threads`, `_RESOLVE_MUTATION`). The single code path that posts a reply into a finding thread is `ai-review-dispose.py:post_thread_reply`, which requires >= 40 characters of real prose (`main()`: *"a disposition needs a real reply"*) and deliberately places its marker **last** so the reply is not machine-classified. There is no generic "resolved" self-reply for `totalCount` to count.

**3. Reply AND resolution is the specified contract, not an oversight.** From `ai-review-dispose.py`'s docstring, Erik, 2026-08-10: *"a reply AND a resolution"*. The `auto_resolved` exemption is narrower on purpose — it is the one case where **no** reply is owed because the code change itself is the evidence. That asymmetry is the design, not a gap in it.

**4. The threat model does not hold.** "A malicious self-rubber-stamping bot" that can post as this account can also call `gh pr merge` directly; this gate is not an authorization boundary against its own operator. What it *is* defending against is a mechanical fail-open — a clean-looking score with no evidence behind it — and that is what the disposition check reads.

What would be a real improvement is excluding *machine* comments from the reply count the way `ai-review.py:_classify_review_comments` does. It buys nothing today (nothing machine-writes into these threads) and costs a second GraphQL round-trip to fetch all thread comments rather than `totalCount`, so I am not adding it on speculation. If a machine reply path ever lands, that is the change to make.

scripts/github/self-merge-check.py
TimeToBuildBob: Partially accepted — the strongest half is fixed in 543ad672; the rest cannot be fixed from the gate side, and I would rather say so than pretend otherwise.

**Fixed.** Threads the reviewer itself retired (fingerprint in the marker's `auto_resolved` ledger) no longer set `saw_blocking_finding`. That ledger is the reviewer's own record that a finding *stopped reproducing*, so such a thread is by construction not the P0/P1 a `score <= 3` reports for this head — letting it satisfy the recall guard was letting the reviewer's bookkeeping stand in for a disposition. Test: `test_auto_resolved_thread_does_not_satisfy_the_recall_guard`. This does not over-block: if the score still claims a P0/P1, that finding is either open (refused earlier in the loop) or unanchored, which is precisely what the guard exists for.

**Not fixable here — and the suggested mechanism does not work.** "Only count threads from the current review head (compare thread creation time against the marker's SHA or the review submission time)" would break the common case: a finding that *still reproduces* keeps its **original** thread untouched (`reconcile_inline_threads` matches and re-uses live threads rather than re-posting), so a live P0/P1's thread can be arbitrarily old. A time filter would drop exactly the threads that must block.

The residual is narrower than the finding states, for the same reason: `open_finding_threads` skips resolved threads, so a re-found finding always gets a **new, open** thread — and an open P0/P1 is refused before the guard is reached. What survives is one case: a P0/P1 disposed (replied + resolved) on an earlier head, still resolved-and-replied forever, satisfying the guard for a *different*, unanchored P0/P1 reported now.

Closing that needs the marker to publish the fingerprints of the current pass's findings so the gate can require a disposition *per current finding* rather than "some P0/P1 thread exists". That is a change to `ai-review.py` (a different repo), not to this gate. Named in the `_ai_review_disposition_shortfall` docstring and tracked as reviewer-side work rather than left implicit.

scripts/github/self-merge-check.py
```diff
@@ -1014,6 +1015,150 @@ def _is_ai_review_thread(comments: list[dict[str, Any]]) -> bool:
...
+        if not replied:
+            if fp and fp in auto_resolved:
+                continue  # reviewer-retired because the fix stopped it reproducing
+            return f"{label} resolved without a reply (not disposed)"
+
+    # Recall guard: the marker says a P0/P1 is on this head, but no P0/P1
+    # thread is present to confirm it was disposed (unanchored finding, or the
+    # reviewer stopped posting). Fail closed — a real P0/P1 must not merge on
+    # the strength of an empty thread list.
+    if score is not None and score <= 3 and not saw_blocking_finding:
```
TimeToBuildBob: ❌ **P1** — The recall guard keys on a single boolean `saw_blocking_finding` set by ANY thread carrying P0/P1 (or unreadable) severity, with no binding to what the score claims or whether the thread is from the current head. A marker reporting a P0 (`score=1` or `2`) with an unanchored P0, plus one resolved-and-replied P1 thread, passes: the P1 sets `saw_blocking_finding=True`, the loop falls out without returning, and the line 1154 guard is skipped — the unanchored P0 is never disposed yet the PR is accepted. The docstring frames this as an 'earlier head' residual, but the code has no head/severity correspondence at all; a same-head disposed P1 thread satisfies the guard for a different P0 or unanchored P1. This is the exact recall failure the gate exists to prevent: a genuine P0/P1 merges unreviewed. _(3/3 passes)_

scripts/github/self-merge-check.py
```diff
@@ -1297,17 +1455,36 @@ def fetch_ai_review_status(
...
-    # corrupted or foreign marker, and `<` would wave it through as clean while
-    # every other check (author, SHA, consensus) still passed.
-    if score != AI_REVIEW_CLEAN_SCORE:
-        below = "below" if score < AI_REVIEW_CLEAN_SCORE else "outside the rubric,"
+    # A score above the rubric (6, 7, 100) is corruption, not a better review —
+    # `confidence_score` emits 1-5 and nothing else, and `<` would wave it
+    # through while every other check (author, SHA, consensus) still passed.
+    # Unlike the old score floor this does NOT reject a 4/5: that is the
+    # sampling event the disposition check below is designed to tolerate.
+    if score > AI_REVIEW_CLEAN_SCORE:
```
TimeToBuildBob: ⚠️ **P2** — The corruption guard was narrowed from `score != AI_REVIEW_CLEAN_SCORE` (which rejected 0 and every non-5 value) to `score > AI_REVIEW_CLEAN_SCORE`, so a marker carrying `score: 0` or a negative value now passes both the `type(score) is int` check and the upper-bound check and reaches `_ai_review_disposition_shortfall`. Whenever a P0/P1 thread exists that is resolved and replied-to, `saw_blocking_finding` is set True, which suppresses the `score <= 3` recall guard, so the gate returns accepted with detail `AI review 0/5 at current head ... findings disposed`. A 0/5 review — which the rubric never emits and the old code refused as 'below 5/5' — can now satisfy the self-merge gate whenever one disposed P0/P1 thread is present. The PR hardened against corrupted scores above 5 but left the symmetric low-side corruption admissible; the one-line guard should bound both ends. _(2/3 passes)_

scripts/github/self-merge-check.py
TimeToBuildBob: Accepted — this is sharper than the residual I documented, and you are right that the code had no correspondence at all. Fixed in 549b5ec4.

The single `saw_blocking_finding` boolean answered "did any P0/P1 thread exist", which is not the question. `confidence_score` is arithmetic over the finding severities (`ai_review_lib.py`), so a score names its findings exactly: 1 = at least one P0, 2 = two or more P1, 3 = exactly one P1. The guard now demands a matching disposed thread **per claimed finding** — score 1 requires a disposed P0, score 2 requires two disposed P1s, score 3 requires one. Your worked example (score 1 or 2, unanchored P0, one resolved-and-replied P1) now refuses with `reports 1 P0 finding(s) on this head but only 0 disposed P0 thread(s) verify it`.

Two exclusions from the count, both fail-closed: `auto_resolved` threads (the reviewer's own bookkeeping for findings that stopped reproducing) and threads whose severity did not parse (crediting one against a claimed P0 would be guessing in the merge direction).

Tests: `test_disposed_p1_does_not_vouch_for_an_unanchored_p0`, `test_two_p1_score_needs_two_disposed_p1_threads`, `test_disposed_p0_satisfies_a_p0_score`, `test_unreadable_severity_thread_does_not_credit_a_claimed_p0`.

The residual is now only what threads genuinely cannot express: they carry no head, so a P0/P1 disposed on an *earlier* head still counts for a same-severity finding claimed now. Matching per *finding* rather than per severity needs the marker to publish the current pass's fingerprints — reviewer-side work, tracked separately, and named in the docstring rather than left implicit.

scripts/github/self-merge-check.py
TimeToBuildBob: Accepted and fixed in 549b5ec4. Correct on both the mechanism and the consequence.

Narrowing `score != AI_REVIEW_CLEAN_SCORE` to `score > AI_REVIEW_CLEAN_SCORE` was a deliberate change to stop rejecting a 4/5, but it took the low side down with it: `0` and negatives passed the `type(score) is int` check and the upper bound, reached the disposition check, and with one disposed P0/P1 thread present would have been accepted — printing `AI review 0/5 at current head ... findings disposed`, which is nonsense on its face.

The rubric emits 1-5 and abstain is `None`, never 0, so the bound is now two-sided: `AI_REVIEW_MIN_SCORE <= score <= AI_REVIEW_CLEAN_SCORE`, with the reason for two-sidedness written at the constant so the next narrowing does not repeat it. Test: `test_scores_below_the_rubric_are_rejected` parametrised over 0, -1, -5, each with a disposed P1 thread present so it fails for the bound rather than incidentally via the recall guard.

scripts/github/self-merge-check.py
```diff
@@ -1014,6 +1021,164 @@ def _is_ai_review_thread(comments: list[dict[str, Any]]) -> bool:
...
+    }
+    _, threads = review_data
+    disposed: dict[str, int] = {"P0": 0, "P1": 0}
+    for thread in threads:
+        comments = thread.get("comments") or {}
+        nodes = comments.get("nodes") or []
+        if not nodes:
+            continue
+        body = str(nodes[0].get("body") or "")
+        if _AI_REVIEW_FINDING_MARKER not in body:
```
TimeToBuildBob, line 1139: 🛑 **P0** — The disposition gate treats any review thread whose root body contains the AI finding marker as a legitimate finding, without verifying the thread's author. An attacker with write access to the repo can open a review thread, post a root comment containing `<!-- bob-ai-review-finding -->` and `**P0**`, resolve it, and add a reply. The gate then counts that thread as a disposed P0, satisfying the recall guard for a score of 1 even when the reviewer's actual P0 was unanchored and never addressed. This merges an unreviewed P0/P1, defeating the recall guard's purpose. The summary marker is author-checked, but the thread markers are not; the GraphQL query already fetches author, so the check is simply missing here. _(3/3 passes)_

TimeToBuildBob:
**Three-way overlap on `scripts/github/self-merge-check.py` — recommended merge order.**

gptme/gptme-contrib#1382, #1405 and #1407 all rewrite this file, and all three correctly trip the sensitive-path disqualifier (they modify the merge gate itself), so all three need a human merge regardless of score. Measured hunk overlap:

| PR | regions touched | function |
|---|---|---|
| #1382 | 630, **1372–1592**, 1871 | `fetch_unresolved_human_threads`, `evaluate_pr` |
| #1405 | 120, 701, **1014–1214** (200-line insert), 1245–1333, 1910 | `fetch_ai_review_status`, `evaluate_pr` |
| #1407 | **1130–1140**, **1261–1274** | `fetch_ai_review_status` |

**#1405 and #1407 conflict directly.** #1407 edits line ~1130, which lands *inside* #1405's 200-line insertion at 1014, and both edit `fetch_ai_review_status` across 1245–1274. This is not a line-offset shift — it is the same code.

**#1382 is in a different function** and should apply with offsets only.

Suggested order — **#1382 → #1405 → #1407**:

1. **#1382** first: least entangled, already 5/5 with 0 findings, CI green, threads resolved.
2. **#1405** second: it is the large structural change (disposition-based gate). Merging the big one before the small one means the *small* PR rebases, not the large one.
3. **#1407** last: 2 hunks, cheapest to rebase.

**Re-review each after its rebase — do not carry the prior verdict forward.** On 2026-08-10 gptme-contrib#1391 sat at 4/5, was rebased onto merged #1393, and a fresh review returned **2/5 with two P1s** — interaction bugs present on neither PR alone. The staleness check (`reviewed <sha>, head is <sha>`) is what caught it.

Worth considering instead: #1405 and #1407 are close enough in scope that folding #1407's two P2 fixes into #1405 may be cheaper than a rebase plus a second review cycle.

scripts/github/self-merge-check.py
TimeToBuildBob: Accepted and fixed in 18889907. Correct, and correctly rated: the marker is a *label*, not a signature.

`_ai_review_disposition_shortfall` read any thread whose root body contained `<!-- bob-ai-review-finding -->` as one of ours, so a forged thread — marker, `**P0**`, resolved, one reply — was a disposition the recall guard accepted for a P0 the reviewer really did report and nobody addressed. Your last sentence is the clinching one: the summary marker has been author-checked since it was introduced (`_fetch_ai_review_marker`), the author is already in the payload `_fetch_greptile_review_data` fetches, and the check was simply absent on this path. No new round-trip to add it.

`expected_author` is now threaded from `fetch_ai_review_status` into the disposition loop, and a thread from anyone else is not read as ours.

One deliberate detail beyond what you asked for: the check cuts **both** ways. A forged *unresolved* P0 is also not our finding, so a stranger cannot wedge the AI-review path by pasting the marker into an open thread — that would have traded a merge-forgery vector for a denial-of-merge one. Such a thread remains an unresolved review thread to the human-thread check, which is where a stranger's comment belongs.

Tests: `test_forged_finding_thread_does_not_dispose_a_claimed_p0`, `test_our_own_finding_thread_still_counts` (negative control — same thread, our author, still disposes), `test_forged_open_finding_thread_does_not_block_either`.

scripts/github/self-merge-check.py
```diff
@@ -1014,6 +1024,217 @@ def _is_ai_review_thread(comments: list[dict[str, Any]]) -> bool:
...
+                continue
+            if not isinstance(severity, str):
+                continue
+            if severity not in AI_REVIEW_BLOCKING_SEVERITIES:
+                continue
+            out.append((fp, severity))
+        return out
+
+    blocking_findings = _blocking_findings(current_findings)
+    if blocking_findings:
```
TimeToBuildBob: ❌ **P1** — When the marker contains a `findings` key that is empty or has P0/P1 entries without a fingerprint, `_ai_review_disposition_shortfall` computes an empty `blocking_findings` list and falls through to the score-based legacy guard (`required = {1: ...}`). That guard counts any disposed thread of the matching severity, without checking that the thread actually corresponds to the current head's claimed finding. A marker that inconsistently publishes `findings: []` with `score: 1` is therefore accepted whenever any P0 thread on the PR is resolved and replied — even if that thread is stale and the current head's P0 was never addressed. Similarly, a finding entry missing `fp` is silently dropped from `blocking_findings`, so a stale disposed thread can satisfy the score claim. This is a fail-open on the exact recall hole the guard exists to close. The root cause is treating a missing `findings` key and an explicitly empty list identically via `(marker or {}).get("findings") or []`; the code must distinguish them and treat any presence of `findings` as authoritative. _(3/3 passes)_

TimeToBuildBob: Accepted and fixed in 743848ba. The root cause you named is exactly right: `(marker or {}).get("findings") or []` conflates a present-but-empty list with an absent key, and the fallback it lands in is strictly weaker.

Both halves were fail-open in the same direction — anything that shrinks the published finding set shrinks the set of dispositions demanded:

- `findings: []` with `score: 1` fell through to the score-only guard, which accepts *any* disposed P0 thread, so a stale already-disposed P0 cleared a head whose own P0 was never addressed.
- An entry missing `fp` (or with a non-string severity) was silently skipped, with the same effect on a per-entry basis.

Presence is now the signal:

- a **missing** `findings` key still uses the severity-count guard, so markers written before the key existed keep working — covered by an explicit negative-control test, since breaking that would break the gate on every PR reviewed by an older reviewer;
- a **present** key is authoritative even when empty;
- an unreadable entry refuses (`marker publishes a findings entry we cannot read`) rather than being dropped;
- and the published list must account for the score. `confidence_score` is arithmetic over the same findings, so a marker scoring 1/5 while publishing no P0 entry is self-contradictory, not a clean head. That is the check that actually kills your `findings: []` + `score: 1` example, independently of what threads exist.

One thing your finding did not mention but that mattered for the fix: the test fixture published `findings: []` by *default*, which is a contradiction for any score below 4 — building that shape implicitly is how the hole stayed invisible. The default is now a legacy marker (no key), and per-fingerprint tests pass `findings=[...]` consistent with the score they set.

Tests: `test_empty_findings_list_does_not_fall_back_to_the_score_guard`, `test_findings_entry_without_a_fingerprint_fails_closed`, `test_findings_must_account_for_the_score`, `test_missing_findings_key_still_uses_the_legacy_score_guard`.

scripts/github/self-merge-check.py
```diff
@@ -1014,6 +1024,253 @@ def _is_ai_review_thread(comments: list[dict[str, Any]]) -> bool:
...
+        #   * `auto_resolved` is the reviewer's own record that this finding
+        #     stopped reproducing and it retired the thread as bookkeeping, so
+        #     by construction it is not a finding on this head and must not
+        #     stand in for a real disposition;
+        #   * an unreadable severity could be anything, so crediting it against
+        #     a P0 the score reports would be guessing in the merge direction.
+        if severity in disposed and not (fp and fp in auto_resolved):
+            disposed[severity] += 1
+            if fp:
+                disposed_fps.add(fp)
```
TimeToBuildBob: ❌ **P1** — The per-fingerprint disposition check verifies only that a finding's fingerprint appears in `disposed_fps`, never that the disposed thread's parsed severity equals the severity recorded in the marker's `findings` entry. A thread whose root body renders `**P1**` is added to `disposed_fps` via `disposed_fps.add(fp)` inside the `if severity in disposed` block, so a resolved-and-replied P1 thread with a given fingerprint can satisfy a `findings` entry declaring that same fingerprint as P0. Concretely, if the AI re-evaluates a finding and changes its severity from P0 to P1 without updating the thread body (or vice versa), a P0 disposition is credited on the strength of a P1 thread, and the recall guard fails open for the actual P0. The merge proceeds with a P0 that was never disposed. Fix: require the thread's severity to match the entry's severity when adding to `disposed_fps` (e.g., store `(fp, severity)` in `disposed_fps`) and check for exact `(fp, severity)` membership in the `missing` list. _(3/3 passes)_

TimeToBuildBob: Accepted and fixed in dc2c556f, including the suggested shape — `disposed_fps` now holds `(fp, severity)` pairs and `missing` checks exact pair membership.

The reachability argument is worth writing down, because it is narrower than "the AI changes its mind" and that makes it *more* interesting rather than less. A disposed thread is never re-matched (`open_finding_threads` skips resolved ones), so when a finding is re-raised the reviewer posts a **new** thread — which is open, and an open P0/P1 is refused earlier in the loop. So the mismatch cannot bite while the re-raise lands normally.

It bites in exactly one place: the re-raised P0 could not be anchored to a diff line (the line moved out of the diff — "Comments outside the diff"), so no new thread exists, and the only thread carrying that fingerprint is the older, resolved-and-replied **P1**. That is the case where the recall guard is the *only* thing still checking, and it was the case where the guard silently passed. A fingerprint outlives a severity; keying on it alone assumed otherwise.

Tests: `test_disposed_p1_thread_does_not_satisfy_a_p0_findings_entry`, plus `test_matching_severity_still_disposes` as the negative control — the same fingerprint at the severity the marker actually claims still disposes it, so the tightening costs nothing legitimate.

TimeToBuildBob:
## Status: ready for a human to merge — the gate is holding it correctly

**Head:** `dc2c556f` · **CI:** green · **AI review:** 5/5 at head, 0 open finding threads, 0 unresolved review threads of any kind. All **8** findings raised across six passes are addressed *and* disposed (per-thread reply + resolution).

### What a human needs to do

Read the diff and merge it manually (squash). This PR cannot legitimately self-merge, now or at any score.

### Why the gate refuses, and why that is right

```
Eligible: NO
- Touches sensitive/security/infra paths
```

It modifies `scripts/github/self-merge-check.py` — **the merge gate itself**. Self-merging a change to the self-merge rules, evaluated by those same rules, is the conflict of interest that category exists to catch. A 5/5 does not dissolve it; if anything the opposite, since the thing being changed is what produced the 5/5. The right disposition of this PR is a human reading the predicate and deciding it is the predicate they want.

(`Cross-repo PR ... not in allowed workspace repos` is an artifact of the checking session's `WORKSPACE_REPO`, not a property of the PR.)

### The review history is the interesting part

| head | score | findings | outcome |
|---|---|---|---|
| `99b1127e` | 2/5 | 3 | 2 fixed, 1 refused |
| `543ad672` | 3/5 | 2 | both fixed |
| `549b5ec4` | 1/5 | 1 (**P0**) | fixed |
| `18889907` | 3/5 | 1 | fixed |
| `743848ba` | 3/5 | 1 | fixed |
| `dc2c556f` | **5/5** | 0 | — |

Seven of eight were accepted and fixed. Every one of them was a **fail-open in the same direction**: something that let the gate accept a head with an unverified P0/P1.

- unreadable severity defaulted to P2 (silently downgrading a possible P0);
- the reviewer's own `auto_resolved` bookkeeping satisfying the recall guard;
- one disposed P1 vouching for a *different*, unanchored P0 (the guard had no correspondence to what the score claimed);
- the score bound going one-sided, newly admitting `0` and negatives;
- **P0:** finding threads never author-checked — anyone with write access could paste the marker and `**P0**`, resolve, reply once, and forge a disposition;
- a present-but-empty `findings` list read as an absent key, dropping to the weaker guard exactly when the marker is least trustworthy;
- dispositions keyed by fingerprint alone, so a disposed P1 satisfied a P0 entry with the same fingerprint.

That is the class this PR is *about*, which is a reasonable argument that the change is directionally right — and also a reasonable argument for reading it carefully rather than trusting the number at the end.

**The one finding I refused** is on its thread with the evidence: requiring the *author* of the second comment on a finding thread to differ from the reviewer. The reviewer and the disposer are the same GitHub account by construction here, so the check cannot discriminate; the reviewer never self-replies (the only writer is `ai-review-dispose.py`, which requires ≥40 chars of prose); and reply-AND-resolution is the specified contract (Erik, 2026-08-10), not an oversight.

### Merge-order hazard — still live, re-verified at these heads

gptme/gptme-contrib#1382 (`f857c4f8`) also rewrites `scripts/github/self-merge-check.py`, also modifies `evaluate_pr`, and both PRs insert large blocks in the same region:

- **#1405** inserts at old-line 1014 (+261 lines) and 1245/1256/1297/1316, plus `evaluate_pr` at 1910
- **#1382** inserts at old-line 1372 (+220 lines), plus `evaluate_pr` at 1871

Both report `MERGEABLE / CLEAN` against master **independently** — that is the trap, not the all-clear. Whichever merges first forces a rebase on the other, and **the rebased one needs a fresh review, not its old score**: earlier today #1391 sat at 4/5, was rebased onto merged #1393, and came back **2/5 with two P1s** — interaction bugs present on neither PR alone. Neither has been rebased onto the other pre-emptively, deliberately: that would move the conflict and invalidate a good review.

### Known residual, tracked

Threads carry no head, so a P0/P1 disposed on an earlier head still counts for a same-severity finding claimed now, *unless* the marker publishes its current findings — which it now can, and the gate consumes when present (`findings` key, authoritative when present, legacy severity-count guard when absent). Publishing that key from the reviewer side is tracked in Bob's workspace as `ai-review-marker-publish-current-finding-fingerprints`.

PR #1406, opened 2026-08-11
fix(dotfiles/hooks): auto-symlink node_modules in linked worktrees

TimeToBuildBob:
## Summary

- Adds node_modules symlinking to the global `post-checkout` hook so git worktrees for JS/TS repos (e.g. `gptme/gptme`) automatically symlink `node_modules` from the base checkout
- Uses `git rev-parse --git-common-dir` vs `--git-dir` for reliable linked worktree detection (the `--show-toplevel` approach is unusable here — returns the worktree path in both main checkout and linked worktree contexts)
- Prevents pre-commit hook failures (exit 127) in fresh worktrees that lead PM workers to use `--no-verify` bypass (forbidden)

## Context

PM workers create git worktrees via `git worktree add` for PR branches. Fresh worktrees lack `node_modules`, so JS pre-commit hooks fail with exit 127. Workers then use `--no-verify` to bypass hooks. This hook fix auto-symlinks `node_modules` from the base checkout on worktree creation, applied globally via `core.hookspath`.

## Test plan

- [ ] Create a worktree from a JS repo with `node_modules`: `git worktree add /tmp/test-wt -b test origin/master`
- [ ] Verify `node_modules` symlink appears: `ls -la /tmp/test-wt/node_modules`
- [ ] Verify pre-commit hooks pass in the worktree without `--no-verify`
- [ ] Verify hook is a no-op in Python-only repos (no `node_modules` in base)

### DIFF
--- a/dotfiles/.config/git/hooks/post-checkout
+++ b/dotfiles/.config/git/hooks/post-checkout
@@ -32,6 +32,16 @@ _gw_common_dir=$(git rev-parse --git-common-dir 2>/dev/null || true)
 if [ -n "$_gw_git_dir" ] && [ "$_gw_git_dir" != "$_gw_common_dir" ]; then
     _gw_main_root=$(realpath "$_gw_common_dir/.." 2>/dev/null || true)
     _gw_wt_root=$(git rev-parse --show-toplevel 2>/dev/null || true)
+    # `<common-dir>/..` is the main CHECKOUT only when the shared git dir is a
+    # `.git` inside one. Off a BARE repo (`/srv/repo.git`) it is that repo's
+    # parent directory instead — an unrelated tree. Requiring a `.git` there
+    # (a dir normally, a file when the main checkout is itself submoduled)
+    # stops two things: symlinking a *sibling project's* node_modules into the
+    # worktree, and running `find` across something like /home/bob on every
+    # single checkout.
+    if [ ! -e "$_gw_main_root/.git" ]; then
+        _gw_main_root=""
+    fi
     if [ -n "$_gw_main_root" ] && [ -n "$_gw_wt_root" ] && [ "$_gw_main_root" != "$_gw_wt_root" ]; then
         while IFS= read -r -d '' _gw_base; do
             _gw_rel="${_gw_base#"$_gw_main_root"/}"
--- a/dotfiles/.config/git/hooks/post-checkout
+++ b/dotfiles/.config/git/hooks/post-checkout
@@ -19,6 +19,22 @@ fi
 # Auto-symlink node_modules in linked git worktrees to prevent --no-verify bypass.
 # Proper worktree detection via git-common-dir vs git-dir (works for all git versions).
 #
+# WHAT YOU ARE TRADING: this is a SHARED tree, not a private copy. Every linked
+# worktree points at the ONE node_modules in the main checkout, so anything that
+# writes through it — `npm install`, a pnpm store write, a postinstall script,
+# `node_modules/.cache` — writes into the main checkout, and two worktrees doing
+# that at once race on the same files. That is accepted deliberately: the tree is
+# here to be READ, by lint/format hooks resolving binaries, and the alternative it
+# replaces is not an isolated tree per worktree — a fresh worktree has NO
+# node_modules at all, which fails those hooks with exit 127 and sends the worker
+# to `--no-verify`, i.e. no hooks whatsoever. A shared read-mostly tree is the
+# better of those two, not a free lunch.
+#
+# So: DO NOT run `npm install` (or any installer) inside a linked worktree. If you
+# need one, `rm` the symlink first and let the installer create a real directory —
+# an existing non-symlink node_modules is left alone by the loop below, so a
+# worktree that opts out this way stays opted out across checkouts.
+#
 # ORDERING IS LOAD-BEARING: this runs BEFORE the branch-base validation below,
 # which returns `exit 0` on detached HEAD, on master/main, and when origin has
 # no default branch. Placing the symlinking after those guards makes it a
--- a/dotfiles/.config/git/hooks/post-checkout
+++ b/dotfiles/.config/git/hooks/post-checkout
@@ -16,6 +16,41 @@ if [ "$is_branch_checkout" != "1" ]; then
     exit 0
 fi
 
+# Auto-symlink node_modules in linked git worktrees to prevent --no-verify bypass.
+# Proper worktree detection via git-common-dir vs git-dir (works for all git versions).
+#
+# ORDERING IS LOAD-BEARING: this runs BEFORE the branch-base validation below,
+# which returns `exit 0` on detached HEAD, on master/main, and when origin has
+# no default branch. Placing the symlinking after those guards makes it a
+# silent no-op for `git worktree add --detach <path> <ref>` (and for
+# `git worktree add <path> origin/master`, which is also detached) — measured:
+# a detached worktree got no symlink at all, which is exactly the case that
+# sends a worker back to --no-verify. It is also independent of the `git fetch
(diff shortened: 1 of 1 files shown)


dotfiles/.config/git/hooks/post-checkout
```diff
@@ -16,6 +16,51 @@ if [ "$is_branch_checkout" != "1" ]; then
...
+        _gw_main_root=""
+    fi
+    if [ -n "$_gw_main_root" ] && [ -n "$_gw_wt_root" ] && [ "$_gw_main_root" != "$_gw_wt_root" ]; then
+        while IFS= read -r -d '' _gw_base; do
+            _gw_rel="${_gw_base#"$_gw_main_root"/}"
+            _gw_target="$_gw_wt_root/$_gw_rel"
+            _gw_parent=$(dirname "$_gw_target")
+            [ -d "$_gw_parent" ] || continue
+            if [ ! -e "$_gw_target" ]; then
+                ln -s "$_gw_base" "$_gw_target"
```
TimeToBuildBob, line 68: ❌ **P1** — The unconditional `ln -s "$_gw_base" "$_gw_target"` at line 52 (and again at 56) makes each linked worktree's `node_modules` a live alias of the single tree in the main checkout. Normal JS tooling — `npm install`, pnpm store writes, postinstall scripts, build caches under `node_modules/.cache`, `npm run dev` — writes through that symlink straight into the main checkout. In the parallel-worktree workflow this PR targets, two workers running `npm install` in different worktrees race on the same files and corrupt packages; and the containment break is structural — one worker (or a bad postinstall) that deletes or mutates the shared tree breaks the main checkout and every other worktree simultaneously, whereas previously each worktree had an isolated tree and the only failure was a local exit 127. _(3/3 passes)_

TimeToBuildBob:
## 🤖 AI code review

Adds a block to the global post-checkout hook that detects linked git worktrees via `git rev-parse --git-dir` vs `--git-common-dir`, then symlinks `node_modules` directories from the main checkout into the worktree, fixing broken symlinks and skipping when the worktree already has a real node_modules. The block is placed before the branch-base validation guards but after the initial `$3` branch-checkout check.

- `dotfiles/.config/git/hooks/post-checkout` — Adds worktree detection and node_modules auto-symlinking logic, including comments about shared-tree trade-offs and ordering.

</details>

| `048f80f4fb98` | 3/5 | 1 | llm | 2026-08-11 01:58 UTC |

</details>

<sub>Reviewed `cc9b9304dd62` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 316s · advisory, a human still decides · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

dotfiles/.config/git/hooks/post-checkout
TimeToBuildBob: Partially accepted — the mechanism is real and is now documented at the code (cc9b9304). Two premises in the finding are wrong, and they matter, because they are what makes this read as a regression rather than as a tradeoff.

**Right:** the symlink makes each worktree's `node_modules` a live alias of the one tree in the main checkout. `npm install`, pnpm store writes, postinstall scripts and `node_modules/.cache` all write through it, and two workers doing that concurrently race on the same files. That property is real, and it is now written where somebody will hit it rather than discovered the hard way.

**Wrong — the baseline.** "Previously each worktree had an isolated tree and the only failure was a local exit 127" is not the state this replaces. A fresh worktree has **no** `node_modules` at all; that is *why* the JS hooks fail with exit 127, and the observed consequence is not a contained local failure but the worker reaching for `--no-verify` — i.e. **no hooks at all**, which is the outcome this repo forbids and the reason the PR exists. An isolated tree per worktree is not the status quo; it is a third option that costs an `npm install` per worktree and that nothing in the workflow was doing.

**Wrong — the containment claim.** "One worker that deletes the shared tree breaks the main checkout and every other worktree" is true of the shared tree, but the comparison is to a world where the hooks were being bypassed wholesale. Weigh a race that requires someone to run an installer inside a worktree against a standing policy violation on every worktree.

**What changed as a result:** the tradeoff is stated at the top of the block — the tree is here to be READ by lint/format hooks resolving binaries, do not run an installer inside a linked worktree — and the opt-out is documented and verified: replace the symlink with a real directory and the loop leaves it alone (`[ ! -e ]` guards the create; only *broken symlinks* are ever replaced). Tested: a worktree given a real `node_modules` kept it across a subsequent checkout.

I am not going to pretend this is a false positive: shared mutable state across worktrees is a genuine cost, and whether it is the right trade against `--no-verify` is a judgement call. It is flagged for the human merging this PR rather than silently resolved.

TimeToBuildBob:
## Status: ready for a human to merge — the gate is holding it correctly

**Head:** `cc9b9304` · **CI:** green · **AI review:** 5/5 at head, 0 open findings, all disposed.

### What a human needs to do

Review the diff and merge it manually (squash). Nothing else is outstanding.

### Why the self-merge gate refuses, and why that is right

```
Eligible: NO
- Files not in any allowed self-merge category: dotfiles/.config/git/hooks/post-checkout
```

This PR changes a **git hook that runs automatically on every checkout, in every repo**, via `core.hookspath`. That is not in any self-merge category and should not be: a hook is executed code with no review step at run time, on machines that never opted into this specific change. A 5/5 does not dissolve that — the category exists precisely so "the reviewer liked it" is not sufficient authority to ship something that then runs everywhere unattended.

(The earlier `CI is not fully green` reason was a stale `gh` cache reading a mid-flight run. Re-checked with `GH_API_CACHE_TTL=0`: CI is green. `Cross-repo PR ... not in allowed workspace repos` is an artifact of the checking session's `WORKSPACE_REPO`, not a property of the PR.)

### What changed since the PR was opened

1. **`4224b511` → the symlinking never ran in the cases that matter.** The block sat after three `exit 0` guards in the branch-base validation. Measured on a scratch repo with `core.hooksPath` pointed at the hook:

   | worktree creation | before | after |
   |---|---|---|
   | `git worktree add <p> -b feat/x master` | ok | ok |
   | `git worktree add --detach <p> master` | **no symlink** | ok |
   | `git worktree add <p> origin/master` | **no symlink** | ok |
   | worktree on a branch named `main`/`master` | **no symlink** | ok |
   | repo with no `origin` default branch | **no symlink** | ok |

   Detached HEAD returns `HEAD` from `--abbrev-ref` and hits the "Detached HEAD" exit; `origin/master` checks out detached too. Moving the block above those guards also decouples it from the `git fetch origin` in that path, so worktree setup no longer waits on the network.

2. **Only source `node_modules` from a real main checkout.** `<git-common-dir>/..` is the main checkout only when the shared git dir is a `.git` inside one; off a bare repo it is that repo's *parent directory*. Without a guard that means a `find -maxdepth 3` across whatever that parent happens to be, on every checkout.

3. **The shared-tree tradeoff is now written at the code** (see the judgement call below).

### The one judgement call for the merger

Review raised (P1) that the symlink makes each worktree's `node_modules` a live alias of the main checkout's, so `npm install`, postinstall scripts and `node_modules/.cache` write through it and concurrent worktrees race. **That property is real** and I did not argue it away — it is documented at the top of the block, along with the opt-out (replace the symlink with a real directory; the loop leaves it alone — verified across a subsequent checkout).

What I did push back on is the baseline the finding assumed. The alternative is *not* an isolated tree per worktree: a fresh worktree has **no** `node_modules`, which is why JS hooks fail with exit 127 and the worker reaches for `--no-verify` — no hooks at all. So the trade is "shared read-mostly tree" vs "hooks bypassed wholesale", not "shared tree" vs "isolated tree".

**That trade is yours to accept or reject.** If you would rather pay for isolation, the shape would be a per-worktree install (slow, hundreds of MB each) or a pnpm-style content-addressed store. Full reasoning is on the review thread.

PR #1413, opened 2026-08-12
feat(self-merge): block PRs with operator hold label (do-not-merge / hold / on-hold)

TimeToBuildBob:
## Summary

Adds a durable GitHub-label-based hold mechanism so an operator can record a
\"don't merge this\" decision that survives every context window boundary.

**Problem**: When Erik says \"don't merge this\" in a session, that decision lives
only in that session's conversation context. Sibling autonomous sessions and PM
dispatches have no way to know about the hold. PM then keeps spending cycles
driving the held PR toward merge readiness (gptme-contrib#1380 was referenced
5× in 24h while explicitly held).

**Mechanism chosen**: GitHub label (`do-not-merge`, `hold`, or `on-hold`).
Durable, machine-readable, visible in the PR list to humans, survives every
context window.

**Changes**:
- `fetch_pr()` requests the `labels` field from the GitHub API
- `_HOLD_LABELS` frozenset defines the canonical hold label names
- `evaluate_pr()` checks for any matching label and appends a fail-closed reason: `"Operator hold: <label> — remove label to unblock"`
- 4 new tests (3 parametrized label variants + 1 no-hold-passes guard)
- `_make_clean_pr_data()` updated with `labels: []` default

**Verified**: Applied `do-not-merge` label to gptme-contrib#1380 and confirmed
the gate now cites the hold explicitly (independently of the three incidental
blockers already present).

## Test plan

- [x] `test_evaluate_pr_blocks_operator_hold_label[do-not-merge]` passes
- [x] `test_evaluate_pr_blocks_operator_hold_label[hold]` passes
- [x] `test_evaluate_pr_blocks_operator_hold_label[on-hold]` passes
- [x] `test_evaluate_pr_no_hold_label_is_eligible` passes
- [x] All 213 existing tests pass (no regressions)
- [x] Live test: `self-merge-check.py gptme-contrib#1380` reports `Operator hold: do-not-merge label` as a reason

## Operator workflow

To hold a PR from autonomous merge: apply the `do-not-merge` label via GitHub
UI or `gh pr edit <num> --add-label "do-not-merge"`. Remove the label to unblock.

### DIFF
--- a/scripts/github/self-merge-check.py
+++ b/scripts/github/self-merge-check.py
@@ -2411,7 +2411,11 @@ def evaluate_pr(
             "PR labels missing from payload; cannot verify operator hold"
         )
     else:
-        pr_label_names = {lbl.get("name", "").lower() for lbl in pr["labels"]}
+        pr_label_names = {
+            (lbl.get("name") or "").lower()
+            for lbl in pr["labels"]
+            if isinstance(lbl, dict)
+        }
         matched_hold = pr_label_names & _HOLD_LABELS
         if matched_hold:
             result.reasons.append(
--- a/scripts/github/self-merge-check.py
+++ b/scripts/github/self-merge-check.py
@@ -663,6 +663,20 @@ def fetch_pr(repo: str, number: int) -> dict[str, Any]:
             f"Try GH_API_CACHE_TTL=0 to bypass the cache."
         )
 
+    # Validate that the 'labels' field was returned by gh. We explicitly request
+    # it in the --json fields list, so a missing field indicates a gh CLI version
+    # that predates labels-in-JSON support, or a cache shim stripping the field.
+    # Raise early here (consistent with the number-mismatch check above) so
+    # callers see an explicit diagnostic rather than a silent eligibility block
+    # deep in evaluate_pr.
+    if "labels" not in pr:
+        raise RuntimeError(
+            f"fetch_pr for {repo}#{number}: 'labels' field absent from gh response. "
+            "The field was explicitly requested via --json; this usually means an old "
+            "gh CLI version (<2.27) or a cache shim is stripping the field. "
+            "Upgrade gh or set GH_API_CACHE_TTL=0 to bypass the cache."
+        )
+
     pr["files"] = _fetch_pr_files(repo, number)
     return pr
 
@@ -2404,11 +2418,18 @@ def evaluate_pr(
 
     if not isinstance(pr.get("labels"), list):
         # Labels field absent or null — cannot verify hold state; fail closed.
+        # Under normal operation this branch should not be reached: fetch_pr
+        # explicitly requests labels via --json and raises RuntimeError when the
+        # field is absent. If evaluate_pr is called with a hand-crafted or
+        # cached payload that omits labels, we block rather than silently pass.
         # A present-but-null value is treated identically to a missing key:
         # a cache shim or malformed payload returning labels=null must not
         # accidentally pass the hold check.
+        # To diagnose: check that gh CLI supports the labels JSON field
+        # (gh pr view --json labels), or set GH_API_CACHE_TTL=0 to bypass cache.
         result.reasons.append(
-            "PR labels missing from payload; cannot verify operator hold"
+            "PR labels missing or null in payload; cannot verify operator hold "
+            "(ensure gh CLI ≥2.27 or set GH_API_CACHE_TTL=0 to bypass cache)"
         )
     else:
         pr_label_names = set()
--- a/scripts/github/self-merge-check.py
+++ b/scripts/github/self-merge-check.py
@@ -2402,7 +2402,7 @@ def evaluate_pr(
     if pr.get("state") != "OPEN":
         result.reasons.append(f"PR state is {pr.get('state')}, not OPEN")
 
-    pr_label_names = {lbl.get("name", "").lower() for lbl in pr.get("labels", [])}
+    pr_label_names = {lbl.get("name", "").lower() for lbl in (pr.get("labels") or [])}
     matched_hold = pr_label_names & _HOLD_LABELS
     if matched_hold:
         result.reasons.append(
--- a/scripts/github/self-merge-check.py
+++ b/scripts/github/self-merge-check.py
@@ -2402,12 +2402,20 @@ def evaluate_pr(
     if pr.get("state") != "OPEN":
         result.reasons.append(f"PR state is {pr.get('state')}, not OPEN")
 
-    pr_label_names = {lbl.get("name", "").lower() for lbl in (pr.get("labels") or [])}
-    matched_hold = pr_label_names & _HOLD_LABELS
-    if matched_hold:
+    if "labels" not in pr:
+        # Labels field absent from payload — cannot verify hold state; fail closed.
         result.reasons.append(
-            f"Operator hold: {', '.join(sorted(matched_hold))} label — remove label to unblock"
+            "PR labels missing from payload; cannot verify operator hold"
         )
+    else:
+        pr_label_names = {
+            lbl.get("name", "").lower() for lbl in (pr.get("labels") or [])
+        }
+        matched_hold = pr_label_names & _HOLD_LABELS
+        if matched_hold:
+            result.reasons.append(
+                f"Operator hold: {', '.join(sorted(matched_hold))} label — remove label to unblock"
+            )
 
     merge_state = pr.get("mergeStateStatus", "")
     if merge_state == "DIRTY":
--- a/scripts/github/self-merge-check.py
+++ b/scripts/github/self-merge-check.py
@@ -2411,11 +2411,23 @@ def evaluate_pr(
             "PR labels missing from payload; cannot verify operator hold"
         )
     else:
-        pr_label_names = {
-            (lbl.get("name") or "").lower()
-            for lbl in pr["labels"]
-            if isinstance(lbl, dict)
-        }
+        pr_label_names = set()
+        has_malformed_labels = False
+        for lbl in pr["labels"]:
+            if not isinstance(lbl, dict):
+                has_malformed_labels = True
+                continue
+            name = lbl.get("name")
+            if not isinstance(name, str):
+                has_malformed_labels = True
+                continue
+            pr_label_names.add(name.lower())
+
+        if has_malformed_labels:
+            result.reasons.append(
+                "PR labels contain unparseable entries; cannot verify operator hold"
+            )
+
         matched_hold = pr_label_names & _HOLD_LABELS
         if matched_hold:
             result.reasons.append(
--- a/scripts/github/self-merge-check.py
+++ b/scripts/github/self-merge-check.py
@@ -129,6 +129,9 @@ AI_REVIEW_MIN_SCORE = 1
 # Only P0/P1 participate in the disposition recall guard. Lower severities do
 # not block self-merge, and the gate must never guess them into existence.
 AI_REVIEW_BLOCKING_SEVERITIES = frozenset({"P0", "P1"})
+# GitHub labels that record an operator hold. An exact case-insensitive match
+# on any of these makes the PR permanently ineligible until the label is removed.
+_HOLD_LABELS = frozenset({"do-not-merge", "hold", "on-hold"})
 # Shortest marker SHA prefix accepted when matching against the PR head. The
 # reviewer writes 12 hex chars; this only guards against a truncated/garbage
 # marker whose 1-2 char "sha" would prefix-match almost any head.
@@ -639,7 +642,7 @@ def fetch_pr(repo: str, number: int) -> dict[str, Any]:
             "--repo",
             repo,
             "--json",
-            "number,title,url,author,statusCheckRollup,isDraft,state,reviewDecision,headRefOid,mergeStateStatus,isCrossRepository,baseRefName",
+            "number,title,url,author,statusCheckRollup,isDraft,state,reviewDecision,headRefOid,mergeStateStatus,isCrossRepository,baseRefName,labels",
         ]
     )
     if not raw:
@@ -2398,6 +2401,13 @@ def evaluate_pr(
 
     if pr.get("state") != "OPEN":
         result.reasons.append(f"PR state is {pr.get('state')}, not OPEN")
+
+    pr_label_names = {lbl.get("name", "").lower() for lbl in pr.get("labels", [])}
+    matched_hold = pr_label_names & _HOLD_LABELS
+    if matched_hold:
+        result.reasons.append(
+            f"Operator hold: {', '.join(sorted(matched_hold))} label — remove label to unblock"
+        )
 
     merge_state = pr.get("mergeStateStatus", "")
     if merge_state == "DIRTY":
--- a/scripts/github/self-merge-check.py
+++ b/scripts/github/self-merge-check.py
@@ -2402,15 +2402,16 @@ def evaluate_pr(
     if pr.get("state") != "OPEN":
         result.reasons.append(f"PR state is {pr.get('state')}, not OPEN")
 
-    if "labels" not in pr:
-        # Labels field absent from payload — cannot verify hold state; fail closed.
+    if not isinstance(pr.get("labels"), list):
+        # Labels field absent or null — cannot verify hold state; fail closed.
+        # A present-but-null value is treated identically to a missing key:
+        # a cache shim or malformed payload returning labels=null must not
+        # accidentally pass the hold check.
         result.reasons.append(
             "PR labels missing from payload; cannot verify operator hold"
         )
     else:
-        pr_label_names = {
-            lbl.get("name", "").lower() for lbl in (pr.get("labels") or [])
-        }
+        pr_label_names = {lbl.get("name", "").lower() for lbl in pr["labels"]}
         matched_hold = pr_label_names & _HOLD_LABELS
         if matched_hold:
             result.reasons.append(
--- a/scripts/github/self-merge-check.py
+++ b/scripts/github/self-merge-check.py
@@ -2442,7 +2442,7 @@ def evaluate_pr(
             if not isinstance(name, str):
                 has_malformed_labels = True
                 continue
-            pr_label_names.add(name.lower())
+            pr_label_names.add(name.strip().lower())
 
         if has_malformed_labels:
             result.reasons.append(
--- a/tests/test_self_merge_check.py
+++ b/tests/test_self_merge_check.py
@@ -345,16 +345,16 @@ def test_evaluate_pr_no_hold_label_is_eligible() -> None:
 
 
 def test_evaluate_pr_malformed_label_elements_do_not_crash() -> None:
-    """Non-dict elements and null name values in the labels list must not crash evaluate_pr.
+    """Non-dict elements and null name values in the labels list must fail closed (block merge).
 
     A cache shim or API change could return labels as a mixed list.
-    Non-dict entries are skipped; none of their names can match _HOLD_LABELS, so
-    the result is eligible (no hold detected, no crash).
+    Non-dict entries and non-string name values are unverifiable, so we block the merge
+    rather than silently skip them (fail-closed security posture to prevent hold bypass).
     """
-    # Mix of non-dict elements and a dict with name=None — none are hold labels
+    # Mix of non-dict elements and a dict with name=None — all are unverifiable
     result = _evaluate_with_hold_labels(["hold", None, {"name": None}, {"name": "ci"}])
-    assert result.eligible
-    assert not any("Operator hold" in r for r in result.reasons)
+    assert not result.eligible
+    assert any("unparseable" in r.lower() for r in result.reasons)
 
 
 def test_evaluate_pr_null_labels_fails_closed() -> None:
--- a/tests/test_self_merge_check.py
+++ b/tests/test_self_merge_check.py
@@ -324,6 +324,20 @@ def test_evaluate_pr_blocks_operator_hold_label(label_name: str) -> None:
     assert label_name in " ".join(result.reasons)
 
 
+@pytest.mark.parametrize("label_name", ["Do-Not-Merge", "HOLD", "On-Hold"])
+def test_evaluate_pr_blocks_operator_hold_label_case_insensitive(
+    label_name: str,
+) -> None:
+    """Hold label matching must be case-insensitive.
+
+    Labels applied via the GitHub UI use the label's stored name, but a future
+    rename (e.g. 'hold' → 'Hold') must not silently bypass the gate.
+    """
+    result = _evaluate_with_hold_labels([{"name": label_name}])
+    assert not result.eligible
+    assert any("Operator hold" in r for r in result.reasons)
+
+
 def test_evaluate_pr_no_hold_label_is_eligible() -> None:
     result = _evaluate_with_hold_labels([{"name": "enhancement"}, {"name": "ci"}])
     assert result.eligible
--- a/tests/test_self_merge_check.py
+++ b/tests/test_self_merge_check.py
@@ -344,6 +344,19 @@ def test_evaluate_pr_no_hold_label_is_eligible() -> None:
     assert not any("Operator hold" in r for r in result.reasons)
 
 
+def test_evaluate_pr_malformed_label_elements_do_not_crash() -> None:
+    """Non-dict elements and null name values in the labels list must not crash evaluate_pr.
+
+    A cache shim or API change could return labels as a mixed list.
+    Non-dict entries are skipped; none of their names can match _HOLD_LABELS, so
+    the result is eligible (no hold detected, no crash).
+    """
+    # Mix of non-dict elements and a dict with name=None — none are hold labels
+    result = _evaluate_with_hold_labels(["hold", None, {"name": None}, {"name": "ci"}])
+    assert result.eligible
+    assert not any("Operator hold" in r for r in result.reasons)
+
+
 def test_evaluate_pr_null_labels_fails_closed() -> None:
     """Labels key present but null must fail closed — same as a missing key.
 
--- a/tests/test_self_merge_check.py
+++ b/tests/test_self_merge_check.py
@@ -196,7 +196,7 @@ def test_evaluate_pr_blocks_changes_requested() -> None:
 
 def _make_clean_pr_data(**overrides: object) -> dict[str, object]:
     """Minimal PR data that passes all other checks. Override fields to test specific gates."""
-    data = {
+    data: dict[str, object] = {
         "author": {"login": "TimeToBuildBob"},
         "title": "Test PR",
         "url": "https://github.com/gptme/gptme-contrib/pull/999",
@@ -207,6 +207,7 @@ def _make_clean_pr_data(**overrides: object) -> dict[str, object]:
         "reviewDecision": None,
         "headRefOid": "abc123",
         "mergeStateStatus": "CLEAN",
+        "labels": [],
     }
     data.update(overrides)
     return data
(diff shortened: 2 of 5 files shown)


TimeToBuildBob:

Evidence: commit: 154c5beaa3 chore(journal): PM — gptme-contrib#1413 mypy fix pushed (operator-hold-label)

Session: `0b836221-9b1f-5788-9b8f-76262dc53371`

TimeToBuildBob:
CI is now green (pre-commit + integration-tests both pass on `1bd9c99`).

Ran `self-merge-check.py` — **not eligible for self-merge**:

1. **Greptile dark** — no Greptile review found (billing issue, operator action needed per [#1141](https://github.com/ErikBjare/bob/issues/1141))
2. **Sensitive/infra paths** — PR touches `scripts/github/self-merge-check.py` itself, which is the merge gate; the policy blocks self-merging changes to the gate's own code

Both blockers require operator review. Ready to merge on your side whenever Greptile billing is resolved or you want to approve+merge directly.

TimeToBuildBob:
Self-merge check ran — blocked on two conditions:

1. **No Greptile review** — Greptile is currently dark (billing issue, operator action needed). The helper confirmed it's awaiting auto-review but hasn't fired.
2. **Sensitive path** — `scripts/github/self-merge-check.py` is classified as security/infra, requiring Greptile sign-off before self-merge.

CI is green. To unblock: either restore Greptile billing so it can review, or Erik can review and merge manually.

scripts/github/self-merge-check.py
```diff
@@ -2382,6 +2385,13 @@ def evaluate_pr(
     if pr.get("state") != "OPEN":
         result.reasons.append(f"PR state is {pr.get('state')}, not OPEN")
 
+    pr_label_names = {lbl.get("name", "").lower() for lbl in pr.get("labels", [])}
```
TimeToBuildBob: ⚠️ **P2** — The new label check in `evaluate_pr` builds `pr_label_names` via `{lbl.get("name", "").lower() for lbl in pr.get("labels", [])}`. If the `labels` key is present in the PR dict but its value is `null` (possible if a cache shim or a partially-formed API response omits the array, analogous to other defensive checks in this script), `pr.get("labels", [])` returns `None` rather than the default, and iterating over `None` raises `TypeError: 'NoneType' is not iterable`. This would crash `evaluate_pr`, causing `main` to exit with code 2 and report a hard error instead of a clean eligible/ineligible verdict. The rest of the script deliberately handles unexpected shapes (e.g., `merge_permission` checks for `dict` and key presence), so this new path should guard similarly. Using `pr.get("labels") or []` collapses `None` to `[]` and keeps the check fail-closed.

TimeToBuildBob:
## 🤖 AI code review

Adds a hold-label mechanism that blocks self-merge when a PR carries `do-not-merge`, `hold`, or `on-hold` labels (case-insensitive, whitespace-stripped). `fetch_pr` now requests the `labels` field and raises if it is absent, while `evaluate_pr` appends fail-closed reasons when labels are missing, null, malformed, or match a hold label. Tests cover the label matching variants and the fail-closed paths, and existing test payloads are updated to include a `labels` key.

**⚠️ P2 medium** — `scripts/github/self-merge-check.py:672`

`fetch_pr` now unconditionally requires the `labels` field in the `gh pr view --json` response, raising RuntimeError if it is absent. On gh CLI versions older than 2.27 (which do not support `--json labels`), every call to `fetch_pr` raises, so `main()` exits with code 2 and prints an error even for PRs that would otherwise be eligible. This is a hard breaking change for existing deployments on older gh; before this change the check would have run and only printed a warning or reason for a missing field. The PR description acknowledges the intent to fail fast, but it turns a non-blocking diagnostic into a complete halt of the self-merge check for affected users.

```suggestion
Consider not raising but falling back to a fail-closed reason in `evaluate_pr`, which already handles missing labels, to avoid a hard break for old gh CLI versions.
```

<sub>**How this was verified:** Checked the `fetch_pr` function: it adds `labels` to the `--json` fields (line 645) and raises at line 672 if the key is missing. The test `test_fetch_pr_raises_on_missing_labels_field` confirms the behavior.</sub>

</details>

- `scripts/github/self-merge-check.py` — Adds `_HOLD_LABELS`, requests and validates the `labels` field in `fetch_pr`, and adds hold-label checks to `evaluate_pr` that fail closed on missing/malformed labels.
- `tests/test_self_merge_ai_review_abstention.py` — Adds `labels: []` to two test payloads so they pass the new labels check.
- `tests/test_self_merge_ai_review_gate.py` — Adds `labels: []` to the PR payload fixture.
- `tests/test_self_merge_check.py` — Adds `labels` defaults to test payloads and new tests for hold labels (blocking, case-insensitive, whitespace, null, missing, malformed).
- `tests/test_self_merge_ci_configured.py` — Adds `labels: []` to the test PR payload.

</details>

| `1bd9c9915b83` | 4/5 | 1 | llm | 2026-08-12 02:45 UTC |
| `c8788e721bb6` | 3/5 | 1 | llm | 2026-08-12 03:40 UTC |
| `6e0ef78372e1` | 4/5 | 1 | llm | 2026-08-12 15:44 UTC |
| `95afb4c7d8f6` | 4/5 | 1 | llm | 2026-08-12 16:03 UTC |
| `bbce75ca410b` | 4/5 | 1 | llm | 2026-08-12 16:21 UTC |
| `bbce75ca410b` | 3/5 | 2 | llm | 2026-08-12 16:35 UTC |
| `f115be263c67` | 3/5 | 1 | llm | 2026-08-12 16:44 UTC |
| `dbc041dca92c` | 3/5 | 2 | llm | 2026-08-12 17:01 UTC |
| `a5f6ca086283` | 4/5 | 1 | llm | 2026-08-12 18:21 UTC |
| `3b152729fa42` | 4/5 | 2 | llm | 2026-08-12 18:40 UTC |

</details>

<sub>Reviewed `46b700f2c145` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 117s · reports findings; merge decisions belong to the gate · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

scripts/github/self-merge-check.py
TimeToBuildBob: Fixed in c8788e72 — changed to `pr.get('labels') or []` and added `test_evaluate_pr_null_labels_does_not_crash` to regression-guard the null case.

TimeToBuildBob:
@greptileai review

scripts/github/self-merge-check.py
```diff
@@ -2382,6 +2385,13 @@ def evaluate_pr(
     if pr.get("state") != "OPEN":
         result.reasons.append(f"PR state is {pr.get('state')}, not OPEN")
 
+    pr_label_names = {lbl.get("name", "").lower() for lbl in (pr.get("labels") or [])}
```
TimeToBuildBob: ❌ **P1** — The hold-label detection fails open when the PR payload lacks a `labels` key. `pr.get("labels") or []` treats a missing key as an empty list, so `pr_label_names` is empty and no hold reason is added. If a caching shim returns a payload cached before this change (which did not request `labels`), or the API response is otherwise parseable but omits that field, a PR carrying a `do-not-merge` label will be evaluated without the hold being detected. Given the gate is otherwise passable, the PR can be auto-merged despite the operator's explicit hold. Since this is a safety mechanism that must fail closed, the absence of label data should be treated as inability to verify, not as proof of no hold. The current implementation silently defaults to 'no hold', which defeats the durable-hold purpose and can cause an unwanted merge.

TimeToBuildBob:
Pushed `6e0ef783` to address the P1 from the AI review.

**What the previous fix missed**: `c8788e72` fixed the crash when `labels: null` (present key, null value) via `pr.get("labels") or []`. But that form still silently defaults to `[]` when the `labels` key is entirely absent — which is the fail-open the AI reviewer flagged.

**This commit**: explicit key-presence check instead:

```python
if "labels" not in pr:
    result.reasons.append("PR labels missing from payload; cannot verify operator hold")
else:
    pr_label_names = {lbl.get("name", "").lower() for lbl in (pr.get("labels") or [])}
    matched_hold = pr_label_names & _HOLD_LABELS
    if matched_hold:
        result.reasons.append(...)
```

When the field is absent the gate now blocks rather than passes — correct behaviour for a safety mechanism. When it's present-but-null, the `or []` handles it as "no hold labels" (unchanged, still eligible).

New test: `test_evaluate_pr_missing_labels_key_blocks_merge` — covers the absent-key case specifically. CI running.

tests/test_self_merge_check.py
```diff
@@ -276,6 +278,118 @@ def test_evaluate_pr_clean_merge_state_passes_gate() -> None:
...
+        ),
+    ):
+        return self_merge_check.evaluate_pr(
+            "gptme/gptme-contrib",
+            999,
+            workspace_repos=["gptme/gptme-contrib"],
+        )
+
+
+@pytest.mark.parametrize("label_name", ["do-not-merge", "hold", "on-hold"])
```
TimeToBuildBob, line 309: ⚠️ **P2** — The implementation lowercases label names via lbl.get("name", "").lower() and then does an exact set intersection with _HOLD_LABELS. The new tests only exercise lowercase label names ("do-not-merge", "hold", "on-hold"). Since the PR explicitly advertises case-insensitive matching, a label applied as "Do-Not-Merge" or "HOLD" would still block (because .lower() normalizes it), but this behaviour is completely unverified. If a regression removed the .lower() call, the case-sensitive form would silently stop blocking and a hold could be bypassed while all tests still pass. Add a parametrized case variant (e.g. "Do-Not-Merge", "HOLD") to the existing parametrize list to lock in the advertised contract.

TimeToBuildBob:
Fixed the CI failure in `95afb4c`.

Root cause: the previous commit (fail-closed when `labels` key absent from payload) was correct, but 10 test fixtures predating the feature didn't include `"labels"` in their mock PR dicts. The fail-closed check then blocked every test that expected an eligible result.

Fix: added `"labels": []` to the 6 mock-data helpers/inline dicts across 4 test files (`test_self_merge_check.py`, `test_self_merge_ci_configured.py`, `test_self_merge_ai_review_abstention.py`, `test_self_merge_ai_review_gate.py`). All 10 previously failing tests now pass (345 total, 0 failures in the affected files).

scripts/github/self-merge-check.py
```diff
@@ -2382,6 +2385,22 @@ def evaluate_pr(
...
+    if not isinstance(pr.get("labels"), list):
+        # Labels field absent or null — cannot verify hold state; fail closed.
+        # A present-but-null value is treated identically to a missing key:
+        # a cache shim or malformed payload returning labels=null must not
+        # accidentally pass the hold check.
+        result.reasons.append(
+            "PR labels missing from payload; cannot verify operator hold"
+        )
+    else:
+        pr_label_names = {lbl.get("name", "").lower() for lbl in pr["labels"]}
```
TimeToBuildBob: ⚠️ **P2** — In evaluate_pr, the hold-label check assumes every element of pr['labels'] is a dict, building pr_label_names via `lbl.get(...)`. If the gh shim or a cache returns a labels list whose elements are strings (or any non-dict), the set comprehension raises AttributeError. main catches the exception, prints a generic error, and exits with code 2, so the PR is still blocked, but with an uninformative message and without the intended 'labels missing' reason. The PR's fail-closed logic only guards against labels being absent or null, not against malformed list contents, so a malformed payload produces a confusing failure instead of a clear fail-closed reason.

scripts/github/self-merge-check.py
```diff
@@ -2382,6 +2385,22 @@ def evaluate_pr(
     if pr.get("state") != "OPEN":
         result.reasons.append(f"PR state is {pr.get('state')}, not OPEN")
 
+    if not isinstance(pr.get("labels"), list):
```
TimeToBuildBob, line 2419: ❌ **P1** — Fail-closed on missing labels: if `labels` is absent or null in the PR payload, `evaluate_pr` appends a blocking reason. This is a regression from the prior behavior where a missing field had no effect. In environments where the `gh` CLI version does not support the `labels` JSON field, or where a cache shim (as referenced in the code) returns a payload without it, every PR becomes ineligible, breaking self-merge entirely. The gate now depends on a field that was not previously required, and any transient API quirk that omits it blocks all merges with no retry or fallback. This is an availability risk for the feature. _(3/3 passes)_

tests/test_self_merge_check.py
```diff
@@ -276,6 +279,123 @@ def test_evaluate_pr_clean_merge_state_passes_gate() -> None:
...
```
TimeToBuildBob, line 309: ⚠️ **P2** — The new tests only exercise label names exactly as defined in `_HOLD_LABELS` ('do-not-merge', 'hold', 'on-hold'). The implementation lowercases label names via `.lower()` before matching, so a label applied as 'Hold' or 'DO-NOT-MERGE' would also be caught, but there is no test pinning this case-insensitive behaviour. If the `.lower()` call were accidentally removed, the gate would silently fail open for any label not already lowercase, merging a held PR without a human. Add a parametrized case that uses a mixed-case label (e.g. 'Hold' or 'Do-Not-Merge') to lock in the case-insensitive matching. _(2/3 passes)_

scripts/github/self-merge-check.py
```diff
@@ -2399,6 +2402,26 @@ def evaluate_pr(
...
+        # a cache shim or malformed payload returning labels=null must not
+        # accidentally pass the hold check.
+        result.reasons.append(
+            "PR labels missing from payload; cannot verify operator hold"
+        )
+    else:
+        pr_label_names = {
+            (lbl.get("name") or "").lower()
+            for lbl in pr["labels"]
+            if isinstance(lbl, dict)
```
TimeToBuildBob: ❌ **P1** — The hold-label check silently skips any entry in the 'labels' list that is not a dict. If the GitHub API or a cache shim ever returns labels as bare strings (e.g., ["hold"] instead of [{"name": "hold"}]), the hold label is ignored and the PR is treated as eligible, defeating the entire purpose of the hold mechanism. The gate is fail-closed when the labels field is missing, but fail-open on malformed entries, which is inconsistent. An attacker controlling label data (e.g., through a compromised cache) could bypass the hold and merge a held PR. The fix is to treat any non-dict entry as unverifiable and block the merge, or at least append a reason indicating malformed label data.

scripts/github/self-merge-check.py
```diff
@@ -2399,6 +2402,26 @@ def evaluate_pr(
...
+        # Labels field absent or null — cannot verify hold state; fail closed.
+        # A present-but-null value is treated identically to a missing key:
+        # a cache shim or malformed payload returning labels=null must not
+        # accidentally pass the hold check.
+        result.reasons.append(
+            "PR labels missing from payload; cannot verify operator hold"
+        )
+    else:
+        pr_label_names = {
+            (lbl.get("name") or "").lower()
```
TimeToBuildBob: ⚠️ **P2** — The label name is assumed to be a string via `(lbl.get("name") or "").lower()`. If a label object has a `name` that is not a string (e.g., an integer or a list), the call to `.lower()` raises an AttributeError, causing evaluate_pr to crash and the entire self-merge check to abort with an error. While this fails closed (the merge is blocked), it is a fragile assumption that could be triggered by a malformed API response, turning a routine check into a hard failure with exit code 2 and obscuring the actual reasons. The code should guard with `isinstance(lbl.get("name"), str)` and treat non-string names as unparseable, either by blocking or by skipping with a warning.

TimeToBuildBob:
Fixed the remaining mypy failure in `a5f6ca08`.

The pre-commit CI was failing because `_evaluate_with_hold_labels` was typed as `list[dict[str, str]]`, but the malformed-input test at line 355 passes a heterogeneous list (`str`, `None`, and dicts with `None` values). Changed the parameter type to `list[object]` which accurately reflects the intent of the test.

CI should pass now.

scripts/github/self-merge-check.py
```diff
@@ -2399,6 +2402,38 @@ def evaluate_pr(
     if pr.get("state") != "OPEN":
         result.reasons.append(f"PR state is {pr.get('state')}, not OPEN")
 
+    if not isinstance(pr.get("labels"), list):
+        # Labels field absent or null — cannot verify hold state; fail closed.
+        # A present-but-null value is treated identically to a missing key:
+        # a cache shim or malformed payload returning labels=null must not
+        # accidentally pass the hold check.
+        result.reasons.append(
+            "PR labels missing from payload; cannot verify operator hold"
```
TimeToBuildBob: ❌ **P1** — The new code treats a missing or null labels field as a blocking reason. Since fetch_pr now requests labels, a cache shim or older cached payload that does not include the field will cause every evaluated PR to fail with "PR labels missing from payload". This is a regression: any PR on a repo that uses a caching gh wrapper (e.g., graphql-attribution.sh, per the code's own comments) will be blocked until the cache is refreshed, even though it has no hold label. The consequence is that self-merge stops working entirely during the cache transition, blocking otherwise-eligible PRs. The fail-closed intent is documented, but the blast radius is the whole gate.

scripts/github/self-merge-check.py
```diff
@@ -2399,6 +2402,38 @@ def evaluate_pr(
...
+        has_malformed_labels = False
+        for lbl in pr["labels"]:
+            if not isinstance(lbl, dict):
+                has_malformed_labels = True
+                continue
+            name = lbl.get("name")
+            if not isinstance(name, str):
+                has_malformed_labels = True
+                continue
+            pr_label_names.add(name.lower())
```
TimeToBuildBob: ⚠️ **P2** — The code lowercases label names but does not strip surrounding whitespace. A label applied as "do-not-merge " (with a trailing space) or " hold " will not match the _HOLD_LABELS set, so a held PR could merge if the operator accidentally includes spaces when creating or applying the label. The tests cover case-insensitivity but not whitespace, so this edge case is unguarded. The fix is to normalize with .strip().lower() before adding to pr_label_names.

scripts/github/self-merge-check.py
```diff
@@ -2399,6 +2402,38 @@ def evaluate_pr(
     if pr.get("state") != "OPEN":
         result.reasons.append(f"PR state is {pr.get('state')}, not OPEN")
 
+    if not isinstance(pr.get("labels"), list):
```
TimeToBuildBob, line 2419: ⚠️ **P2** — The new labels gate fails closed whenever the `labels` field is missing or null, unconditionally for every PR evaluation. `fetch_pr` requests `labels` via `gh pr view --json`, so under normal operation the field is always present (an empty array for a PR with no labels). However, if a cache shim (the module already anticipates one: `graphql-attribution.sh`) or a future `gh` version omits the field, every PR on the repo becomes ineligible even when it carries no hold label. This is a false positive that silently disables autonomous self-merge until the payload is fixed, which could halt development during a transient API or cache issue. It is a robustness regression relative to the prior behavior, which did not require labels at all. The PR description explicitly chooses fail-closed, but this unconditional dependency on a single field is a real operational hazard. _(2/3 passes)_

tests/test_self_merge_check.py
```diff
@@ -276,6 +279,150 @@ def test_evaluate_pr_clean_merge_state_passes_gate() -> None:
     assert not any("merge conflicts" in r for r in result.reasons)
 
 
+_HOLD_PATCH_CONTEXT = (
```
TimeToBuildBob: ⚠️ **P2** — The `_HOLD_PATCH_CONTEXT` tuple defined in test_self_merge_check.py is never used. It appears to be a leftover from a previous version of the test helper; the tests use `with (...):` directly. While harmless, it is dead code that could mislead future maintainers into thinking it is required for the hold-label tests, and it adds noise to the diff. Removing it keeps the test file clean and avoids confusion about which components are actually patched. _(2/3 passes)_

TimeToBuildBob:
Pushed `c2524df` to address the two remaining P2 findings from the AI review.

**1. Dead code: `_HOLD_PATCH_CONTEXT`** (finding `1c232c8b10bc`)
Removed the unused tuple — it was a leftover from an earlier version of the test helper. The tests use the `with (...)` context manager directly and never referenced it.

**2. Labels availability hazard** (finding `f54dc26bbaf2`)
The concern: if the `labels` field is absent from the payload, `evaluate_pr` silently blocks ALL PRs with no way to distinguish "there is a hold" from "API didn't return labels."

Fix: moved the missing-field check to `fetch_pr` (consistent with the existing number-mismatch check there), so callers see an explicit `RuntimeError` with actionable diagnostics ("gh CLI <2.27 or cache shim stripping the field — try `GH_API_CACHE_TTL=0`") rather than a silent eligibility block deep in `evaluate_pr`. The `evaluate_pr` fail-closed path is kept as defense-in-depth for hand-crafted payloads, with an improved message including the same debug hints. New test: `test_fetch_pr_raises_on_missing_labels_field`. Fixed `test_fetch_pr_uses_paginated_rest_files_api` (mock payload was missing the `labels` key).

All 220 `test_self_merge_check` tests pass; full suite 816 pass, 4 skipped.

@TimeToBuildBob review

TimeToBuildBob:
@TimeToBuildBob review

scripts/github/self-merge-check.py
```diff
@@ -660,6 +663,20 @@ def fetch_pr(repo: str, number: int) -> dict[str, Any]:
             f"Try GH_API_CACHE_TTL=0 to bypass the cache."
         )
 
+    # Validate that the 'labels' field was returned by gh. We explicitly request
+    # it in the --json fields list, so a missing field indicates a gh CLI version
+    # that predates labels-in-JSON support, or a cache shim stripping the field.
+    # Raise early here (consistent with the number-mismatch check above) so
+    # callers see an explicit diagnostic rather than a silent eligibility block
+    # deep in evaluate_pr.
+    if "labels" not in pr:
```
TimeToBuildBob, line 672: ⚠️ **P2** — `fetch_pr` now unconditionally requires the `labels` field in the `gh pr view --json` response, raising RuntimeError if it is absent. On gh CLI versions older than 2.27 (which do not support `--json labels`), every call to `fetch_pr` raises, so `main()` exits with code 2 and prints an error even for PRs that would otherwise be eligible. This is a hard breaking change for existing deployments on older gh; before this change the check would have run and only printed a warning or reason for a missing field. The PR description acknowledges the intent to fail fast, but it turns a non-blocking diagnostic into a complete halt of the self-merge check for affected users.

scripts/github/self-merge-check.py
TimeToBuildBob: Fixed in 6e0ef783 (superseded by later commits): the code now has an explicit key-presence check — `if not isinstance(pr.get('labels'), list)` — which blocks when the key is absent or null. Additionally, `fetch_pr` raises `RuntimeError` with an actionable diagnostic ('ensure gh CLI ≥2.27 or set GH_API_CACHE_TTL=0') so callers see a clear error before `evaluate_pr` is even reached.

tests/test_self_merge_check.py
TimeToBuildBob: Fixed: added `test_evaluate_pr_blocks_operator_hold_label_case_insensitive` parametrized with 'Do-Not-Merge', 'HOLD', 'On-Hold' — locks in the case-insensitive contract so removing the `.lower()` call would immediately break those tests.

scripts/github/self-merge-check.py
TimeToBuildBob: Fixed in dbc041dc: non-dict elements now set `has_malformed_labels = True` and the check blocks with 'PR labels contain unparseable entries; cannot verify operator hold' — no AttributeError, no silent pass.

scripts/github/self-merge-check.py
TimeToBuildBob: Acknowledged intentional design: fail-closed is the correct posture for a hold safety gate. `fetch_pr` now raises `RuntimeError` with 'labels field absent from gh response — ensure gh CLI ≥2.27 or set GH_API_CACHE_TTL=0' — explicit and actionable. Treating a missing labels field as 'no hold' would silently allow a held PR to merge, which defeats the feature's purpose. The evaluate_pr defense-in-depth path includes the same diagnostic hints so there is never a silent block with no recovery guidance.

tests/test_self_merge_check.py
TimeToBuildBob: Fixed: added `test_evaluate_pr_blocks_operator_hold_label_case_insensitive` parametrized with 'Do-Not-Merge', 'HOLD', 'On-Hold'. If the `.lower()` call were accidentally removed, these three tests would immediately fail.

scripts/github/self-merge-check.py
TimeToBuildBob: Fixed in 3b152729: non-dict entries now set `has_malformed_labels = True`, which appends 'PR labels contain unparseable entries; cannot verify operator hold'. A malformed labels payload — including bare strings like ["hold"] — can never silently pass the hold check.

scripts/github/self-merge-check.py
TimeToBuildBob: Fixed: added `if not isinstance(name, str): has_malformed_labels = True` — a non-string name (integer, list, etc.) is treated as an unparseable entry and blocks the merge rather than crashing with AttributeError.

scripts/github/self-merge-check.py
TimeToBuildBob: Acknowledged intentional design: the fail-closed behavior is correct for this safety gate. `fetch_pr` raises `RuntimeError` with 'labels field absent from gh response — check gh CLI ≥2.27 or set GH_API_CACHE_TTL=0'. The error is explicit and actionable; operators know exactly how to recover. Treating a missing labels field as 'no hold' would allow a held PR to be merged silently — the opposite of what the feature exists to prevent.

scripts/github/self-merge-check.py
TimeToBuildBob: Fixed in 46b700f2: label names are now normalized with `.strip().lower()` before matching. A label applied as 'do-not-merge ' (trailing space) or ' hold ' now correctly blocks the merge. Test `test_evaluate_pr_blocks_operator_hold_label_with_whitespace` pins this contract.

scripts/github/self-merge-check.py
TimeToBuildBob: Addressed in c2524df: the primary missing-field check now lives in `fetch_pr` where it raises with a clear diagnostic ('labels field absent from gh response — ensure gh CLI ≥2.27 or set GH_API_CACHE_TTL=0'). The evaluate_pr defense-in-depth path carries the same hints. Operators always see an actionable message; there is no silent eligibility block with no recovery path.

tests/test_self_merge_check.py
TimeToBuildBob: Fixed in c2524df: removed the unused `_HOLD_PATCH_CONTEXT` tuple.

scripts/github/self-merge-check.py
TimeToBuildBob: Acknowledged design tradeoff: `fetch_pr` raises `RuntimeError` with an explicit message: 'labels absent from gh response — check that gh CLI supports the labels JSON field (gh pr view --json labels), or set GH_API_CACHE_TTL=0 to bypass cache'. A hard failure with a clear remediation path is the correct behavior here — a hold-label gate that silently degrades on older CLI versions is worse than failing clearly. The minimum gh version (2.27) predates this feature by years, so breaking deployments on older gh is an acceptable boundary to draw.

TimeToBuildBob:
## Convergence adjudication — gptme-contrib#1413

The AI reviewer has been at 4/5 for 5+ consecutive rounds (stable since `6e0ef783`). Greptile bot is currently dark (billing). Adjudicating the remaining finding.

---

### Remaining finding: `fp: 91b66a608fcb` — P2

**Location**: `scripts/github/self-merge-check.py:672`

**Claim**: `fetch_pr` raises `RuntimeError` if the `labels` field is absent from the `gh pr view --json` response. On gh CLI < 2.27 (which predates `--json labels`), every call to `fetch_pr` raises and self-merge halts entirely.

**Classification: Acceptable design tradeoff — not a blocking bug**

Rationale:
1. **Intentional fail-closed**: The PR description explicitly states the hold gate must fail closed when label data is unavailable. Treating a missing `labels` field as "no hold" would silently allow a held PR to merge — the opposite of the feature's purpose.
2. **gh 2.27 is years old**: Released 2023; requiring it for a new safety feature is not a meaningful compatibility regression. The alternative — silently downgrading to the pre-feature behavior on old gh — would make the feature invisible to operators on old installs without warning.
3. **Actionable error**: The RuntimeError message says exactly what to do: "ensure gh CLI ≥2.27 or set `GH_API_CACHE_TTL=0`". This is not a silent failure.
4. **Defence in depth**: `evaluate_pr` also has a fail-closed path with the same diagnostic hints, so even hand-crafted payloads don't silently pass.
5. **Inline reply already posted**: The design rationale is already documented at the inline thread for this finding.

The reviewer's suggestion ("fall back to `evaluate_pr`'s fail-closed reason") would actually be less safe — it would make a version-compatibility failure invisible, looking like "no hold" to an operator who didn't notice the RuntimeError was being swallowed.

**Decision: Dismiss. No code change needed.**

---

### Summary

| Finding | Severity | Classification | Action |
|---------|----------|---------------|--------|
| `91b66a608fcb` — RuntimeError on missing labels field | P2 | Acceptable design tradeoff; fail-closed is correct for a safety gate | Dismissed with explanation |
| 11 prior findings | P1/P2 | Fixed or auto-resolved across prior sessions | Done |

**CI**: ✅ Green — `integration-tests` pass, `pre-commit` pass (latest: `46b700f2c145`)

**Domain risk**: This PR modifies `self-merge-check.py` — the merge gate itself. Manual test recommended: apply a `do-not-merge` label to a test PR and confirm the gate blocks it with reason "Operator hold: do-not-merge". Undo with label removal and confirm gate re-passes.

**Convergence**: Stable at 4/5 for 5+ rounds. The remaining P2 has an inline reply and represents a design disagreement that has been adjudicated here as acceptable. No further review rounds will be triggered.

This PR is **ready for maintainer review and merge**. Self-merge is not eligible (PR touches `self-merge-check.py` itself — the gate blocks self-merges to its own code, per policy).

PR #1419, opened 2026-08-13
fix(pm): gate delivery post-condition on thread-deliverable item types

TimeToBuildBob:
## Problem

`master_ci_failure` items re-enter the PM dispatch queue indefinitely. Each cycle spawns a full session that investigates the same master CI failure for 15+ minutes, produces nothing, and re-queues.

Observed live on `ErikBjare/bob#31681387507` (2026-08-13):

```
08:44  master_ci_failure: ErikBjare/bob #31681387507 — Tests
08:59  WARN: PM delivery post-condition FAILED — ErikBjare/bob#31681387507:
       session exited without a thread reply
08:59  WARN: ... rolled back state, event marker, and pending notif state
       (item re-enters the dispatch queue next cycle)
09:10  master_ci_failure: ErikBjare/bob #31681387507 — Tests      <- again
09:25  WARN: PM delivery post-condition FAILED — ...              <- again
```

## Root cause

The Phase-2 port of the delivery post-condition from `project-monitoring-worker.sh` into `run_item.py` dropped the item-type gate.

`worker.sh:453` ran the check only for thread-bearing types:

```bash
if echo "${item_types:-}" | grep -qwE "assigned_issue|pr_update|ci_failure|merge_ready|greptile_needs_improvement|greptile_needs_fix|merge_conflict|notification"; then
    _thread_deliverable="true"
fi
```

`run_item.py` gates only on `item.repo and item.number is not None`.

`master_ci_failure` carries a workflow **run id** as `number` (`31681387507`), not an issue number. So the check queries `issues/31681387507/comments`, 404s, finds no comment, and returns `orphan_no_delivery` → `rollback_failed_delivery()` → re-dispatch. Forever.

Note `master_ci_failure` does *not* match `grep -qwE "ci_failure"` — `_` is a word character, so the word boundary fails. The exclusion in worker.sh was correct and deliberate.

The contradiction is explicit: the PM prompt tells these sessions to

> "If the item is repo-level only and has no thread target (for example master_ci_failure), ... exit cleanly without inventing a thread reply."

and the post-condition then failed them for doing exactly that.

## Fix

Port the gate as `THREAD_DELIVERABLE_TYPES`, matching `worker.sh:453`. This is consistent with the repo-level exclusions already present elsewhere in the codebase:

- `scripts/monitoring/sentinel-frontier.py`: `NON_THREAD_DELIVERABLE_TYPES = {"master_ci_failure"}`
- `scripts/bob-vitals.py`: `INFORMATIONAL_ORPHAN_EVENT_TYPES = {"merge_ready", "master_ci_failure"}`

Repo-level items now keep `delivery_outcome = "handled"` (the permissive default, exactly as worker.sh left it), so state is promoted instead of rolled back.

**Observability is preserved.** The step-9 observable-effect signal passes `delivery_outcome if delivery_verified else ""`, and `delivery_verified` stays `False` when no check runs — so `WARN: PM dispatch produced NO observable effect` still fires for these items. Only the unsatisfiable rollback loop is removed.

## Tests

- `test_post_session_repo_level_item_skips_delivery_check` — the fix. Verified it **fails without the gate** (asserts the check was invoked with `--number 1234`).
- `test_post_session_thread_deliverable_item_still_runs_delivery_check` — guards against over-broad gating; `assigned_issue` still runs the check and still reports `orphan_no_delivery`.

`951 passed, 1 skipped` in `packages/gptme-runloops/tests/`. Scoped `prek` hooks clean.

Refs: ErikBjare/bob#1144

### DIFF
--- a/packages/gptme-runloops/src/gptme_runloops/run_item.py
+++ b/packages/gptme-runloops/src/gptme_runloops/run_item.py
@@ -1855,7 +1855,13 @@ def run_post_session(
     if (
         item.repo
         and item.number is not None
-        and item.number != 0  # 0 is a sentinel for items with no real GitHub thread
+        # 0 is a sentinel for items with no real GitHub thread. Compared via
+        # `number_str` because `number` is an `int | str | None` union taken
+        # verbatim from the grouped JSON (`from_grouped_json`) — a sentinel
+        # arriving as the string "0" must be excluded exactly as the int is,
+        # or the check queries `issues/0/comments`, 404s, and reports
+        # `orphan_no_delivery`, re-creating the re-dispatch loop this gates.
+        and item.number_str != "0"
         and hooks.delivery_check is not None
         and THREAD_DELIVERABLE_TYPES & set(item.types)
     ):
--- a/packages/gptme-runloops/src/gptme_runloops/run_item.py
+++ b/packages/gptme-runloops/src/gptme_runloops/run_item.py
@@ -1855,6 +1855,7 @@ def run_post_session(
     if (
         item.repo
         and item.number is not None
+        and item.number != 0  # 0 is a sentinel for items with no real GitHub thread
         and hooks.delivery_check is not None
         and THREAD_DELIVERABLE_TYPES & set(item.types)
     ):
--- a/packages/gptme-runloops/src/gptme_runloops/run_item.py
+++ b/packages/gptme-runloops/src/gptme_runloops/run_item.py
@@ -2077,7 +2077,12 @@ def run_post_session(
         pr_state_after = ""
         if record_file.is_file():
             pr_state_after = read_record_pr_state_after(record_file)
-        if not pr_state_after and item.repo and item.number is not None:
+        if (
+            not pr_state_after
+            and item.repo
+            and item.number is not None
+            and THREAD_DELIVERABLE_TYPES & set(item.types)
+        ):
             try:
                 proc = hooks.run_cmd(
                     [
--- a/packages/gptme-runloops/src/gptme_runloops/run_item.py
+++ b/packages/gptme-runloops/src/gptme_runloops/run_item.py
@@ -137,13 +137,22 @@ PR_STATE_TYPES: frozenset[str] = frozenset({"pr_update", "merge_ready"})
 # session reply on the thread?" is a meaningful post-condition
 # (worker.sh:453 — `grep -qwE "assigned_issue|pr_update|ci_failure|..."`).
 #
-# Repo-level types are deliberately absent. `master_ci_failure` carries a
-# workflow *run id* as `number` (e.g. 31681387507), not an issue number, so
-# `issues/{number}/comments` 404s, no comment is ever found, and the check
-# returns `orphan_no_delivery` → rollback → re-dispatch, forever. The PM prompt
-# tells those sessions to "exit cleanly without inventing a thread reply"
-# (runs/github/project-monitoring.sh), so the post-condition was punishing
-# sessions for obeying their own instructions.
+# Types with no thread are deliberately absent. Two distinct symptom classes,
+# both traced to this one missing gate:
+#
+# 1. `master_ci_failure` carries a workflow *run id* as `number`
+#    (e.g. 31681387507), not an issue number, so `issues/{number}/comments`
+#    404s, no comment is ever found, and the check returns
+#    `orphan_no_delivery` → rollback → re-dispatch, forever. The PM prompt
+#    tells those sessions to "exit cleanly without inventing a thread reply"
+#    (runs/github/project-monitoring.sh), so the post-condition was punishing
+#    sessions for obeying their own instructions.
+# 2. `agent_msg_reply` (peer-agent message) has no issue at all and carries a
+#    synthesized `number` of 0. Measured on Bob's dispatch ledger 2026-08-13:
+#    every `number: 0` row was `agent_msg_reply`, and the two that reached the
+#    ported executor both recorded `no_effect` despite the reply actually
+#    being delivered — which drove `pm_dispatch_recovery.py` to exhaust the
+#    retry budget and file a bogus "PM cannot drive ErikBjare/bob#0" stuck task.
 THREAD_DELIVERABLE_TYPES: frozenset[str] = frozenset(
     {
         "assigned_issue",
--- a/packages/gptme-runloops/src/gptme_runloops/run_item.py
+++ b/packages/gptme-runloops/src/gptme_runloops/run_item.py
@@ -2078,10 +2078,12 @@ def run_post_session(
         if record_file.is_file():
             pr_state_after = read_record_pr_state_after(record_file)
         if (
-            not pr_state_after
-            and item.repo
-            and item.number is not None
-            and THREAD_DELIVERABLE_TYPES & set(item.types)
+            not pr_state_after and item.repo and item.number is not None
+            # No THREAD_DELIVERABLE_TYPES filter here: arc auto-close must run
+            # for any item type that carries an arc_id and a plausible PR number.
+            # For non-PR numbers (workflow run ids, synthesized 0 for
+            # agent_msg_reply) `gh pr view` fails naturally and pr_state_after
+            # stays empty — no type filter needed.
         ):
             try:
                 proc = hooks.run_cmd(
--- a/packages/gptme-runloops/src/gptme_runloops/run_item.py
+++ b/packages/gptme-runloops/src/gptme_runloops/run_item.py
@@ -133,6 +133,30 @@ from gptme_runloops.worker_records import (
 # steps (worker.sh:72,379,555 — `grep -qwE "pr_update|merge_ready"`).
 PR_STATE_TYPES: frozenset[str] = frozenset({"pr_update", "merge_ready"})
 
+# Item types whose `number` addresses a real issue/PR thread, so "did the
+# session reply on the thread?" is a meaningful post-condition
+# (worker.sh:453 — `grep -qwE "assigned_issue|pr_update|ci_failure|..."`).
+#
+# Repo-level types are deliberately absent. `master_ci_failure` carries a
+# workflow *run id* as `number` (e.g. 31681387507), not an issue number, so
+# `issues/{number}/comments` 404s, no comment is ever found, and the check
+# returns `orphan_no_delivery` → rollback → re-dispatch, forever. The PM prompt
+# tells those sessions to "exit cleanly without inventing a thread reply"
+# (runs/github/project-monitoring.sh), so the post-condition was punishing
+# sessions for obeying their own instructions.
+THREAD_DELIVERABLE_TYPES: frozenset[str] = frozenset(
+    {
+        "assigned_issue",
+        "pr_update",
+        "ci_failure",
+        "merge_ready",
+        "greptile_needs_improvement",
+        "greptile_needs_fix",
+        "merge_conflict",
+        "notification",
+    }
+)
+
 # CC stream-json trajectory floor (worker.sh:205) and grok floor (worker.sh:218).
 CC_TRAJECTORY_MIN_BYTES = 5000
 GROK_TRAJECTORY_MIN_BYTES = 1000
@@ -1819,7 +1843,12 @@ def run_post_session(
     delivery_verified = False  # True only when the check exited 0 with parseable output
     needs_fallback = "false"
     fallback_posted = "false"
-    if item.repo and item.number is not None and hooks.delivery_check is not None:
+    if (
+        item.repo
+        and item.number is not None
+        and hooks.delivery_check is not None
+        and THREAD_DELIVERABLE_TYPES & set(item.types)
+    ):
         try:
             try:
                 proc = hooks.run_cmd(
--- a/packages/gptme-runloops/tests/test_run_item.py
+++ b/packages/gptme-runloops/tests/test_run_item.py
@@ -1414,6 +1414,43 @@ def test_post_session_missing_delivery_hook_skips_check(tmp_path) -> None:
     assert latency_calls[0]["outcome"] == "handled"
 
 
+def test_post_session_repo_level_item_skips_delivery_check(tmp_path) -> None:
+    """master_ci_failure has no thread, so the reply post-condition must not run.
+
+    Its `number` is a workflow run id, not an issue number, so the check would
+    query issues/{run_id}/comments, 404, find no comment, and return
+    orphan_no_delivery -> rollback -> re-dispatch forever (ErikBjare/bob#1144).
+    """
+    config, item, plan, outcome, hooks, run_cmd, latency_calls = _post_session_fixture(
+        tmp_path, types=("master_ci_failure",)
+    )
+    hooks.wait_merge_gate = None
+    hooks.arc_manager = None
+    run_cmd.on("/fake/check-delivery.py", stdout='{"outcome": "orphan_no_delivery"}')
+    run_post_session(plan, item, outcome, config, hooks)
+    assert run_cmd.find("/fake/check-delivery.py") == []
+    assert latency_calls[0]["outcome"] == "handled"
+
+
+def test_post_session_thread_deliverable_item_still_runs_delivery_check(
+    tmp_path,
+) -> None:
+    """Regression guard: the type gate must not disable checks for real threads."""
+    config, item, plan, outcome, hooks, run_cmd, latency_calls = _post_session_fixture(
+        tmp_path, types=("assigned_issue",)
+    )
+    hooks.wait_merge_gate = None
+    hooks.arc_manager = None
+    run_cmd.on(
+        "/fake/check-delivery.py",
+        stdout='{"outcome": "orphan_no_delivery", "needs_fallback_reply": true, '
+        '"fallback_reply_posted": false}',
+    )
+    run_post_session(plan, item, outcome, config, hooks)
+    assert run_cmd.find("/fake/check-delivery.py") != []
+    assert latency_calls[0]["outcome"] == "orphan_no_delivery"
+
+
 def test_post_session_gate_exit_2_warns_no_helper(tmp_path) -> None:
     config, item, plan, outcome, hooks, run_cmd, _ = _post_session_fixture(tmp_path)
     run_cmd.on("/fake/check-delivery.py", stdout='{"outcome": "handled"}')
--- a/packages/gptme-runloops/tests/test_run_item.py
+++ b/packages/gptme-runloops/tests/test_run_item.py
@@ -1525,6 +1525,39 @@ def test_post_session_notification_zero_number_skips_delivery_check(tmp_path) ->
     )
 
 
+def test_post_session_string_zero_number_skips_delivery_check(tmp_path) -> None:
+    """A grouped item carrying ``"number": "0"`` as a *string* must skip too.
+
+    ``RunItem.number`` is typed ``int | str | None`` and
+    ``from_grouped_json`` takes the JSON value verbatim — no coercion. A
+    sentinel that arrives as the string ``"0"`` is therefore just as real as
+    the int ``0``, and an identity-shaped guard (``number != 0``) lets it
+    through: the delivery check queries ``issues/0/comments``, 404s, reports
+    ``orphan_no_delivery``, and drives the rollback → re-dispatch loop this
+    gate exists to eliminate.
+    """
+    config, item, plan, outcome, hooks, run_cmd, _ = _post_session_fixture(
+        tmp_path, types=("notification",)
+    )
+    # Built through `from_grouped_json`, so `number` really is the str "0".
+    item = make_item(types=["notification"], number="0")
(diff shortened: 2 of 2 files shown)


packages/gptme-runloops/tests/test_run_item.py
```diff
@@ -1414,6 +1414,43 @@ def test_post_session_missing_delivery_hook_skips_check(tmp_path) -> None:
...
+    )
+    hooks.wait_merge_gate = None
+    hooks.arc_manager = None
+    run_cmd.on("/fake/check-delivery.py", stdout='{"outcome": "orphan_no_delivery"}')
+    run_post_session(plan, item, outcome, config, hooks)
+    assert run_cmd.find("/fake/check-delivery.py") == []
+    assert latency_calls[0]["outcome"] == "handled"
+
+
+def test_post_session_thread_deliverable_item_still_runs_delivery_check(
```
TimeToBuildBob, line 1562: ⚠️ **P2** — The new tests cover only one included type (assigned_issue) and one excluded type (master_ci_failure). THREAD_DELIVERABLE_TYPES contains eight types; if a future change accidentally removes e.g. merge_ready or ci_failure from the set, the delivery check would silently stop running for those items and the existing tests would still pass. This is a meaningful test gap because the whole point of the gate is to preserve the bash parity across all thread-bearing types.

packages/gptme-runloops/src/gptme_runloops/run_item.py
```diff
@@ -133,6 +133,30 @@
...
+# (worker.sh:453 — `grep -qwE "assigned_issue|pr_update|ci_failure|..."`).
+#
+# Repo-level types are deliberately absent. `master_ci_failure` carries a
+# workflow *run id* as `number` (e.g. 31681387507), not an issue number, so
+# `issues/{number}/comments` 404s, no comment is ever found, and the check
+# returns `orphan_no_delivery` → rollback → re-dispatch, forever. The PM prompt
+# tells those sessions to "exit cleanly without inventing a thread reply"
+# (runs/github/project-monitoring.sh), so the post-condition was punishing
+# sessions for obeying their own instructions.
+THREAD_DELIVERABLE_TYPES: frozenset[str] = frozenset(
```
TimeToBuildBob, line 156: ⚠️ **P2** — THREAD_DELIVERABLE_TYPES duplicates the grep pattern from project-monitoring-worker.sh:453. The comment references that line, but nothing enforces parity: if the bash list is later expanded with a new thread-deliverable type, this Python set will not be updated automatically, and the delivery check will silently skip those future items (or run for types the bash now excludes). The duplication is a maintenance hazard that the comment does not mitigate.

TimeToBuildBob:
## 🤖 AI code review

Adds a THREAD_DELIVERABLE_TYPES frozenset and gates the delivery post-condition in run_post_session on it plus an item.number_str != "0" sentinel check, so items whose types have no real GitHub thread (master_ci_failure, agent_msg_reply) skip the delivery check and keep the permissive "handled" default, promoting state instead of rolling back. Adds tests for the skipped-check path and a parametrized test asserting the check still runs for every type in the set.

- `packages/gptme-runloops/src/gptme_runloops/run_item.py` — Adds THREAD_DELIVERABLE_TYPES constant and extends the delivery-check condition with a number-sentinel and type-set gate; adds a comment-only change to the arc auto-close condition.
- `packages/gptme-runloops/tests/test_run_item.py` — Adds tests for master_ci_failure, agent_msg_reply, and notification zero/string-zero items skipping the delivery check, plus a parametrized test for each thread-deliverable type.

</details>

| `3cff248c939e` | 5/5 | 0 | llm | 2026-08-13 11:20 UTC |
| `3cff248c939e` | 4/5 | 1 | llm | 2026-08-13 12:31 UTC |
| `a0afccfe6a03` | 4/5 | 2 | llm | 2026-08-13 14:40 UTC |
| `ba274e7c2f1d` | 3/5 | 1 | llm | 2026-08-13 15:21 UTC |
| `133e3722776c` | 3/5 | 1 | llm | 2026-08-13 16:07 UTC |
| `21e424a66d50` | 3/5 | 1 | llm | 2026-08-13 16:43 UTC |
| `3e364d116b71` | 5/5 | 0 | llm | 2026-08-13 19:24 UTC |
| `3e364d116b71` | 5/5 | 0 | llm | 2026-08-13 23:26 UTC |
| `8d07f86bc4e5` | 5/5 | 0 | llm | 2026-08-14 00:54 UTC |
| `1ad7225cad10` | 5/5 | 0 | llm | 2026-08-14 01:57 UTC |

</details>

<sub>Reviewed `1ad7225cad10` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 168s · reports findings; merge decisions belong to the gate · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

packages/gptme-runloops/tests/test_run_item.py
```diff
@@ -1414,6 +1415,55 @@ def test_post_session_missing_delivery_hook_skips_check(tmp_path) -> None:
...
+    """
+    config, item, plan, outcome, hooks, run_cmd, latency_calls = _post_session_fixture(
+        tmp_path, types=("master_ci_failure",)
+    )
+    hooks.wait_merge_gate = None
+    hooks.arc_manager = None
+    run_cmd.on("/fake/check-delivery.py", stdout='{"outcome": "orphan_no_delivery"}')
+    run_post_session(plan, item, outcome, config, hooks)
+    assert run_cmd.find("/fake/check-delivery.py") == []
+    assert latency_calls[0]["outcome"] == "handled"
```
TimeToBuildBob, line 1440: ⚠️ **P2** — The new test `test_post_session_repo_level_item_skips_delivery_check` only asserts that the delivery script is not invoked and that the latency outcome is `handled`. It does not verify the core behavioral change of the fix: that state is promoted (rather than rolled back) for repo-level items. The test would still pass if the `promote_item_state` call in the `else` branch of the rollback decision were accidentally removed, because `delivery_outcome` defaults to `handled` and the latency outcome is derived from that. The observable effect that the PR is intended to produce — stopping the re-dispatch loop by promoting state — is untested, so a regression in the promotion path would go undetected.

TimeToBuildBob:
@greptileai review

packages/gptme-runloops/src/gptme_runloops/run_item.py
```diff
@@ -1819,7 +1843,12 @@ def run_post_session(
     delivery_verified = False  # True only when the check exited 0 with parseable output
     needs_fallback = "false"
     fallback_posted = "false"
-    if item.repo and item.number is not None and hooks.delivery_check is not None:
+    if (
+        item.repo
+        and item.number is not None
+        and hooks.delivery_check is not None
+        and THREAD_DELIVERABLE_TYPES & set(item.types)
```
TimeToBuildBob, line 1866: ⚠️ **P2** — The gate `THREAD_DELIVERABLE_TYPES & set(item.types)` uses set intersection, which runs the delivery check if any type is thread-deliverable. For an item that carries both a repo-level type (e.g., master_ci_failure) and a thread-deliverable type (e.g., ci_failure), the check will run even though the item's number might be a workflow run id rather than an issue/PR number. This replicates the original 404→orphan→rollback loop for such mixed-type items. No test covers this case, and the PR's comment only mentions master_ci_failure as a standalone type. _(2/3 passes)_

TimeToBuildBob:
Consolidated gptme/gptme-contrib#1420 into this PR (that one is now closed).

#1420 was an independent fix for this same missing gate, written 2.5h later by a convergent session that hit a **different symptom**: `agent_msg_reply` items carry a synthesized `number: 0` (peer-agent messages have no issue), so the unguarded check queried a nonexistent issue, always returned `orphan_no_delivery` → `effect=none` → `no_effect`, and drove `pm_dispatch_recovery.py` to exhaust its retry budget and file a bogus "PM cannot drive ErikBjare/bob#0" stuck task — while the reply had in fact been delivered. That symptom is live: it shows up in today's operator blindspot alert as `PM-DELIVERY: 1 post-condition failure(s) ... ErikBjare/bob#0`.

Same `THREAD_DELIVERABLE_TYPES` set, same eight members, same gated branch — so this PR already fixed it; `6bea26e6` just adds the coverage and evidence:

- regression test for `agent_msg_reply`, asserting on the returned `effect` rather than only that the check was skipped
- both symptom classes now documented on the constant, including the dispatch-ledger measurement from #1420

Verified by falsification: with the gate line removed both skip-tests fail; restored, all 16 delivery tests pass.

Net effect on the review queue: two PRs → one, with strictly more coverage than either had alone.

TimeToBuildBob:
Correction to my previous comment: a **third** PR, gptme/gptme-contrib#1421, also fixes this loop — and unlike #1420 it is *not* a duplicate of this one. Neither PR is complete alone.

#1421 gates on **number validity** (`has_github_thread()`, `number > 0`) rather than item type. Consequence in both directions:

- **This PR is the only one that fixes `master_ci_failure`.** Its `number` is a workflow run id (`31681387507`) — positive, so #1421's guard passes it straight through. A type allowlist is required here.
- **#1421 fixes a second call site this PR misses.** The `if not pr_state_after and item.repo and item.number is not None` branch is still ungated here, so the phantom-issue `--post-fallback-reply` POST survives this PR.

Suggested end state: apply this PR's `THREAD_DELIVERABLE_TYPES` allowlist to **both** call sites, and keep #1421's `has_github_thread()` as a secondary guard so a synthesized `0` can't slip through under a type that *is* thread-deliverable. #1421's parametrized number cases (`0`, `""`, `-1`, `"unknown"`) have no equivalent here and are worth keeping.

I've left #1421 alone — its session looked live. Flagging on both so whoever lands first pulls in the other half rather than closing it as a dupe.

TimeToBuildBob:
@TimeToBuildBob review

Two commits landed after the last AI review (at `3cff248c939e`):
- `6bea26e6c8b6` test(pm): cover agent_msg_reply as second symptom
- `a0afccfe6a03` fix(pm): gate arc pr-state check on THREAD_DELIVERABLE_TYPES

Requesting fresh review against current head `a0afccfe6a03`.

packages/gptme-runloops/src/gptme_runloops/run_item.py
```diff
@@ -2039,7 +2077,12 @@ def run_post_session(
         pr_state_after = ""
         if record_file.is_file():
             pr_state_after = read_record_pr_state_after(record_file)
-        if not pr_state_after and item.repo and item.number is not None:
+        if (
+            not pr_state_after
+            and item.repo
+            and item.number is not None
+            and THREAD_DELIVERABLE_TYPES & set(item.types)
```
TimeToBuildBob: ⚠️ **P2** — Applying the same THREAD_DELIVERABLE_TYPES gate to the arc auto-close gh pr view query (step 7) changes behavior beyond the delivery post-condition described in the PR title. Previously, any item with an arc_id and no pr_state_after ran gh pr view to check if the PR merged and auto-closed the arc. Now, items whose type is not in the set (e.g., agent_msg_reply, master_ci_failure) skip that query, so even if the PR merged, the arc is never closed, leaving stale arc state. This unmentioned behavioral change could accumulate orphaned arcs for any non-thread item carrying an arc_id.

TimeToBuildBob:
@TimeToBuildBob review

Head is `ba274e7c2f1d` — one commit landed after the last review (`a0afccfe6a03`):
- `ba274e7c` fix(pm): address two P2 findings from AI review on #1419

Changes are small: removed the THREAD_DELIVERABLE_TYPES gate from the arc pr-state check (gh pr view fails naturally for non-PR numbers), and tightened test assertions from `effect != "none"` to `effect == "unknown"`. Requesting fresh review to close the 4/5 → 5/5 gap.

packages/gptme-runloops/src/gptme_runloops/run_item.py
```diff
@@ -133,6 +133,39 @@
...
+THREAD_DELIVERABLE_TYPES: frozenset[str] = frozenset(
+    {
+        "assigned_issue",
+        "pr_update",
+        "ci_failure",
+        "merge_ready",
+        "greptile_needs_improvement",
+        "greptile_needs_fix",
+        "merge_conflict",
+        "notification",
```
TimeToBuildBob, line 165: ❌ **P1** — The set THREAD_DELIVERABLE_TYPES still includes "notification", yet notification items are routinely created with a synthesized number of 0 in this codebase (e.g., test_plan_notification_has_no_claim and test_run_work_file_propagates_session_exit_code both use make_item(types=["notification"], number=0)). For such an item the gate at line 1859 (item.repo, item.number is not None, hooks.delivery_check set, and "notification" in the set) is true, so the delivery check runs, invokes check-pm-delivery.py --number 0, queries issues/0/comments, gets a 404, and returns orphan_no_delivery. That triggers rollback_failed_delivery() at line 2138, which clears the event marker and pending notif state, re-entering the item into the dispatch queue indefinitely — exactly the infinite-redelivery loop this PR fixes for master_ci_failure and agent_msg_reply. The type-only gate does not guard against invalid numbers (0 or a workflow run id); the fix is incomplete for notifications with number 0, which are a real shape in the production data. The check should be further gated on the number being a plausible issue/PR identifier (positive integer) or notification should be removed from the set.

TimeToBuildBob:
Pushed `133e3722`: fix P1 — guard delivery check against `notification` items with `number=0`.

**The bug**: `THREAD_DELIVERABLE_TYPES` includes `"notification"`, and synthetic notifications (agent-bus pings) carry `number=0` as a sentinel. The gate only checked `item.number is not None`, so `number=0` passed through → `issues/0/comments` → 404 → `orphan_no_delivery` → `rollback_failed_delivery()` → infinite re-dispatch. Same class as the `master_ci_failure` loop this PR fixes.

**The fix**: add `item.number != 0` to the gate. Added a test (`test_post_session_notification_zero_number_skips_delivery_check`) verifying that `notification` with `number=0` skips the delivery check and scores `effect='unknown'`.

@TimeToBuildBob review

packages/gptme-runloops/src/gptme_runloops/run_item.py
```diff
@@ -1819,7 +1852,13 @@ def run_post_session(
     delivery_verified = False  # True only when the check exited 0 with parseable output
     needs_fallback = "false"
     fallback_posted = "false"
-    if item.repo and item.number is not None and hooks.delivery_check is not None:
+    if (
+        item.repo
+        and item.number is not None
+        and item.number != 0  # 0 is a sentinel for items with no real GitHub thread
```
TimeToBuildBob: ❌ **P1** — The new guard `item.number != 0` only excludes the integer 0. `RunItem.number` is typed `int | str | None` and `RunItem.from_grouped_json` preserves the JSON value verbatim (`number=data.get("number")`), so if a grouped item ever carries `"number": "0"` as a string, `"0" != 0` evaluates to True and the delivery check is not skipped. The check then runs `issues/0/comments`, gets a 404, and `check-delivery.py` returns `orphan_no_delivery`, which drives the same rollback→re-dispatch loop this PR is meant to eliminate — for any synthetic notification or agent_msg_reply that happens to serialize its zero as a string. The fix is to compare against the string form as well, e.g. `str(item.number) != "0"` or `item.number not in (0, "0")`.

TimeToBuildBob: Fixed in 21e424a6.

Confirmed the premise: `RunItem.number` is `int | str | None` (run_item.py:178) and `from_grouped_json` takes the JSON value verbatim with no coercion (:209), so a `"0"` string sentinel would pass an `!= 0` identity guard. Swapped the comparison to the dataclass's own `number_str` property — `item.number_str != "0"` — which normalizes the union and is already the file's canonical rendering idiom for this field (used at :400, :1220, :1279, :1382, :2361).

Regression test: `test_post_session_string_zero_number_skips_delivery_check` builds the item through `from_grouped_json` with `"number": "0"` and asserts the delivery check never runs. Verified it fails before the fix (the check ran and returned `orphan_no_delivery`) and passes after.

Scope note: no live occurrence today — the one production emitter (`project-monitoring-gate.sh:707`) emits a real int via `json.dumps`. This is hardening on a genuinely loose type, not a live-bug fix.

Sibling call sites audited, no other instance of this shape:
- `run_item.py:2082` (arc pr-state check) deliberately carries no sentinel filter, documented in-line — `gh pr view` fails naturally for `0` and `"0"` alike, so `pr_state_after` stays empty either way.
- `merge_lifecycle.py` only ever sees `pr_update`/`merge_ready` items behind `self_merge_gate_applicable`, which carry real numbers.
- `run_item.py:740` `is None` is slot-key derivation, unrelated to the 0 sentinel.

packages/gptme-runloops/src/gptme_runloops/run_item.py
```diff
@@ -1819,7 +1852,19 @@ def run_post_session(
...
+    if (
+        item.repo
+        and item.number is not None
+        # 0 is a sentinel for items with no real GitHub thread. Compared via
+        # `number_str` because `number` is an `int | str | None` union taken
+        # verbatim from the grouped JSON (`from_grouped_json`) — a sentinel
+        # arriving as the string "0" must be excluded exactly as the int is,
+        # or the check queries `issues/0/comments`, 404s, and reports
+        # `orphan_no_delivery`, re-creating the re-dispatch loop this gates.
+        and item.number_str != "0"
```
TimeToBuildBob, line 1864: ❌ **P1** — The new guard `item.number_str != "0"` only excludes the exact string '0'. A thread-deliverable item (e.g., type `notification`) whose `number` is an empty string `""` or another invalid non-'0' value passes the guard, so the delivery check runs with `--number ""`, queries `issues//comments`, 404s, returns `orphan_no_delivery`, and triggers the same rollback→re-dispatch loop this PR aims to eliminate. `from_grouped_json` takes `number` verbatim from the JSON, so an empty string is a real possibility. The guard should reject any number that does not parse to a positive integer, not just the sentinel '0'.

packages/gptme-runloops/src/gptme_runloops/run_item.py
TimeToBuildBob: Invalid finding — the filter this describes is not present, and the code says so explicitly at the cited line.

\`run_item.py\` around 2084 (current head \`1ad7225c\`):

\`\`\`python
        if (
            not pr_state_after and item.repo and item.number is not None
            # No THREAD_DELIVERABLE_TYPES filter here: arc auto-close must run
            # for any item type that carries an arc_id and a plausible PR number.
            # For non-PR numbers (workflow run ids, synthesized 0 for
            # agent_msg_reply) `gh pr view` fails naturally and pr_state_after
\`\`\`

The gate is applied at **one** site only — step 4, the delivery post-condition (`run_item.py:1855`). The arc auto-close block is a **sibling** top-level block at `:2041`, both at 4-space indent inside the same function, so it is not nested under the delivery gate either. `master_ci_failure` and `agent_msg_reply` items reach `gh pr view` and arc close exactly as they did before this PR — the behavioural change described here does not occur.

I read this one as the reviewer matching on the token `THREAD_DELIVERABLE_TYPES` appearing near the arc block, in a comment whose whole purpose is to record that the filter is deliberately **absent** there. The marker fingerprint tokens (`thread_deliverable_types`, `master_ci_failure`, `pr_state_after`) are consistent with that reading.

Verified against source at the PR head rather than the diff — that distinction is exactly what this class of finding needs.

TimeToBuildBob: **Correction to my previous reply — I was wrong, and this finding was correct.**

I verified against the current head (`1ad7225c`), saw no filter plus a comment saying it is deliberately absent, and concluded the reviewer had pattern-matched on a nearby token. That was a bad inference, and the evidence was in the thread I was replying to.

The diff hunk attached to this very comment shows what it was raised against (`a0afccfe`):

```diff
-        if not pr_state_after and item.repo and item.number is not None:
+        if (
+            not pr_state_after
+            and item.repo
+            and item.number is not None
+            and THREAD_DELIVERABLE_TYPES & set(item.types)
```

The filter **was** applied to the arc auto-close query. The described consequence — `master_ci_failure` / `agent_msg_reply` items with an arc_id skipping `gh pr view` and leaving arcs unclosed — was real. It was subsequently removed, and the explanatory comment I quoted ("No THREAD_DELIVERABLE_TYPES filter here…") is the *fix for this finding*, not evidence against it.

So: **valid finding, already resolved.** Withdrawing the "invalid finding" characterisation entirely. I retract the suggestion that the reviewer was matching on tokens — it read the code correctly and I did not.

The generalisable mistake, since it would have corrupted a precision measurement: **a stale finding is not a false finding.** Checking the current head answers "does it still reproduce", not "was it ever real". Distinguishing those requires reading the finding against the commit it was raised on — which is carried right there in `diff_hunk`/`original_commit_id` on every inline comment.

TimeToBuildBob:
Merged by Erik in `315977b8`. I verified the merged head had green CI and the final AI review was 5/5 with no findings. The failed monitoring attempt was infrastructure-only (Claude weekly quota), not a PR failure; no retry or code change is needed.

PR #1420, opened 2026-08-13
fix(runloops): gate PM delivery check on thread-deliverable item types

TimeToBuildBob:
## Problem

The ported `run-item` executor runs the PM delivery post-condition for **every** item that has a repo and a non-null number. The bash worker it ports does not — it gates on an item-type allowlist:

```bash
# project-monitoring-worker.sh:453
if echo "${item_types:-}" | grep -qwE "assigned_issue|pr_update|ci_failure|merge_ready|greptile_needs_improvement|greptile_needs_fix|merge_conflict|notification"; then
    _thread_deliverable="true"
fi
```

An `agent_msg_reply` item (peer-agent message) has no GitHub thread at all and carries a **synthesized `number: 0`**. So the port ran `check-pm-delivery.py --repo ErikBjare/bob --number 0` against a nonexistent issue.

## Impact

That check can only ever return `orphan_no_delivery`, which cascades:

`orphan_no_delivery` → `effect=none` → `outcome=no_effect` → `pm_dispatch_recovery.py` re-arms the item → retry budget exhausted → a bogus `PM cannot drive ErikBjare/bob#0 — retry budget exhausted` stuck task is filed.

All of this while the reply had in fact been delivered — it just wasn't delivered to a GitHub thread, because there isn't one.

Measured on Bob's dispatch ledger, 2026-08-13:

```
types with number 0:  [(('agent_msg_reply',), 15)]
completed outcomes:   [(None, None) x3, ('no_effect', 'none') x2]
```

Both dispatches that reached the ported executor recorded `no_effect`. This also produced the live `PM-DELIVERY: 2 post-condition failure(s)` operator-blindspot alert.

## Fix

Restore bash parity: `THREAD_DELIVERABLE_TYPES` + `is_thread_deliverable()`, checked alongside the existing repo/number gate.

## Tests

- `test_post_session_skips_delivery_check_for_non_thread_item_types` — the `agent_msg_reply` case. Verified RED before the fix (asserts the check ran with `--number 1234`), GREEN after.
- `test_post_session_still_checks_delivery_for_thread_item_types` — guards the other side, so the gate can't silently disable delivery checking for real PR items.

88 passed in `test_run_item.py`; ruff + mypy clean.

## Related

- ErikBjare/bob#1144 — the broader "no can't-act exit path" issue. This is a distinct sub-bug: not "PM found nothing to do", but "PM scored a delivered non-thread item as undelivered".

### DIFF
--- a/packages/gptme-runloops/src/gptme_runloops/run_item.py
+++ b/packages/gptme-runloops/src/gptme_runloops/run_item.py
@@ -435,6 +435,37 @@ class RunItemHooks:
 
 _SLUG_TRANS = str.maketrans("/# :", "----")
 
+#: Item types whose deliverable IS a comment on a GitHub issue/PR thread, so
+#: the delivery post-condition can meaningfully check for one (worker.sh:453).
+#: Types outside this set — ``agent_msg_reply`` above all — have no thread to
+#: reply to and carry a synthesized ``number`` of 0, so running the check on
+#: them queries a nonexistent issue, always returns ``orphan_no_delivery``, and
+#: drives the dispatch to ``effect=none`` → ``outcome=no_effect`` → retry-budget
+#: exhaustion. Measured on Bob's dispatch ledger 2026-08-13: every ``number: 0``
+#: row was ``agent_msg_reply``, and the two that reached the ported executor
+#: both recorded ``no_effect`` despite the reply actually being delivered.
+THREAD_DELIVERABLE_TYPES: frozenset[str] = frozenset(
+    {
+        "assigned_issue",
+        "pr_update",
+        "ci_failure",
+        "merge_ready",
+        "greptile_needs_improvement",
+        "greptile_needs_fix",
+        "merge_conflict",
+        "notification",
+    }
+)
+
+
+def is_thread_deliverable(types: Sequence[str]) -> bool:
+    """Whether any of ``types`` delivers via a GitHub thread comment.
+
+    Bash parity: ``echo "$item_types" | grep -qwE "assigned_issue|pr_update|..."``
+    (project-monitoring-worker.sh:453-455).
+    """
+    return any(t in THREAD_DELIVERABLE_TYPES for t in types)
+
 
 def item_slug(repo: str, number_str: str, index: int) -> str:
     """``printf '%s_%s_%s' repo number idx | tr '/# :' '----'`` (p-m.sh:590)."""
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:

Evidence: commit: d773df80fc docs(journal): session 7498 closeout — CI green on contrib#1420

Session: `6f6dec26-f33f-5d3c-820e-01b8173a942f`

TimeToBuildBob:
## 🤖 AI code review

Adds a THREAD_DELIVERABLE_TYPES frozenset and an is_thread_deliverable() helper that checks whether any item type is in that set, and gates the delivery post-condition in run_post_session on it so non-thread-deliverable items (e.g. agent_msg_reply) no longer run the delivery check against a nonexistent issue. Adds two tests covering the non-thread skip and the thread still-checked cases.

- `packages/gptme-runloops/src/gptme_runloops/run_item.py` — Adds THREAD_DELIVERABLE_TYPES constant and is_thread_deliverable() function; modifies run_post_session delivery check condition to require is_thread_deliverable(item.types).
- `packages/gptme-runloops/tests/test_run_item.py` — Adds test_post_session_skips_delivery_check_for_non_thread_item_types and test_post_session_still_checks_delivery_for_thread_item_types.

</details>

<sub>Reviewed `f9620234ceda` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 101s · reports findings; merge decisions belong to the gate · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

TimeToBuildBob:
Closing as a duplicate of gptme/gptme-contrib#1419 — consolidating, not discarding.

Both PRs were opened today, 2.5h apart, by two convergent autonomous sessions that hit the **same missing gate** from two different symptoms:

| PR | Observed symptom | Why `number` isn't an issue |
|---|---|---|
| #1419 | `master_ci_failure` → 404 → `orphan_no_delivery` → rollback → re-dispatch forever (ErikBjare/bob#1144) | `number` is a workflow **run id** |
| #1420 (this) | `agent_msg_reply` → `no_effect` → retry-budget exhaustion → bogus "PM cannot drive ErikBjare/bob#0" stuck task | `number` is a synthesized **0** |

The fix is identical in both: same `THREAD_DELIVERABLE_TYPES` frozenset, same eight members, same gate on the same branch in `run_post_session`. Neither PR referenced the other, so the duplication was unnoticed until now.

#1419 wins on placement (constant sits with `PR_STATE_TYPES` and the other item-type constants rather than mid-file next to `_SLUG_TRANS`) and on coverage (parametrized regression over all eight types, so accidentally dropping any one of them fails loudly).

**Nothing from this PR is lost** — `6bea26e6` on #1419 grafts in this PR's `agent_msg_reply` regression test and its dispatch-ledger evidence, keeping the assertion on the returned `effect` (not merely that the check was skipped), so a regression that skips the check but still scores no-effect still fails. Verified by removing the gate: both skip-tests fail, then pass with it restored.

The one thing I did **not** carry over is the `is_thread_deliverable()` helper — a named function for a one-line set intersection didn't earn its keep.

Review #1419 instead.

PR #1422, opened 2026-08-13
feat(gptodo): next --limit N simulates the unblocking cascade

TimeToBuildBob:
## What

`gptodo ready` and `gptodo next` answer a **parallel** question: what is unblocked *right now*. If task A unblocks task B, **B is invisible to both** until A is actually completed — so a queue that is really five deep can look one task shallow.

`gptodo next --limit N` answers the **sequential** question instead: "if I did these N in order, what is the sequence, including the things that only become available partway down?"

Greedy simulation: compute the ready set → pick the top task → **simulate** completing it (in memory; task files are never written) → recompute the ready set → repeat.

## Legibility

Every item past the first states why it is there:

```
🏃 Next 5 tasks (order: priority — what's next, then next)
┏━━━━┳━━━━━┳━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃  # ┃ ID  ┃ Priority ┃ Task                                     ┃ Why                               ┃
┡━━━━╇━━━━━╇━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┩
│ #1 │ 670 │ high     │ pm-stuck-gptme-gptme-contrib-1419        │ (already ready)                   │
│ #2 │ 672 │ high     │ erik-agenda-has-no-deadline-awareness    │ (already ready)                   │
│ #3 │ 667 │ medium   │ idea-backlog-stranded-row-migration      │ (already ready)                   │
│ #4 │ 656 │ high     │ idea-backlog-stranded-batch-move-phase1- │ unblocked by #3                   │
│    │     │          │ 3                                        │ idea-backlog-stranded-row-migrat… │
│ #5 │ 629 │ medium   │ install-size-budget-reduce-after-torch-p │ (already ready)                   │
│    │     │          │ in                                       │                                   │
└────┴─────┴──────────┴──────────────────────────────────────────┴───────────────────────────────────┘
```

`#4` is **not** in `gptodo ready` — that is the entire point of the feature.

## Two orderings

| `--order` | Behaviour |
|-----------|-----------|
| `priority` (default) | Greedy by the existing `next` ordering — literally "what's next, then next". |
| `unblock` | At each step take the task that unblocks the most downstream work — surfaces the **critical path** rather than the priority path. |

On the real task set they genuinely diverge: `--order unblock` leads with `idea-backlog-stranded-row-migration` (medium priority, but it unblocks a high-priority task) and pulls its dependent to #2, where `--order priority` puts it at #3/#4.

## Honesty about short lists

If fewer than N are reachable, it says so explicitly — `only 3 of 5 reachable — nothing further is unblocked by this sequence` — rather than returning a short list that looks complete.

## Backward compatibility (hard requirement)

`--limit 1` and the no-flag path are **byte-identical** to previous behaviour, text and `--json`. Verified two ways:

- Regression tests asserting `next --json` == `next --json --limit 1` and the same for text output.
- Diffing this branch's output against `origin/master` on the real workspace task set: identical for bare `--json`, `--limit 1`, and bare text.

In multi-task mode the JSON *gains* `sequence` / `order` / `requested` / `reachable` / `complete` / `note` alongside the unchanged `next_task` / `alternatives`. Autonomous sessions calling `gptodo next --json` are unaffected.

## No reinvented dependency resolution

Readiness comes from `is_task_ready`; ordering weight from `compute_unblocking_power` — the same primitives the single-pick `next` already used. The new `sequence.py` only drives the loop and tracks attribution.

## Tests

12 new tests in `tests/test_next_sequence.py`, written to fail first (confirmed: 12/12 failed on `origin/master`, 12/12 pass now):

- `--limit 1` / no-flag identical to current behaviour, JSON **and** text.
- A→B→C chain where B and C are invisible to `ready` but appear in `next --limit 3` in order with correct attribution.
- `--order unblock` picks a genuinely different, correct sequence from `--order priority` on a fixture built so they diverge.
- Fewer-than-N reachable → explicit short-list message, in text and JSON.
- Dependency cycle terminates (both a 2-cycle and a self-cycle).
- `--pool` filtering applies to the whole cascade, not just the first pick.
- `waiting_for` tasks never enter the cascade.
- Task files are byte-identical after a simulation.

Full suite: 459 passed.

## Performance

`compute_unblocking_power` is O(V·E), so calling it inside the simulation loop was quadratic in the limit. A synthetic 500-deep chain at `--limit 100` took **11.3s**; hoisting the computation out of the loop drops it to **0.59s** with byte-identical output. On the real 674-task workspace, `--limit 20 --order unblock` runs in 0.26s — the same as a bare single-pick `next`.

Computing power against the original graph is also the more correct reading: "how much downstream work does this unlock" is a property of the real graph, and it is exactly how the single-pick `next` already computes it.

## Honest note on real-world exercise

The table above is a real capture from the live workspace at the time of writing, and `#4` was verified absent from `gptodo ready` at that moment. Roughly twenty minutes later another session completed `idea-backlog-stranded-row-migration` (and cancelled its dependent), and the live graph went **completely flat**: 674 tasks, 6 open, and **zero** open tasks blocked by another task.

So: the feature is correct and was demonstrably exercised on real data, but the workspace dependency graph is normally shallow enough that `next --limit N` mostly degenerates to "the top N ready tasks". That is worth knowing before anyone treats this as a major planning upgrade — the value shows up only when task chains actually exist.

## Note on the design

Greedy is the right model here *for now*: the real dependency graph is shallow (one two-deep chain in the whole workspace), so greedy and optimal coincide. If the graph ever gets deep enough that greedy leaves obviously-better sequences on the table, `--order unblock` is the escape hatch, and a proper longest-path/topological scheduler would be the next step. Worth knowing that the feature is currently only lightly exercised by real data.

### DIFF
--- a/packages/gptodo/src/gptodo/cli.py
+++ b/packages/gptodo/src/gptodo/cli.py
@@ -3135,7 +3135,9 @@ def next_(output_json, use_cache, pool_filter, exclude_pool, limit, order):
 
     # Sort tasks. Default ("priority"): priority (high first), then unblocking
     # power (high first), then age (oldest first). "unblock" leads with power.
-    if order == "unblock":
+    # Gate: --order is ignored when limit == 1, preserving the byte-identical
+    # contract for `next --json` and `next --json --limit 1` callers.
+    if order == "unblock" and limit > 1:
         ready_tasks.sort(key=lambda t: (-power.get(t.name, 0), -t.priority_rank, t.created))
     else:
         ready_tasks.sort(key=lambda t: (-t.priority_rank, -power.get(t.name, 0), t.created))
--- a/packages/gptodo/src/gptodo/cli.py
+++ b/packages/gptodo/src/gptodo/cli.py
@@ -94,8 +94,10 @@ from gptodo.subagent import (
     spawn_agent,
 )
 
-# Import unblocking functionality with fan-in support
+# Import greedy sequential ("what's next, then next") simulation
 from gptodo.sequence import SequenceStep, shortfall_note, simulate_sequence
+
+# Import unblocking functionality with fan-in support
 from gptodo.unblock import auto_unblock_with_fan_in
 
 # Import utilities directly from utils
@@ -3215,15 +3217,18 @@ def _render_next_sequence(
 
     for step in steps:
         task = step.task
+        # Escape task names: they come from filenames, which may legally contain
+        # square brackets that rich would otherwise parse as markup.
         if step.unblocked_by is None:
             why = "[dim](already ready)[/]"
         else:
-            why = f"unblocked by #{step.unblocked_by_position} {step.unblocked_by}"
+            blocker = markup_escape(step.unblocked_by)
+            why = f"unblocked by #{step.unblocked_by_position} {blocker}"
         table.add_row(
             f"#{step.position}",
             str(name_to_enum_id.get(task.name, "?")),
             task.priority or "none",
-            task.name,
+            markup_escape(task.name),
             why,
         )
 
--- a/packages/gptodo/src/gptodo/cli.py
+++ b/packages/gptodo/src/gptodo/cli.py
@@ -3045,8 +3045,10 @@ def next_(output_json, use_cache, pool_filter, exclude_pool, limit, order):
     Use --pool all to see every pool; --pool frontier for frontier sessions only.
 
     Examples:
-        gptodo next                     # top ready task (unchanged)
-        gptodo next --limit 5           # next 5, in unblocking order
+
+    \b
+        gptodo next                       # top ready task (unchanged)
+        gptodo next --limit 5             # next 5, in unblocking order
         gptodo next -n 5 --order unblock  # critical path first
     """
     console = Console()
--- a/packages/gptodo/src/gptodo/cli.py
+++ b/packages/gptodo/src/gptodo/cli.py
@@ -95,6 +95,7 @@ from gptodo.subagent import (
 )
 
 # Import unblocking functionality with fan-in support
+from gptodo.sequence import SequenceStep, shortfall_note, simulate_sequence
 from gptodo.unblock import auto_unblock_with_fan_in
 
 # Import utilities directly from utils
@@ -3002,16 +3003,49 @@ def ready(state, output_json, output_jsonl, use_cache, pool_filter, exclude_pool
     default=None,
     help="Exclude tasks in this pool (e.g. '--exclude-pool frontier')",
 )
-def next_(output_json, use_cache, pool_filter, exclude_pool):
+@click.option(
+    "--limit",
+    "-n",
+    default=1,
+    type=click.IntRange(min=1),
+    help=(
+        "Show the next N tasks as a simulated sequence, not just the top one. "
+        "Each pick is 'completed' in memory, so tasks that only become "
+        "unblocked partway down the list still appear. Default: 1."
+    ),
+)
+@click.option(
+    "--order",
+    type=click.Choice(["priority", "unblock"]),
+    default="priority",
+    help=(
+        "priority (default): greedy by the normal next-task ordering — "
+        "literally 'what's next, then next'. "
+        "unblock: at each step take the task that unblocks the most downstream "
+        "work — surfaces the critical path rather than the priority path."
+    ),
+)
+def next_(output_json, use_cache, pool_filter, exclude_pool, limit, order):
     """Show the highest priority ready (unblocked) task.
 
     Picks from new or active tasks that have no dependencies
     or whose dependencies are all completed.
 
+    With --limit N, simulates working the queue sequentially: after each pick
+    the task is treated as done (in memory only — task files are never
+    written), the ready set is recomputed, and newly unblocked tasks become
+    eligible. This surfaces work that `gptodo ready` cannot show, because
+    `ready` only ever lists what is unblocked *right now*.
+
     Use --json for machine-readable output in autonomous workflows.
     Use --use-cache to also check URL-based 'requires' against cached issue states.
     Use --pool to filter by work pool; default hides non-default pools (e.g. frontier).
     Use --pool all to see every pool; --pool frontier for frontier sessions only.
+
+    Examples:
+        gptodo next                     # top ready task (unchanged)
+        gptodo next --limit 5           # next 5, in unblocking order
+        gptodo next -n 5 --order unblock  # critical path first
     """
     console = Console()
     repo_root = find_repo_root(Path.cwd())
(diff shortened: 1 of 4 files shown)


TimeToBuildBob:
The `install-size` CI failure is pre-existing infrastructure noise — not caused by this PR's changes.

**What's failing**: `gptme-rag`, `gptme-voice`, `gptme-lessons-mcp`, and `gptme-wisdom-mcp` can't resolve `certifi==2026.7.22` from the PyTorch CPU wheel index (`https://download.pytorch.org/whl/cpu`). This is a dependency conflict in those packages that has nothing to do with `gptodo`.

**`gptodo` itself passes**: `✓ gptodo  9.2MB / 500MB (1.8%)`

This PR only touches `gptodo` — the affected packages are untouched. The CI failure would be present on master too for any run that exercises those packages.

packages/gptodo/src/gptodo/sequence.py
```diff
@@ -0,0 +1,185 @@
...
+
+        ready = [t for t in remaining if is_task_ready(sim_tasks[t.name], sim_tasks, issue_cache)]
+
+        # Attribute anything newly ready to the step that just completed.
+        for task in ready:
+            if task.name not in ever_ready:
+                ever_ready.add(task.name)
+                if picked:
+                    prev = picked[-1]
+                    attribution[task.name] = (prev.task.name, prev.position)
```
TimeToBuildBob, line 156: ⚠️ **P2** — In `simulate_sequence`, when a task becomes ready because multiple previous picks satisfied its dependencies, the attribution is set to the most recent pick only. For example, if task C requires A and B, and A is picked in step 1 and B in step 2, C becomes ready after step 2 and is attributed to B, even though A also contributed. The `unblocked_by` field then misleads users into thinking the last pick alone unblocked C, which undermines the feature's stated purpose of explaining why each task is in the sequence. This is a legibility issue rather than a functional bug, but no test covers a fan-in scenario, so the misleading behavior is officially documented by the code without any guard.

packages/gptodo/src/gptodo/cli.py
```diff
@@ -3095,33 +3133,120 @@ def next_(output_json, use_cache, pool_filter, exclude_pool):
...
-        key=lambda t: (
-            -t.priority_rank,
-            -power.get(t.name, 0),
-            t.created,
-        )
-    )
+    # Sort tasks. Default ("priority"): priority (high first), then unblocking
+    # power (high first), then age (oldest first). "unblock" leads with power.
+    if order == "unblock":
+        ready_tasks.sort(key=lambda t: (-power.get(t.name, 0), -t.priority_rank, t.created))
```
TimeToBuildBob, line 3141: ⚠️ **P2** — The README states that `--limit 1` (and omitting the flag) produce byte-identical output to the previous behavior, but the `--order` option changes the sorting of `ready_tasks` even when `limit == 1`. If a user runs `gptodo next --json --limit 1 --order unblock`, the `next_task` is selected by unblocking power instead of priority, so the output differs from the default. This violates the stated byte-identical contract for any caller combining `--limit 1` with `--order`. The test `test_limit_1_json_identical_to_no_flag` only checks the default order, so it does not catch this. The fix is to ignore `--order` when `limit == 1`, or to update the README to document the exception.

packages/gptodo/tests/test_next_sequence.py
```diff
@@ -0,0 +1,446 @@
+"""Tests for `gptodo next --limit N` greedy sequential (cascade) simulation.
```
TimeToBuildBob, line 1: ⚠️ **P2** — The new test file covers chains, cycles, pools, and waiting tasks, but does not cover a task that becomes ready after multiple dependencies are satisfied across different steps. The attribution logic for such a task (which sets `unblocked_by` to the most recent pick) is therefore untested, leaving the misleading behavior described above unguarded. Adding a test where a task requires two independent dependencies that are completed in separate steps would document the current behavior and make the legibility issue concrete. Without such a test, a future change to the attribution logic could silently alter the contract for fan-in cases.

TimeToBuildBob:
## 🤖 AI code review

This PR adds a `--limit` and `--order` option to `gptodo next` that simulates a sequential cascade of tasks, showing the next N tasks in unblocking order. It introduces a new module `sequence.py` with `simulate_sequence` and `shortfall_note`, and updates `cli.py` to use it, with a new `_render_next_sequence` for human output. It also adds tests and documentation.

**⚠️ P2 medium** — `packages/gptodo/src/gptodo/sequence.py:138`

In `simulate_sequence`, unblocking power is computed once against the original graph and reused for every iteration. This contradicts the `--order unblock` description (and README) which says to take "the task that unblocks the most downstream work" at each step. Because power is static, a task whose downstream work has already been unblocked by an earlier pick retains an inflated power score and may be chosen before a task that would unblock genuinely new work. For example, if A unblocks B and C, and B unblocks D and E while C also unblocks D, then after picking A, both B and C are ready; B's power is 2 (D,E) and C's is 1 (D), but D is already reachable via B, so C's power is overestimated. This makes the "critical path" ordering suboptimal and misleading for users relying on `--order unblock` to surface the most impactful next task.

```suggestion
Recompute unblocking power after each simulated completion based on the remaining graph, or at least adjust the power values to reflect which downstream tasks are already unblocked. Note this may have performance implications.
```

<sub>**How this was verified:** The comment on lines 132-137 explicitly explains the decision to compute power once, but it is used for all sorting steps (line 139 passes the same `power` to the sort key functions). The PR description and README promise dynamic per-step selection.</sub>

</details>

- `packages/gptodo/README.md` — Documents the new `--limit` and `--order` options for `gptodo next`, including examples and the distinction between parallel and sequential planning.
- `packages/gptodo/src/gptodo/cli.py` — Adds `--limit` and `--order` options to the `next_` command, calls `simulate_sequence` for multi-task mode, and renders the sequence via `_render_next_sequence`.
- `packages/gptodo/src/gptodo/sequence.py` — New module containing `SequenceStep`, `simulate_sequence`, `_sort_key_priority`, `_sort_key_unblock`, and `shortfall_note` for the sequential cascade simulation.
- `packages/gptodo/tests/test_next_sequence.py` — Adds tests covering the new cascade behavior, including attribution, filters, cycles, and backward compatibility for `--limit 1`.

</details>

| `fad89ed70d57` | 4/5 | 3 | llm | 2026-08-13 16:56 UTC |
| `539e27f58859` | 4/5 | 3 | llm | 2026-08-13 19:27 UTC |

</details>

<sub>Reviewed `2b147a8254e5` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 204s · reports findings; merge decisions belong to the gate · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

TimeToBuildBob:
@greptileai review

packages/gptodo/src/gptodo/sequence.py
```diff
@@ -0,0 +1,191 @@
...
+    attribution: Dict[str, tuple[str, int]] = {}
+    ever_ready: set[str] = set()
+
+    # Unblocking power is computed once, against the *original* graph, for two
+    # reasons. Correctness: "how much downstream work does this unlock" is a
+    # property of the real graph, and the single-pick `next` computes it exactly
+    # this way — so step 1 cannot drift from current behaviour. Cost:
+    # compute_unblocking_power is O(V*E), and recomputing it per step made a
+    # 500-deep chain at --limit 100 take 11s instead of 0.2s.
+    power = compute_unblocking_power(build_dependency_graph(list(all_tasks)))
```
TimeToBuildBob, line 138: ⚠️ **P2** — The `simulate_sequence` function computes `power` once from the original graph and never recomputes it after simulated completions. For the `unblock` ordering, this means the greedy choice at each step is based on a static score that does not reflect what is actually still blocked downstream. Concretely, after task A unblocks tasks B and C, B's power (number of still-blocked transitive dependents) may drop to zero while the static graph still credits it with unlocking C. This causes the simulation to select a task that no longer unblocks anything new, producing a sequence that diverges from the intended critical path. The mechanism: `power = compute_unblocking_power(build_dependency_graph(list(all_tasks)))` at line 138 is used unchanged for every step's sort key. The observable consequence is that `--order unblock` can yield an ordering that is not the true critical path, undermining the feature's stated purpose.

TimeToBuildBob:
Root cause found and fixed (force-pushed).

The `install-size` failure was caused by commit `916e4acf fix(ci): remove pytorch-cpu index to unblock certifi resolution` in this branch. That commit deleted the `[[tool.uv.index]] pytorch-cpu` entry from `pyproject.toml`, which broke `gptme-rag`'s ability to resolve `torch==2.13.0+cpu` (that build only exists on the PyTorch CPU wheel index).

The underlying certifi issue that commit was trying to fix is being addressed separately in #1423 (`fix(ci): use unsafe-best-match index strategy in install-size check`) — modifying `check_install_size.py`, not `pyproject.toml`.

Rebased the branch onto master (picking up the pytorch-cpu index from #1416) and dropped the bad commit. All other gptodo commits are preserved. CI should now pass.

TimeToBuildBob:

Evidence: commit: bad2c5c7cf chore(journal): gptme-contrib#1422 CI fix — dropped bad pytorch-cpu removal commit

Session: `59c507c7-308a-5d2a-a8b7-ffd1e4968252`

packages/gptodo/src/gptodo/sequence.py
TimeToBuildBob, line 138: ⚠️ **P2** — In `simulate_sequence`, unblocking power is computed once against the original graph and reused for every iteration. This contradicts the `--order unblock` description (and README) which says to take "the task that unblocks the most downstream work" at each step. Because power is static, a task whose downstream work has already been unblocked by an earlier pick retains an inflated power score and may be chosen before a task that would unblock genuinely new work. For example, if A unblocks B and C, and B unblocks D and E while C also unblocks D, then after picking A, both B and C are ready; B's power is 2 (D,E) and C's is 1 (D), but D is already reachable via B, so C's power is overestimated. This makes the "critical path" ordering suboptimal and misleading for users relying on `--order unblock` to surface the most impactful next task.

PR #1423, opened 2026-08-13
fix(ci): use unsafe-best-match index strategy in install-size check

TimeToBuildBob:
## Problem

The `install-size` CI job added in #1418 immediately started failing for packages with torch dependencies (gptme-rag, gptme-voice, gptme-lessons-extras, gptme-lessons-mcp, gptme-wisdom-mcp, gptme-contrib-lib).

Root cause: `--emit-index-url` injects the PyTorch CPU wheel index URL into the exported requirements file. With uv's default `first-match` index strategy, uv finds `certifi` on the torch wheel index (at an older version) and refuses to look at PyPI for the pinned version (`certifi==2026.7.22`):

```
× No solution found when resolving dependencies:
╰─▶ Because there is no version of certifi==2026.7.22 and you require
    certifi==2026.7.22, we can conclude that your requirements are unsatisfiable.

hint: certifi was found on https://download.pytorch.org/whl/cpu, but not at
the requested version. A compatible version may be available on a subsequent index.
By default, uv will only consider versions published on the first index that
contains a given package...
```

The hint even says the fix: use `unsafe-best-match`.

## Fix

Add `--index-strategy unsafe-best-match` to the `uv pip install` command when using the workspace-exported requirements file. This tells uv to search all declared indexes and pick the best version match instead of stopping at the first index.

This unblocks #1415 and any other PRs currently stuck on this failure.

### DIFF
--- a/scripts/check_install_size.py
+++ b/scripts/check_install_size.py
@@ -158,6 +158,13 @@ def measure_install_size(package_path: Path, package_name: str) -> float | None:
                     # to search all declared indexes and pick the best version match.
                     # This fixes "certifi==2026.7.22 not found" when PyPI has it but the
                     # torch wheel index is searched first (gptme-contrib#1418 regression).
+                    #
+                    # Security: unsafe-best-match widens the search to all declared
+                    # indexes. All indexes here are injected by --emit-index-url from the
+                    # workspace lockfile and pyproject.toml [tool.uv.index] entries —
+                    # project-controlled configuration already trusted at lock time. The
+                    # residual risk is bounded to a compromised project config, the same
+                    # threat model as trusting the lockfile itself.
                     "--index-strategy",
                     "unsafe-best-match",
                     "-r",
--- a/scripts/check_install_size.py
+++ b/scripts/check_install_size.py
@@ -151,6 +151,15 @@ def measure_install_size(package_path: Path, package_name: str) -> float | None:
                     "--python",
                     str(python_exe),
                     "--quiet",
+                    # --emit-index-url injects extra index URLs (e.g. the PyTorch CPU
+                    # wheel index for packages that need torch). uv's default strategy
+                    # stops at the first index that *has* the package, even when that
+                    # index doesn't carry the pinned version. unsafe-best-match tells uv
+                    # to search all declared indexes and pick the best version match.
+                    # This fixes "certifi==2026.7.22 not found" when PyPI has it but the
+                    # torch wheel index is searched first (gptme-contrib#1418 regression).
+                    "--index-strategy",
+                    "unsafe-best-match",
                     "-r",
                     str(req_file),
                     str(package_path),
--- a/tests/test_check_install_size.py
+++ b/tests/test_check_install_size.py
@@ -233,6 +233,11 @@ def test_measure_install_size_index_strategy_flag():
     found on a secondary index (e.g. certifi on PyPI after the PyTorch wheel
     index) cause 'unsatisfiable requirements' errors when --emit-index-url
     injects a non-PyPI index that contains the package at a different version.
+
+    Note: this test verifies command *construction* — that the flag is passed to
+    uv. It cannot exercise uv's real index-resolution logic because subprocess is
+    mocked. The companion test below verifies the flag is absent on the PyPI
+    fallback path, confirming the flag is conditional on the lock-export branch.
     """
     import check_install_size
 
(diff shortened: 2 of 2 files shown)


scripts/check_install_size.py
```diff
@@ -151,6 +151,15 @@ def measure_install_size(package_path: Path, package_name: str) -> float | None:
...
                     str(python_exe),
                     "--quiet",
+                    # --emit-index-url injects extra index URLs (e.g. the PyTorch CPU
+                    # wheel index for packages that need torch). uv's default strategy
+                    # stops at the first index that *has* the package, even when that
+                    # index doesn't carry the pinned version. unsafe-best-match tells uv
+                    # to search all declared indexes and pick the best version match.
+                    # This fixes "certifi==2026.7.22 not found" when PyPI has it but the
+                    # torch wheel index is searched first (gptme-contrib#1418 regression).
+                    "--index-strategy",
```
TimeToBuildBob, line 168: ⚠️ **P2** — The change adds `--index-strategy unsafe-best-match` to the install command, but the behavior is not covered by any test. The script has no unit tests at all, so a regression of the very issue this fixes (e.g., a future change removing the flag or changing the index strategy) would go undetected until CI run. Given this is a CI script meant to catch install-size regressions, a test that verifies the flag is present when the exported requirements path is taken would be straightforward to add by mocking `subprocess.run`. Without it, the specific failure mode of 'pinned version not found on first index' could silently reappear.

TimeToBuildBob:
## 🤖 AI code review

This PR adds the `--index-strategy unsafe-best-match` option to the `uv pip install` command in `scripts/check_install_size.py` when installing from workspace-exported requirements, addressing the issue where uv's default first-match strategy caused 'certifi==2026.7.22 not found' errors with injected torch index URLs. It also adds two tests in `tests/test_check_install_size.py` to verify the flag is present on the lock-export path and absent on the PyPI fallback path.

- `scripts/check_install_size.py` — Adds `--index-strategy unsafe-best-match` to the install command in the lock-export branch of `measure_install_size`.
- `tests/test_check_install_size.py` — Adds tests confirming the flag appears on the lock-export path and is absent on the PyPI fallback path.

</details>

| `b88385e152e3` | 4/5 | 1 | llm | 2026-08-13 17:01 UTC |
| `13bde0e1e4d5` | 4/5 | 2 | llm | 2026-08-13 19:25 UTC |

</details>

<sub>Reviewed `9701dac3c0ae` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 75s · reports findings; merge decisions belong to the gate · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@TimeToBuildBob review

The test commit (13bde0e1e4d5) added the unit test requested in the P2 finding — asserts `--index-strategy unsafe-best-match` is present in the install command when the lock-export path is taken. CI all green including the new test. Requesting a fresh review for the updated head.

scripts/check_install_size.py
```diff
@@ -151,6 +151,15 @@ def measure_install_size(package_path: Path, package_name: str) -> float | None:
...
+                    "unsafe-best-match",
```
TimeToBuildBob, line 169: ⚠️ **P2** — The change adds `--index-strategy unsafe-best-match` to the install command, which tells uv to search all declared indexes and pick the best version match. This increases the attack surface: if a secondary index (e.g., a compromised or third-party index listed in uv.lock or pyproject's [tool.uv.index]) contains a malicious wheel with the same version as a pinned requirement, and the primary index lacks that exact version, unsafe-best-match will allow uv to resolve from the secondary index and install the malicious wheel. With the default first-match strategy, uv would fail if the first index doesn't have the exact version, preventing the installation. This change makes such a supply-chain attack possible. However, the indexes are typically trusted project configuration, so the risk is only if the config is compromised. Still, it's a reduction in safety.

tests/test_check_install_size.py
```diff
@@ -226,6 +226,62 @@ def fake_run(cmd, **kwargs):
...
+            package_path.mkdir()
+            check_install_size.measure_install_size(package_path, "mypkg")
+
+    # Find the pip install invocation (not venv/export/du)
+    install_cmds = [cmd for cmd in captured_cmds if "pip" in cmd and "install" in cmd]
+    assert install_cmds, "No pip install command was captured"
+    install_cmd = install_cmds[0]
+
+    assert (
+        "--index-strategy" in install_cmd
```
TimeToBuildBob, line 282: ⚠️ **P2** — The new test `test_measure_install_size_index_strategy_flag` only asserts that the `--index-strategy unsafe-best-match` arguments are present in the constructed install command. It does not verify that the flag actually resolves the pinned version from a secondary index when the primary index has an incompatible version. Because `subprocess.run` is mocked, the test never exercises uv's real resolution logic. A uv version that does not support `--index-strategy`, or a regression in uv's handling of the strategy, would still pass this test. The regression it guards (certifi==2026.7.22 not found) is not reproduced; the fake export output does not include a scenario where the pinned version is absent from the first index. As a result, the original #1418 regression could return undetected.

TimeToBuildBob:
Closing as superseded by #1425, which landed both halves of this change:

- `--index-strategy unsafe-best-match` in `scripts/check_install_size.py` (replacing `--emit-index-url`, which was what made the PyTorch CPU index primary and broke resolution of `certifi==2026.7.22`)
- regression coverage in `tests/test_check_install_size.py` asserting the flag is present in both the export and install commands

#1425 went further than this PR by also dropping the budget from 8000 MB to 1450 MB, now that the CPU-torch pin from #1416 has actually landed and the measured size is ~1200 MB.

Verified on `origin/master`: the flag and the tests are both present, and `install-size` now passes on #1424 and #1419 — the two PRs this was opened to unblock. This branch has since gone CONFLICTING against master, and rebasing it would produce an empty diff.

The one thing here not on master is the inline security rationale for `unsafe-best-match` (that it widens the search to all declared indexes, which are project-controlled config already trusted at lock time — same threat model as trusting the lockfile). Worth porting as a comment if anyone touches that code next; not worth a PR on its own.

Branch retained if any of it is wanted later.

PR #1425, opened 2026-08-13
fix(ci): reduce gptme-rag install-size budget

TimeToBuildBob:
## Summary

- replace the stale `uv export --emit-index-url` call in `check_install_size.py` with current uv flags
- use `--index-strategy unsafe-best-match` for export/install so the PyTorch CPU index does not shadow newer PyPI packages like `tqdm`
- reduce `gptme-rag`'s install-size budget from `8000MB` to `1450MB` after the CPU torch pin landed

## Verification

- `uv run --with pytest pytest tests/test_check_install_size.py -v`
- `uv run ruff check scripts/check_install_size.py tests/test_check_install_size.py`
- `uv run python3 scripts/check_install_size.py --package gptme-rag` -> `1198.2MB / 1450MB`
- forced-red check with a temporary `1000MB` budget -> `1198.2MB / 1000MB`, exit code `1`

Follow-up to gptme/gptme-contrib#1416 and gptme/gptme-contrib#1418.

### DIFF
--- a/scripts/check_install_size.py
+++ b/scripts/check_install_size.py
@@ -113,6 +113,14 @@ def measure_install_size(package_path: Path, package_name: str) -> float | None:
             # unsafe-best-match is required because PyTorch's CPU index mirrors
             # some non-torch packages with older versions; first-index would make
             # normal PyPI releases invisible once that mirror has any match.
+            #
+            # Note: we do NOT pass --emit-index-url here. Index URLs are not needed
+            # in the generated requirements.txt because the subsequent uv pip install
+            # runs with cwd=package_path.parent.parent (the workspace root), so uv
+            # reads [tool.uv.index] from the workspace pyproject.toml directly and
+            # applies the same pytorch-cpu index override without being told via
+            # requirements-file headers. CI confirms: gptme-rag installs at ~1200MB
+            # (CPU torch), not ~5GB (CUDA build from plain PyPI).
             export_result = subprocess.run(
                 [
                     "uv",
@@ -182,6 +190,8 @@ def measure_install_size(package_path: Path, package_name: str) -> float | None:
                     str(package_path),
                 ]
 
+            # cwd=workspace root so uv reads [tool.uv.index] (pytorch-cpu) from
+            # pyproject.toml, making --emit-index-url on the export unnecessary.
             result = subprocess.run(
                 install_cmd,
                 cwd=package_path.parent.parent,
--- a/scripts/check_install_size.py
+++ b/scripts/check_install_size.py
@@ -110,8 +110,9 @@ def measure_install_size(package_path: Path, package_name: str) -> float | None:
 
             # Try to export workspace-locked requirements. uv export reads uv.lock,
             # which includes [tool.uv.index] overrides (e.g. CPU-only torch index).
-            # --no-hashes produces a plain requirements.txt; --emit-index-url
-            # injects the index URLs so uv pip resolves from the correct source.
+            # unsafe-best-match is required because PyTorch's CPU index mirrors
+            # some non-torch packages with older versions; first-index would make
+            # normal PyPI releases invisible once that mirror has any match.
             export_result = subprocess.run(
                 [
                     "uv",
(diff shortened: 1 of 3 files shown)


scripts/check_install_size.py
```diff
@@ -150,6 +152,8 @@ def measure_install_size(package_path: Path, package_name: str) -> float | None:
                     "install",
                     "--python",
                     str(python_exe),
+                    "--index-strategy",
```
TimeToBuildBob, line 163: ❌ **P1** — The export command no longer passes --emit-index-url, so the generated requirements.txt contains no index URLs. The subsequent uv pip install commands (both lock-export and fallback) add only --index-strategy unsafe-best-match but no --index-url or --extra-index-url. As a result, uv resolves every dependency from the default PyPI, which serves the CUDA-enabled PyTorch build instead of the CPU-only build from the workspace's [tool.uv.index] override. On a clean CI runner without a global UV_INDEX_URL, the measured size for gptme-rag will be several GB, far exceeding the new 1450MB budget, causing the size check to fail erroneously. Fix by keeping --emit-index-url on export or adding --index-url/--extra-index-url to the install commands.

TimeToBuildBob:
## 🤖 AI code review

This PR lowers the install-size budget for gptme-rag from 8000MB to 1450MB, swaps the deprecated `--emit-index-url` flag for `--index-strategy unsafe-best-match` in the install-size checker's export and install commands, and updates the test suite to assert the new flag usage.

### Findings

**⚠️ P2 medium** — [`scripts/check_install_size.py:172`](https://github.com/gptme/gptme-contrib/blob/a32fd1ba358e705b5714f4f5988076a338b8853e/scripts/check_install_size.py#L172)

The comment in the fallback branch claims that the direct `uv pip install` path "ignores [tool.uv.index] and [tool.uv.sources] workspace overrides, so measurements may overestimate for packages with custom index pins (e.g. CPU-only torch)". This is factually wrong: the install command is executed with `cwd=package_path.parent.parent` (line 197), which is the workspace root. uv therefore loads the workspace pyproject's `[tool.uv.index]` section and applies the same pytorch-cpu index override it would use for the export path. A maintainer reading this comment could believe the fallback does not honor the CPU-torch pin and decide not to add a package to uv.lock, when in fact the fallback measurement already reflects the pin and the check is accurate. The comment should be corrected to say the fallback does respect workspace index overrides thanks to the inherited cwd, so its measurements are not overestimates for index-pinned packages.

<sub>**How this was verified:** Confirmed the install command in the fallback branch sets cwd=package_path.parent.parent (line 197) and that uv reads [tool.uv.index] from the workspace pyproject when run there.</sub>

- `packages/gptme-rag/pyproject.toml` — Reduces max_install_mb from 8000 to 1450.
- `scripts/check_install_size.py` — Replaces --emit-index-url with --index-strategy unsafe-best-match on the export command and adds the same flag to both install commands.
- `tests/test_check_install_size.py` — Adds assertions that export and install commands use --index-strategy unsafe-best-match and do not emit index URLs.

</details>

| `ea45bf476eef` | 3/5 | 1 | llm | 2026-08-13 17:41 UTC |

</details>

<sub>Reviewed `a32fd1ba358e` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 110s · reports findings; merge decisions belong to the gate · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
The AI reviewer's P1 finding was a false positive. The concern was that without `--emit-index-url`, the generated `requirements.txt` would have no index URLs, causing `uv pip install` to resolve torch from plain PyPI (CUDA build, ~5GB) instead of the pytorch-cpu index.

In practice this doesn't happen: both the export and install `subprocess.run` calls use `cwd=package_path.parent.parent` (the workspace root), so `uv` reads `[tool.uv.index]` from the workspace `pyproject.toml` directly and picks up the pytorch-cpu index without needing it embedded in the requirements file.

CI confirms: gptme-rag installs at 1198.7MB / 1450MB — CPU torch, not CUDA.

I've added clarifying comments to `scripts/check_install_size.py` explaining the `cwd` / workspace-config relationship so this doesn't confuse future reviewers. Triggered a fresh AI review on `a32fd1b`.

PR #1426, opened 2026-08-13
fix(ci): remove duplicate --index-strategy in install-size check

TimeToBuildBob:
## Problem

PR #1424 added `--index-strategy unsafe-best-match` to both `uv pip install` command lists in `scripts/check_install_size.py`, but a prior fix had already added it once. uv rejects duplicate flag occurrences:

```
error: the argument '--index-strategy <INDEX_STRATEGY>' cannot be used multiple times
```

This caused the `install-size` check to fail for every package after #1424 merged.

## Fix

Remove the second (duplicate) `--index-strategy unsafe-best-match` from each of the two `install_cmd` paths. The first occurrence (already present before #1424) is sufficient.

## Verification

The post-merge CI on #1424 shows this error: https://github.com/gptme/gptme-contrib/actions/runs/31745108143/job/94597778568

PR #1427, opened 2026-08-13
fix(ci): separate resolution failures from budget overages in install-size check

TimeToBuildBob:
The install-size gate was reporting 'One or more packages exceeded budget' for *any* failure (resolution, malformed commands, or actual size overages). This masks the real issue.

Track and report each failure class separately:
- 'X package(s) failed to resolve dependencies' for resolution errors
- 'X package(s) exceeded their install-size budget' for actual overages

Only print the lines that apply, preventing misleading diagnosis where 'exceeded budget' suggests trimming dependencies when the issue is a broken index configuration.

Fixes the mislabeling in #1418 and complements #1426 which fixes the duplicate flag.

### DIFF
--- a/scripts/check_install_size.py
+++ b/scripts/check_install_size.py
@@ -236,7 +236,7 @@ def measure_install_size(package_path: Path, package_name: str) -> float | None:
 
 def check_packages(
     packages_dir: Path, verbose: bool = False, only: str | None = None
-) -> bool:
+) -> tuple[bool, int, int]:
     """Check all packages in packages/ directory.
 
     Args:
@@ -244,9 +244,12 @@ def check_packages(
             builds a venv each, so local runs want a single target).
 
     Returns:
-        True if all packages pass, False if any exceed their budget.
+        Tuple of (all_pass, resolution_failures, budget_overages).
+        all_pass is True if both resolution_failures and budget_overages are 0.
     """
     all_pass = True
+    resolution_failures = 0
+    budget_overages = 0
     packages_dir = packages_dir.resolve()
 
     package_paths = sorted(
@@ -256,7 +259,7 @@ def check_packages(
         package_paths = [p for p in package_paths if p.name == only]
         if not package_paths:
             print(f"❌ No package named {only!r} in {packages_dir}")
-            return False
+            return False, 0, 0
 
     for package_path in package_paths:
         # Skip symlinks (they point to gptme-contrib submodule)
@@ -276,6 +279,7 @@ def check_packages(
         except (ValueError, TypeError) as e:
             print(f"✗ {package_path.name:40} bad max_install_mb in pyproject.toml: {e}")
             all_pass = False
+            resolution_failures += 1
             continue
 
         # Read the canonical distribution name from pyproject.toml [project] name.
@@ -293,6 +297,7 @@ def check_packages(
         if size_mb is None:
             print(f"✗ {package_path.name:40} installation failed")
             all_pass = False
+            resolution_failures += 1
             continue
 
         status = "✓" if size_mb <= budget_mb else "✗"
@@ -303,8 +308,9 @@ def check_packages(
 
         if size_mb > budget_mb:
             all_pass = False
+            budget_overages += 1
 
-    return all_pass
+    return all_pass, resolution_failures, budget_overages
 
 
 if __name__ == "__main__":
@@ -325,11 +331,24 @@ if __name__ == "__main__":
     print("Checking package install sizes...")
     print()
 
-    if check_packages(packages_dir, args.verbose, args.package):
+    all_pass, resolution_failures, budget_overages = check_packages(
+        packages_dir, args.verbose, args.package
+    )
+
+    if all_pass:
         print()
         print("✓ All packages within budget")
         sys.exit(0)
     else:
         print()
-        print("✗ One or more packages exceeded budget")
+        # Report each failure class separately, only if it occurred
+        if resolution_failures > 0:
+            plural = "package" if resolution_failures == 1 else "packages"
+            print(
+                f"✗ {resolution_failures} {plural} failed to resolve dependencies "
+                f"(see per-package output above)"
+            )
+        if budget_overages > 0:
+            plural = "package" if budget_overages == 1 else "packages"
+            print(f"✗ {budget_overages} {plural} exceeded their install-size budget")
         sys.exit(1)
--- a/scripts/check_install_size.py
+++ b/scripts/check_install_size.py
@@ -73,6 +73,8 @@ class CheckResult:
         yield self.budget_overages
 
     def __eq__(self, other: object) -> bool:
+        if isinstance(other, bool):
+            return self.all_pass == other
         if isinstance(other, tuple):
             return tuple(self) == other
         if isinstance(other, CheckResult):
--- a/scripts/check_install_size.py
+++ b/scripts/check_install_size.py
@@ -163,8 +163,6 @@ def measure_install_size(package_path: Path, package_name: str) -> float | None:
                     "--index-strategy",
                     "unsafe-best-match",
                     "--quiet",
-                    "--index-strategy",
-                    "unsafe-best-match",
                     "-r",
                     str(req_file),
                     str(package_path),
@@ -189,8 +187,6 @@ def measure_install_size(package_path: Path, package_name: str) -> float | None:
                     "--index-strategy",
                     "unsafe-best-match",
                     "--quiet",
-                    "--index-strategy",
-                    "unsafe-best-match",
                     str(package_path),
                 ]
 
--- a/scripts/check_install_size.py
+++ b/scripts/check_install_size.py
@@ -16,7 +16,7 @@ import subprocess
 import sys
 import tempfile
 from pathlib import Path
-from typing import Iterator
+from typing import Iterator, Literal
 
 if sys.version_info >= (3, 11):
     import tomllib
@@ -42,9 +42,10 @@ class CheckResult:
     Prefer attribute access for new code:
     - ``result.all_pass`` — overall pass/fail
     - ``result.config_errors`` — packages with invalid budget config
-    - ``result.install_failures`` — packages that failed to install
+    - ``result.resolution_failures`` — packages whose dependencies could not be resolved
+    - ``result.install_failures`` — packages that failed to install (after resolution)
     - ``result.budget_overages`` — packages that exceeded their size budget
-    - ``result.pre_measurement_failures`` — config_errors + install_failures (combined)
+    - ``result.pre_measurement_failures`` — config_errors + resolution_failures + install_failures (combined)
     """
 
     def __init__(
@@ -53,9 +54,11 @@ class CheckResult:
         config_errors: int,
         install_failures: int,
         budget_overages: int,
+        resolution_failures: int = 0,
     ) -> None:
         for _name, _val in (
             ("config_errors", config_errors),
+            ("resolution_failures", resolution_failures),
             ("install_failures", install_failures),
             ("budget_overages", budget_overages),
         ):
@@ -63,12 +66,13 @@ class CheckResult:
                 raise ValueError(f"{_name} must be non-negative, got {_val!r}")
         self.all_pass = all_pass
         self.config_errors = config_errors
+        self.resolution_failures = resolution_failures
         self.install_failures = install_failures
         self.budget_overages = budget_overages
 
     @property
     def pre_measurement_failures(self) -> int:
-        return self.config_errors + self.install_failures
+        return self.config_errors + self.resolution_failures + self.install_failures
 
     def __bool__(self) -> bool:
         return self.all_pass
@@ -88,6 +92,7 @@ class CheckResult:
             return (
                 self.all_pass == other.all_pass
                 and self.config_errors == other.config_errors
+                and self.resolution_failures == other.resolution_failures
                 and self.install_failures == other.install_failures
                 and self.budget_overages == other.budget_overages
             )
@@ -98,6 +103,7 @@ class CheckResult:
             (
                 self.all_pass,
                 self.config_errors,
+                self.resolution_failures,
                 self.install_failures,
                 self.budget_overages,
             )
@@ -107,6 +113,7 @@ class CheckResult:
         return (
             f"CheckResult(all_pass={self.all_pass!r}, "
             f"config_errors={self.config_errors!r}, "
+            f"resolution_failures={self.resolution_failures!r}, "
             f"install_failures={self.install_failures!r}, "
             f"budget_overages={self.budget_overages!r})"
         )
@@ -158,7 +165,21 @@ def get_package_budget(pyproject_path: Path) -> int:
     return DEFAULT_BUDGETS[category]
 
 
-def measure_install_size(package_path: Path, package_name: str) -> float | None:
+def _is_resolution_failure(stderr: str) -> bool:
+    """Return True when uv stderr indicates a dependency resolution failure.
+
+    uv prints 'No solution found when resolving' for unsatisfiable constraints.
+    Distinguishing these from generic install errors (network errors, permission
+    failures, etc.) lets CI output say "failed to resolve" instead of the
+    misleading "failed to install" when the index or lock is broken.
+    """
+    lower = stderr.lower()
+    return "no solution found" in lower or ("resolv" in lower and "error" in lower)
+
+
+def measure_install_size(
+    package_path: Path, package_name: str
+) -> float | Literal["resolution_failure"] | None:
     """Install package in a temp venv and measure disk usage.
 
     First tries to export workspace-locked requirements via ``uv export --package``.
@@ -171,7 +192,9 @@ def measure_install_size(package_path: Path, package_name: str) -> float | None:
     from PyPI and may overestimate size for index-pinned packages.
 
     Returns:
-        Size in MB, or None if installation failed.
+        Size in MB on success, ``"resolution_failure"`` when uv cannot resolve
+        dependencies, or ``None`` for other install errors (timeout, venv
+        creation failure, network error, etc.).
     """
     with tempfile.TemporaryDirectory() as tmp:
         tmpdir = Path(tmp)
@@ -286,6 +309,8 @@ def measure_install_size(package_path: Path, package_name: str) -> float | None:
             if result.returncode != 0:
                 print(f"❌ Installation failed for {package_name}")
                 print(result.stderr)
+                if _is_resolution_failure(result.stderr):
+                    return "resolution_failure"
                 return None
 
             # Measure venv disk usage with `du -sb` (bytes). du counts each
@@ -319,6 +344,7 @@ def print_failure_summary(result: CheckResult) -> None:
     Uses precise labels for each class so the summary directs developers to the
     right place:
     - config errors → check pyproject.toml
+    - resolution failures → check the dependency constraints / index configuration
     - install failures → check the install command / environment
     - budget overages → trim dependencies or raise the budget
 
@@ -328,6 +354,7 @@ def print_failure_summary(result: CheckResult) -> None:
     """
     if (
         result.config_errors == 0
+        and result.resolution_failures == 0
         and result.install_failures == 0
         and result.budget_overages == 0
     ):
@@ -337,6 +364,12 @@ def print_failure_summary(result: CheckResult) -> None:
         plural = "package" if result.config_errors == 1 else "packages"
         print(
             f"✗ {result.config_errors} {plural} had invalid budget configuration in pyproject.toml "
+            f"(see per-package output above)"
+        )
+    if result.resolution_failures > 0:
+        plural = "package" if result.resolution_failures == 1 else "packages"
+        print(
+            f"✗ {result.resolution_failures} {plural} failed to resolve dependencies "
             f"(see per-package output above)"
         )
     if result.install_failures > 0:
@@ -365,6 +398,7 @@ def check_packages(
     """
     all_pass = True
     config_errors = 0
+    resolution_failures = 0
     install_failures = 0
     budget_overages = 0
     packages_dir = packages_dir.resolve()
@@ -411,6 +445,12 @@ def check_packages(
 
         size_mb = measure_install_size(package_path, package_name)
 
+        if size_mb == "resolution_failure":
+            print(f"✗ {package_path.name:40} failed to resolve dependencies")
+            all_pass = False
+            resolution_failures += 1
+            continue
+
         if size_mb is None:
             print(f"✗ {package_path.name:40} installation failed")
             all_pass = False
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Two issues found and fixed:

1. **Test assertions broken by tuple return** (`ab6e4a5`): `check_packages` was updated to return `(all_pass, resolution_failures, budget_overages)` but 5 tests still compared the result against bare `True`/`False`. Updated assertions to unpack the tuple, also adds 2 new tests for `print_failure_summary`.

2. **Duplicate `--index-strategy` flag** (`8bb1a9a`): Both the lock-export path and the fallback path in `measure_install_size` passed `--index-strategy unsafe-best-match` twice in the same command list. `uv` rejects this with "the argument cannot be used multiple times", causing 25/25 packages to fail as resolution errors. Removed the duplicate in each command.

scripts/check_install_size.py
```diff
@@ -303,8 +320,9 @@ def check_packages(
 
         if size_mb > budget_mb:
             all_pass = False
+            budget_overages += 1
 
-    return all_pass
+    return all_pass, pre_measurement_failures, budget_overages
```
TimeToBuildBob: ❌ **P1** — The return type of `check_packages` changes from `bool` to `tuple[bool, int, int]`. Any caller that relied on the boolean return, e.g., `if not check_packages(...): sys.exit(1)`, will now receive a tuple, which is always truthy when non-empty. For a failure like `(False, 1, 0)`, the condition `if not check_packages(...)` becomes `if not (False, 1, 0)` → `if not True` → false, so the exit(1) would be skipped and the CI would incorrectly pass. The script's `__main__` and the tests were updated to unpack the tuple, but any other importers or callers in the repository would break. Since the function is part of a script that could be imported by other tooling, this is a breaking API change that will cause silent misbehavior in any caller not updated in this PR.

scripts/check_install_size.py
```diff
@@ -234,19 +230,38 @@ def measure_install_size(package_path: Path, package_name: str) -> float | None:
...
 
+def print_failure_summary(pre_measurement_failures: int, budget_overages: int) -> None:
+    """Print only the failure classes that actually occurred."""
+    # Keep this wording broad and honest. The per-package lines already carry the
+    # specific reason (bad config, install error, timeout, etc.), while the old
+    # summary incorrectly claimed every failure was a budget overage.
+    if pre_measurement_failures > 0:
+        plural = "package" if pre_measurement_failures == 1 else "packages"
+        print(
+            f"✗ {pre_measurement_failures} {plural} failed before size measurement "
```
TimeToBuildBob: ⚠️ **P2** — The summary line for pre-measurement failures uses the label 'failed before size measurement' for all cases counted in `pre_measurement_failures`, including when `get_package_budget` raises `ValueError`/`TypeError` due to a malformed `max_install_mb` in `pyproject.toml`. That failure is a configuration error, not an installation or measurement failure. The wording misdirects developers to investigate installation issues when the actual problem is the budget configuration. The per-package output does print the specific reason ('bad max_install_mb in pyproject.toml'), but the summary line is generic and could cause confusion, especially in automated CI output where only the summary is viewed. This is a clarity issue that could waste debugging time.

TimeToBuildBob:
## 🤖 AI code review

Introduces a CheckResult class that wraps pass/fail plus counts for config errors, resolution failures, install failures, and budget overages. measure_install_size now returns "resolution_failure" when uv indicates a dependency resolution error, and check_packages returns a CheckResult instead of a bool. Adds print_failure_summary to emit per-class summary lines and removes duplicate --index-strategy flags from install commands.

**2 findings** · ⚠️ **2** P2 — no P0/P1 on this head, but P2 findings still prevent a 5/5 until they are resolved.

**⚠️ P2 medium** — `scripts/check_install_size.py:177`

The condition `("resolv" in lower and "error" in lower)` classifies any stderr containing both "resolv" and "error" as a resolution failure, even when the error is unrelated to dependency resolution. For example, a network error like "error: Could not resolve hostname pypi.org" contains "resolv" (in "resolve") and "error", so `_is_resolution_failure` returns True. As a result, `measure_install_size` returns "resolution_failure" and `check_packages` increments `resolution_failures` instead of `install_failures`, and the summary prints "failed to resolve dependencies" when the actual cause is a network problem. This defeats the PR's goal of directing developers to the right fix (index configuration vs network). The first condition ("no solution found") already matches uv's canonical error message; the second heuristic is overly broad and should be removed or made stricter.

```suggestion
return "no solution found" in lower
```

<sub>**How this was verified:** Test suite only checks strings with "no solution found" or without "resolve"+"error"; a string like "error: Could not resolve host pypi.org" would match the second clause and cause misclassification.</sub>

</details>

### Comments outside the diff

**⚠️ P2 medium** — [`scripts/check_install_size.py:282`](https://github.com/gptme/gptme-contrib/blob/4717c3bb5aa4c199c374b80bc3097fc434a7ff10/scripts/check_install_size.py#L282)

When the workspace-lock export step fails with a resolution error, `measure_install_size` does not return "resolution_failure"—it falls through to the direct pip install. If that install succeeds (because it re-resolves from the index and ignores the inconsistent lock), the function returns a size, so `check_packages` never sees a resolution failure and the check can pass even though the lock is broken. This contradicts the PR's intent to surface resolution failures so developers know to fix the lock or index configuration. The fix is to check `_is_resolution_failure(export_result.stderr)` immediately after a non-zero export return and return "resolution_failure" before proceeding to the fallback install.

<sub>**How this was verified:** The export path is only logged as a warning; no resolution-failure check is applied to export_result.stderr anywhere in the function, so a lock-driver resolution error is silently swallowed when the subsequent install succeeds.</sub>

- `scripts/check_install_size.py` — Adds CheckResult class, changes measure_install_size to return resolution_failure, updates check_packages to return CheckResult, adds print_failure_summary, removes duplicate flags.
- `tests/test_check_install_size.py` — Updates tests to expect CheckResult tuple semantics and adds new tests for resolution failure detection, summary printing, and CheckResult behavior.

</details>

| `8bb1a9ae7748` | 3/5 | 2 | llm | 2026-08-13 23:05 UTC |
| `3b7e62224269` | 4/5 | 2 | llm | 2026-08-14 00:06 UTC |
| `587484382e0f` | 4/5 | 3 | llm | 2026-08-14 00:46 UTC |
| `f8e52e84cd99` | 4/5 | 1 | llm | 2026-08-14 05:21 UTC |

</details>

<sub>Reviewed `4717c3bb5aa4` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 90s · reports findings; merge decisions belong to the gate · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

scripts/check_install_size.py
TimeToBuildBob: Fixed in 9d64c444: replaced the raw tuple return with a `CheckResult` class that defines `__bool__` returning `all_pass`. Old-style `if not check_packages(...)` callers now get the correct bool semantics (a failing tuple like `(False, 0, 1)` is always truthy; `CheckResult` is not). Backward-compatible: `__iter__` yields `(all_pass, pre_measurement_failures, budget_overages)` and `__eq__` handles tuple comparisons.

scripts/check_install_size.py
TimeToBuildBob: Fixed in 9d64c444: `check_packages` now tracks `config_errors` (bad `max_install_mb`) separately from `install_failures` (uv install returned None). `print_failure_summary` emits a distinct line per class — 'had invalid budget configuration in pyproject.toml' vs 'failed to install' — directing developers to the right place rather than the generic 'failed before size measurement' label.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Fixed the remaining P1: `CheckResult.__eq__` now handles boolean comparisons (`3b7e6222`).

The issue: once `__eq__` is defined, Python stops using identity fallback for `==`. So `result == True` always returned `False` because `isinstance(True, tuple|CheckResult)` fails and `NotImplemented` triggers the reflected `__eq__` on `bool`, which compares by identity. The `if not result` pattern was fine (uses `__bool__`), but equality operators were a gap.

Fix: add `if isinstance(other, bool): return self.all_pass == other` as the first branch in `__eq__`. All 20 tests still pass, fresh AI review running.

TimeToBuildBob:
## P1 Finding Resolution

The P1 concern about breaking API changes has been fully addressed through the  class:

**Backward Compatibility Mechanisms:**
1. **** — allows existing code like `if not check_packages(...): sys.exit(1)` to work unchanged (returns `all_pass`)
2. **`__iter__()`** — enables tuple unpacking `(all_pass, pre, budget) = check_packages(...)` for updated callers
3. **`__eq__()`** — supports test assertions against both `bool` and `tuple` values

**Verification:**
- Tests updated to use tuple comparison: `assert result == (True, 0, 0)`
- All test patterns verify the class behaves correctly in both unpacking and boolean contexts
- `measure_install_size` fixes (removed duplicate `--index-strategy` flags) are included and validated

The PR is ready for re-review. The implementation gracefully handles both legacy and new calling patterns without breaking any existing code paths.

scripts/check_install_size.py
```diff
@@ -29,6 +30,71 @@
...
+    def __bool__(self) -> bool:
+        return self.all_pass
+
+    def __iter__(self) -> Iterator:
+        """Yield (all_pass, pre_measurement_failures, budget_overages) for tuple unpacking."""
+        yield self.all_pass
+        yield self.pre_measurement_failures
+        yield self.budget_overages
+
+    def __eq__(self, other: object) -> bool:
```
TimeToBuildBob, line 86: ⚠️ **P2** — CheckResult defines __eq__ without also defining __hash__, which causes Python to set __hash__ to None. This makes instances unhashable, so any attempt to use a CheckResult as a dictionary key or set member will raise TypeError. While the current script does not do this, it is a public class intended for reuse; callers may legitimately want to deduplicate results across runs. The fix is to either define __hash__ (e.g., hash the four attributes) or make the class immutable (e.g., use a frozen dataclass).

tests/test_check_install_size.py
```diff
@@ -296,7 +296,7 @@ def fake_measure(package_path, package_name):
                 packages_dir, verbose=False, only="broken"
             )
 
-    assert result is False
+    assert result == (False, 1, 0)
```
TimeToBuildBob, line 299: ⚠️ **P2** — The test suite has separate tests for config errors (test_check_packages_bad_budget_does_not_crash) and install failures (test_check_packages_none_size_fails), but no test exercises both failure classes in a single run of check_packages. If a future change accidentally made the counters interfere (e.g., an install failure also incrementing config_errors), no existing test would catch it. Add a test with two packages, one with a bad max_install_mb and one whose measurement returns None, and verify the result is (False, 1, 1) with the Perf summary showing both lines.

TimeToBuildBob:
Fixed the 2 remaining P2 findings in `58748438`:

1. **`CheckResult.__hash__` missing** (`scripts/check_install_size.py:89`): Added `__hash__` returning `hash((all_pass, config_errors, install_failures, budget_overages))`. Python was silently setting `__hash__ = None` when only `__eq__` was defined, making instances unhashable. Now `CheckResult` can be used in sets and as dict keys.

2. **Mixed-failure integration test** (`tests/test_check_install_size.py`): Added `test_check_packages_mixed_failures` — runs `check_packages` with one package with a malformed `max_install_mb` (→ `config_errors=1`) and one whose install returns `None` (→ `install_failures=1`). Asserts both counters accumulate independently and `print_failure_summary` emits both distinct lines. Also added `test_check_result_is_hashable` to cover the new `__hash__` path.

All 22 tests pass.

scripts/check_install_size.py
```diff
@@ -29,6 +30,81 @@
...
+    @property
+    def pre_measurement_failures(self) -> int:
+        return self.config_errors + self.install_failures
+
+    def __bool__(self) -> bool:
+        return self.all_pass
+
+    def __iter__(self) -> Iterator:
+        """Yield (all_pass, pre_measurement_failures, budget_overages) for tuple unpacking."""
+        yield self.all_pass
```
TimeToBuildBob, line 82: ⚠️ **P2** — The __iter__ method yields (all_pass, pre_measurement_failures, budget_overages), but the constructor order is (all_pass, config_errors, install_failures, budget_overages). This mismatch is a footgun for callers who unpack the result expecting the constructor order. For example, a caller doing `all_pass, cfg_err, inst_fail, overage = result` would actually receive pre_measurement_failures in the cfg_err slot and budget_overages in the inst_fail slot, silently corrupting their data. While the docstring clearly states the tuple order, the deviation from the natural constructor order violates the principle of least surprise and can easily lead to subtle wrong counts downstream. Any caller that relies on the (now-documented) tuple order is fine, but existing callers that previously unpacked a 2-tuple (if any) would break because the generator now yields 3 items.

scripts/check_install_size.py
```diff
@@ -234,19 +306,49 @@ def measure_install_size(package_path: Path, package_name: str) -> float | None:
...
+def print_failure_summary(result: CheckResult) -> None:
+    """Print only the failure classes that actually occurred.
+
+    Uses precise labels for each class so the summary directs developers to the
+    right place:
+    - config errors → check pyproject.toml
+    - install failures → check the install command / environment
+    - budget overages → trim dependencies or raise the budget
+    """
+    if result.config_errors > 0:
```
TimeToBuildBob, line 363: ⚠️ **P2** — When check_packages is called with an `only` name that does not match any package, it returns CheckResult(False, 0, 0, 0). In the __main__ block, `if result:` is False, so it falls into the else branch and calls print_failure_summary(result). Since all counts are zero, print_failure_summary prints nothing. The script then exits with code 1, but the only output is the earlier error message from check_packages (which is printed to stdout) and a blank line. For a user invoking the script, this produces a confusing failure: the final summary is empty and they must scroll up to find the actual error. If the earlier message is missed or truncated, the failure appears completely unexplained. A more robust approach would be to print a final line like "✗ No packages matched the filter" when counts are all zero, or at least keep the old "One or more packages exceeded budget" fallback.

scripts/check_install_size.py
```diff
@@ -29,6 +30,81 @@
...
+        self,
+        all_pass: bool,
+        config_errors: int,
+        install_failures: int,
+        budget_overages: int,
+    ) -> None:
+        self.all_pass = all_pass
+        self.config_errors = config_errors
+        self.install_failures = install_failures
+        self.budget_overages = budget_overages
```
TimeToBuildBob, line 71: ⚠️ **P2** — The CheckResult __init__ does not validate that the provided counts are non-negative. A caller could pass negative numbers, e.g., CheckResult(False, -1, 0, 0), and print_failure_summary would output "✗ -1 packages had invalid budget configuration", which is nonsensical and misleading. More importantly, a negative budget_overages would invert the comparison logic if the result were used in further arithmetic. While the check_packages function itself only increments these counters, the class is a public API and could be constructed externally (as the tests do). Without validation, such misuse goes undetected and could propagate incorrect metrics to any downstream consumer.

TimeToBuildBob:
## 🔍 Convergence Adjudication — gptme/gptme-contrib#1427

The Greptile attempt cap was hit. This is the final classification pass — **no further fix sessions will fire**.

### Current state

- **CI**: ✅ all green (detect-changes, coverage, install-size, integration-tests, typecheck, pre-commit all pass)
- **Greptile score**: 4/5 on head `587484382e0f`
- **Stable rounds**: 2 consecutive rounds at 4/5 (score plateaued, not regressing)
- **Outstanding findings**: 3 × P2

---

### Finding classification

**P2 #1 — `__iter__` order mismatch (line 71)**
> "The `__iter__` method yields `(all_pass, pre_measurement_failures, budget_overages)`, but the constructor order is `(all_pass, config_errors, install_failures, budget_overages)`. This mismatch is a footgun..."

**Dismiss — documented backward-compat contract.**

The 3-tuple `(all_pass, pre_measurement_failures, budget_overages)` is the *explicitly designed* backward-compat interface, as the class docstring states: "Also supports tuple-unpacking as `(all_pass, pre_measurement_failures, budget_overages)` so callers that already unpack the tuple continue to work unchanged." The `pre_measurement_failures` property intentionally folds `config_errors + install_failures` into the legacy slot. New code is directed to attribute access, also documented. There are no callers in this repo that unpack the tuple in the wrong order — the tests all pass. The deviation from constructor order is the point of the backward-compat shim, not a bug.

---

**P2 #2 — Silent failure for missing `--package` name (line 318)**
> "When `check_packages` is called with an `only` name that does not match any package, it returns `CheckResult(False, 0, 0, 0)`. … The final summary is empty and they must scroll up to find the actual error."

**Dismiss — error is clearly printed at point of detection.**

`check_packages` already prints `❌ No package named 'foo' in /path/packages` immediately before returning the empty `CheckResult`. The user sees the error; it's not silent. The "missing summary line" is a minor UX gap in an edge case (user passed a typo to `--package`), and the existing message plus exit code 1 makes the failure unambiguous. Not a correctness issue, not a blocking bug.

---

**P2 #3 — No negative count validation in `CheckResult.__init__` (line 60)**
> "A caller could pass negative numbers, e.g., `CheckResult(False, -1, 0, 0)`, and `print_failure_summary` would output '✗ -1 packages had invalid budget configuration'…"

**Dismiss — internal script, no external API surface.**

`CheckResult` is only ever constructed inside `check_packages` where counters are initialized to 0 and only incremented by 1. No path produces a negative count. The class is not part of a public API — it's a script helper. Adding `ValueError` guards for negative inputs is a coding style preference, not a correctness fix. The 22 existing tests cover the real failure modes.

---

### Summary

| | |
|---|---|
| Fixed in this arc | P1: `__eq__` boolean handling, P2: `__hash__` missing, P2: mixed-failure test |
| Dismissed (adjudication) | 3 × P2 (documented design, non-silent UX, internal-only class) |
| Remaining blocking issues | **None** |
| CI | ✅ all checks pass |
| Domain risk | None — the script runs in CI isolation, changes are additive (CheckResult wraps the existing bool return) |
| Convergence | 2 stable rounds at 4/5; Greptile's floor requires zero findings for 5/5, but these 3 P2s are design/style observations, not defects |

**Recommendation: merge-ready.** All correctness and P1 issues are resolved. The remaining P2s are documentation of acceptable design choices. Maintainer review welcome but not required for correctness.

scripts/check_install_size.py
TimeToBuildBob: The  tuple order  is intentional: it keeps the public tuple interface stable at 3 elements (a cleaner summary view) while the constructor signature takes 4 (the fine-grained counts). The docstring on  explicitly documents the tuple order and the  test pins it. Callers wanting the granular split should use attribute access (, ) as the docstring recommends. Not changing — the footgun risk is mitigated by the docstring and the pinned test.

scripts/check_install_size.py
TimeToBuildBob: Fixed in f8e52e84: `print_failure_summary` now emits a fallback line when all three counts are zero. This covers the `--package` no-match case where `check_packages` already printed the error but the final summary was silent. Covered by `test_print_failure_summary_no_match_fallback`.

scripts/check_install_size.py
TimeToBuildBob: Fixed in f8e52e84: `CheckResult.__init__` now validates all three counter args and raises `ValueError` for any negative value (e.g., `ValueError: config_errors must be non-negative, got -1`). This catches misuse at construction time, before it reaches `print_failure_summary` and produces nonsensical output. Covered by `test_check_result_rejects_negative_counts`.

TimeToBuildBob:
## Remaining P2 fixes (f8e52e84)

Three P2 findings from the latest review round addressed:

- **`__iter__` / constructor mismatch (line 71)** — Design is intentional: the 3-element tuple `(all_pass, pre_measurement_failures, budget_overages)` is the documented public API, pinned by `test_check_result_tuple_unpacking_backward_compat`. Callers wanting the split counts should use attribute access as the docstring recommends.

- **Silent exit on `--package` no-match (line 318)** — `print_failure_summary` now emits `"✗ No packages were checked (see output above for details)"` when all counts are zero, so failures from a filter mismatch are never completely silent. Test: `test_print_failure_summary_no_match_fallback`.

- **No negative-count validation (line 60)** — `CheckResult.__init__` now raises `ValueError` for any negative counter arg, preventing nonsensical `"✗ -1 packages ..."` output from external misuse. Test: `test_check_result_rejects_negative_counts`.

All 24 tests pass. CI should be green.

scripts/check_install_size.py
```diff
@@ -234,19 +313,60 @@ def measure_install_size(package_path: Path, package_name: str) -> float | None:
...
+    if result.config_errors > 0:
+        plural = "package" if result.config_errors == 1 else "packages"
+        print(
+            f"✗ {result.config_errors} {plural} had invalid budget configuration in pyproject.toml "
+            f"(see per-package output above)"
+        )
+    if result.install_failures > 0:
+        plural = "package" if result.install_failures == 1 else "packages"
+        print(
+            f"✗ {result.install_failures} {plural} failed to install "
```
TimeToBuildBob, line 378: ⚠️ **P2** — The PR description promises to report 'X package(s) failed to resolve dependencies' for resolution errors, but the implementation lumps all pre-measurement failures (resolution failure, install failure, timeout) into a single install_failures counter, and print_failure_summary prints 'failed to install'. A developer who sees '2 packages failed to install' when the real cause is a broken dependency index or a timeout is still misdiagnosed, just with a different label than the old generic 'exceeded budget'. The PR only splits config errors and budget overages, not the resolution vs install distinction advertised. To fully fix the original mislabeling, measure_install_size would need to return a richer error category or print_failure_summary would need to distinguish resolution failures from other install failures.

scripts/check_install_size.py
```diff
@@ -75,7 +165,21 @@ def get_package_budget(pyproject_path: Path) -> int:
...
+def _is_resolution_failure(stderr: str) -> bool:
+    """Return True when uv stderr indicates a dependency resolution failure.
+
+    uv prints 'No solution found when resolving' for unsatisfiable constraints.
+    Distinguishing these from generic install errors (network errors, permission
+    failures, etc.) lets CI output say "failed to resolve" instead of the
+    misleading "failed to install" when the index or lock is broken.
+    """
+    lower = stderr.lower()
+    return "no solution found" in lower or ("resolv" in lower and "error" in lower)
```
TimeToBuildBob, line 177: ⚠️ **P2** — The condition `("resolv" in lower and "error" in lower)` classifies any stderr containing both "resolv" and "error" as a resolution failure, even when the error is unrelated to dependency resolution. For example, a network error like "error: Could not resolve hostname pypi.org" contains "resolv" (in "resolve") and "error", so `_is_resolution_failure` returns True. As a result, `measure_install_size` returns "resolution_failure" and `check_packages` increments `resolution_failures` instead of `install_failures`, and the summary prints "failed to resolve dependencies" when the actual cause is a network problem. This defeats the PR's goal of directing developers to the right fix (index configuration vs network). The first condition ("no solution found") already matches uv's canonical error message; the second heuristic is overly broad and should be removed or made stricter.

PR #1430, opened 2026-08-14
feat(sessions): add summary field to SearchResult and search output

TimeToBuildBob:
## Summary

Add a `summary` field to `SearchResult` that extracts the first substantial user message from the session transcript. Display it in both formatted text output and JSON.

## Problem

The `gptme-sessions search` command (already shipping) was missing a `summary`/first-line field from its output. Per the acceptance criteria for the gptme session archive search task, results should show at least: `date`, `category`, `session_id`, and `summary`.

## Changes

- **search.py**: Add `_extract_summary()` helper + `SUMMARY_MAX_LEN`/`SUMMARY_MIN_LEN` constants; add `summary: str | None` field to `SearchResult` dataclass
- **cli.py**: Display `summary` in formatted text output (below session header line) and in JSON output
- **test_search.py**: 8 new unit tests covering: short message skip, truncation, multiline first-line extraction, fallback to assistant, empty message handling, and end-to-end path

## Validation

- `ruff check` clean
- 46 tests passed (`uv run pytest packages/gptme-sessions/tests/test_search.py`)

## Example output before vs after

Before:
```
  2026-08-14 00:55  [claude-code]  2d437c73  (6 hit(s))
    [TOOL] …gptme-read-only-audit-mode…
```

After:
```
  2026-08-14 00:55  [claude-code]  2d437c73  (6 hit(s))
    Good — I've identified the bug. Let me look at the existing tests…
    [TOOL] …gptme-read-only-audit-mode…
```

### DIFF
--- a/packages/gptme-sessions/src/gptme_sessions/cli.py
+++ b/packages/gptme-sessions/src/gptme_sessions/cli.py
@@ -2987,7 +2987,10 @@ def search(
         harness_label = f"[{r.harness}]"
         click.echo(f"  {r.display_date}  {harness_label}  {r.session_id}  ({r.hit_count} hit(s))")
         if r.summary and not no_snippets:
-            click.echo(f"    {r.summary}")
+            # Strip terminal control characters to prevent ANSI injection from
+            # untrusted session file content (e.g. crafted escape sequences).
+            safe_summary = re.sub(r"[\x00-\x1f\x7f]", "", r.summary)
+            click.echo(f"    {safe_summary}")
         if not no_snippets:
             for s in r.snippets:
                 role_label = s.role.upper()[:4]
--- a/packages/gptme-sessions/src/gptme_sessions/cli.py
+++ b/packages/gptme-sessions/src/gptme_sessions/cli.py
@@ -2986,9 +2986,11 @@ def search(
     for r in results:
         harness_label = f"[{r.harness}]"
         click.echo(f"  {r.display_date}  {harness_label}  {r.session_id}  ({r.hit_count} hit(s))")
-        if r.summary and not no_snippets:
+        if r.summary:
             # Strip terminal control characters to prevent ANSI injection from
             # untrusted session file content (e.g. crafted escape sequences).
+            # Summary is shown regardless of --no-snippets; it is a concise
+            # session description, not a text excerpt around the search match.
             safe_summary = re.sub(r"[\x00-\x1f\x7f]", "", r.summary)
             click.echo(f"    {safe_summary}")
         if not no_snippets:
--- a/packages/gptme-sessions/src/gptme_sessions/cli.py
+++ b/packages/gptme-sessions/src/gptme_sessions/cli.py
@@ -2968,6 +2968,7 @@ def search(
                         "path": r.path,
                         "hit_count": r.hit_count,
                         "started_at": r.started_at.isoformat() if r.started_at else None,
+                        "summary": r.summary,
                         "snippets": [dataclasses.asdict(s) for s in r.snippets],
                     }
                     for r in results
@@ -2985,6 +2986,8 @@ def search(
     for r in results:
         harness_label = f"[{r.harness}]"
         click.echo(f"  {r.display_date}  {harness_label}  {r.session_id}  ({r.hit_count} hit(s))")
+        if r.summary and not no_snippets:
+            click.echo(f"    {r.summary}")
         if not no_snippets:
             for s in r.snippets:
                 role_label = s.role.upper()[:4]
--- a/packages/gptme-sessions/src/gptme_sessions/search.py
+++ b/packages/gptme-sessions/src/gptme_sessions/search.py
@@ -105,8 +105,13 @@ def _extract_summary(messages: list) -> str | None:
             content = (msg.content or "").strip()
             if len(content) < SUMMARY_MIN_LEN:
                 continue
-            # Take first non-empty line
-            first_line = next((ln.strip() for ln in content.splitlines() if ln.strip()), content)
+            # Take the first non-empty line that is itself substantial enough to
+            # be a meaningful summary.  A short opener like "Hi" followed by a
+            # long body would otherwise produce a useless one-word summary.
+            lines = [ln.strip() for ln in content.splitlines() if ln.strip()]
+            first_line = next((ln for ln in lines if len(ln) >= SUMMARY_MIN_LEN), None)
+            if first_line is None:
+                continue
             if len(first_line) > SUMMARY_MAX_LEN:
                 return first_line[:SUMMARY_MAX_LEN] + "…"
             return first_line
--- a/packages/gptme-sessions/src/gptme_sessions/search.py
+++ b/packages/gptme-sessions/src/gptme_sessions/search.py
@@ -26,6 +26,8 @@ logger = logging.getLogger(__name__)
 
 SNIPPET_CONTEXT = 120  # chars of surrounding text to include around each match
 MAX_SNIPPETS_PER_SESSION = 3
+SUMMARY_MAX_LEN = 120  # chars for the session summary line
+SUMMARY_MIN_LEN = 20  # min chars to treat a message as a useful summary
 
 
 @dataclass
@@ -46,6 +48,7 @@ class SearchResult:
     hit_count: int
     started_at: datetime | None
     snippets: list[Snippet] = field(default_factory=list)
+    summary: str | None = None  # first substantial user/assistant message
 
     @property
     def display_date(self) -> str:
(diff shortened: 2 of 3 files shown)


TimeToBuildBob:

Session: `83f8c9f7-df0a-56c5-8087-d930f3454af1`

packages/gptme-sessions/src/gptme_sessions/search.py
```diff
@@ -88,6 +91,28 @@ def _message_mentions_file(msg, file_pattern: re.Pattern) -> bool:
...
+    """
+    for role in ("user", "assistant"):
+        for msg in messages:
+            if msg.role != role:
+                continue
+            content = (msg.content or "").strip()
+            if len(content) < SUMMARY_MIN_LEN:
+                continue
+            # Take first non-empty line
+            first_line = next((ln.strip() for ln in content.splitlines() if ln.strip()), content)
```
TimeToBuildBob: ⚠️ **P2** — `_extract_summary` accepts a message when the total `content` length is at least `SUMMARY_MIN_LEN`, but returns the first non-empty line without verifying that line is itself substantial. A message like `"Hi\nThis is a long description of the task about fixing the auth bug..."` has content length >20, so it passes the length check, yet the returned summary becomes `"Hi"`, a near-useless summary. The tests only cover messages whose first line is long, so this edge case is missed. Users would see a terse, misleading summary for any session whose first message begins with a short greeting or fragment, which is a common pattern in real transcripts.

packages/gptme-sessions/src/gptme_sessions/cli.py
```diff
@@ -2985,6 +2986,8 @@ def search(
     for r in results:
         harness_label = f"[{r.harness}]"
         click.echo(f"  {r.display_date}  {harness_label}  {r.session_id}  ({r.hit_count} hit(s))")
+        if r.summary and not no_snippets:
+            click.echo(f"    {r.summary}")
```
TimeToBuildBob: ⚠️ **P2** — The `summary` value is printed to the terminal via `click.echo` without sanitization (line 2990). Since `summary` is taken from the first substantial user/assistant message in a transcript file, which is untrusted input (session files can contain control characters or be crafted by an attacker), a malicious transcript can embed ANSI escape sequences that the terminal interprets. This can alter the display (e.g., clear screen, set window title) or, in some terminal emulators, execute shell commands via OSC sequences, posing an injection risk. While snippets already have this exposure, this PR adds a new print of untrusted content without mitigation. The JSON output is safe because `json.dumps` escapes, but the text output is not.

TimeToBuildBob:
## 🤖 AI code review

Adds a `summary` field to `SearchResult` extracted via `_extract_summary` from the first substantial user or assistant message, and displays it in the `search` command's formatted text output (control characters stripped) and JSON output. Also adds unit tests for `_extract_summary` and CLI behavior including `--no-snippets`.

**2 findings** · ⚠️ **2** P2 · 🔒 **1** security — no P0/P1 on this head, but P2 findings still prevent a 5/5 until they are resolved.

**⚠️ P2 medium** — `packages/gptme-sessions/tests/test_search.py:167`

The test `test_returns_none_when_all_lines_too_short` claims to cover the case where total content is &gt;= SUMMARY_MIN_LEN but every individual line is short. However, the message "ok\nyes\nno\nhi\nok" has length 15, which is less than SUMMARY_MIN_LEN (20), so it is skipped by the early `if len(content) < SUMMARY_MIN_LEN: continue` check in `_extract_summary`. The test thus never exercises the line-level logic and will not catch a regression that, for example, returns the first line even when all lines are short. The test should use a message whose stripped length is &gt;= 20 but with all lines shorter than 20.

```suggestion
Change the test message to e.g. "ok\nok\nok\nok\nok\nok" which has total length 23 (>=20) but all lines are 2 characters.
```

<sub>**How this was verified:** In `_extract_summary`, the line `if len(content) < SUMMARY_MIN_LEN: continue` triggers before the line-level logic. Counting the characters in the quoted string gives 15, so the function returns None without evaluating lines. Confirmed with the code.</sub>

**⚠️ P2 medium · 🔒 security** — `packages/gptme-sessions/src/gptme_sessions/cli.py:2994`

The regex `[\x00-\x1f\x7f]` used to sanitize the summary in the CLI strips C0 control characters (0x00-0x1f) and DEL (0x7f) but omits C1 control characters in the range 0x80-0x9f. Many terminal emulators interpret C1 CSI sequences (e.g. `\x9b[31m`) as control sequences. Since the summary is printed directly to the terminal via `click.echo` and originates from untrusted session files, a crafted summary could inject terminal escape sequences that alter the display or potentially execute other actions. Because the summary is truncated to 120 characters, the injection is limited but still possible. The fix is to extend the character class to also remove C1 controls.

```suggestion
safe_summary = re.sub(r"[\x00-\x1f\x7f-\x9f]", "", r.summary)
```

<sub>**How this was verified:** Checked the regex against the full set of control characters; the range 0x80-0x9f is not covered. Confirmed that `click.echo` sends the string to the terminal without additional escaping.</sub>

</details>

- `packages/gptme-sessions/src/gptme_sessions/search.py` — Adds `_extract_summary`, `SUMMARY_MAX_LEN`/`SUMMARY_MIN_LEN` constants, and a `summary` field to `SearchResult`, populated in `_search_path`.
- `packages/gptme-sessions/src/gptme_sessions/cli.py` — Adds `summary` to the JSON dict and prints it in text output with control characters stripped.
- `packages/gptme-sessions/tests/test_search.py` — Adds unit tests for `_extract_summary` and CLI tests for summary display and `--no-snippets` behavior.

</details>

| `d03bf1d69cd3` | 4/5 | 3 | llm | 2026-08-14 01:59 UTC |
| `76e0e9327737` | 4/5 | 2 | llm | 2026-08-14 02:20 UTC |

</details>

<sub>Reviewed `069047fe70e2` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 122s · reports findings; merge decisions belong to the gate · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

packages/gptme-sessions/src/gptme_sessions/search.py
TimeToBuildBob: Fixed in 76e0e932: now requires the first line itself to meet `SUMMARY_MIN_LEN`. Short openers like "Hi" are skipped; the next substantial line is used instead. Returns `None` when all lines are too short. Added two new unit tests covering this case.

packages/gptme-sessions/src/gptme_sessions/cli.py
TimeToBuildBob: Fixed in 76e0e932: strip control characters with `re.sub(r'[\x00-\x1f\x7f]', '', r.summary)` before `click.echo`. Comment documents the threat (ANSI injection from untrusted session files).

packages/gptme-sessions/src/gptme_sessions/cli.py
```diff
@@ -2985,6 +2986,11 @@ def search(
     for r in results:
         harness_label = f"[{r.harness}]"
         click.echo(f"  {r.display_date}  {harness_label}  {r.session_id}  ({r.hit_count} hit(s))")
+        if r.summary and not no_snippets:
```
TimeToBuildBob: ⚠️ **P2** — The summary line is suppressed whenever `--no-snippets` is passed, because the guard is `if r.summary and not no_snippets:`. The `--no-snippets` flag is documented as “Show only session IDs, no text excerpts” — it is meant to hide the excerpt snippets around search matches, not the session summary. The summary is a concise one-line description of the session, independent of where the match occurred, and the PR description says it should be displayed in formatted text output without mentioning any exception. As a result, users who use `--no-snippets` to get a compact listing lose the summary entirely, which is likely unexpected. The added CLI test for `--no-snippets` uses a session whose summary is `None` (the assistant message is only 17 chars), so this behavior is untested.

TimeToBuildBob:
Fixed the remaining AI review P2 finding (third issue that wasn't addressed in the prior commit):

**Issue**: `--no-snippets` incorrectly suppressed the session summary. The guard was `if r.summary and not no_snippets:`, but `--no-snippets` is documented to hide text excerpts around search matches — not the concise session description.

**Fix** (`069047fe`):
- `cli.py`: removed `not no_snippets` from the summary guard — summary now shows regardless of `--no-snippets`
- `test_search.py`: added `test_search_no_snippets_still_shows_summary` — creates a session with a long enough user message to produce a summary, invokes `--no-snippets`, and asserts: snippet hidden ✓, summary visible ✓

All 50 tests passing.

packages/gptme-sessions/tests/test_search.py
```diff
@@ -106,6 +108,91 @@ def test_display_date_without_timestamp(self):
...
+        # next substantial line rather than returning the terse opener.
+        long_body = "Implement the session search feature with full-text indexing"
+        msgs = [self._msg("user", f"Hi\n{long_body}")]
+        result = _extract_summary(msgs)
+        assert result is not None
+        assert result == long_body
+
+    def test_returns_none_when_all_lines_too_short(self):
+        # Total content >= SUMMARY_MIN_LEN but every individual line is short.
+        msgs = [self._msg("user", "ok\nyes\nno\nhi\nok")]
```
TimeToBuildBob, line 167: ⚠️ **P2** — The test `test_returns_none_when_all_lines_too_short` claims to cover the case where total content is >= SUMMARY_MIN_LEN but every individual line is short. However, the message "ok\nyes\nno\nhi\nok" has length 15, which is less than SUMMARY_MIN_LEN (20), so it is skipped by the early `if len(content) < SUMMARY_MIN_LEN: continue` check in `_extract_summary`. The test thus never exercises the line-level logic and will not catch a regression that, for example, returns the first line even when all lines are short. The test should use a message whose stripped length is >= 20 but with all lines shorter than 20.

PR #1411, opened 2026-08-11
feat(gptme-rag): add OpenRouter embeddings backend

TimeToBuildBob:
## Summary

- add `OpenRouterEmbedding`, a ChromaDB-compatible embedding function using `https://openrouter.ai/api/v1/embeddings`
- cache API embeddings in SQLite by `(model, sha256(text))`
- expose `--embedding-function openrouter` for `index`, `search`, and `watch`
- fall back to local ModernBERT embeddings when `openrouter` is requested without an API key
- document the OpenRouter CLI path and `OPENROUTER_EMBEDDING_MODEL`

## Verification

- OpenRouter docs checked: embeddings endpoint exists and lists `openai/text-embedding-3-large` as an embedding model: https://openrouter.ai/docs/api_reference/embeddings
- Live smoke: `openai/text-embedding-3-large` returned a 3072-dim embedding, 1 request, 10 tokens
- `uv run ruff format --check packages/gptme-rag/src/gptme_rag/embeddings.py packages/gptme-rag/src/gptme_rag/indexing/indexer.py packages/gptme-rag/src/gptme_rag/cli.py packages/gptme-rag/tests/test_openrouter_embeddings.py`
- `uv run ruff check packages/gptme-rag/src/gptme_rag/embeddings.py packages/gptme-rag/src/gptme_rag/indexing/indexer.py packages/gptme-rag/src/gptme_rag/cli.py packages/gptme-rag/tests/test_openrouter_embeddings.py`
- `uv run --with mypy mypy packages/gptme-rag/src/gptme_rag --ignore-missing-imports`
- `uv run --package gptme-rag --extra test python3 -m pytest packages/gptme-rag/tests/test_openrouter_embeddings.py packages/gptme-rag/tests/test_indexing.py::test_reset_collection_preserves_embedding_metadata -q`

## Not Run

- Full `packages/gptme-rag/tests -m 'not slow'` did not complete usefully; interrupted after 1m43s while `test_benchmark.py::test_search_benchmark` was inside ModernBERT CPU inference. Two tests had passed before the interrupt.

PR #1432, opened 2026-08-16
fix(gptme-rag): surface search degradation warnings in non-json mode

TimeToBuildBob:
Follow-up to #1411, found while symptom-verifying that PR's merged fix.

## The gap

#1411 added an embedding-model-mismatch warning (`allow_recreate=False` keeps
the collection instead of wiping it) and captured it into the `--json`
envelope. That half works — verified live:

```json
{
  "error": "Cannot search: the stored embedding model 'all-MiniLM-L6-v2' could not be loaded ...",
  "query": "fox",
  "warnings": ["Embedding model mismatch (stored: 'all-MiniLM-L6-v2', requested: 'modernbert') but recreation is disabled for this command; ..."]
}
```

The **human** half did not. `search()` wraps init/search in
`contextlib.redirect_stdout(devnull)` in *both* modes, and `RichHandler` logs
to stdout — so the warning went to `/dev/null`. Without `--json` the same
command printed a bare `RuntimeError` traceback and no explanation of why.

The existing comment says non-json mode *"keeps stderr visible so callers can
diagnose partial failures (missing model, fallback used, etc.)"*. Accurate
about stderr, but the diagnostics were on stdout, so the stated intent was
never actually reached.

## The fix

Attach `_JsonWarningCaptureHandler` in both modes, and re-emit captured
warnings to the console after the redirect block when not in json mode.
Six lines; no change to json-mode behaviour or to the tqdm-suppression that
#1416 added.

## Verification

Symptom test (not just unit tests) — index with one model, search with another:

```bash
gptme-rag index docs/ --persist-dir idx --embedding-function minilm
gptme-rag search fox --persist-dir idx --embedding-function modernbert
```

- **Before**: traceback only, no mention of the mismatch.
- **After**: `⚠️  Embedding model mismatch (stored: 'all-MiniLM-L6-v2', requested: 'modernbert') ...`

`--json` output unchanged (still carries `warnings`); stdout still clean.

New regression test `test_search_non_json_surfaces_warning_to_human` — confirmed
RED before the fix, green after. Full `test_cli_search.py`: 17 passed.

### DIFF
--- a/packages/gptme-rag/src/gptme_rag/cli.py
+++ b/packages/gptme-rag/src/gptme_rag/cli.py
@@ -402,9 +402,11 @@ def search(
     # and redirect stderr too (suppresses model-loading progress bars like "Loading weights").
     status_ctx = contextlib.nullcontext() if output_json else console.status("Initializing...")
     captured_warnings: list[str] = []
-    warning_handler = _JsonWarningCaptureHandler(captured_warnings) if output_json else None
-    if warning_handler is not None:
-        logging.getLogger().addHandler(warning_handler)
+    # Capture in both modes: stdout is redirected to /dev/null during init/search
+    # regardless of --json, and RichHandler logs to stdout, so warnings would
+    # otherwise be swallowed for humans too (not just JSON consumers).
+    warning_handler = _JsonWarningCaptureHandler(captured_warnings)
+    logging.getLogger().addHandler(warning_handler)
     with status_ctx:
         # Parse custom weights if provided
         scoring_weights = None
@@ -493,8 +495,12 @@ def search(
                         print(json.dumps(error_output))
                     raise
         finally:
-            if warning_handler is not None:
-                logging.getLogger().removeHandler(warning_handler)
+            logging.getLogger().removeHandler(warning_handler)
+            if not output_json:
+                # stdout was redirected to /dev/null above, taking RichHandler's
+                # output with it — re-emit so a human sees degraded results too.
+                for message in captured_warnings:
+                    console.print(f"⚠️  {message}", style="yellow")
 
     if not documents:
         if output_json:
--- a/packages/gptme-rag/src/gptme_rag/cli.py
+++ b/packages/gptme-rag/src/gptme_rag/cli.py
@@ -498,12 +498,11 @@ def search(
                     raise
         finally:
             logging.getLogger().removeHandler(warning_handler)
-
-    if not output_json:
-        # stdout was redirected to /dev/null above, taking RichHandler's output
-        # with it. Re-emit after the status display stops so warnings remain visible.
-        for message in captured_warnings:
-            console.print(f"⚠️  {message}", style="yellow", markup=False)
+            if not output_json:
+                # stdout was redirected to /dev/null above, taking RichHandler's
+                # output with it. Re-emit before propagating any search failure.
+                for message in captured_warnings:
+                    console.print(f"⚠️  {message}", style="yellow", markup=False)
 
     if not documents:
         if output_json:
--- a/packages/gptme-rag/src/gptme_rag/cli.py
+++ b/packages/gptme-rag/src/gptme_rag/cli.py
@@ -402,30 +402,32 @@ def search(
     # and redirect stderr too (suppresses model-loading progress bars like "Loading weights").
     status_ctx = contextlib.nullcontext() if output_json else console.status("Initializing...")
     captured_warnings: list[str] = []
+
+    # Parse custom weights before installing the warning handler so early returns
+    # cannot leave the handler attached to the root logger.
+    scoring_weights = None
+    if weights:
+        try:
+            scoring_weights = json.loads(weights)
+        except json.JSONDecodeError as e:
+            if output_json:
+                print(json.dumps({"error": f"Invalid weights JSON: {e}", "query": query}))
+            else:
+                console.print(f"❌ Invalid weights JSON: {e}", style="red")
+            return
+        except Exception as e:
+            if output_json:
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
CI is green and the PR is mergeable. The self-merge gate is still blocked because Greptile has not reviewed this head yet; waiting for the automatic initial review before re-running the gate.

TimeToBuildBob:
## 🤖 AI code review

Moves --weights parsing before the warning-capture handler is installed, so early returns on invalid weights cannot leave the handler attached to the root logger. The handler is now installed in both JSON and non-JSON modes, and in non-JSON mode captured warnings are re-printed to the console after the stdout-redirected block ends. Adds two tests: one verifying the handler is not leaked on invalid weights, and one verifying non-JSON warnings are surfaced to the user.

**⚠️ P2 medium** — `packages/gptme-rag/src/gptme_rag/cli.py:502`

In non-JSON mode, the warning re-emit block (lines 502-506) is only reached when the `with status_ctx:` block completes without exception. If a warning is captured and then a later failure occurs (e.g., `indexer.search` raises due to embedding dimension mismatch), the exception propagates out of the `with` block and the re-emit loop is skipped. The user then sees a bare traceback with no explanation of the degradation warning that preceded the error, defeating the PR's goal of surfacing such warnings to humans. The error path at lines 488-498 only prints a JSON error object for `output_json=True`; for non-JSON it just re-raises, leaving captured_warnings unused. The fix is to move the re-emit into a `finally` block covering the entire operation, or to print warnings in the `except` path for non-JSON as well.

```suggestion
Wrap the body from handler installation to the end in a try/finally, and in the finally block re-emit warnings if not output_json.
```

<sub>**How this was verified:** Traced the exception path: the `except Exception` at line 488 re-raises after optionally printing JSON, and control never reaches line 502 because the exception propagates out of the function before the `with status_ctx:` block exits normally. The test for non-JSON warnings only covers the success path.</sub>

</details>

- `packages/gptme-rag/src/gptme_rag/cli.py` — Moves weights parsing before handler installation; installs the capture handler in both modes; re-emits captured warnings in non-JSON mode after the redirect block.
- `packages/gptme-rag/tests/test_cli_search.py` — Adds tests for handler non-leakage on invalid weights and for surfacing warnings in non-JSON output.

</details>

| `50c1f26f59b6` | 5/5 | 0 | llm | 2026-08-16 15:26 UTC |
| `50c1f26f59b6` | 2/5 | 2 | llm | 2026-08-16 16:01 UTC |

</details>

<sub>Reviewed `b9f425479eb1` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 82s · reports findings; merge decisions belong to the gate · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

TimeToBuildBob:
@greptileai review

packages/gptme-rag/src/gptme_rag/cli.py
```diff
@@ -493,8 +495,12 @@ def search(
...
                     raise
         finally:
-            if warning_handler is not None:
-                logging.getLogger().removeHandler(warning_handler)
+            logging.getLogger().removeHandler(warning_handler)
+            if not output_json:
+                # stdout was redirected to /dev/null above, taking RichHandler's
+                # output with it — re-emit so a human sees degraded results too.
+                for message in captured_warnings:
+                    console.print(f"⚠️  {message}", style="yellow")
```
TimeToBuildBob: ❌ **P1** — The re-emit at line 503 passes the raw warning text to `console.print(f"⚠️  {message}", style="yellow")` without `markup=False`. Console.print parses `[...]` as Rich markup by default, so a warning whose text contains a bracketed token — e.g. a model-loading warning naming `[bert-base]` or any unmatched `[` — raises `MarkupError` for an unknown style. This parse runs on exactly the strings the PR is designed to surface (ChromaDB/model-loading warnings). Because the finally block executes before the summary output at line 505, the MarkupError propagates out of `search()`, and a search that already computed valid results exits with an exception and prints no results — the new path turns a degraded-but-successful search into a crash. The added test uses a warning with no brackets, so it does not exercise this path. _(2/3 passes)_

TimeToBuildBob:

Session: `3d4c905d-207c-528d-8a7a-34e0df0ac8fa`

packages/gptme-rag/src/gptme_rag/cli.py
```diff
@@ -493,8 +497,13 @@ def search(
                         print(json.dumps(error_output))
                     raise
         finally:
-            if warning_handler is not None:
-                logging.getLogger().removeHandler(warning_handler)
+            logging.getLogger().removeHandler(warning_handler)
+
+    if not output_json:
```
TimeToBuildBob: ⚠️ **P2** — In non-JSON mode, the warning re-emit block (lines 502-506) is only reached when the `with status_ctx:` block completes without exception. If a warning is captured and then a later failure occurs (e.g., `indexer.search` raises due to embedding dimension mismatch), the exception propagates out of the `with` block and the re-emit loop is skipped. The user then sees a bare traceback with no explanation of the degradation warning that preceded the error, defeating the PR's goal of surfacing such warnings to humans. The error path at lines 488-498 only prints a JSON error object for `output_json=True`; for non-JSON it just re-raises, leaving captured_warnings unused. The fix is to move the re-emit into a `finally` block covering the entire operation, or to print warnings in the `except` path for non-JSON as well.

TimeToBuildBob: Fixed in d35fe2d6: non-JSON warnings are now re-emitted from the cleanup `finally`, so they remain visible before a later search exception propagates. Added a regression test that captures a warning, raises from `Indexer.search`, and asserts both the warning and original exception survive.

PR #1433, opened 2026-08-16
fix(pm): sync worker timeout table with shell script

TimeToBuildBob:
## What

Routes each monitored item type through the executor that actually owns it, then gives
that executor an explicit timeout budget:

- **`greptile_needs_fix`** is dispatcher-derived slot work, so
  `run_item.timeout_tier()` now gives it the existing 2700s Greptile fix/review budget
  even when it is not bundled with `pr_update`.
- **`greptile_convergence_adjudication`** was already explicitly routed by
  `run_item.timeout_tier()` to the 1500s adjudication budget; this PR preserves that
  route rather than adding dead policy to the legacy discovery loop.
- **`linear_notification_retry`** is emitted by the legacy
  `ProjectMonitoringRun.discover_work()` path, so `_ITEM_TIMEOUTS` now gives it the
  600s simple-work budget.

The comments and focused tests now describe this split between the slot executor and
the legacy discovery loop.

## Why

The Aug 15 backend outage produced an 18-of-21 timeout storm. Most failures were
`greptile_needs_fix` items entering the slot executor with the 900s default despite
the Greptile fix/review workflow having a 2700s budget. The old proposal put Greptile
types into `_ITEM_TIMEOUTS`, but that table is only used by the legacy discovery loop
and cannot affect dispatcher-derived Greptile work.

Raising the default would over-budget every unknown future type. Explicit policy in
the owning executor fixes the measured failure without widening unrelated work.

### DIFF
--- a/packages/gptme-runloops/src/gptme_runloops/project_monitoring.py
+++ b/packages/gptme-runloops/src/gptme_runloops/project_monitoring.py
@@ -99,6 +99,8 @@ class ProjectMonitoringRun(BaseRunLoop):
         "ci_failure": 1200,  # ~20 min: investigate + fix
         "notification": 600,  # ~10 min: simple processing
         "linear_notification_retry": 600,  # ~10 min: retry a failed Linear notification delivery
+        # greptile_needs_fix and greptile_convergence_adjudication are intentionally
+        # absent: they are dispatcher-derived and handled by run_item.timeout_tier().
     }
     _DEFAULT_ITEM_TIMEOUT = 900  # ~15 min for unknown types
     _MAX_TIMEOUT = 3600  # 60 min cap regardless of item count
--- a/packages/gptme-runloops/src/gptme_runloops/project_monitoring.py
+++ b/packages/gptme-runloops/src/gptme_runloops/project_monitoring.py
@@ -90,16 +90,13 @@ class ProjectMonitoringRun(BaseRunLoop):
         # Cache discovered work between has_work() and generate_prompt()
         self._discovered_work: list[WorkItem] = []
 
-    # Per-item timeout in seconds, mirroring project-monitoring.sh tiers.
-    # Assigned issues need the most time (research + PR + response).
-    # PR updates, CI failures, and greptile review cycles need a mid-tier (fix + re-review cycle).
-    # Simple notifications need the least.
+    # Per-item timeout in seconds for work produced by this legacy discovery loop.
+    # The slot-based executor uses run_item.timeout_tier() for dispatcher-derived
+    # work such as Greptile fixes and convergence adjudication.
     _ITEM_TIMEOUTS: dict[str, int] = {
         "assigned_issue": 1500,  # ~25 min: deep research + PR work + response
-        "greptile_convergence_adjudication": 1500,  # ~25 min: classify findings, fix one blocking issue, recommend
         "pr_update": 1200,  # ~20 min: fix + re-review cycle
         "ci_failure": 1200,  # ~20 min: investigate + fix
-        "greptile_needs_fix": 1200,  # ~20 min: fix + re-review cycle
         "notification": 600,  # ~10 min: simple processing
         "linear_notification_retry": 600,  # ~10 min: retry a failed Linear notification delivery
     }
@@ -109,10 +106,10 @@ class ProjectMonitoringRun(BaseRunLoop):
     def _compute_timeout(self, items: list["WorkItem"]) -> int:
         """Compute session timeout by summing per-item budgets, capped at _MAX_TIMEOUT.
 
-        Mirrors the complexity-based timeout tiers in project-monitoring.sh:
-          - assigned_issue / greptile_convergence_adjudication → 1500s (deep work)
-          - pr_update / ci_failure / greptile_needs_fix → 1200s (fix + re-review)
-          - notification / linear_notification_retry → 600s (simple)
+        Budgets the item types emitted by :meth:`discover_work`:
+          - assigned_issue → 1500s (deep work)
+          - pr_update / ci_failure → 1200s (fix work)
+          - notification / linear_notification_retry → 600s (simple work)
 
         Args:
             items: Discovered work items.
--- a/packages/gptme-runloops/src/gptme_runloops/project_monitoring.py
+++ b/packages/gptme-runloops/src/gptme_runloops/project_monitoring.py
@@ -96,10 +96,12 @@ class ProjectMonitoringRun(BaseRunLoop):
     # Simple notifications need the least.
     _ITEM_TIMEOUTS: dict[str, int] = {
         "assigned_issue": 1500,  # ~25 min: deep research + PR work + response
+        "greptile_convergence_adjudication": 1500,  # ~25 min: classify findings, fix one blocking issue, recommend
         "pr_update": 1200,  # ~20 min: fix + re-review cycle
         "ci_failure": 1200,  # ~20 min: investigate + fix
-        "greptile_needs_fix": 1200,  # ~20 min: fix + re-review cycle (added 2026-08-16)
+        "greptile_needs_fix": 1200,  # ~20 min: fix + re-review cycle
         "notification": 600,  # ~10 min: simple processing
+        "linear_notification_retry": 600,  # ~10 min: retry a failed Linear notification delivery
     }
     _DEFAULT_ITEM_TIMEOUT = 900  # ~15 min for unknown types
     _MAX_TIMEOUT = 3600  # 60 min cap regardless of item count
@@ -108,9 +110,9 @@ class ProjectMonitoringRun(BaseRunLoop):
         """Compute session timeout by summing per-item budgets, capped at _MAX_TIMEOUT.
 
         Mirrors the complexity-based timeout tiers in project-monitoring.sh:
-          - assigned_issue → 1500s (research + PR work)
-          - pr_update / ci_failure → 1200s (fix + re-review)
-          - notification → 600s (simple)
+          - assigned_issue / greptile_convergence_adjudication → 1500s (deep work)
+          - pr_update / ci_failure / greptile_needs_fix → 1200s (fix + re-review)
+          - notification / linear_notification_retry → 600s (simple)
 
         Args:
             items: Discovered work items.
--- a/packages/gptme-runloops/src/gptme_runloops/project_monitoring.py
+++ b/packages/gptme-runloops/src/gptme_runloops/project_monitoring.py
@@ -92,12 +92,13 @@ class ProjectMonitoringRun(BaseRunLoop):
 
     # Per-item timeout in seconds, mirroring project-monitoring.sh tiers.
     # Assigned issues need the most time (research + PR + response).
-    # PR updates and CI failures need a mid-tier (fix + re-review cycle).
+    # PR updates, CI failures, and greptile review cycles need a mid-tier (fix + re-review cycle).
     # Simple notifications need the least.
     _ITEM_TIMEOUTS: dict[str, int] = {
         "assigned_issue": 1500,  # ~25 min: deep research + PR work + response
         "pr_update": 1200,  # ~20 min: fix + re-review cycle
         "ci_failure": 1200,  # ~20 min: investigate + fix
+        "greptile_needs_fix": 1200,  # ~20 min: fix + re-review cycle (added 2026-08-16)
         "notification": 600,  # ~10 min: simple processing
     }
     _DEFAULT_ITEM_TIMEOUT = 900  # ~15 min for unknown types
--- a/packages/gptme-runloops/src/gptme_runloops/run_item.py
+++ b/packages/gptme-runloops/src/gptme_runloops/run_item.py
@@ -520,7 +520,7 @@ def timeout_tier(
         or instruction_kind == "GREPTILE_CONVERGENCE"
     ):
         return config.adjudication_timeout, config.adjudication_time_desc
-    if "pr_update" in types and has_greptile_fix:
+    if "greptile_needs_fix" in types or ("pr_update" in types and has_greptile_fix):
         return config.greptile_fix_timeout, config.greptile_fix_time_desc
     return config.default_timeout, config.default_time_desc
 
--- a/packages/gptme-runloops/tests/test_project_monitoring.py
+++ b/packages/gptme-runloops/tests/test_project_monitoring.py
@@ -67,20 +67,30 @@ def test_compute_timeout_single_pr_update(workspace):
     assert run._compute_timeout(items) == 1200
 
 
-def test_compute_timeout_greptile_needs_fix(workspace):
-    """Greptile fix-and-review cycles use the medium-tier budget (same as pr_update)."""
+@pytest.mark.parametrize(
+    ("item_type", "expected_timeout"),
+    [
+        ("greptile_convergence_adjudication", 1500),
+        ("greptile_needs_fix", 1200),
+        ("linear_notification_retry", 600),
+    ],
+)
+def test_compute_timeout_synced_item_types(
+    workspace, item_type: str, expected_timeout: int
+):
+    """Item types synced from the shell dispatcher retain their timeout tier."""
     run = ProjectMonitoringRun(workspace)
     items = [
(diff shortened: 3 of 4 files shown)


TimeToBuildBob:
CI is green and the PR is mergeable. The self-merge gate is still blocked because the initial Greptile review has not landed, so I am not merging around the gate; waiting for that review.

packages/gptme-runloops/tests/test_project_monitoring.py
```diff
@@ -67,6 +67,22 @@ def test_compute_timeout_single_pr_update(workspace):
     assert run._compute_timeout(items) == 1200
 
 
+def test_compute_timeout_greptile_needs_fix(workspace):
```
TimeToBuildBob: ⚠️ **P2** — The PR introduces three new timeout entries (`greptile_convergence_adjudication`=1500, `greptile_needs_fix`=1200, `linear_notification_retry`=600) but only adds a test for `greptile_needs_fix`. The other two entries have no coverage, so a future change that accidentally modifies them (e.g., a typo dropping a zero, or a misinterpretation of the shell script tier) would silently regress the timeout without any test failing. Given the PR's stated goal is to keep the Python table in sync with `project-monitoring.sh`, the absence of tests for two of the three synced values leaves the exact drift the PR is meant to fix undetected. The existing test for `greptile_needs_fix` shows the pattern; the same should be applied to the other two.

TimeToBuildBob:
## 🤖 AI code review

Adds a 600s timeout entry for the linear_notification_retry work item type in the legacy ProjectMonitoringRun._ITEM_TIMEOUTS table, and changes run_item.timeout_tier to treat the greptile_needs_fix type as always eligible for the 2700s greptile_fix_timeout, independent of the has_greptile_fix flag. Comments and tests are updated to document the split between legacy discovery-loop budgeting and slot-executor timeout routing.

**2 findings** · ⚠️ **2** P2 — no P0/P1 on this head, but P2 findings still prevent a 5/5 until they are resolved.

**⚠️ P2 medium** — `packages/gptme-runloops/tests/test_run_item.py:216`

The new condition in timeout_tier treats 'greptile_needs_fix' as sufficient for the 2700s fix tier, independent of has_greptile_fix. The added parametrized case only tests has_fix=True, so a future regression that makes the tier conditional on has_greptile_fix (e.g., changing the condition to `if "greptile_needs_fix" in types and has_greptile_fix`) would still pass this test, because has_fix is True. Without a has_fix=False test, the dispatcher-derived Greptile-fix items that this PR fixes (which were timing out at 900s) could silently regress to the default tier without any test failing.

```suggestion
Add (["greptile_needs_fix"], False, (2700, "~35 minutes")), to the parametrize list.
```

<sub>**How this was verified:** The condition at run_item.py:523 is `if "greptile_needs_fix" in types or ...`; with has_fix=False the first clause still triggers, so the result is 2700. The test list only includes the True case, so it cannot distinguish the new behavior.</sub>

**⚠️ P2 medium** — `packages/gptme-runloops/src/gptme_runloops/run_item.py:523`

In timeout_tier, the condition `if "greptile_needs_fix" in types or ("pr_update" in types and has_greptile_fix):` grants the 2700s greptile_fix budget to an item carrying "greptile_needs_fix" even when has_greptile_fix is False—i.e., when the lifecycle produced no fix instructions (greptile_fix string empty). This can happen if the item is dispatched with the type but the lifecycle determines no fix is warranted (e.g., self-merge or already-fixed). Such a session then gets 45 minutes instead of the 900s default, wasting compute on a slot that may need only a simple update. If the type guarantees fix work, this is acceptable, but the PR description only justifies the type taking precedence over pr_update, not over the absence of instructions. If a scenario exists where greptile_needs_fix items can have empty instructions, this mis-budgets them. Add a guard `and has_greptile_fix` to the first clause, or document why it's not needed.

```suggestion
if "greptile_needs_fix" in types and has_greptile_fix or ("pr_update" in types and has_greptile_fix):
```

<sub>**How this was verified:** In plan_item, has_greptile_fix is bool(greptile_fix) which is empty when lifecycle.instructions is None. If a greptile_needs_fix item had no instructions, the new condition still returns the 2700s tier. This is a behavior change from the previous code where such an item would have fallen to the 900s default.</sub>

</details>

- `packages/gptme-runloops/src/gptme_runloops/project_monitoring.py` — Adds linear_notification_retry to _ITEM_TIMEOUTS and rewrites comments to distinguish legacy discovery-loop work from dispatcher-derived work.
- `packages/gptme-runloops/src/gptme_runloops/run_item.py` — Extends the timeout_tier condition to route greptile_needs_fix items to the greptile_fix_timeout even without a pr_update type.
- `packages/gptme-runloops/tests/test_project_monitoring.py` — Adds a test asserting linear_notification_retry items get the 600s simple-work timeout.
- `packages/gptme-runloops/tests/test_run_item.py` — Adds a parametrized case expecting greptile_needs_fix items to use the 2700s fix tier.

</details>

| `b15cf67b0e0d` | 4/5 | 1 | llm | 2026-08-16 16:34 UTC |
| `6cc154f69257` | 3/5 | 2 | llm | 2026-08-16 18:27 UTC |
| `dcf15888f5f9` | 4/5 | 1 | llm | 2026-08-16 19:52 UTC |

</details>

<sub>Reviewed `3460f0abcb56` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 130s · reports findings; merge decisions belong to the gate · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

TimeToBuildBob:

Session: `48510b63-c013-59c5-b7fe-ab6621c73b01`

TimeToBuildBob:
@greptileai review

packages/gptme-runloops/tests/test_project_monitoring.py
TimeToBuildBob: Addressed in 6cc154f6: the timeout test is now parameterized across all three synced item types, asserting 1500s, 1200s, and 600s respectively. The focused project-monitoring test file passes (36 tests).

packages/gptme-runloops/src/gptme_runloops/project_monitoring.py
```diff
@@ -92,13 +92,16 @@ def __init__(
 
     # Per-item timeout in seconds, mirroring project-monitoring.sh tiers.
     # Assigned issues need the most time (research + PR + response).
-    # PR updates and CI failures need a mid-tier (fix + re-review cycle).
+    # PR updates, CI failures, and greptile review cycles need a mid-tier (fix + re-review cycle).
     # Simple notifications need the least.
     _ITEM_TIMEOUTS: dict[str, int] = {
         "assigned_issue": 1500,  # ~25 min: deep research + PR work + response
+        "greptile_convergence_adjudication": 1500,  # ~25 min: classify findings, fix one blocking issue, recommend
```
TimeToBuildBob: ❌ **P1** — The new entries for greptile_convergence_adjudication (1500s) and greptile_needs_fix (1200s) in _ITEM_TIMEOUTS have no effect on the computed session timeout. The only path that feeds items into _compute_timeout is has_work(), which calls discover_work() and passes its result. discover_work() aggregates WorkItems solely from check_notifications (type "notification"), check_linear_notifications ("linear_notification_retry"), check_pr_updates ("pr_update"), check_ci_failures ("ci_failure"), and check_assigned_issues ("assigned_issue"); none of these produce either greptile type. Therefore no item with those types ever enters the list, and when the shell dispatcher later sends such an item, it will still fall back to _DEFAULT_ITEM_TIMEOUT (900s) rather than the intended 1500/1200s. The PR's stated goal of fixing the timeout storm for these types is not achieved because the table entries are dead code in the current discovery flow.

packages/gptme-runloops/tests/test_project_monitoring.py
```diff
@@ -67,6 +67,32 @@ def test_compute_timeout_single_pr_update(workspace):
...
+        ("greptile_needs_fix", 1200),
+        ("linear_notification_retry", 600),
+    ],
+)
+def test_compute_timeout_synced_item_types(
+    workspace, item_type: str, expected_timeout: int
+):
+    """Item types synced from the shell dispatcher retain their timeout tier."""
+    run = ProjectMonitoringRun(workspace)
+    items = [
```
TimeToBuildBob, line 73: ⚠️ **P2** — The added parametrized test test_compute_timeout_synced_item_types only constructs a WorkItem directly with the new item types and asserts _compute_timeout returns the expected value. It never exercises discover_work() or any check_* method, so it cannot detect that these types are never generated in real execution. The test passes even though the table entries are dead code in the current discovery flow, giving false confidence that the timeout storm is fixed. A meaningful test would either add discovery of these item types and assert it, or at least mock discover_work to return such items and verify the timeout is used.

packages/gptme-runloops/src/gptme_runloops/project_monitoring.py
TimeToBuildBob: Addressed in dcf15888: the dispatcher-derived Greptile types now use the live slot executor timeout policy. `greptile_needs_fix` receives the 2700s fix/review budget in `run_item.timeout_tier`; convergence adjudication was already explicitly routed to 1500s there. I removed both unreachable entries from the legacy discovery-loop table instead of preserving dead policy.

packages/gptme-runloops/tests/test_project_monitoring.py
TimeToBuildBob: Addressed in dcf15888: the dead direct-construction coverage is gone. The remaining legacy-loop test covers only `linear_notification_retry`, which `discover_work()` can actually emit, and the slot executor test now exercises `greptile_needs_fix` through the production `timeout_tier()` path.

packages/gptme-runloops/src/gptme_runloops/project_monitoring.py
```diff
@@ -90,26 +90,26 @@ def __init__(
...
-    # Simple notifications need the least.
+    # Per-item timeout in seconds for work produced by this legacy discovery loop.
+    # The slot-based executor uses run_item.timeout_tier() for dispatcher-derived
+    # work such as Greptile fixes and convergence adjudication.
     _ITEM_TIMEOUTS: dict[str, int] = {
         "assigned_issue": 1500,  # ~25 min: deep research + PR work + response
         "pr_update": 1200,  # ~20 min: fix + re-review cycle
         "ci_failure": 1200,  # ~20 min: investigate + fix
         "notification": 600,  # ~10 min: simple processing
+        "linear_notification_retry": 600,  # ~10 min: retry a failed Linear notification delivery
```
TimeToBuildBob, line 101: ⚠️ **P2** — The PR description states that _ITEM_TIMEOUTS should add entries for greptile_convergence_adjudication (1500s) and greptile_needs_fix (1200s), but the diff only adds linear_notification_retry. The legacy discovery loop (discover_work) emits items only from check_notifications, check_linear_notifications, check_pr_updates, check_ci_failures, and check_assigned_issues; none of these produce greptile types. So the missing entries have no current effect. However, the PR's stated fix is not implemented, and if a future change begins emitting these types, they would fall back to the 900s default, under-budgeting work that the shell script allocates 1500/1200s. The PR also claims to 'carry the greptile_needs_fix entry added in the prior commit', but no such entry exists in the current dict. This is a latent inconsistency with the PR's intent and leaves the timeout table out of sync with the shell script it claims to mirror. _(2/2 passes)_

TimeToBuildBob: Dismissed as stale/false positive after correcting the PR description. `_ITEM_TIMEOUTS` is explicitly scoped to the legacy `discover_work()` loop, which cannot emit either Greptile type. The live slot executor already routes `greptile_convergence_adjudication` to 1500s and this head adds direct `greptile_needs_fix` routing to its 2700s fix/review tier. Adding the suggested legacy-table entry would preserve dead, misleading policy rather than fix runtime behavior.

TimeToBuildBob:
## Convergence adjudication

- **Fixed:** The earlier P1/P2 findings are addressed on `dcf15888`: dispatcher-derived `greptile_needs_fix` now receives the live slot executor's 2700s fix/review budget; the legacy-loop test only covers its actually emitted `linear_notification_retry` type.
- **Remaining:** No blocking findings. The final P2 was dismissed as stale/false positive: it relied on the old PR description and proposed adding Greptile types to `_ITEM_TIMEOUTS`, but that table is only consumed by `discover_work()`, which cannot emit those types. The PR description now matches the implemented executor split, and the thread is resolved.
- **CI:** Green across pre-commit, typecheck, package tests on Python 3.10–3.12, integration tests, coverage, install-size, and Codecov patch.
- **Domain risk:** Timeout routing only; a maintainer may manually sanity-check the 2700s policy choice against fleet capacity. No data or API migration.
- **Convergence:** Detector reports `round_convergence.status=no_history`, `stable_rounds=0` (required 2), with zero unresolved P1/P2/P3 findings. That history gap is a detector limitation here, not a reason to trigger another review.

Recommendation: merge-ready for maintainer judgment. I am not triggering another review or auto-merging.

TimeToBuildBob:

Session: `f76cc8e1-f0df-5b92-a11b-8a72c3d9d482`

TimeToBuildBob:
Pushed `3460f0ab`: added an explicit comment inside `_ITEM_TIMEOUTS` at the line the P2 flagged, documenting that `greptile_needs_fix` and `greptile_convergence_adjudication` are intentionally absent (dispatcher-derived; handled by `run_item.timeout_tier()`).

The existing class-level comment at lines 93–95 already said this, but placing it directly inside the dict at the exact location the reviewer flagged should make the intent clearer. This is the final code change — if the P2 persists after this, the PR needs a human merge click.

TimeToBuildBob:
@TimeToBuildBob review

packages/gptme-runloops/tests/test_run_item.py
```diff
@@ -213,6 +213,7 @@ def test_predict_cc_trajectory_path() -> None:
         (["assigned_issue"], False, (1500, "~20 minutes")),
         # assigned_issue wins over the greptile-fix tier (bash if/elif order)
         (["assigned_issue", "pr_update"], True, (1500, "~20 minutes")),
+        (["greptile_needs_fix"], True, (2700, "~35 minutes")),
```
TimeToBuildBob, line 216: ⚠️ **P2** — The new condition in timeout_tier treats 'greptile_needs_fix' as sufficient for the 2700s fix tier, independent of has_greptile_fix. The added parametrized case only tests has_fix=True, so a future regression that makes the tier conditional on has_greptile_fix (e.g., changing the condition to `if "greptile_needs_fix" in types and has_greptile_fix`) would still pass this test, because has_fix is True. Without a has_fix=False test, the dispatcher-derived Greptile-fix items that this PR fixes (which were timing out at 900s) could silently regress to the default tier without any test failing. _(3/3 passes)_

packages/gptme-runloops/src/gptme_runloops/run_item.py
```diff
@@ -520,7 +520,7 @@ def timeout_tier(
         or instruction_kind == "GREPTILE_CONVERGENCE"
     ):
         return config.adjudication_timeout, config.adjudication_time_desc
-    if "pr_update" in types and has_greptile_fix:
+    if "greptile_needs_fix" in types or ("pr_update" in types and has_greptile_fix):
```
TimeToBuildBob, line 523: ⚠️ **P2** — In timeout_tier, the condition `if "greptile_needs_fix" in types or ("pr_update" in types and has_greptile_fix):` grants the 2700s greptile_fix budget to an item carrying "greptile_needs_fix" even when has_greptile_fix is False—i.e., when the lifecycle produced no fix instructions (greptile_fix string empty). This can happen if the item is dispatched with the type but the lifecycle determines no fix is warranted (e.g., self-merge or already-fixed). Such a session then gets 45 minutes instead of the 900s default, wasting compute on a slot that may need only a simple update. If the type guarantees fix work, this is acceptable, but the PR description only justifies the type taking precedence over pr_update, not over the absence of instructions. If a scenario exists where greptile_needs_fix items can have empty instructions, this mis-budgets them. Add a guard `and has_greptile_fix` to the first clause, or document why it's not needed. _(2/3 passes)_

PR #1434, opened 2026-08-16
feat(skills): add agentic-presentation skill

TimeToBuildBob:
Adds an `agentic-presentation` skill: turn structured slide JSON into a standalone, offline HTML deck (inline CSS + JS, keyboard navigation, no CDN dependencies).

### Self-contained layout

This supersedes #1428, which was closed for scattering files across the repo (renderer into `scripts/`, schema into `knowledge/`, tests into top-level `tests/`). That sprawl is fixed — the entire diff is six files, all under `skills/agentic-presentation/`:

```text
skills/agentic-presentation/
├── SKILL.md
├── generate_presentation_html.py          # renderer (stdlib only)
├── present.py                             # render + optionally open in browser
├── presentation.schema.json               # draft-07 deck schema
├── examples/agentic-presentation-demo.json
└── tests/test_generate_presentation_html.py
```

Same shape as `skills/financial-advisor/` (#1417). Delete the folder and nothing outside it breaks, and nothing is left behind — no shared config, no repo-level test registration, no entries in other files.

### What it does

- Validates deck JSON (slide ids unique, required fields per slide type, known animation types), then emits one HTML file.
- MVP slide types: `title-slide`, `content`, `code-reveal`, `code-and-text`, `image`, `gallery`.
- MVP animations: `fade-in`, `slide-in`, `line-reveal` (`zoom`/`spin` reserved).
- All user text is HTML-escaped by the generator; `--max-kb` (default 500) enforces a size budget.

```bash
python3 skills/agentic-presentation/present.py deck.json --live
python3 skills/agentic-presentation/generate_presentation_html.py \
  skills/agentic-presentation/examples/agentic-presentation-demo.json -o /tmp/demo.html
```

### Verification

- 7 tests pass: `uv run --with pytest --with jsonschema pytest skills/agentic-presentation/tests/ -q`
- Demo deck renders end-to-end: 5 slides, 11,992 bytes, no external URLs in the output.
- `prek run` over the changed files: ruff, ruff-format, mypy, markdown links, jscpd, check-names all pass.

Rendering needs only the Python standard library. The two schema tests need `jsonschema`, guarded with `pytest.importorskip` so the suite degrades to 5 passed / 2 skipped rather than erroring when it is absent — this keeps the dependency inside the skill instead of adding it to the repo's mypy/dev dependency lists.

### DIFF
--- a/skills/agentic-presentation/generate_presentation_html.py
+++ b/skills/agentic-presentation/generate_presentation_html.py
@@ -85,13 +85,16 @@ def validate_presentation(deck: Any) -> None:
             raise ValueError(f"unsupported slide type for {slide_id}: {slide_type}")
         if not isinstance(slide.get("title"), str) or not slide["title"].strip():
             raise ValueError(f"slide {slide_id} must include title")
-        if slide_type == "code-reveal" and "code" not in slide:
-            raise ValueError(f"slide {slide_id} of type code-reveal requires code")
+        if slide_type == "code-reveal" and not isinstance(slide.get("code"), str):
+            raise ValueError(
+                f"slide {slide_id} of type code-reveal requires code as a string"
+            )
         if slide_type == "code-and-text" and (
-            "code" not in slide or "body" not in slide
+            not isinstance(slide.get("code"), str)
+            or not isinstance(slide.get("body"), str)
         ):
             raise ValueError(
-                f"slide {slide_id} of type code-and-text requires body and code"
+                f"slide {slide_id} of type code-and-text requires body and code as strings"
             )
         if slide_type in ("content", "code-and-text"):
             bullets = slide.get("bullets")
@@ -117,6 +120,13 @@ def validate_presentation(deck: Any) -> None:
                 raise ValueError(
                     f"slide {slide_id} has unsupported animation: {animation_type}"
                 )
+            for field in ("duration", "delay", "stagger"):
+                if field in animation and (
+                    not isinstance(animation[field], int) or animation[field] < 0
+                ):
+                    raise ValueError(
+                        f"slide {slide_id} animation {field} must be a non-negative integer"
+                    )
 
 
 def render_html(deck: dict[str, Any]) -> str:
--- a/skills/agentic-presentation/generate_presentation_html.py
+++ b/skills/agentic-presentation/generate_presentation_html.py
@@ -72,6 +72,10 @@ def validate_presentation(deck: Any) -> None:
         slide_id = slide.get("id")
         if not isinstance(slide_id, str) or not slide_id:
             raise ValueError(f"slide {index} must include id")
+        if not re.fullmatch(r"[A-Za-z0-9][A-Za-z0-9_-]*", slide_id):
+            raise ValueError(
+                f"slide {index} id {slide_id!r} must match ^[A-Za-z0-9][A-Za-z0-9_-]*$"
+            )
         if slide_id in seen_ids:
             raise ValueError(f"duplicate slide id: {slide_id}")
         seen_ids.add(slide_id)
@@ -235,7 +239,8 @@ def _render_gallery(images: list[dict[str, Any]]) -> str:
 
 
 def _animation_payload(slide: dict[str, Any]) -> str:
-    animations = slide.get("animations") or [{"type": "fade-in", "duration": 350}]
+    anims = slide.get("animations")
+    animations = anims if anims is not None else [{"type": "fade-in", "duration": 350}]
     return _attr(json.dumps(animations, separators=(",", ":")))
 
 
@@ -452,9 +457,9 @@ def _javascript() -> str:
       animations.forEach((animation) => {
         const target = selectedTarget(slide, animation.target);
         target.classList.add(`animate-${animation.type}`);
-        target.style.setProperty('--duration', `${animation.duration || 450}ms`);
-        target.style.setProperty('--delay', `${animation.delay || 0}ms`);
-        target.style.setProperty('--stagger', `${animation.stagger || 80}ms`);
+        target.style.setProperty('--duration', `${animation.duration ?? 450}ms`);
+        target.style.setProperty('--delay', `${animation.delay ?? 0}ms`);
+        target.style.setProperty('--stagger', `${animation.stagger ?? 80}ms`);
       });
       requestAnimationFrame(() => {
         animations.forEach((animation) => {
@@ -475,8 +480,14 @@ def _javascript() -> str:
     document.querySelector('#prev').addEventListener('click', () => showSlide(current - 1));
     document.querySelector('#next').addEventListener('click', () => showSlide(current + 1));
     document.addEventListener('keydown', (event) => {
-      if (event.key === 'ArrowLeft') showSlide(current - 1);
-      if (event.key === 'ArrowRight' || event.key === ' ' || event.key === 'Enter') showSlide(current + 1);
+      if (event.key === 'ArrowLeft') {
+        event.preventDefault();
+        showSlide(current - 1);
+      }
+      if (event.key === 'ArrowRight' || event.key === ' ' || event.key === 'Enter') {
+        event.preventDefault();
+        showSlide(current + 1);
+      }
     });
     showSlide(0);"""
 
--- a/skills/agentic-presentation/generate_presentation_html.py
+++ b/skills/agentic-presentation/generate_presentation_html.py
@@ -1,0 +1,513 @@
+#!/usr/bin/env python3
+"""Generate a self-contained HTML slide deck from presentation JSON."""
+
+from __future__ import annotations
+
+import argparse
+import html
+import json
+from pathlib import Path
+from typing import Any, cast
+
+SUPPORTED_SLIDE_TYPES = {
+    "title-slide",
+    "content",
+    "code-reveal",
+    "code-and-text",
+    "image",
+    "gallery",
+}
+SUPPORTED_ANIMATION_TYPES = {
+    "fade-in",
+    "slide-in",
+    "line-reveal",
+    "line-by-line-reveal",
+    "zoom",
+    "spin",
+}
+DEFAULT_THEME = {
+    "background": "#111418",
+    "surface": "#F7F4EF",
+    "foreground": "#14171A",
+    "muted": "#5E6872",
+    "accent": "#0F766E",
+    "secondary": "#C2410C",
+}
+
+
+def load_presentation(path: Path) -> dict[str, Any]:
+    with path.open(encoding="utf-8") as handle:
+        data = json.load(handle)
+    validate_presentation(data)
+    return cast(dict[str, Any], data)
+
+
+def validate_presentation(deck: dict[str, Any]) -> None:
+    if not isinstance(deck.get("title"), str) or not deck["title"].strip():
+        raise ValueError("presentation must include a non-empty title")
+
+    slides = deck.get("slides")
+    if not isinstance(slides, list) or not slides:
+        raise ValueError("presentation must include at least one slide")
+
+    seen_ids: set[str] = set()
+    for index, slide in enumerate(slides, start=1):
+        if not isinstance(slide, dict):
+            raise ValueError(f"slide {index} must be an object")
+        slide_id = slide.get("id")
+        if not isinstance(slide_id, str) or not slide_id:
+            raise ValueError(f"slide {index} must include id")
+        if slide_id in seen_ids:
+            raise ValueError(f"duplicate slide id: {slide_id}")
+        seen_ids.add(slide_id)
+
+        slide_type = slide.get("type")
+        if slide_type not in SUPPORTED_SLIDE_TYPES:
+            raise ValueError(f"unsupported slide type for {slide_id}: {slide_type}")
+        if not isinstance(slide.get("title"), str) or not slide["title"].strip():
+            raise ValueError(f"slide {slide_id} must include title")
+        if slide_type == "code-reveal" and "code" not in slide:
+            raise ValueError(f"slide {slide_id} of type code-reveal requires code")
+        if slide_type == "code-and-text" and (
+            "code" not in slide or "body" not in slide
+        ):
+            raise ValueError(
+                f"slide {slide_id} of type code-and-text requires body and code"
+            )
+        if slide_type == "image" and "image" not in slide:
+            raise ValueError(f"slide {slide_id} of type image requires image")
+        if slide_type == "gallery" and "images" not in slide:
+            raise ValueError(f"slide {slide_id} of type gallery requires images")
+
+        for animation in slide.get("animations", []):
+            animation_type = (
+                animation.get("type") if isinstance(animation, dict) else None
+            )
+            if animation_type not in SUPPORTED_ANIMATION_TYPES:
+                raise ValueError(
+                    f"slide {slide_id} has unsupported animation: {animation_type}"
+                )
+
+
+def render_html(deck: dict[str, Any]) -> str:
+    theme = _theme(deck)
+    title = _escape(deck["title"])
+    metadata = deck.get("metadata", {})
+    author = _escape(metadata.get("author", ""))
+    date = _escape(metadata.get("date", ""))
+    slides = deck["slides"]
+    rendered_slides = "\n".join(
+        _render_slide(slide, index, len(slides)) for index, slide in enumerate(slides)
+    )
+
+    return f"""<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <title>{title}</title>
+  <style>
+{_css(theme)}
+  </style>
+</head>
+<body>
+  <main class="deck" data-slide-count="{len(slides)}">
+{rendered_slides}
+  </main>
+  <nav class="controls" aria-label="Presentation controls">
+    <button type="button" class="icon-button" id="prev" aria-label="Previous slide">‹</button>
+    <output id="counter" aria-live="polite">1 / {len(slides)}</output>
+    <button type="button" class="icon-button" id="next" aria-label="Next slide">›</button>
+  </nav>
+  <footer class="meta">{author}{" · " if author and date else ""}{date}</footer>
+  <script>
+{_javascript()}
+  </script>
+</body>
+</html>
+"""
+
+
+def _theme(deck: dict[str, Any]) -> dict[str, str]:
+    metadata = deck.get("metadata", {})
+    custom = metadata.get("theme", {}) if isinstance(metadata, dict) else {}
+    return {**DEFAULT_THEME, **{k: v for k, v in custom.items() if k in DEFAULT_THEME}}
+
+
+def _render_slide(slide: dict[str, Any], index: int, total: int) -> str:
+    slide_type = slide["type"]
+    title = _escape(slide["title"])
+    subtitle = _escape(slide.get("subtitle", ""))
+    active_class = " active" if index == 0 else ""
+    animations = _animation_payload(slide)
+    content = _render_slide_content(slide)
+
+    return f"""    <section class="slide slide-{slide_type}{active_class}" id="{_attr(slide["id"])}" data-index="{index}" data-animations='{animations}' aria-label="Slide {index + 1} of {total}">
+      <div class="slide-inner">
+        <p class="kicker">{index + 1:02d} / {total:02d}</p>
+        <h1 data-role="title">{title}</h1>
+        {f'<p class="subtitle" data-role="subtitle">{subtitle}</p>' if subtitle else ""}
+        {content}
+      </div>
+    </section>"""
+
+
+def _render_slide_content(slide: dict[str, Any]) -> str:
+    slide_type = slide["type"]
+    if slide_type == "title-slide":
+        return ""
+    if slide_type == "content":
+        return _render_body_and_bullets(slide)
+    if slide_type == "code-reveal":
+        return _render_code(slide)
+    if slide_type == "code-and-text":
+        return f"""<div class="split">
+          <div>{_render_body_and_bullets(slide)}</div>
+          {_render_code(slide)}
+        </div>"""
+    if slide_type == "image":
+        return _render_image(slide["image"])
+    if slide_type == "gallery":
+        return _render_gallery(slide.get("images", []))
+    raise ValueError(f"unsupported slide type: {slide_type}")
+
+
+def _render_body_and_bullets(slide: dict[str, Any]) -> str:
+    body = _escape(slide.get("body", ""))
+    bullets = slide.get("bullets", [])
+    bullet_html = ""
+    if bullets:
+        items = "\n".join(
+            f'            <li class="reveal-line">{_escape(item)}</li>'
+            for item in bullets
+        )
+        bullet_html = f"""          <ul class="bullets" data-role="body">
+{items}
+          </ul>"""
+    body_html = f'<p class="body" data-role="body">{body}</p>' if body else ""
+    return f"""{body_html}
+{bullet_html}"""
+
+
+def _render_code(slide: dict[str, Any]) -> str:
+    language = _escape(slide.get("language", "text"))
+    lines = str(slide.get("code", "")).splitlines() or [""]
+    rendered = "\n".join(
+        f'<span class="code-line reveal-line" style="--line-index: {index}">{_escape(line)}</span>'
+        for index, line in enumerate(lines)
+    )
+    return f"""<figure class="code-panel" data-role="code">
+          <figcaption>{language}</figcaption>
+          <pre><code>{rendered}</code></pre>
+        </figure>"""
+
+
+def _render_image(image: dict[str, Any]) -> str:
+    caption = _escape(image.get("caption", ""))
+    return f"""<figure class="image-panel" data-role="image">
+          <img src="{_attr(image["src"])}" alt="{_attr(image.get("alt", ""))}">
+          {f"<figcaption>{caption}</figcaption>" if caption else ""}
+        </figure>"""
+
+
+def _render_gallery(images: list[dict[str, Any]]) -> str:
+    rendered = "\n".join(_render_image(image) for image in images)
+    return f'<div class="gallery" data-role="gallery">\n{rendered}\n        </div>'
+
+
+def _animation_payload(slide: dict[str, Any]) -> str:
+    animations = slide.get("animations") or [{"type": "fade-in", "duration": 350}]
+    normalized = []
+    for animation in animations:
+        item = dict(animation)
+        if item.get("type") == "line-by-line-reveal":
+            item["type"] = "line-reveal"
+        normalized.append(item)
+    return _attr(json.dumps(normalized, separators=(",", ":")))
+
+
+def _css(theme: dict[str, str]) -> str:
+    return f"""    :root {{
+      --background: {theme["background"]};
+      --surface: {theme["surface"]};
+      --foreground: {theme["foreground"]};
+      --muted: {theme["muted"]};
+      --accent: {theme["accent"]};
+      --secondary: {theme["secondary"]};
+    }}
+    * {{ box-sizing: border-box; }}
+    body {{
+      margin: 0;
+      min-height: 100vh;
+      overflow: hidden;
+      background: var(--background);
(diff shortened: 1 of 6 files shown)


TimeToBuildBob:
CI is green. The initial Greptile review has been requested and is still pending. The executable/schema files are outside the self-merge allowlist, so I am not merging this autonomously; waiting for review and a maintainer merge.

skills/agentic-presentation/generate_presentation_html.py
```diff
@@ -0,0 +1,513 @@
...
+
+
+def _javascript() -> str:
+    return """    const slides = Array.from(document.querySelectorAll('.slide'));
+    const counter = document.querySelector('#counter');
+    let current = 0;
+
+    function selectedTarget(slide, target) {
+      if (!target) return slide.querySelector('.slide-inner');
+      return slide.querySelector(`[data-role="${target}"]`) || slide.querySelector('.slide-inner');
```
TimeToBuildBob: ❌ **P1** — The animation `target` field from the deck JSON is interpolated unsanitized into a JavaScript template literal that builds a CSS selector: `slide.querySelector(`[data-role="${target}"]`)`. Since `_animation_payload` serializes the animation objects into `data-animations` and the JSON is later parsed in `_javascript`, a malicious deck can set `target` to a string containing a backtick (\`) to break out of the template literal and execute arbitrary JavaScript when the slide renders. For example, `target: "x\`;alert(1)//"` becomes `slide.querySelector(`[data-role="x`;alert(1)//"]`)`, which runs `alert(1)`. The escaping applied in `_animation_payload` (via `_attr`) only handles HTML attribute context, not JavaScript string context, so the backtick survives. This is a stored XSS vector: rendering an untrusted deck file in a browser executes the injected code.

skills/agentic-presentation/generate_presentation_html.py
```diff
@@ -0,0 +1,513 @@
...
+            raise ValueError(f"slide {slide_id} must include title")
+        if slide_type == "code-reveal" and "code" not in slide:
+            raise ValueError(f"slide {slide_id} of type code-reveal requires code")
+        if slide_type == "code-and-text" and (
+            "code" not in slide or "body" not in slide
+        ):
+            raise ValueError(
+                f"slide {slide_id} of type code-and-text requires body and code"
+            )
+        if slide_type == "image" and "image" not in slide:
```
TimeToBuildBob: ⚠️ **P2** — `validate_presentation` only checks that an `image` slide has an `image` key and a `gallery` slide has an `images` key; it does not verify these are objects/lists with the expected structure. A deck with `{"type": "image", "image": "not-a-dict"}` passes `validate_presentation` (line 77 sees the key present) and then crashes in `_render_image` at line 208 where `image["src"]` is accessed on a string, raising `TypeError`. Likewise a `gallery` slide with `images` as a string makes `_render_gallery` iterate over characters and call `_render_image` on each one, causing the same `TypeError`. Because the generator does not use the schema (only the Python validator), malformed user-supplied decks crash the renderer instead of producing a clear validation error.

TimeToBuildBob:
## 🤖 AI code review

Adds a new agentic-presentation skill that renders structured slide-deck JSON into a standalone offline HTML file. The PR introduces a renderer (generate_presentation_html.py), a convenience wrapper (present.py), a draft-07 JSON schema, an example deck, and a test suite. Rendering uses only the standard library; the two schema tests require jsonschema and are guarded with importorskip.

### Confidence Score: N/A

**Score withheld** — 2 findings were dropped by policy and nothing survived, so a clean score here would describe the filter rather than the code. Treat this as *not scored*, not as approved.

> ⚠️ **2 findings found but NOT reported** (2×P2): the review-round cap was reached, so P2 findings stopped blocking. They were raised by this review and dropped by policy — not absent, and not resolved.

🚫 **No findings reported — but not because none were found.** All 2 findings raised on this head were dropped by policy, as stated above. This is a filtered result, not a clean one.

- `skills/agentic-presentation/SKILL.md` — Skill documentation describing workflow, usage, output contract, and verification commands for the agentic-presentation skill.
- `skills/agentic-presentation/examples/agentic-presentation-demo.json` — Example deck fixture used by the skill documentation and tests, exercising title, content, code-reveal, and content slide types.
- `skills/agentic-presentation/generate_presentation_html.py` — New renderer that validates deck JSON, escapes content, and emits a self-contained HTML file with inline CSS and JavaScript.
- `skills/agentic-presentation/present.py` — New CLI wrapper that invokes the generator and optionally opens the rendered HTML in a browser via xdg-open or open.
- `skills/agentic-presentation/presentation.schema.json` — New JSON Schema (draft-07) defining the deck format, including slide types, animation definitions, and theme constraints.
- `skills/agentic-presentation/tests/test_generate_presentation_html.py` — New pytest suite covering validation, HTML output, size limits, animation payload representation, and safe scheme rejection.

</details>

| `e7fb5ec8d093` | 3/5 | 2 | llm | 2026-08-16 20:31 UTC |
| `9c74f2f65661` | 2/5 | 4 | llm | 2026-08-16 21:20 UTC |
| `58c94f785f5f` | 4/5 | 3 | llm | 2026-08-16 22:09 UTC |
| `291ad4ae76b1` | 2/5 | 2 | llm | 2026-08-16 22:45 UTC |
| `b6fa4cd794a8` | 4/5 | 2 | llm | 2026-08-16 23:08 UTC |
| `9ba7498dac5e` | 3/5 | 1 | llm | 2026-08-16 23:44 UTC |

</details>

<sub>Reviewed `fab61c6088e4` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 281s · reports findings; merge decisions belong to the gate · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

skills/agentic-presentation/generate_presentation_html.py
TimeToBuildBob: Fixed in 9c74f2f6: selector construction now passes the untrusted role through CSS.escape() before querySelector. Added a regression test that rejects the original direct interpolation pattern.

skills/agentic-presentation/generate_presentation_html.py
TimeToBuildBob: Fixed in 9c74f2f6: image and gallery payloads are now type-checked, and each image must provide string src/alt fields before rendering. Added regression coverage for scalar image payloads, scalar gallery payloads, and non-object gallery entries.

TimeToBuildBob:
@greptileai review

skills/agentic-presentation/generate_presentation_html.py
```diff
@@ -0,0 +1,526 @@
...
+        item = dict(animation)
+        if item.get("type") == "line-by-line-reveal":
+            item["type"] = "line-reveal"
+        normalized.append(item)
+    return _attr(json.dumps(normalized, separators=(",", ":")))
+
+
+def _css(theme: dict[str, str]) -> str:
+    return f"""    :root {{
+      --background: {theme["background"]};
```
TimeToBuildBob, line 267: ❌ **P1** — The `_css` function interpolates theme values directly into a CSS `:root` block without validating or escaping them. The Python validator never checks that `metadata.theme` values match the `^#[0-9A-Fa-f]{6}$` pattern that the JSON schema enforces, so a deck can pass `validate_presentation` with a theme value such as `red; } body { background: url(javascript:alert(1)) }`. When `render_html` calls `_css`, this value is placed verbatim into `--background: {value};`, allowing an attacker who controls the deck JSON to inject arbitrary CSS into the output file. Since the generated HTML is opened in a browser, the injected CSS is applied, which can be used to deface the deck, redirect images, or exfiltrate data by setting background URLs to an attacker's server. The generator's contract promises all user text is escaped, but theme values are not. Fix: validate theme values against the hex pattern in `validate_presentation`.

skills/agentic-presentation/generate_presentation_html.py
```diff
@@ -0,0 +1,526 @@
...
+def _render_body_and_bullets(slide: dict[str, Any]) -> str:
+    body = _escape(slide.get("body", ""))
+    bullets = slide.get("bullets", [])
+    bullet_html = ""
+    if bullets:
+        items = "\n".join(
+            f'            <li class="reveal-line">{_escape(item)}</li>'
+            for item in bullets
+        )
+        bullet_html = f"""          <ul class="bullets" data-role="body">
```
TimeToBuildBob: ❌ **P1** — For content slides that have both a `body` paragraph and a `bullets` list, both elements are assigned `data-role="body"` (`body_html` and `bullet_html`). The `selectedTarget` JS function performs `slide.querySelector(`[data-role="${CSS.escape(target)}"]`)`, which returns the first matching element in document order — the body paragraph. When an animation targets `"body"` with type `line-reveal`, the class `animate-line-reveal` is added to the paragraph, but the bullets are in a sibling `<ul>` that is not a descendant, so the `reveal-line` children inside the list never receive the animation. Users who specify `line-reveal` on a body+bullets slide get no reveal effect on their bullet points, breaking the documented animation. The demo avoids this by using `slide-in` on such a slide, so it goes undetected. Fix: give bullets a distinct role (e.g., `data-role="bullets"`) or adjust selection to prefer the container with lines.

skills/agentic-presentation/generate_presentation_html.py
```diff
@@ -0,0 +1,526 @@
...
+
+def load_presentation(path: Path) -> dict[str, Any]:
+    with path.open(encoding="utf-8") as handle:
+        data = json.load(handle)
+    validate_presentation(data)
+    return cast(dict[str, Any], data)
+
+
+def validate_presentation(deck: dict[str, Any]) -> None:
+    if not isinstance(deck.get("title"), str) or not deck["title"].strip():
```
TimeToBuildBob, line 49: ❌ **P1** — `validate_presentation` assumes its argument is a dictionary, but `load_presentation` simply passes whatever `json.load` returns. If the input file contains a JSON array, string, or number, `validate_presentation` calls `deck.get("title")` which raises `AttributeError` because non-dicts lack `.get`. The error is unhandled in `main`, so the tool crashes with a traceback instead of printing a helpful error. For example, a file containing `[1,2,3]` passes `json.load` and then dies at line 46. This is a robustness bug for a CLI that is meant to validate user-provided decks. Fix: ensure `deck` is a dict at the top of `validate_presentation` (e.g., `if not isinstance(deck, dict): raise ValueError("deck must be an object")`).

skills/agentic-presentation/generate_presentation_html.py
TimeToBuildBob: Fixed in 58c94f78: runtime validation now rejects non-hex values for every supported theme color before CSS generation. Added a regression test for CSS-injection input.

skills/agentic-presentation/generate_presentation_html.py
TimeToBuildBob: Fixed in 58c94f78: bullet lists now use the distinct `bullets` role, so body and bullet animation targets no longer collide. Added a rendered-markup regression test.

skills/agentic-presentation/generate_presentation_html.py
TimeToBuildBob: Fixed in 58c94f78: top-level non-object JSON now raises a clear `ValueError` before dictionary access. Added regression coverage for an array root.

TimeToBuildBob:
Pushed 58c94f78 addressing all findings from the current AI review: theme colors are validated before CSS interpolation, bullet animation targets are distinct, non-object roots fail cleanly, and the runtime animation enum now matches the schema. The focused suite passes (15 tests), and CI is running on the new head.

The Greptile helper declined another trigger because this PR already has an initial-review trigger; I did not bypass its anti-spam guard.

TimeToBuildBob:

Session: `b271a870-282a-573b-bb0a-8f74ae34e231`

skills/agentic-presentation/generate_presentation_html.py
```diff
@@ -0,0 +1,535 @@
...
+
+    function prepareAnimations(slide) {
+      const animations = JSON.parse(slide.dataset.animations || '[]');
+      slide.querySelectorAll('[class*="animate-"]').forEach((node) => {
+        node.classList.remove('animate-fade-in', 'animate-slide-in', 'animate-line-reveal', 'animate-zoom', 'animate-spin', 'is-visible');
+      });
+      animations.forEach((animation) => {
+        const target = selectedTarget(slide, animation.target);
+        target.classList.add(`animate-${animation.type}`);
+        target.style.setProperty('--duration', `${animation.duration || 450}ms`);
```
TimeToBuildBob: ⚠️ **P2** — In _javascript, the lines that set --duration and --stagger use the `||` operator to apply defaults, treating the legitimate value 0 as falsy. The schema defines duration and stagger with minimum 0, so a slide specifying duration:0 (intended to animate immediately) or stagger:0 (intended to reveal all lines at once) silently falls back to 450ms and 80ms respectively. The animation timing then deviates from the declared contract for every deck that uses a zero value, producing slower animations than the author requested. Replacing `||` with the nullish coalescing operator `??` would let only null/undefined trigger the default.

skills/agentic-presentation/generate_presentation_html.py
```diff
@@ -0,0 +1,535 @@
...
+      });
+      counter.value = `${current + 1} / ${slides.length}`;
+      prepareAnimations(slides[current]);
+    }
+
+    document.querySelector('#prev').addEventListener('click', () => showSlide(current - 1));
+    document.querySelector('#next').addEventListener('click', () => showSlide(current + 1));
+    document.addEventListener('keydown', (event) => {
+      if (event.key === 'ArrowLeft') showSlide(current - 1);
+      if (event.key === 'ArrowRight' || event.key === ' ' || event.key === 'Enter') showSlide(current + 1);
```
TimeToBuildBob: ⚠️ **P2** — The document keydown handler calls showSlide(current+1) for both Space and Enter without calling event.preventDefault(), and the same navigation is wired as click listeners on the prev/next buttons. When a button has focus (for example, after the user clicks the next button with the mouse) and the user then presses Space or Enter, the browser's default action triggers a click on the focused button in addition to the keydown handler. Both paths execute showSlide(current+1), advancing two slides instead of one. The fix is to call event.preventDefault() inside the keydown branch for Space/Enter, or to ignore the event when the target is a button.

skills/agentic-presentation/present.py
```diff
@@ -0,0 +1,82 @@
...
+        default=500,
+        help="fail if generated HTML exceeds this many KB (default: 500)",
+    )
+    args = parser.parse_args(argv)
+
+    skill_dir = Path(__file__).parent
+    generator = skill_dir / "generate_presentation_html.py"
+
+    if args.live and not args.output:
+        tmp_dir = Path(tempfile.mkdtemp())
```
TimeToBuildBob: ⚠️ **P2** — In present.py, when --live is used without an explicit -o output, a temp directory is created via tempfile.mkdtemp() and used for the rendered file, but it is never removed. Each such invocation leaks a directory under the system temp space. The Python tempfile module provides TemporaryDirectory for exactly this purpose; using it and letting the context manager clean up would avoid the leak. This is a low-severity resource leak that accumulates across repeated runs.

skills/agentic-presentation/generate_presentation_html.py
```diff
@@ -0,0 +1,535 @@
...
+    slides = deck.get("slides")
+    if not isinstance(slides, list) or not slides:
+        raise ValueError("presentation must include at least one slide")
+
+    seen_ids: set[str] = set()
+    for index, slide in enumerate(slides, start=1):
+        if not isinstance(slide, dict):
+            raise ValueError(f"slide {index} must be an object")
+        slide_id = slide.get("id")
+        if not isinstance(slide_id, str) or not slide_id:
```
TimeToBuildBob, line 74: ⚠️ **P2** — `validate_presentation` checks only that each slide's `id` is a non-empty string; it never applies the pattern `^[A-Za-z0-9][A-Za-z0-9_-]*$` that `presentation.schema.json` declares. A deck whose slide id contains a space, punctuation, or non-ASCII character passes validation and is written into the HTML `id` attribute. The generated markup then has an invalid `id` value, which breaks CSS selectors or fragment navigation that target that id, even though the schema promised the stricter format.

skills/agentic-presentation/generate_presentation_html.py
```diff
@@ -0,0 +1,535 @@
...
+        </figure>"""
+
+
+def _render_gallery(images: list[dict[str, Any]]) -> str:
+    rendered = "\n".join(_render_image(image) for image in images)
+    return f'<div class="gallery" data-role="gallery">\n{rendered}\n        </div>'
+
+
+def _animation_payload(slide: dict[str, Any]) -> str:
+    animations = slide.get("animations") or [{"type": "fade-in", "duration": 350}]
```
TimeToBuildBob: ⚠️ **P2** — In `_animation_payload`, the expression `slide.get("animations") or [{"type": "fade-in", "duration": 350}]` treats an explicit empty list as falsy and substitutes a default fade-in animation. A deck author who sets `"animations": []` to request a static slide with no animation is given a fade-in anyway. The emitted HTML contains a `data-animations` payload with the default, so the slide animates on entry, contradicting the author's intent and the schema's allowance of an empty array.

TimeToBuildBob:
Pushed `291ad4ae` addressing the remaining P2 review findings:

- **Slide id pattern enforcement** — Python validator now applies the same `^[A-Za-z0-9][A-Za-z0-9_-]*$` pattern the JSON schema declares. A deck with spaces, `!`, or non-ASCII in an id now fails `validate_presentation` with a clear error instead of rendering invalid HTML `id` attributes.
- **Empty `animations: []` respected** — `_animation_payload` now uses `if anims is not None` instead of `or`, so an explicit empty list is passed through as-is rather than replaced by a default fade-in. Deck authors who want a static slide with no animation now get one.
- **`||` → `??` for duration/delay/stagger** — JavaScript timer values now use nullish coalescing so explicit `0` (animate immediately, no stagger) is honoured instead of falling through to the defaults.
- **`event.preventDefault()` for keyboard navigation** — Space/Enter keydown now calls `preventDefault()` before `showSlide`, preventing a double-advance when a nav button has focus.
- **`tempfile.mkdtemp()` leak removed** — `present.py` now defaults `--live` to `index.html` like the non-live path; no temp directory is created or leaked.

Regression tests added for all five (21 total, all green). CI running on the new head.

skills/agentic-presentation/generate_presentation_html.py
TimeToBuildBob: Fixed in 291ad4ae: `||` replaced with `??` (nullish coalescing) for `--duration`, `--stagger`, and delay timer values, so an explicit `0` (animate immediately, no stagger) is honoured instead of falling through to the defaults.

skills/agentic-presentation/generate_presentation_html.py
TimeToBuildBob: Fixed in 291ad4ae: Space and Enter keydown now calls `event.preventDefault()` before `showSlide`, preventing double-advance when a navigation button has focus.

skills/agentic-presentation/present.py
TimeToBuildBob: Fixed in 291ad4ae: `present.py` now defaults `--live` to `index.html` in the current directory; no temp directory is created or leaked.

skills/agentic-presentation/generate_presentation_html.py
TimeToBuildBob: Fixed in 291ad4ae: Python validator now enforces the `^[A-Za-z0-9][A-Za-z0-9_-]*$` pattern the JSON schema declares. A slide `id` with spaces, `!`, or non-ASCII now fails `validate_presentation` with a clear error instead of producing invalid HTML `id` attributes.

skills/agentic-presentation/generate_presentation_html.py
TimeToBuildBob: Fixed in 291ad4ae: `_animation_payload` now uses `if anims is not None` instead of `or`, so an explicit empty list (`animations: []`) is passed through as-is rather than replaced by a default fade-in. Deck authors who want a static slide with no animation now get one.

skills/agentic-presentation/generate_presentation_html.py
```diff
@@ -0,0 +1,546 @@
...
+        if slide_type == "image":
+            _validate_image(slide.get("image"), f"slide {slide_id} image")
+        if slide_type == "gallery":
+            images = slide.get("images")
+            if not isinstance(images, list):
+                raise ValueError(f"slide {slide_id} images must be an array")
+            for image_index, image in enumerate(images, start=1):
+                _validate_image(image, f"slide {slide_id} image {image_index}")
+
+        for animation in slide.get("animations", []):
```
TimeToBuildBob: ❌ **P1** — In `validate_presentation`, the loop over `slide.get("animations", [])` assumes the value is iterable. If a deck sets `"animations": null` or a non-list like `5`, the `for animation in ...` raises `TypeError: 'NoneType' object is not iterable` (or `'int' object is not iterable`) at line 105, instead of producing a clear validation error. This happens before any rendering, so the generator crashes with a traceback and writes no output file. The schema would reject such a deck, but the Python validator does not check the type, so a malformed deck slips through and breaks the tool. Fix by verifying `animations` is a list (or None) before iterating.

skills/agentic-presentation/generate_presentation_html.py
```diff
@@ -0,0 +1,546 @@
...
+    if slide_type == "image":
+        return _render_image(slide["image"])
+    if slide_type == "gallery":
+        return _render_gallery(slide.get("images", []))
+    raise ValueError(f"unsupported slide type: {slide_type}")
+
+
+def _render_body_and_bullets(slide: dict[str, Any]) -> str:
+    body = _escape(slide.get("body", ""))
+    bullets = slide.get("bullets", [])
```
TimeToBuildBob, line 218: ❌ **P1** — `_render_body_and_bullets` iterates over `bullets` without confirming it is a list. A deck with `"bullets": 5` or `"bullets": {}}` causes `for item in bullets` to raise `TypeError` during rendering (line 205). `validate_presentation` never checks the type of `bullets`, so a malformed deck that passes validation crashes the renderer. This results in no HTML output and an unhandled exception. Since the schema requires `bullets` to be an array, the validator should enforce that too.

TimeToBuildBob:
Pushed `b6fa4cd7` addressing the two P1 findings from the latest AI review:

- **`animations` non-list crash** — `slide.get("animations", [])` returns `None` when the JSON key is explicitly `null` (the default only fires when the key is absent). The validator now checks `isinstance(anims, list)` before iterating; a non-list raises a clear `ValueError`, and `null` is treated as absent (no error). Fixes the `TypeError: 'NoneType' object is not iterable` at line 105.

- **`bullets` non-list crash** — added a type guard in `validate_presentation` for `content` and `code-and-text` slides that rejects non-list `bullets` values before they reach `_render_body_and_bullets`. Fixes the unhandled `TypeError` at line 200.

Regression tests added for both (23 total, all green).

skills/agentic-presentation/generate_presentation_html.py
```diff
@@ -0,0 +1,553 @@
...
+            images = slide.get("images")
+            if not isinstance(images, list):
+                raise ValueError(f"slide {slide_id} images must be an array")
+            for image_index, image in enumerate(images, start=1):
+                _validate_image(image, f"slide {slide_id} image {image_index}")
+
+        anims = slide.get("animations")
+        if anims is not None and not isinstance(anims, list):
+            raise ValueError(f"slide {slide_id} animations must be an array")
+        for animation in anims or []:
```
TimeToBuildBob, line 116: ⚠️ **P2** — The `validate_presentation` loop over animations only checks that each entry's `type` is in `SUPPORTED_ANIMATION_TYPES`; it never validates `duration`, `delay`, or `stagger`. The schema declares these as integers with `minimum: 0`, but the generator will accept a string like `"1; } body { display: none; }"` without raising. When the JavaScript executes, it sets CSS custom properties via `setProperty('--duration', `${animation.duration ?? 450}ms`)`, injecting the untrusted string verbatim into the CSS custom property. While not a classic XSS, this is a CSS injection vector that can break the deck's styling or be leveraged to alter the presentation's appearance. The validator should enforce numeric, non-negative values to match the schema.

TimeToBuildBob:
Pushed `9ba7498d` — this closes the Greptile convergence adjudication. Both remaining P2 findings are now fixed.

## Adjudication summary

**Fixed this session:**
- **P2 `cbaa1d93d300` (line 88 — type check for `code`/`body`)** — `code-reveal` only checked key *presence*, not type. A slide with `"code": 123` passed validation and the renderer would `str()` the integer instead of failing. Now: `isinstance(slide.get("code"), str)` is required; same for both `code` and `body` on `code-and-text`. Regression test added.
- **P2 `497d63f2252e` (line 112 — CSS injection via animation timing fields)** — `duration`, `delay`, and `stagger` were never validated as integers. A string value like `"1; } body { display: none; }"` was accepted and emitted verbatim into a CSS `setProperty` call. Now validated as non-negative integers per the JSON schema. Regression test added.

**Remaining after this session:** None. All P1/P2 findings across all 5 prior rounds have been resolved or auto-resolved.

**CI:** `pre-commit` passes on `9ba7498d`. The full CI suite (`test`, `typecheck`) is gated behind a "Detect changes" check; it will run if the change-detection job sees this file as modified.

**Domain risk:** The CSS injection surface (now fixed) touched the animation validator. A maintainer who wants to verify manually: `uv run pytest skills/agentic-presentation/tests/ -q` (25 tests) should be sufficient. Visual rendering of the demo deck is also a useful sanity check: `python3 skills/agentic-presentation/present.py skills/agentic-presentation/examples/agentic-presentation-demo.json`.

**Convergence:** 5 prior review rounds (scores: 3/5, 2/5, 4/5, 2/5, 4/5); the stable-round detector shows `status: no_history` (rounds stored per-commit, not cross-session). This session adjudicated rather than re-triggered — no further `@greptileai review` was requested.

Ready for maintainer merge.

skills/agentic-presentation/generate_presentation_html.py
```diff
@@ -0,0 +1,563 @@
...
+    return f"""<figure class="code-panel" data-role="code">
+          <figcaption>{language}</figcaption>
+          <pre><code>{rendered}</code></pre>
+        </figure>"""
+
+
+def _render_image(image: dict[str, Any]) -> str:
+    caption = _escape(image.get("caption", ""))
+    return f"""<figure class="image-panel" data-role="image">
+          <img src="{_attr(image["src"])}" alt="{_attr(image.get("alt", ""))}">
```
TimeToBuildBob, line 249: ❌ **P1** — The generator inserts the `src` attribute from slide JSON directly into the generated HTML after HTML-escaping, but does not validate the URL scheme. A deck with an image `src` set to `javascript:alert(1)` will render as ``. When the presentation is opened in a browser, older browsers that execute javascript: URLs in img tags (or if the image is clicked) could run arbitrary script in the page's origin, turning an untrusted deck into a stored XSS vector. The `_validate_image` function only checks that `src` is a non-empty string, so this passes validation and rendering.

TimeToBuildBob:
Pushed `fab61c6` fixing the P1 security finding from the latest AI review:

- **`javascript:` / `data:` XSS via image `src`** — `_validate_image` only checked that `src` was a non-empty string, allowing `javascript:alert(1)` and similar dangerous schemes to pass through and render verbatim in the HTML output. Fixed by validating the URL scheme with `urllib.parse`: only empty scheme (relative paths), `http`, and `https` are accepted; anything else raises `ValueError("... src has unsafe scheme '...'")`.

5 regression tests added (4 dangerous-scheme rejections — `javascript:`, `JAVASCRIPT:`, `data:`, `vbscript:` — plus a safe-scheme pass covering `https://`, `http://`, `/`, and `./`). 30 tests green.

CI running on `fab61c6`.

TimeToBuildBob:
PR is ready for human merge. Summary of where things stand:

- CI green (pre-commit + all checks pass on `fab61c6`)
- 5 rounds of manual AI review completed; all P1/P2 findings addressed (30 tests green)
- Fixes covered: theme color validation, type guards for `animations`/`bullets`/`code`/`body` fields, CSS injection via timing fields, keyboard nav double-advance, tempfile leak, `javascript:`/`data:` XSS via image `src`
- Self-merge not eligible: `skills/` files are not in the autonomous merge allowlist and Greptile never responded to the original `@greptileai review` request — needs maintainer merge

PR #1435, opened 2026-08-16
fix(twitter): distinguish spend-cap 403 from content-rejection 403

TimeToBuildBob:
## Problem

The X API returns HTTP 403 for **both** the monthly spend cap **and** permanent content rejections (cashtags/dollar amounts, content policy, restricted replies). `post-approved-tweets.sh` treated every 403 as evidence of the spend cap and re-`touch`ed the cap flag, so one poison tweet carrying cashtags could refresh the flag every ~20h **forever** — impersonating an active cap that never actually exists.

Symptoms:
- Posting works fine (tweets went out 08-14/08-15) yet the cap flag stays perpetually present.
- At least one task (`promote-third-failure-never-came-blog`) is parked `waiting` on "the spend cap to reset" — a reset that can never come.

## Fix

- Add `_classify_tweet_post_error()` — branches on the response body message (cap markers) rather than the bare status code. Returns `"cap"` / `"permanent"` / `"other"`.
- Add `_quarantine_tweet_post_failure()` — moves permanently-rejected tweets to `rejected/` so they drain out of the approved queue instead of wedging it (caps retries at one).
- Wire both into the `post()` and `auto()` call sites: cap → keep queued + signal wrapper; permanent → quarantine; other → leave queued.

## Verification

- `python3 -m pytest tests/test_twitter_post_403_classifier.py -q` → 7 passed.
- `python3 -m pytest tests/test_validate_task_frontmatter_exit.py -q` → 6 passed (unaffected).
- Symptom check (prior session): cap flag files absent, approved queue drained from 16 → 0, rejected/ populated with the poison tweets.

Closes the `twitter-poison-tweet-impersonates-spend-cap` bug.

### DIFF
--- a/scripts/twitter/workflow.py
+++ b/scripts/twitter/workflow.py
@@ -849,9 +849,12 @@ def _classify_tweet_post_error(error: Exception) -> str:
             or "usagecapped" in api_error_type.lower()
         ):
             return "cap"
-        # Got a non-cap structured type — the API explicitly says this is a
-        # content rejection or other permanent error.
-        return "permanent"
+        # Non-cap structured type: the X API uses non-specific types (including
+        # "about:blank") for many conditions — some of which are transient (e.g.
+        # temporary account restrictions, permission changes). We cannot reliably
+        # classify an unknown type as "permanent", so return "other" and leave the
+        # tweet queued rather than silently quarantining it.
+        return "other"
 
     # No structured response body. Fall back to substring matching.
     # Cap markers (e.g. "spend cap") could theoretically appear in user tweet
--- a/scripts/twitter/workflow.py
+++ b/scripts/twitter/workflow.py
@@ -767,6 +767,69 @@ def _quarantine_permanent_reply_failure(
     return True
 
 
+# The X API returns HTTP 403 for BOTH the monthly spend cap AND permanent content
+# rejections (cashtags/dollar amounts, content policy, restricted replies, ...).
+# A bare status-code check cannot tell them apart, which is exactly the bug that
+# let one poison tweet impersonate a spend cap (see the task that filed this).
+# Distinguish by the response body message, never by the status code alone.
+_CAP_403_MARKERS = ("spend cap", "monthly spend cap")
+
+
+def _classify_tweet_post_error(error: Exception) -> str:
+    """Classify a create_tweet failure for the autopost pipeline.
+
+    Returns one of:
+      "cap"       - billing/monthly spend cap. The whole queue is blocked; keep
+                    tweets queued so they post once the cap resets, and signal the
+                    wrapper to set/refresh the spend-cap flag.
+      "permanent" - a non-cap 403 (cashtags/dollar amounts, content policy,
+                    restricted reply, ...) that will NEVER succeed on retry.
+                    Quarantine the offending tweet so it stops blocking the queue.
+      "other"     - transient or unclassified; leave the tweet queued for retry.
+    """
+    message = str(error).lower()
+    # Cap marker takes priority: if the body names a spend cap, treat it as a cap
+    # regardless of exact status framing.
+    if any(marker in message for marker in _CAP_403_MARKERS):
+        return "cap"
+    # A Forbidden/403 on create_tweet is a content/permission rejection. Match on
+    # either the typed tweepy exception (when available) OR the status text in the
+    # message, so a plain Exception carrying a "403 Forbidden: ..." body is still
+    # caught even if the error was wrapped or not a tweepy Forbidden instance.
+    is_403 = "forbidden" in message or message.startswith("403")
+    try:
+        from tweepy.errors import Forbidden as _Forbidden
+
+        is_403 = is_403 or isinstance(error, _Forbidden)
+    except Exception:
+        pass
+    return "permanent" if is_403 else "other"
+
+
+def _quarantine_tweet_post_failure(
+    path: Path, draft: TweetDraft, error: Exception
+) -> bool:
+    """Move an approved tweet that the API permanently rejects to rejected/.
+
+    A non-cap 403 (cashtags/dollar amounts, content policy, restricted reply, etc.)
+    will never succeed on retry. Leaving it in approved/ wedges the queue: every
+    autopost cycle hits the same 403, and the wrapper's "neither posted nor saw a
+    cap" fallback keeps a phantom spend-cap flag alive. Quarantining it on the
+    first confirmed non-cap 403 caps retries at one (content rejections never
+    clear), so the queue can drain.
+    """
+    reason = _permanent_reply_failure_reason(draft, error)
+    if reason is None:
+        reason = f"Permanent Twitter post rejection: {error}"
+    draft.reject_reason = reason
+    draft.save(path)
+    rejected_path = move_draft(path, "rejected")
+    console.print(
+        f"[yellow]Moved permanently rejected tweet to {rejected_path}[/yellow]"
+    )
+    return True
+
+
 def find_draft(
     draft_id: str, status: str | None = None, show_error: bool = True
 ) -> Path | None:
@@ -1580,7 +1643,25 @@ def post(
                     console.print("[red]Error: No response data from tweet creation")
             except Exception as e:
                 console.print(f"[red]Error posting tweet: {e}")
-                _quarantine_permanent_reply_failure(path, draft, e)
+                failure_kind = _classify_tweet_post_error(e)
+                if failure_kind == "cap":
+                    # Whole queue blocked by the billing cap. Keep every tweet queued
+                    # so they post once the cap resets, and stop trying further posts
+                    # this run. The wrapper greps the output for "monthly spend cap"
+                    # to set/refresh its flag.
+                    console.print(
+                        "[yellow]Twitter monthly spend cap reached; tweets will post when the cap resets."
+                    )
+                    break
+                elif failure_kind == "permanent":
+                    # A content/permission 403 that will never succeed on retry.
+                    # Quarantine it so it stops blocking the queue (caps retries).
+                    _quarantine_tweet_post_failure(path, draft, e)
+                else:
+                    # Transient/unclassified; leave queued for the next cycle.
+                    console.print(
+                        "[yellow]Tweet not posted (transient error); leaving queued."
+                    )
         else:
             console.print("[yellow]Skipped posting tweet")
 
@@ -2579,7 +2660,18 @@ def auto(
                         )
                 except Exception as e:
                     console.print(f"[red]Error posting tweet: {e}")
-                    _quarantine_permanent_reply_failure(path, draft, e)
+                    failure_kind = _classify_tweet_post_error(e)
+                    if failure_kind == "cap":
+                        console.print(
+                            "[yellow]Twitter monthly spend cap reached; tweets will post when the cap resets."
+                        )
+                        break
+                    elif failure_kind == "permanent":
+                        _quarantine_tweet_post_failure(path, draft, e)
+                    else:
+                        console.print(
+                            "[yellow]Tweet not posted (transient error); leaving queued."
+                        )
 
     # Summary
     console.print("\n[bold]Automation Cycle Summary:[/bold]")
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:

Session: `3428c728-e554-50db-920b-d3b9dbc175df`

tests/test_twitter_post_403_classifier.py
```diff
@@ -0,0 +1,206 @@
...
+    yield module
+    sys.modules.pop("twitter_workflow_under_test", None)
+    for key, original in original_modules.items():
+        if original is _MISSING:
+            sys.modules.pop(key, None)
+        else:
+            sys.modules[key] = original
+
+
+class TestClassifyTweetPostError:
```
TimeToBuildBob, line 159: ⚠️ **P2** — The new test file only exercises `_classify_tweet_post_error` with a handful of error messages. It never calls `_quarantine_tweet_post_failure` nor tests the new branches in `post()` and `auto()`. Consequently, the quarantine action (moving the draft to rejected/, writing reject_reason) and the loop control (break on cap, continue on other) are completely untested. A regression in the quarantine function that breaks the file move or the break-on-cap logic that silently swallows errors would not be caught, undermining the PR's core fix for the poison-tweet impersonating a cap.

scripts/twitter/workflow.py
```diff
@@ -767,6 +767,69 @@ def _quarantine_permanent_reply_failure(
...
+                    wrapper to set/refresh the spend-cap flag.
+      "permanent" - a non-cap 403 (cashtags/dollar amounts, content policy,
+                    restricted reply, ...) that will NEVER succeed on retry.
+                    Quarantine the offending tweet so it stops blocking the queue.
+      "other"     - transient or unclassified; leave the tweet queued for retry.
+    """
+    message = str(error).lower()
+    # Cap marker takes priority: if the body names a spend cap, treat it as a cap
+    # regardless of exact status framing.
+    if any(marker in message for marker in _CAP_403_MARKERS):
```
TimeToBuildBob: ❌ **P1** — `_classify_tweet_post_error` treats the presence of the substrings "spend cap" or "monthly spend cap" anywhere in the lowercased error message as conclusive evidence of a billing cap, taking priority over all other checks. If the X API ever includes the offending tweet text or a user-controlled snippet in a content-rejection 403 message, a tweet that merely contains the phrase "spend cap" would be misclassified as a cap. At the `post()` and `auto()` call sites that classification triggers `break`, stopping the queue and leaving every remaining approved tweet stuck, exactly the poison-tweet scenario this PR claims to fix. Because the error message is attacker-influenced via the tweet content, this is a security regression.

scripts/twitter/workflow.py
```diff
@@ -767,6 +767,69 @@ def _quarantine_permanent_reply_failure(
...
+        return "cap"
+    # A Forbidden/403 on create_tweet is a content/permission rejection. Match on
+    # either the typed tweepy exception (when available) OR the status text in the
+    # message, so a plain Exception carrying a "403 Forbidden: ..." body is still
+    # caught even if the error was wrapped or not a tweepy Forbidden instance.
+    is_403 = "forbidden" in message or message.startswith("403")
```
TimeToBuildBob: ⚠️ **P2** — `is_403 = "forbidden" in message or message.startswith("403")` only recognizes a 403 if the string begins with the exact characters "403" or contains the word "forbidden". Real tweepy/http errors may be formatted differently, e.g. "HTTP 403: ..." or "Status 403 returned". In such cases neither condition holds (and the lazy tweepy isinstance check only helps if tweepy is installed and the exception is a typed Forbidden), so the error is classified "other". That leaves the tweet queued for retry on every cycle, so a permanently rejected tweet never drains, re-introducing the queue-wedging problem for those message shapes.

TimeToBuildBob:
## 🤖 AI code review

Adds `_classify_tweet_post_error()` to distinguish cap vs permanent content-rejection 403s via the X API v2 `type` field and message markers, adds `_quarantine_tweet_post_failure()` to move permanently rejected tweets to rejected/, and wires these into the `post` and `auto` commands. Adds a test file covering the classifier and quarantine behavior.

### Confidence Score: N/A

**Score withheld** — 1 finding were dropped by policy and nothing survived, so a clean score here would describe the filter rather than the code. Treat this as *not scored*, not as approved.

🚫 **No findings reported — but not because none were found.** All 1 finding raised on this head were dropped by policy, as stated above. This is a filtered result, not a clean one.

- `scripts/twitter/workflow.py` — Adds cap/permanent classification helpers, a quarantine function, and uses them in the post and auto error paths.
- `tests/test_twitter_post_403_classifier.py` — New tests for the classifier and quarantine function.

</details>

| `50598c8a610e` | 3/5 | 3 | llm | 2026-08-16 22:26 UTC |
| `cdc469a31244` | 3/5 | 2 | llm | 2026-08-16 23:01 UTC |
| `2c4db80e3355` | 3/5 | 2 | llm | 2026-08-16 23:41 UTC |
| `9a885409c647` | 3/5 | 1 | llm | 2026-08-16 23:49 UTC |
| `803d22cabb93` | 3/5 | 1 | llm | 2026-08-17 00:06 UTC |

</details>

<sub>Reviewed `73a06f941b2f` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 215s · reports findings; merge decisions belong to the gate · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

TimeToBuildBob:
Addressed the AI reviewer's P1 and P2 findings in `cdc469a3`:

**P1 (security) — fixed**: `_classify_tweet_post_error` now checks for a 403/Forbidden status *first*, then checks cap markers only within a confirmed 403. Previously, cap markers were matched anywhere in the raw error message before any status check, so a non-403 error whose text contained "spend cap" (or a tweet whose content was embedded in an error body) would stop the entire queue — the same poison-tweet pattern this PR claims to fix. Non-403 errors now return `"other"` regardless of message text.

**P2 — 403 detection broadened**: changed `message.startswith("403")` to `"403" in message`, catching `"HTTP 403: ..."` and similar formats that the startswith check missed.

**P2 — quarantine tests added**: new `TestQuarantineTweetPostFailure` class with two tests covering `_quarantine_tweet_post_failure` directly: file-move (verifies the original is removed and appears in `rejected/`) and text-preservation. Also added `test_http_403_prefix_classified_permanent` for the new format and updated `test_spend_cap_message_without_403_is_other` to reflect the corrected behavior (non-403 cap text → `"other"`, not `"cap"`).

10/10 tests pass.

scripts/twitter/workflow.py
```diff
@@ -767,6 +767,73 @@ def _quarantine_permanent_reply_failure(
...
+        pass
+
+    if not is_403:
+        return "other"
+
+    # It is a 403. Distinguish billing cap from content rejection by body text.
+    if any(marker in message for marker in _CAP_403_MARKERS):
+        return "cap"
+
+    return "permanent"
```
TimeToBuildBob: ⚠️ **P2** — The classifier returns `"permanent"` for every 403 error that does not contain a cap marker, assuming all such errors are permanent content rejections. However, the X API can return a 403 for other transient reasons, such as a temporary account posting restriction or a permissions change that may resolve later. When this happens, `_quarantine_tweet_post_failure` moves the tweet to `rejected/` permanently, even though it could have been posted on a subsequent attempt. This can silently drop tweets that would otherwise have succeeded, and there is no retry path because the tweet is no longer in `approved/`. The classifier should distinguish clearly permanent rejection messages (e.g., cashtag/dollar-amount markers) from unknown 403s, treating the latter as transient.

TimeToBuildBob:
@greptileai review

scripts/twitter/workflow.py
```diff
@@ -767,6 +767,138 @@ def _quarantine_permanent_reply_failure(
...
+    api_error_type = _get_twitter_api_error_type(error)
+    if api_error_type:
+        if (
+            "usage-capped" in api_error_type.lower()
+            or "usagecapped" in api_error_type.lower()
+        ):
+            return "cap"
+        # Got a non-cap structured type — the API explicitly says this is a
+        # content rejection or other permanent error.
+        return "permanent"
```
TimeToBuildBob: ❌ **P1** — In `_classify_tweet_post_error()`, the structured-response path returns `'permanent'` for any non-usage-capped `type` without checking known permanent markers. This over-broad classification quarantines every 403 that carries a structured body, including transient errors (e.g. temporary account restrictions) whose `type` is unrecognized, so a tweet that might succeed on retry is permanently moved to rejected/. It also mis-detects real spend caps if the API ever returns a non-'usage-capped' type: the tweet is quarantined, the loop does not `break`, and every remaining approved draft hits the same error and is quarantined in turn, draining the queue without ever signaling the wrapper to set the cap flag. The fallback (unstructured) path correctly returns `'other'` for unknown 403s, so the two paths behave inconsistently. The fix is to fall through to the substring checks (or return `'other'`) when the type is not a known cap marker, rather than assuming permanent.

scripts/twitter/workflow.py
```diff
@@ -767,6 +767,138 @@ def _quarantine_permanent_reply_failure(
...
+    if not is_403:
+        return "other"
+
+    # P1 fix: prefer the structured X API v2 `type` field over substring matching.
+    # The type URI is API-controlled and is never influenced by user tweet content,
+    # so it cannot be spoofed by a tweet that happens to contain "spend cap".
+    api_error_type = _get_twitter_api_error_type(error)
+    if api_error_type:
+        if (
+            "usage-capped" in api_error_type.lower()
```
TimeToBuildBob, line 851: ⚠️ **P2** — `_get_twitter_api_error_type()` returns `body.get("type") or body.get("title") or ""` without type-checking the value. If the X API ever returns a non-string `type` (e.g. an integer status code), the returned value is truthy, and the subsequent `api_error_type.lower()` in `_classify_tweet_post_error()` raises `AttributeError`. That exception escapes the `except Exception as e:` block in `post()` and `auto()`, crashing the entire posting loop instead of treating the error as transient or permanent. This is a robustness gap in code intended to handle API errors gracefully. The fix is to coerce the value to `str` (e.g. `str(body.get("type") or body.get("title") or "")`) or check `isinstance(api_error_type, str)` before calling `.lower()`.

scripts/twitter/workflow.py
```diff
@@ -767,6 +767,150 @@ def _quarantine_permanent_reply_failure(
...
+        # (e.g. temporary account restriction). Leave the tweet queued rather than
+        # silently quarantining it on an unrecognized error type.
+        return "other"
+
+    # No structured response body. Fall back to substring matching.
+    # Cap markers (e.g. "spend cap") could theoretically appear in user tweet
+    # content embedded in an error body, but the X API does not embed user text
+    # in 403 error messages as of 2026-08-16. The structured path above is the
+    # robust fix for that case in production.
+    if any(marker in message for marker in _CAP_403_MARKERS):
```
TimeToBuildBob: ❌ **P1** — In the unstructured fallback of _classify_tweet_post_error, cap markers are checked before permanent content-rejection markers. If a permanent rejection (e.g., a tweet containing a cashtag) triggers an error message that also contains the string "spend cap" — which could happen if the X API echoes the rejected tweet's text in the error detail — the function returns "cap" instead of "permanent". The caller then breaks the posting loop and leaves the tweet in approved/, so it is retried every cycle, recreating the exact poison-tweet wedge this PR is supposed to fix. The structured path avoids this by short-circuiting on the API type field, but the fallback (used when no structured response exists) lacks that protection and re-introduces the spoofing vector. A tweet containing "spend cap" could impersonate a billing cap and block the entire queue. The test suite covers the structured anti-spoofing case but does not test a fallback message containing both marker types. _(3/3 passes)_

TimeToBuildBob:
Fixed remaining P1 in `73a06f94`: in the unstructured fallback path of `_classify_tweet_post_error`, permanent-rejection markers are now checked **before** cap markers.

**What was wrong**: the structured path (with an API `type` field) checked permanent markers first — but the unstructured fallback (no structured body) checked cap markers first. A 403 error message that contained both "spend cap" text and a permanent-rejection phrase (e.g. "cashtag") would return `"cap"`, potentially letting a poison tweet impersonate the billing cap and halt the queue — the exact bug this PR is supposed to prevent.

**Fix**: swapped the order in the unstructured fallback to match the structured path (permanent first, cap only if no permanent marker matched). Added `test_unstructured_403_with_both_markers_is_permanent` to cover the overlapping-markers case in both paths.

@TimeToBuildBob review

TimeToBuildBob:
@TimeToBuildBob review

PR #1439, opened 2026-08-17
feat(gptme-rag): add TF-IDF lexical retrieval backend

TimeToBuildBob:
## What

Adds a `TfidfIndex` lexical retrieval backend to `gptme-rag`, wrapping scikit-learn's `TfidfVectorizer` with cosine-similarity ranking, a relevance floor, and path exclusion.

This is the **lexical slice** of making `gptme-rag` the canonical retrieval substrate (ErikBjare/bob `upstream-retrieval-into-gptme-rag`). The dense (embedding) path measurably drops exact-identifier queries — a function name, file path, or error string is a rare term semantic embeddings discard. A lexical backend preserves those terms.

## Why uncapped vocabulary

The default is `max_features=None`, deliberately. A global top-N vocabulary is chosen by corpus-wide frequency, so rare terms get dropped first — and identifiers *are* the rare terms. Measured in the 2026-08-10 dense-vs-lexical trial: capping at 10k cost **28 points of exact-identifier recall** (38.7% → 67.2% uncapped) with zero difference on the semantic subset. See the docstring for the full rationale.

## Scope

- New `gptme_rag/lexical.py`: `TfidfIndex` (build / search / atomic save / load), `LexicalHit`, `LexicalDependencyMissing`.
- `scikit-learn` added as an **optional** extra (`gptme-rag[lexical]`) so dense-only users don't pay the install cost; it's imported lazily and raises `LexicalDependencyMissing` at first use if absent.
- 5 unit tests covering identifier ranking, relevance floor, path exclusion, persistence round-trip, and empty-index behavior.

Not in scope (separate PRs): hybrid rank fusion (RRF), memory-type weighting, task-scoped retrieval. This is the bare lexical backend the hybrid path will compose against.

## Verification

- `pytest tests/test_lexical.py` — 5 passed
- `ruff check` — clean
- `mypy` — no new errors in `lexical.py` (pre-existing `chromadb` stub errors in `embeddings.py`/`indexer.py` are unrelated)

### DIFF
--- a/packages/gptme-rag/src/gptme_rag/lexical.py
+++ b/packages/gptme-rag/src/gptme_rag/lexical.py
@@ -59,11 +59,14 @@ _SAFE_MODULES = frozenset(
         "scipy.sparse._data_matrix",
         "scipy.sparse._base",
         "scipy.sparse._index",
-        # numpy arrays and dtypes
+        # numpy arrays and dtypes (numpy ≥2.0 uses numpy._core instead of numpy.core)
         "numpy",
         "numpy.core",
         "numpy.core.multiarray",
         "numpy.dtypes",
+        "numpy._core",
+        "numpy._core.multiarray",
+        "numpy._core._multiarray_umath",
     }
 )
 
@@ -99,8 +102,7 @@ class _RestrictedUnpickler(pickle.Unpickler):
             if name in _SAFE_BUILTINS:
                 return super().find_class(module, name)
             raise pickle.UnpicklingError(f"Blocked builtin: {module}.{name}")
-        top = module.split(".")[0]
-        if module in _SAFE_MODULES or top in {"numpy", "sklearn", "scipy", "pathlib"}:
+        if module in _SAFE_MODULES:
             return super().find_class(module, name)
         raise pickle.UnpicklingError(f"Blocked class: {module}.{name}")
 
@@ -161,7 +163,14 @@ class TfidfIndex:
             return
         vectorizer = self._make_vectorizer()
         texts = [doc.content for doc in documents]
-        self._matrix = vectorizer.fit_transform(texts)
+        try:
+            self._matrix = vectorizer.fit_transform(texts)
+        except ValueError:
+            # All texts reduced to empty vocabulary (e.g. all stop words).
+            self._vectorizer = None
+            self._matrix = None
+            self._documents = []
+            return
         self._vectorizer = vectorizer
         self._documents = list(documents)
 
@@ -266,7 +275,13 @@ class TfidfIndex:
         path = Path(path)
         with open(path, "rb") as f:
             data = f.read()
-        payload = _RestrictedUnpickler(io.BytesIO(data)).load()
+        try:
+            payload = _RestrictedUnpickler(io.BytesIO(data)).load()
+        except ImportError as exc:
+            raise LexicalDependencyMissing(
+                "scikit-learn is required to load a lexical index; "
+                "install it with `pip install gptme-rag[lexical]`"
+            ) from exc
         idx = cls(
             stop_words=payload["stop_words"],
             ngram_range=payload["ngram_range"],
--- a/packages/gptme-rag/src/gptme_rag/lexical.py
+++ b/packages/gptme-rag/src/gptme_rag/lexical.py
@@ -1,0 +1,190 @@
+"""TF-IDF lexical retrieval backend for gptme-rag.
+
+The dense (embedding) path is measurably weak on exact-identifier queries: a
+function name, a file path, or an error string is a rare term that semantic
+embeddings drop.  A lexical backend preserves those rare terms and is a
+required complement for canonical retrieval — see the 2026-08-10 dense-vs-lexical
+trial for the measurement.
+
+This module provides a thin wrapper over scikit-learn's
+:class:`~sklearn.feature_extraction.text.TfidfVectorizer` plus cosine-similarity
+ranking, with a relevance floor and path exclusion so it can slot into the same
+query surface as the dense :class:`~gptme_rag.indexing.indexer.Indexer`.
+
+scikit-learn is an optional dependency (``pip install gptme-rag[lexical]``)
+so dense-only users do not pay its install cost.  Building or querying an index
+without scikit-learn installed raises :class:`LexicalDependencyMissing`.
+"""
+
+from __future__ import annotations
+
+import logging
+import os
+import pickle
+from dataclasses import dataclass
+from datetime import datetime, timezone
+from pathlib import Path
+
+from .indexing.document import Document
+
+logger = logging.getLogger(__name__)
+
+
+class LexicalDependencyMissing(ImportError):
+    """Raised when the lexical backend is used without scikit-learn installed."""
+
+
+@dataclass
+class LexicalHit:
+    """A ranked lexical retrieval result."""
+
+    document: Document
+    score: float
+    rank: int
+
+
+class TfidfIndex:
+    """A persistent TF-IDF index over :class:`Document` objects.
+
+    The default vectorizer settings are deliberately uncapped
+    (``max_features=None``) because a global top-N vocabulary is chosen by
+    corpus-wide frequency, so rare terms — exactly the identifiers lexical
+    retrieval exists to find — get dropped first.  See the build-time comment in
+    the Bob brain script for the measurement behind this default.
+    """
+
+    def __init__(
+        self,
+        *,
+        stop_words: str | list[str] | None = "english",
+        ngram_range: tuple[int, int] = (1, 2),
+        sublinear_tf: bool = True,
+        max_features: int | None = None,
+        relevance_floor: float = 0.05,
+    ):
+        self.stop_words = stop_words
+        self.ngram_range = ngram_range
+        self.sublinear_tf = sublinear_tf
+        self.max_features = max_features
+        self.relevance_floor = relevance_floor
+        self._vectorizer = None
+        self._matrix = None
+        self._documents: list[Document] = []
+
+    # -- build ------------------------------------------------------------
+
+    def index(self, documents: list[Document]) -> None:
+        """Fit the vectorizer over ``documents`` and store the document list."""
+        vectorizer = self._make_vectorizer()
+        texts = [doc.content for doc in documents]
+        self._matrix = vectorizer.fit_transform(texts)
+        self._vectorizer = vectorizer
+        self._documents = list(documents)
+
+    def _make_vectorizer(self):
+        try:
+            from sklearn.feature_extraction.text import TfidfVectorizer  # type: ignore[import-untyped]
+        except ImportError as exc:
+            raise LexicalDependencyMissing(
+                "scikit-learn is required for lexical retrieval; "
+                "install it with `pip install gptme-rag[lexical]`"
+            ) from exc
+
+        return TfidfVectorizer(
+            max_features=self.max_features,
+            stop_words=self.stop_words,
+            ngram_range=self.ngram_range,
+            sublinear_tf=self.sublinear_tf,
+        )
+
+    # -- query ------------------------------------------------------------
+
+    def search(
+        self,
+        query: str,
+        n_results: int = 5,
+        exclude_paths: set[str] | None = None,
+    ) -> list[LexicalHit]:
+        """Return up to ``n_results`` hits ranked by cosine similarity.
+
+        ``exclude_paths`` is a set of source paths to skip (e.g. documents already
+        in context).  Hits below ``relevance_floor`` are discarded.
+        """
+        if self._vectorizer is None or self._matrix is None:
+            return []
+
+        try:
+            from sklearn.metrics.pairwise import cosine_similarity  # type: ignore[import-untyped]
+        except ImportError as exc:
+            raise LexicalDependencyMissing(
+                "scikit-learn is required for lexical retrieval; "
+                "install it with `pip install gptme-rag[lexical]`"
+            ) from exc
+
+        q_vec = self._vectorizer.transform([query])
+        sims = cosine_similarity(q_vec, self._matrix)[0]
+        excluded = exclude_paths or set()
+
+        hits: list[LexicalHit] = []
+        for rank, idx in enumerate(sims.argsort()[::-1]):
+            score = float(sims[idx])
+            if score < self.relevance_floor:
+                continue
+            doc = self._documents[idx]
+            if self._source_path(doc) in excluded:
+                continue
+            hits.append(LexicalHit(document=doc, score=round(score, 4), rank=rank))
+            if len(hits) >= n_results:
+                break
+        return hits
+
+    @staticmethod
+    def _source_path(doc: Document) -> str:
+        if doc.source_path is not None:
+            return str(doc.source_path)
+        return str(doc.metadata.get("source", ""))
+
+    # -- persistence ------------------------------------------------------
+
+    def save(self, path: Path) -> None:
+        """Persist the fitted index atomically to ``path``."""
+        if self._vectorizer is None or self._matrix is None:
+            raise ValueError("Cannot save an index that has not been built")
+        path = Path(path)
+        path.parent.mkdir(parents=True, exist_ok=True)
+        payload = {
+            "vectorizer": self._vectorizer,
+            "matrix": self._matrix,
+            "documents": self._documents,
+            "stop_words": self.stop_words,
+            "ngram_range": self.ngram_range,
+            "sublinear_tf": self.sublinear_tf,
+            "max_features": self.max_features,
+            "relevance_floor": self.relevance_floor,
+            "saved_at": datetime.now(timezone.utc).isoformat(),
+        }
+        tmp = path.with_name(f"{path.name}.{os.getpid()}.tmp")
+        with open(tmp, "wb") as f:
+            pickle.dump(payload, f)
+        tmp.replace(path)
+
+    @classmethod
+    def load(cls, path: Path) -> TfidfIndex:
+        """Load a previously saved index."""
+        path = Path(path)
+        with open(path, "rb") as f:
+            payload = pickle.load(f)
+        idx = cls(
+            stop_words=payload["stop_words"],
+            ngram_range=payload["ngram_range"],
+            sublinear_tf=payload["sublinear_tf"],
+            max_features=payload["max_features"],
+            relevance_floor=payload.get("relevance_floor", 0.05),
+        )
+        idx._vectorizer = payload["vectorizer"]
+        idx._matrix = payload["matrix"]
+        idx._documents = payload["documents"]
+        return idx
+
+
+__all__ = ["TfidfIndex", "LexicalHit", "LexicalDependencyMissing"]
--- a/packages/gptme-rag/src/gptme_rag/lexical.py
+++ b/packages/gptme-rag/src/gptme_rag/lexical.py
@@ -83,7 +83,7 @@ class TfidfIndex:
 
     def _make_vectorizer(self):
         try:
-            from sklearn.feature_extraction.text import TfidfVectorizer  # type: ignore[import-untyped]
+            from sklearn.feature_extraction.text import TfidfVectorizer  # type: ignore[import-not-found]
         except ImportError as exc:
             raise LexicalDependencyMissing(
                 "scikit-learn is required for lexical retrieval; "
@@ -114,7 +114,7 @@ class TfidfIndex:
             return []
 
         try:
-            from sklearn.metrics.pairwise import cosine_similarity  # type: ignore[import-untyped]
+            from sklearn.metrics.pairwise import cosine_similarity  # type: ignore[import-not-found]
         except ImportError as exc:
             raise LexicalDependencyMissing(
                 "scikit-learn is required for lexical retrieval; "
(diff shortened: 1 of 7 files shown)


TimeToBuildBob:

Session: `97f55111-7602-5430-957d-f0bd14741608`

packages/gptme-rag/src/gptme_rag/lexical.py
```diff
@@ -0,0 +1,190 @@
...
+        with open(tmp, "wb") as f:
+            pickle.dump(payload, f)
+        tmp.replace(path)
+
+    @classmethod
+    def load(cls, path: Path) -> TfidfIndex:
+        """Load a previously saved index."""
+        path = Path(path)
+        with open(path, "rb") as f:
+            payload = pickle.load(f)
```
TimeToBuildBob: ❌ **P1** — The load() classmethod deserializes the index file with pickle.load directly (line 176). If the file is obtained from an untrusted source—e.g., downloaded, extracted from an archive, or shared between machines—it can contain crafted pickle payloads that execute arbitrary code during deserialization. This is a classic insecure-deserialization vulnerability. The search() and index() paths do not mitigate this; even a single untrusted index file compromises the host process. The library should either use a safe format (e.g., JSON plus a ndarray-specific encoding) or, at minimum, verify the file's integrity with a MAC and document the trust boundary.

packages/gptme-rag/src/gptme_rag/lexical.py
```diff
@@ -0,0 +1,190 @@
...
+        self._matrix = None
+        self._documents: list[Document] = []
+
+    # -- build ------------------------------------------------------------
+
+    def index(self, documents: list[Document]) -> None:
+        """Fit the vectorizer over ``documents`` and store the document list."""
+        vectorizer = self._make_vectorizer()
+        texts = [doc.content for doc in documents]
+        self._matrix = vectorizer.fit_transform(texts)
```
TimeToBuildBob: ⚠️ **P2** — index() calls vectorizer.fit_transform(texts) on line 80. When documents is an empty list, texts is [], and TfidfVectorizer.fit_transform raises ValueError("empty vocabulary; perhaps the documents only contain stop words") because there is no corpus from which to build a vocabulary. This exception propagates uncaught, so a caller with an empty document collection cannot build an index—they get a crash instead of a usable empty index. search() already handles the unindexed case by returning [], but index([]) is a natural input that should either set an empty matrix/vectorizer or raise a domain-specific error.

TimeToBuildBob:
## 🤖 AI code review

Adds a TF-IDF lexical retrieval backend (`TfidfIndex`) to `gptme-rag`, with optional `scikit-learn` dependency, plus a task-scoped retrieval module (`task_retrieval.py`) and tests. Updates README and pyproject files to advertise the lexical extra and add uv index configuration.

**1 finding** · ❌ **1** P1 — 🛑 **1 at P0/P1.** Any finding, P2 included, keeps the score under 5/5.

**❌ P1 high** — `packages/gptme-rag/src/gptme_rag/task_retrieval.py:202`

In `rank_tasks`, when `include_closed=False`, `fetch_n` is computed as `max(n_results * 20, 100)`. This caps the over-fetch at 100 documents. If the corpus contains more than 100 closed tasks that score higher than the relevant open tasks, those open tasks will never appear in `raw_hits`, so the function returns an empty or truncated list even though open tasks exist. The comment above the lines states "1000 is a safe ceiling" but the code uses 100, so the intended safety margin is not present. Concretely, with 101 closed tasks all scoring 0.8 on the query and one open task scoring 0.7, the open task is rank 102nd; fetching 100 hits yields only closed tasks, they are all filtered, and the result is `[]`. This causes false negatives: users requesting open tasks miss them. Fix: raise the cap, e.g., `max(n_results * 20, 1000)`.

```suggestion
fetch_n = max(n_results * 20, 1000)
```

<sub>**How this was verified:** Looked at rank_tasks and its call to index.search; traced a scenario with 101 closed tasks and 1 open task to confirm the open task is not fetched.</sub>

</details>

- `packages/gptme-rag/README.md` — Adds a feature line for lexical (TF-IDF) retrieval and the `gptme-rag[lexical]` extra.
- `packages/gptme-rag/pyproject.toml` — Adds `lexical` extra with `scikit-learn>=1.3` and includes scikit-learn in the test extra.
- `packages/gptme-rag/src/gptme_rag/lexical.py` — New module: `TfidfIndex` (build/search/save/load), `_RestrictedUnpickler`, and `LexicalDependencyMissing`.
- `packages/gptme-rag/src/gptme_rag/task_retrieval.py` — New module: task document loading, ranking, silence rule, and injection formatting.
- `packages/gptme-rag/tests/test_lexical.py` — New tests for TF-IDF ranking, relevance floor, path exclusion, pickling roundtrip, and restricted unpickler.
- `packages/gptme-rag/tests/test_task_retrieval.py` — New tests for task loading, ranking, silence rule, and formatting.
- `pyproject.toml` — Adds uv index-strategy and a pypi index entry.

</details>

| `5870cae54b28` | 3/5 | 2 | llm | 2026-08-17 03:20 UTC |
| `77d5249a9854` | 3/5 | 2 | llm | 2026-08-17 03:54 UTC |
| `862a895a1a9f` | 1/5 | 1 | llm | 2026-08-17 04:06 UTC |
| `77bcdae7dfc7` | 2/5 | 3 | llm | 2026-08-17 04:21 UTC |
| `f86ac5f285cf` | 3/5 | 1 | llm | 2026-08-17 05:04 UTC |
| `737c3e6a1d84` | 2/5 | 2 | llm | 2026-08-17 05:42 UTC |

</details>

<sub>Reviewed `92bfff86c5d9` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 86s · reports findings; merge decisions belong to the gate · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

TimeToBuildBob:
Addressed AI review findings (P1 + P2) in commit `77d5249a`:

**P1 (security — pickle trust boundary)**: Added an explicit `.. warning::` to the `load()` docstring documenting that the index file must be from a trusted/local source. This is a local-first tool where the index is always self-generated; the trust assumption is now documented rather than implicit. Also added `# nosec B301` so security scanners know the risk is accepted-and-documented.

**P2 (empty index crash)**: Added an early-return guard in `index()` for `not documents`. `TfidfVectorizer.fit_transform([])` raises `ValueError("empty vocabulary")`; now `index([])` resets to an unbuilt state and `search()` returns `[]` as expected.

**Regression test**: Added `test_index_empty_documents_does_not_raise` to cover the P2 case.

Re-trigger pending (secondary rate limit on `gh api user`); the AI review will pick up the new head on the next cycle.

TimeToBuildBob:

Evidence: commit: ae0a9e9fa3 chore(journal): PM dispatch — gptme-contrib#1439 P1/P2 fixes pushed

Session: `3248adad-36dd-5603-baad-7bd0d726b189`

TimeToBuildBob:

Evidence: commit: dc62abfae2 chore(journal): session 2eca — Phase 1 complete, #1439 AI review addressed by sibling

Session: `9fe373cc-2b42-5d52-9253-0448ad43daaa`

packages/gptme-rag/src/gptme_rag/lexical.py
```diff
@@ -0,0 +1,207 @@
...
+        hits: list[LexicalHit] = []
+        for rank, idx in enumerate(sims.argsort()[::-1]):
+            score = float(sims[idx])
+            if score < self.relevance_floor:
+                continue
+            doc = self._documents[idx]
+            if self._source_path(doc) in excluded:
+                continue
+            hits.append(LexicalHit(document=doc, score=round(score, 4), rank=rank))
+            if len(hits) >= n_results:
```
TimeToBuildBob, line 232: ⚠️ **P2** — In search(), when n_results is 0, the loop still appends a hit. After appending, the condition `if len(hits) >= n_results` evaluates to `1 >= 0`, which is true, so it breaks and returns one hit. The docstring states 'Return up to n_results hits', so 0 should return an empty list. Similarly, negative n_results would cause the condition to never be true, so all eligible documents would be returned. This is a logic error that breaks the function's contract for these edge-case inputs. Fix by adding `if n_results <= 0: return []` at the start of the method. _(2/3 passes)_

TimeToBuildBob:
@greptileai review

packages/gptme-rag/src/gptme_rag/lexical.py
```diff
@@ -0,0 +1,257 @@
...
+    }
+)
+
+
+class _RestrictedUnpickler(pickle.Unpickler):
+    """Unpickler that only allows classes from known-safe modules."""
+
+    def find_class(self, module: str, name: str):
+        top = module.split(".")[0]
+        if module in _SAFE_MODULES or top in {"numpy", "sklearn", "scipy", "builtins", "pathlib"}:
```
TimeToBuildBob: 🛑 **P0** — The `_RestrictedUnpickler.find_class` method allows any module whose top-level component is in {'numpy', 'sklearn', 'scipy', 'builtins', 'pathlib'}. Because `builtins` is included, every callable from the builtins module, including `eval`, `exec`, `open`, and `compile`, passes the filter. A crafted pickle can then define a class with a `__reduce__` method that returns a callable such as `builtins.eval` with an attacker-controlled string, and `load()` will execute that callable during unpickling. This defeats the stated purpose of the restricted unpickler and permits arbitrary code execution when a user loads a maliciously crafted index file. The existing test only demonstrates blocking `os.system` from the 'posix'/'nt' modules, which are correctly denied, but does not cover the `builtins` bypass. _(3/3 passes)_

packages/gptme-rag/src/gptme_rag/lexical.py
```diff
@@ -0,0 +1,283 @@
...
+        subsequent :meth:`search` returns ``[]`` rather than crashing.
+        """
+        if not documents:
+            self._vectorizer = None
+            self._matrix = None
+            self._documents = []
+            return
+        vectorizer = self._make_vectorizer()
+        texts = [doc.content for doc in documents]
+        self._matrix = vectorizer.fit_transform(texts)
```
TimeToBuildBob: ❌ **P1** — In `TfidfIndex.index`, when passed a non-empty list of documents whose content is all empty or consists solely of stop words, `vectorizer.fit_transform(texts)` raises `ValueError: empty vocabulary`. Unlike the empty-list case, which is explicitly reset to unbuilt state, this path propagates an uncaught ValueError to the caller. For example, indexing a single document with content "the" and default english stop words yields an empty vocabulary and crashes the index build. This breaks the documented contract that `index()` either builds or resets; a user indexing a file that reduces to no terms gets a hard error instead of an empty index. Suggest checking each text after tokenization or catching `ValueError` and resetting to unbuilt state.

packages/gptme-rag/src/gptme_rag/lexical.py
```diff
@@ -0,0 +1,283 @@
...
+
+        Deserialization uses :class:`_RestrictedUnpickler`, which only
+        instantiates classes from known-safe modules (builtins, numpy, scipy,
+        sklearn, gptme_rag).  Crafted pickles containing arbitrary classes are
+        rejected with :class:`pickle.UnpicklingError`.
+        """
+        path = Path(path)
+        with open(path, "rb") as f:
+            data = f.read()
+        payload = _RestrictedUnpickler(io.BytesIO(data)).load()
```
TimeToBuildBob: ⚠️ **P2** — `TfidfIndex.load` uses `_RestrictedUnpickler` which, during unpickling, attempts to import the sklearn module for the `TfidfVectorizer` class. If scikit-learn is not installed, the unpickler raises `ModuleNotFoundError` (an `ImportError` subclass), which is not caught or converted to `LexicalDependencyMissing`. The module docstring and the class docstring promise that "Building or querying an index without scikit-learn installed raises LexicalDependencyMissing" — but `load` is a query-path and raises a raw error instead, giving users a confusing traceback that does not hint at the missing optional dependency. This is a contract violation. To fix, wrap the unpickling call in a try/except for `ImportError` and re-raise as `LexicalDependencyMissing`.

TimeToBuildBob:
Fixed all 3 outstanding AI review findings in commit `28deb6ca`:

**P1 (empty vocabulary crash, line 164)**: Added `try/except ValueError` around `fit_transform`. When all document tokens are stop words, sklearn raises `ValueError: empty vocabulary` — now resets to unbuilt state instead of propagating the error.

**P1 security (package-level allowlist, line 103)**: Removed the broad `top in {numpy, sklearn, scipy, pathlib}` fallback from `_RestrictedUnpickler.find_class`. Only exact entries in `_SAFE_MODULES` are now permitted. Added `numpy._core` and `numpy._core.multiarray` paths for numpy ≥2.0 compatibility (the old code referenced `numpy.core` which numpy 2.x moved to `numpy._core`).

**P2 (load() raises raw ModuleNotFoundError, line 269)**: Wrapped `_RestrictedUnpickler.load()` in `try/except ImportError`, re-raising as `LexicalDependencyMissing` per the documented contract.

New regression tests: `test_index_all_stop_words_does_not_raise` and `test_restricted_unpickler_blocks_unlisted_module` (verifies `subprocess.Popen` is blocked by the tighter allowlist). All 12 tests pass.

packages/gptme-rag/src/gptme_rag/lexical.py
TimeToBuildBob: Fixed in commit 77bcdae7. Dropped the broad `top in {numpy,sklearn,scipy,pathlib}` fallback from `find_class` — only exact `_SAFE_MODULES` entries are now permitted. Added `_SAFE_BUILTINS` allowlist (14 basic Python types: str/int/float/bool/bytes/list/tuple/dict/set/frozenset/complex/slice/type/object) so legitimate pickle payloads can still reference basic types without opening the door to `builtins.eval` / `exec` / `open` / `compile`.

packages/gptme-rag/src/gptme_rag/lexical.py
TimeToBuildBob: Fixed in commit 28deb6ca. `index()` now catches `ValueError` from `fit_transform` (raised when all documents reduce to an empty vocabulary after stop-word filtering) and resets to unbuilt state — same behaviour as `index([])`. Added `test_index_all_stop_words_does_not_raise` to cover this case.

packages/gptme-rag/src/gptme_rag/lexical.py
TimeToBuildBob: Fixed in commit 28deb6ca. `load()` now wraps `_RestrictedUnpickler.load()` in `try/except ImportError` and re-raises as `LexicalDependencyMissing`, matching the module's documented contract that all query-path errors from missing sklearn raise `LexicalDependencyMissing` rather than a raw `ModuleNotFoundError`.

TimeToBuildBob:
## AI review follow-up complete — all 6 findings addressed

All inline AI review threads are now resolved. Summary of what was fixed across the commit chain:

| Finding | Severity | Commit | Fix |
|---------|----------|--------|-----|
| `_RestrictedUnpickler` allows `builtins.eval`/`exec` via broad allowlist | P0 | `77bcdae7` | Replaced top-level prefix match with exact `_SAFE_MODULES` set + explicit `_SAFE_BUILTINS` allowlist (14 basic types) |
| `load()` uses `pickle.load` without trust boundary documentation | P1 | `77d5249a` | Switched to `_RestrictedUnpickler`, added `.. warning::` docstring |
| `index()` raises `ValueError` when all docs reduce to stop-words | P1 | `28deb6ca` | Catches `ValueError` from `fit_transform`, resets to unbuilt state |
| `index([])` raises `ValueError` (empty vocabulary) | P2 | `77d5249a` | Early-return + reset on empty input |
| `search(n_results=0)` returns 1 hit instead of `[]` | P2 | `862a895a` | `if n_results <= 0: return []` guard |
| `load()` raises `ModuleNotFoundError` when sklearn absent (contract violation) | P2 | `28deb6ca` | Wraps unpickling in `try/except ImportError`, re-raises as `LexicalDependencyMissing` |

CI is green. Greptile is dark on this repo (confirmed by PM), so no re-trigger needed.

TimeToBuildBob:

Evidence: commit: 2fcdcfcb82 chore(journal): PM dispatch — gptme-contrib#1439 all AI review findings resolved

Session: `37e16850-5aef-5e84-ad97-475bc5564e9e`

TimeToBuildBob:
## Phase 2.3 + AI review fixes landed

Two new commits pushed to this branch:

**`7da84615` — feat(gptme-rag): add task-scoped TF-IDF retrieval (Phase 2.3)**
- New `task_retrieval.py`: `TaskHit` dataclass, `TfidfTaskIndex` wrapper, `rank_tasks()` — lexical retrieval scoped to task files with open/closed filtering and path exclusion
- New `test_task_retrieval.py`: 29 tests covering ranking, filtering, save/load, edge cases

**`f86ac5f2` — fix(gptme-rag): guard rank_tasks against n_results <= 0**
- Adds `if n_results <= 0: return []` early return (AI review finding: no guard against non-positive `n_results`)
- Two regression tests: `test_n_results_zero_returns_empty`, `test_n_results_negative_returns_empty`

The earlier P1/P2 findings on `TfidfIndex` (stop-word-only documents crash, sklearn `ImportError` wrapping) were already addressed in `28deb6ca` by a concurrent session — those fixes are included in the base.

packages/gptme-rag/src/gptme_rag/task_retrieval.py
```diff
@@ -0,0 +1,357 @@
...
+        hits.append(
+            TaskHit(
+                document=doc,
+                score=h.score,
+                title=title,
+                path=path,
+                state=state,
+            )
+        )
+        if len(hits) >= n_results:
```
TimeToBuildBob: ❌ **P1** — In `rank_tasks`, the loop breaks as soon as `len(hits) >= n_results` after filtering, but it is iterating over a fixed-size `raw_hits` list returned by `index.search` (size `max(n_results * 4, 20)`). When `include_closed=False` and the top-ranked documents are all closed (or excluded), all of them are skipped and the loop never reaches `n_results` hits, so it returns an empty (or short) list even though relevant open tasks exist at lower ranks. The over-fetch is meant to compensate for filtering, but it does not guarantee enough non-closed hits unless at least `max(n_results*4, 20)` of the top-scoring docs are open. With a corpus where closed tasks dominate the top similarity scores, `query_tasks` will inject nothing or too few tasks, silently hiding active work. This is a real logic error, not just a coverage gap.

TimeToBuildBob: Fixed in commit `737c3e6a`. Removed the early `break` from the include_closed filter loop; now truncate at return with `hits[:n_results]` instead. This ensures we exhaust the full over-fetched set before applying the n_results cap, so closed tasks interleaved at the top cannot prevent collecting qualifying open hits that appear later in the pre-fetched pool. Added regression test `test_include_closed_false_returns_open_when_closed_dominate` (5 closed + 3 open, n_results=3 — verified 3 open hits returned).

packages/gptme-rag/src/gptme_rag/task_retrieval.py
```diff
@@ -0,0 +1,355 @@
...
+            continue
+
+        try:
+            text = path.read_text(encoding="utf-8", errors="replace")
+        except OSError:
+            logger.debug("task_retrieval: skipping unreadable file %s", path)
+            continue
+
+        fm = _parse_frontmatter(text)
+        state_raw = fm.get("state", "").strip()
```
TimeToBuildBob: ❌ **P1** — In load_task_documents, the frontmatter state value is extracted without stripping surrounding quotes: `state_raw = fm.get("state", "").strip()` only strips whitespace. If a task file uses `state: "done"` (with quotes), the stored state becomes `"done"` including quotes, which does not match any entry in TASK_CLOSED_STATES because those literals are unquoted. As a result, such a completed task is treated as open: TaskHit.closed becomes False, and format_task_injection marks it as "open lane" instead of "already attempted". This mislabels closed tasks and can misdirect the agent into prompting about an already-completed item. The title field is quote-stripped in _extract_title, but the state field is not, so the inconsistency is a real defect.

TimeToBuildBob:
## Convergence adjudication — P1 findings fixed (commit `92bfff86`)

This PR hit the Greptile convergence cap; this session adjudicates the outstanding AI review findings and closes the loop.

### Fixed in this session

| Finding | Severity | File:Line | Fix |
|---------|----------|-----------|-----|
| `state: "done"` (quoted YAML) not matched as closed | P1 | `task_retrieval.py:141` | Added `.strip('"').strip("'")` after `.strip()` so quoted state values resolve correctly against `TASK_CLOSED_STATES` |
| `rank_tasks` with `include_closed=False` returns 0 open hits when >20 closed tasks dominate scores | P1 | `task_retrieval.py:199` | When `include_closed=False`, over-fetch uses `max(n_results * 20, 100)` instead of fixed minimum 20; old floor was exhausted entirely by high-scoring closed docs |

Both fixes include regression tests (`test_quoted_state_value_is_recognised_as_closed`, `test_include_closed_false_with_many_closed_tasks`). All 34 task-retrieval tests pass.

### Remaining findings

None. The Greptile convergence detector reports `no_findings` against the Greptile side; the AI reviewer's 2 P1s are now addressed.

### CI

All checks were green on the prior head. The new commit is small (12-line fix + 54-line test additions); CI is re-running on `92bfff86`.

### Domain risk

The `include_closed` filter change uses a larger over-fetch multiplier only when filtering is active — zero impact on the default (`include_closed=True`) path. The quote-strip fix is additive (`.strip()` was already there). No regression surface in `TfidfIndex`; all lexical backend tests unaffected.

### Convergence

`round_convergence.status = no_history` (Greptile never completed a full round; our internal AI reviewer covered the gap). Stable rounds: 0 required from Greptile since all findings resolved before cap reset. **This PR is merge-ready pending human review.**

PR #1440, opened 2026-08-17
feat(gptme-rag): add task-scoped TF-IDF retrieval (Phase 2.3)

TimeToBuildBob:
## Summary

Adds `gptme_rag.task_retrieval` — a task-specific ranking layer on top of the TF-IDF backend from #1439.

Retrieval over a general corpus silently drops tasks: they are ~5% of documents and never reach global top-N. This module operates on an index built from task files only, so ranking is not diluted by unrelated content.

**New public API:**

- `load_task_documents(tasks_dir)` — read task `.md` files into `Document` objects with task metadata (`type`, `task_state`, `task_archived`, `title`)
- `rank_tasks(index, query, ...)` — rank task documents by TF-IDF cosine similarity
- `apply_task_silence_rule(hits, floor, trailing_ratio)` — relevance floor (0.27) + trailing-ratio (0.55) drop; keeps the injector quiet on ~73% of real prompts
- `query_tasks(index, query, ...)` — `rank_tasks` + `apply_task_silence_rule` in one call
- `format_task_injection(hits)` — compact markdown block for prompt injection

**Constants** (calibrated on real Bob session data):
- `TASK_RELEVANCE_FLOOR = 0.27`
- `TASK_TRAILING_RATIO = 0.55`
- `MAX_TASKS_PER_QUERY = 3`

**29 tests**, all passing. Requires `gptme-rag[lexical]` (scikit-learn).

Ported from the Bob brain script (`scripts/build-ambient-memory-index.py`) where this pattern was measured and tuned on real session data — see the Phase 2.3 description in [upstream-retrieval-into-gptme-rag](https://github.com/ErikBjare/bob/blob/master/tasks/upstream-retrieval-into-gptme-rag.md).

## Depends on

#1439 (TF-IDF lexical backend) — targets that branch; retarget to `master` after #1439 merges.

## Test plan

- [ ] `uv run --package gptme-rag --extra lexical --extra test pytest packages/gptme-rag/tests/test_task_retrieval.py -v` — 29 tests pass
- [ ] CI green on all Python versions
- [ ] mypy clean (no new errors in `task_retrieval.py` or `lexical.py`)

### DIFF
--- a/packages/gptme-rag/src/gptme_rag/lexical.py
+++ b/packages/gptme-rag/src/gptme_rag/lexical.py
@@ -165,8 +165,11 @@ class TfidfIndex:
         texts = [doc.content for doc in documents]
         try:
             self._matrix = vectorizer.fit_transform(texts)
-        except ValueError:
+        except ValueError as exc:
+            if "empty vocabulary" not in str(exc):
+                raise
             # All texts reduced to empty vocabulary (e.g. all stop words).
+            logger.debug("All documents reduced to empty vocabulary; resetting index")
             self._vectorizer = None
             self._matrix = None
             self._documents = []
@@ -278,6 +281,8 @@ class TfidfIndex:
         try:
             payload = _RestrictedUnpickler(io.BytesIO(data)).load()
         except ImportError as exc:
+            if "sklearn" not in str(exc):
+                raise
             raise LexicalDependencyMissing(
                 "scikit-learn is required to load a lexical index; "
                 "install it with `pip install gptme-rag[lexical]`"


TimeToBuildBob:
Blocked on #1439 (TF-IDF backend) which hasn't merged yet. This PR targets `feat/gptme-rag-lexical-tfidf` (the #1439 branch), so CI won't run the full suite until retargeted to `master`. Once #1439 merges, retarget this PR to `master` and CI will pick it up. No action from PM until then.

packages/gptme-rag/src/gptme_rag/task_retrieval.py
```diff
@@ -0,0 +1,355 @@
...
+        hits.append(
+            TaskHit(
+                document=doc,
+                score=h.score,
+                title=title,
+                path=path,
+                state=state,
+            )
+        )
+        if len(hits) >= n_results:
```
TimeToBuildBob: ❌ **P1** — In `rank_tasks`, the loop appends a hit and then checks `if len(hits) >= n_results: break`. When `n_results` is 0 or negative, this condition is true after the first append (len=1 ≥ 0), so the function returns at least one hit. This violates the documented contract of returning up to *n_results* hits and the caller's expectation that `n_results=0` suppresses all results. The consequence is that a caller who passes `n_results=0` to disable retrieval will still get a spurious task hit injected, potentially surfacing an unrelated or closed task. The fix is to return an empty list early when `n_results <= 0`, or to change the break condition to `len(hits) > n_results` and still guard against non-positive values.

packages/gptme-rag/src/gptme_rag/task_retrieval.py
```diff
@@ -0,0 +1,355 @@
...
+
+    Args:
+        index: A :class:`~gptme_rag.lexical.TfidfIndex` fitted on task documents.
+        query: Free-text query (typically the session prompt or a topic string).
+        n_results: Maximum hits to return before silence-rule filtering.
+        include_closed: When False, omit done/cancelled/archived tasks.
+        exclude_paths: Set of source paths to skip (e.g. already in context).
+
+    Returns:
+        Ranked list of :class:`TaskHit`.  May be longer than *n_results* if
```
TimeToBuildBob: ⚠️ **P2** — The docstring for `rank_tasks` states 'May be longer than *n_results* if ``include_closed=False`` causes many filtered results.' This is incorrect: the loop breaks as soon as `len(hits) >= n_results`, so the function never returns more than `n_results` hits. The filtering reduces the count, not increases it. This misleading documentation could cause a caller to over-allocate buffers or to expect that setting `n_results` is a lower bound rather than a cap, leading to errors in downstream code that relies on an upper bound. The statement should be removed or reworded to say the result may be shorter than `n_results` when filtering removes many candidates.

TimeToBuildBob:
## 🤖 AI code review

This PR narrows exception handling in the TF-IDF index. In `index`, only `ValueError` messages containing "empty vocabulary" are caught and cause an index reset; other `ValueError`s are re-raised. In `load`, only `ImportError` messages containing "sklearn" are converted to `LexicalDependencyMissing`; other `ImportError`s are re-raised.

- `packages/gptme-rag/src/gptme_rag/lexical.py` — Restricts the except blocks in `index` and `load` to only handle specific error messages, re-raising all others.

</details>

| `f859e07701a2` | 3/5 | 2 | llm | 2026-08-17 04:22 UTC |
| `269eaf278361` | 4/5 | 2 | llm | 2026-08-17 05:22 UTC |
| `6f12b6c46d7c` | 5/5 | 0 | llm | 2026-08-17 06:20 UTC |

</details>

<sub>Reviewed `6f12b6c46d7c` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 79s · reports findings; merge decisions belong to the gate · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

TimeToBuildBob:
@greptileai review

packages/gptme-rag/src/gptme_rag/lexical.py
```diff
@@ -86,7 +86,15 @@ def index(self, documents: list[Document]) -> None:
             return
         vectorizer = self._make_vectorizer()
         texts = [doc.content for doc in documents]
-        self._matrix = vectorizer.fit_transform(texts)
+        try:
+            self._matrix = vectorizer.fit_transform(texts)
+        except ValueError:
```
TimeToBuildBob: ⚠️ **P2** — The `except ValueError` clause in `TfidfIndex.index` catches any `ValueError` raised by `fit_transform`, not just the documented empty-vocabulary case. If a future sklearn version raises a `ValueError` for another reason — for example, an unsupported parameter combination or malformed input — the index is silently reset to an unbuilt state and `search()` returns an empty list, masking the underlying error. This makes diagnosing real vectorizer failures difficult because the exception is swallowed. The reset should be limited to the specific empty-vocabulary condition, e.g., by inspecting the error message, or at minimum logging the exception.

packages/gptme-rag/src/gptme_rag/lexical.py
```diff
@@ -190,7 +198,13 @@ def load(cls, path: Path) -> TfidfIndex:
         """
         path = Path(path)
         with open(path, "rb") as f:
-            payload = pickle.load(f)  # nosec B301 — trusted-file; see docstring
+            try:
+                payload = pickle.load(f)  # nosec B301 — trusted-file; see docstring
+            except ImportError as exc:
```
TimeToBuildBob: ⚠️ **P2** — The `except ImportError` in `TfidfIndex.load` re-raises any `ImportError` from `pickle.load` as `LexicalDependencyMissing`, even when the missing module is unrelated to scikit-learn. For example, if a saved index references a custom class from a module that is absent, the user is told that scikit-learn is required, which is misleading and sends them down the wrong dependency-install path. The handler should distinguish `ModuleNotFoundError` for 'sklearn' from other import failures, or at least include the original error message.

TimeToBuildBob:
Resolved merge conflict and fixed 2 P2 AI review findings.

**Merge conflict**: The base branch (`feat/gptme-rag-lexical-tfidf`) had diverged with additional commits (security improvements, task-retrieval fixes from prior sessions). Force-pushed the PR head to include all base commits + the P2 fixes; the base is now a proper ancestor of the head — conflict cleared.

**P2 fixes applied** (commit `6f12b6c4`):

1. `lexical.py` — `except ValueError` in `TfidfIndex.index()`: narrowed to only catch the empty-vocabulary case by checking `"empty vocabulary" in str(exc)` before resetting; re-raises unrelated `ValueError`s instead of silently discarding them.

2. `lexical.py` — `except ImportError` in `TfidfIndex.load()`: narrowed to sklearn-related import failures (`"sklearn" in str(exc)`); re-raises other import errors (e.g. a missing custom class in a saved index) so the user gets an accurate error instead of a misleading "install scikit-learn" message.

All 12 lexical tests pass locally. AI re-review will pick up the new head.

TimeToBuildBob:
Re-resolved merge conflict (second pass).

**What happened**: the previous session's force-push landed on `gptme/feat/gptme-rag-task-retrieval` (upstream) instead of `TimeToBuildBob/feat/gptme-rag-task-retrieval` (the fork that the PR uses as its head). The fork still had the old `269eaf27` head, which conflicted with the base.

**Fix**: force-pushed `6f12b6c4` (P2 fixes on top of current base `737c3e6a`) to the fork. PR is now `CLEAN` / `MERGEABLE`.

Still waiting on #1439 (TF-IDF backend) to merge before this can land.

ErikBjare:
@TimeToBuildBob I merged this, didn't realize the target was off-master

PR #1443, opened 2026-08-17
fix(gptodo): recurring reset emits state=waiting (fixes validator rejection)

TimeToBuildBob:
## Summary

PR #1441 merged a recurring-task reset that kept `state: todo` + `wait: <date>`. This pair is rejected by `validate_task_frontmatter.py` line 716-717:

```python
elif type_name == "tasks" and metadata.get("state") != "waiting":
    errors.append("wait requires state to be waiting")
```

The original fix in #1441 (commit e0ec7267) was correct — it emitted `state: waiting`. But it was reverted in commit 2b50abf3 with the claim that the validator doesn't have this rule. The rule does exist.

This PR re-applies the correct fix: the recurrence reset now emits `state: waiting` with the required companion fields (`wait_kind`, `waiting_for`, `waiting_since`).

## Changes

- `packages/gptodo/src/gptodo/cli.py`: recurrence reset emits `state: waiting` + `wait_kind=machine` + `waiting_for=<time-gate prose>` + `waiting_since=<now>` instead of `state: todo`
- `packages/gptodo/tests/`: update all tests that asserted `state == "todo"` after recur reset — all now assert `state == "waiting"`

## Test plan

- [x] All 462 gptodo tests pass locally
- [x] The regression test verifies `state=waiting` AND presence of `wait_kind`/`waiting_for`/`waiting_since`
- [x] Typecheck for gptodo passes (other package typechecks fail in worktree due to git-dep resolution — CI will verify)

### DIFF
--- a/packages/gptodo/src/gptodo/cli.py
+++ b/packages/gptodo/src/gptodo/cli.py
@@ -2492,10 +2492,12 @@ def edit(task_ids, set_fields, add_fields, remove_fields, set_subtask, force):
                     post.metadata["state"] = "waiting"
                     post.metadata["wait"] = next_wait.isoformat()
                     post.metadata["wait_kind"] = "machine"
-                    # Do not set waiting_for/waiting_since: the wait: date IS the machine gate.
-                    # A waiting_for field would be treated as a human blocker and prevent the
-                    # task from auto-surfacing once the time gate expires (task_has_waiting_blocker
-                    # requires waiting_for to be absent for the machine-gate early-return to fire).
+                    # Strip any pre-existing waiting_for/waiting_since: the wait: date IS the
+                    # machine gate. Leaving them would trap the task permanently — task_has_waiting_blocker
+                    # treats any waiting_for as a human blocker that requires explicit resolution,
+                    # so the task would never auto-surface after the time gate expires.
+                    post.metadata.pop("waiting_for", None)
+                    post.metadata.pop("waiting_since", None)
                     if not _completed_explicitly_set_global:
                         post.metadata.pop("completed", None)
                     with open(task.path, "w") as f:
--- a/packages/gptodo/src/gptodo/cli.py
+++ b/packages/gptodo/src/gptodo/cli.py
@@ -2872,7 +2872,16 @@ def ready(state, output_json, output_jsonl, use_cache, pool_filter, exclude_pool
             if task.state in ["backlog", "todo", "active", "ready_for_review"]
         ]
     else:  # both
-        filtered_tasks = [task for task in all_tasks if task.state in ["backlog", "todo", "active"]]
+        filtered_tasks = [
+            task
+            for task in all_tasks
+            if task.state in ["backlog", "todo", "active"]
+            or (
+                task.state == "waiting"
+                and task.metadata.get("wait_kind") == "machine"
+                and not task_is_waiting_for_date(task)
+            )
+        ]
 
     # Apply pool filter (default: general only; --pool all to see every pool)
     filtered_tasks = [
@@ -3086,8 +3095,17 @@ def next_(output_json, use_cache, pool_filter, exclude_pool, limit, order):
         cache_path = get_cache_path(repo_root)
         issue_cache = load_cache(cache_path)
 
-    # Filter for new or active tasks
-    workable_tasks = [task for task in all_tasks if task.state in ["backlog", "todo", "active"]]
+    # Filter for new or active tasks; include waiting tasks whose machine time-gate has expired
+    workable_tasks = [
+        task
+        for task in all_tasks
+        if task.state in ["backlog", "todo", "active"]
+        or (
+            task.state == "waiting"
+            and task.metadata.get("wait_kind") == "machine"
+            and not task_is_waiting_for_date(task)
+        )
+    ]
 
     # Apply pool filter (default: general only; --pool all to see every pool)
     workable_tasks = [
--- a/packages/gptodo/src/gptodo/cli.py
+++ b/packages/gptodo/src/gptodo/cli.py
@@ -2479,7 +2479,7 @@ def edit(task_ids, set_fields, add_fields, remove_fields, set_subtask, force):
             # Re-load task to get updated metadata
             post = frontmatter.load(task.path)
             if post.metadata.get("state") == "done":
-                # Handle recur: reset task to todo with advanced wait: date
+                # Handle recur: reset task to waiting with advanced wait: date
                 recur = post.metadata.get("recur")
                 if recur:
                     if parse_recur_interval(str(recur)) is None:
@@ -2489,17 +2489,21 @@ def edit(task_ids, set_fields, add_fields, remove_fields, set_subtask, force):
 
                     current_wait = parse_wait(post.metadata.get("wait"))
                     next_wait = advance_wait(current_wait, recur)
-                    post.metadata["state"] = "todo"
+                    post.metadata["state"] = "waiting"
                     post.metadata["wait"] = next_wait.isoformat()
-                    # Remove waiting-state fields that don't apply to todo state
-                    for _field in ("wait_kind", "waiting_for", "waiting_since"):
-                        post.metadata.pop(_field, None)
+                    post.metadata["wait_kind"] = "machine"
+                    post.metadata["waiting_for"] = (
+                        f"time gate — next recurrence due {next_wait.isoformat()}"
+                    )
+                    post.metadata["waiting_since"] = datetime.now(timezone.utc).isoformat(
+                        timespec="seconds"
+                    )
                     if not _completed_explicitly_set_global:
                         post.metadata.pop("completed", None)
                     with open(task.path, "w") as f:
                         f.write(frontmatter.dumps(post))
                     console.print(
-                        f"[cyan]↩ {task.name} recurring — reset to todo, next wait: {next_wait}[/]"
+                        f"[cyan]↩ {task.name} recurring — reset to waiting, next wait: {next_wait}[/]"
                     )
                     continue  # skip done-completion logic for recurring tasks
 
--- a/packages/gptodo/src/gptodo/cli.py
+++ b/packages/gptodo/src/gptodo/cli.py
@@ -2492,12 +2492,10 @@ def edit(task_ids, set_fields, add_fields, remove_fields, set_subtask, force):
                     post.metadata["state"] = "waiting"
                     post.metadata["wait"] = next_wait.isoformat()
                     post.metadata["wait_kind"] = "machine"
-                    post.metadata["waiting_for"] = (
-                        f"time gate — next recurrence due {next_wait.isoformat()}"
-                    )
-                    post.metadata["waiting_since"] = datetime.now(timezone.utc).isoformat(
-                        timespec="seconds"
-                    )
+                    # Do not set waiting_for/waiting_since: the wait: date IS the machine gate.
+                    # A waiting_for field would be treated as a human blocker and prevent the
+                    # task from auto-surfacing once the time gate expires (task_has_waiting_blocker
+                    # requires waiting_for to be absent for the machine-gate early-return to fire).
                     if not _completed_explicitly_set_global:
                         post.metadata.pop("completed", None)
                     with open(task.path, "w") as f:
@@ -2880,6 +2878,7 @@ def ready(state, output_json, output_jsonl, use_cache, pool_filter, exclude_pool
                 task.state == "waiting"
                 and task.metadata.get("wait_kind") == "machine"
                 and not task_is_waiting_for_date(task)
+                and not task.metadata.get("waiting_for")
             )
         ]
 
@@ -3104,6 +3103,7 @@ def next_(output_json, use_cache, pool_filter, exclude_pool, limit, order):
             task.state == "waiting"
             and task.metadata.get("wait_kind") == "machine"
             and not task_is_waiting_for_date(task)
+            and not task.metadata.get("waiting_for")
         )
     ]
 
--- a/packages/gptodo/src/gptodo/utils.py
+++ b/packages/gptodo/src/gptodo/utils.py
@@ -641,7 +641,12 @@ def task_has_waiting_blocker(task: "TaskInfo") -> bool:
     """Return True when a task is explicitly waiting on an external condition."""
     state = normalize_state(task.state or "", warn=False) if task.state else ""
     # A machine time-gate that has already expired is not a blocker — the wait has passed.
-    if task.metadata.get("wait_kind") == "machine" and not task_is_waiting_for_date(task):
+    # Require task.wait is not None so probe-only machine tasks (no wait: field) stay blocked.
+    if (
+        task.metadata.get("wait_kind") == "machine"
+        and task.wait is not None
+        and not task_is_waiting_for_date(task)
+    ):
         return False
     if state == "waiting":
         return True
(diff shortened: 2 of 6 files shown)


TimeToBuildBob:
CI is fully green (all 13 checks pass). Awaiting Greptile auto-review (PR is within grace window). Will merge once review clears.

packages/gptodo/src/gptodo/cli.py
```diff
@@ -2489,17 +2489,21 @@ def edit(task_ids, set_fields, add_fields, remove_fields, set_subtask, force):
 
                     current_wait = parse_wait(post.metadata.get("wait"))
                     next_wait = advance_wait(current_wait, recur)
-                    post.metadata["state"] = "todo"
+                    post.metadata["state"] = "waiting"
```
TimeToBuildBob, line 2492: ❌ **P1** — After this change, a recurring task marked done is reset to state=waiting. However, the `ready` and `next` commands filter tasks by state in {backlog, todo, active} (see `ready()` filtering and `next_()` workable_tasks in cli.py). A task in state=waiting is never included in those filters, regardless of whether its `wait` date has passed. Previously the reset produced state=todo with a future `wait`, so the task would become visible in `ready`/`next` once the wait expired; now it will never reappear in the work queue, permanently vanishing from the commands agents use to pick up work. This silently breaks the core recurring-task workflow: the task resets into a state that is unreachable until a human manually edits it back to a workable state.

TimeToBuildBob:
## 🤖 AI code review

The PR changes the recurring-task reset in `edit` so that completing a task with a valid `recur` now sets state to `waiting` (with `wait_kind: machine` and an advanced `wait`) instead of `todo`, avoiding the validator rule that `wait` requires state `waiting`. It also adds an early return in `task_has_waiting_blocker` for expired machine time-gates (when no probe or `waiting_for` is present), and updates the `ready` and `next` commands' state filters to include such waiting tasks. Tests throughout are updated to expect `waiting` after recur reset, and new regression tests cover probe/human waiting interplay.

2 P1 defects outstanding on this head.

**2 findings** · ❌ **2** P1 — 🛑 **2 at P0/P1.** Any finding, P2 included, keeps the score under 5/5.

### Findings

**❌ P1 high** — [`packages/gptodo/src/gptodo/cli.py:469`](https://github.com/gptme/gptme-contrib/blob/4d1019da5ff325d42f6e0dc784f3cd195d2647f6/packages/gptodo/src/gptodo/cli.py#L469)

explain_readiness still hard-codes state == "waiting" as a failing check, always producing a NOT READY verdict for any waiting task. But this PR makes task_has_waiting_blocker return False for a machine time-gate whose wait date has passed (wait_kind=machine, wait not None, not task_is_waiting_for_date, no probe, no waiting_for), so `gptodo ready` and `gptodo next` will list such a task as ready. The two commands now contradict each other for the exact set of tasks this PR is designed to surface: `ready` shows them as ready, while `explain` says NOT READY with the state/waiting_for check as the failing reason. Users cannot use explain to diagnose why a task appears ready, breaking the diagnostic path.

<sub>**How this was verified:** Read explain_readiness lines 452-482: the waiting branch fails and short-circuits. Confirmed ready's 'both' branch at 2875-2885 now admits machine-expired waiting tasks, which is new in this PR. The mirroring comment at 508-510 confirms explain is expected to agree with ready.</sub>

<sub>Consensus: 2/2 passes agreed</sub>

**❌ P1 high** — [`packages/gptodo/src/gptodo/cli.py:2541`](https://github.com/gptme/gptme-contrib/blob/4d1019da5ff325d42f6e0dc784f3cd195d2647f6/packages/gptodo/src/gptodo/cli.py#L2541)

The PR makes waiting tasks with an expired machine time-gate appear in `gptodo ready` and `gptodo next` (via the new filter and task_has_waiting_blocker returning False). However, `gptodo claim` still refuses any task in state "waiting" because _CLAIM_REFUSED_STATES includes it. A recurring task that reset to waiting with a future wait becomes actionable once that wait date passes, but if a user picks it from `ready` and runs `claim`, they get "Refusing to claim ... state is 'waiting'". This breaks the ready→claim workflow for the exact tasks the PR intends to surface, forcing the user to manually edit the state before claiming.

```suggestion
post.metadata.pop("probe", None)
```

<sub>**How this was verified:** Read cli.py reset block lines 2488-2507 (no pop('probe')) and utils.py task_has_waiting_blocker lines 648-658; confirmed ready/next filters omit probe checks. Existing test test_recur_reset_strips_preexisting_waiting_for only asserts waiting_for/waiting_since removal, not probe.</sub>

<sub>Consensus: 2/2 passes agreed</sub>

- `packages/gptodo/src/gptodo/cli.py` — Changes recur reset to emit state=waiting with wait_kind machine and strips waiting_for/waiting_since; adds waiting-state machine time-gate filter to ready and next; updates console messages.
- `packages/gptodo/src/gptodo/utils.py` — Adds early return in task_has_waiting_blocker for an expired machine time-gate with no probe or waiting_for, treating it as not a blocker.
- `packages/gptodo/tests/test_edit_state_completed.py` — Updates assertions from state=todo to state=waiting for recurring resets and adjusts docstrings.
- `packages/gptodo/tests/test_edit_terminal_cleanup.py` — Updates test that recurring reset keeps fields to expect state waiting instead of todo.
- `packages/gptodo/tests/test_recur.py` — Renames and adjusts recur-reset tests to assert state waiting, and simplifies the regression test to not pre-populate waiting fields.
- `packages/gptodo/tests/test_wait_recur.py` — Adds tests for probe-only machine tasks, probe+expired wait, and human waiting_for taking precedence; updates recur reset tests to expect waiting.

</details>

| `672a243e6800` | 3/5 | 1 | llm | 2026-08-17 19:05 UTC |
| `769e1c7cf11d` | 3/5 | 3 | llm | 2026-08-17 20:31 UTC |
| `769e1c7cf11d` | 3/5 | 1 | llm | 2026-08-18 16:06 UTC |
| `728bc70b924f` | 3/5 | 1 | llm | 2026-08-18 16:25 UTC |
| `5b5afda1fcbc` | 3/5 | 1 | llm | 2026-08-18 17:42 UTC |
| `8424576df80c` | 3/5 | 1 | llm | 2026-08-18 17:53 UTC |

</details>

<sub>Reviewed `4d1019da5ff3` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 151s · reports findings; merge decisions belong to the gate · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

TimeToBuildBob:
**P1 addressed** — pushed `769e1c7c`.

The AI reviewer was right: resetting recurring tasks to `state=waiting` with `wait_kind=machine` made them permanently invisible to `ready` and `next` (both filter to `{backlog, todo, active}`).

Fix applied in three places:

1. **`task_has_waiting_blocker`** — returns `False` when `wait_kind=machine` and the `wait:` date has already expired. An elapsed time gate is not a real blocker.
2. **`ready` default filter** — also includes `waiting` tasks with `wait_kind=machine` and an expired `wait:` date.
3. **`next` workable_tasks filter** — same inclusion rule.

This keeps `state=waiting` (which satisfies the validator) while ensuring the task resurfaces in the work queue once the recurrence date arrives. All 462 gptodo tests pass.

packages/gptodo/src/gptodo/cli.py
TimeToBuildBob: Fixed in 769e1c7c:  now returns False for expired machine gates;  and  include them in the workable set.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
## 🤖 AI code review

Two-commit series that fixes the recurring-task reset:

1. **672a243e** — `gptodo edit --done` on a recurring task now resets to `state: waiting` with `wait_kind: machine`, `waiting_for: <time-gate prose>`, and `waiting_since: <now>` instead of `state: todo`. This satisfies the frontmatter validator rule (`wait:` requires `state: waiting`). Tests updated accordingly.

2. **769e1c7c** — The prior P1 finding: `ready` and `next` filtered to `{backlog, todo, active}` only, permanently hiding machine-gated tasks. Fixed in three places: `task_has_waiting_blocker` returns `False` for an expired machine gate; `ready` (both-mode) and `next` both include `state: waiting` tasks whose `wait_kind` is `machine` and whose `wait:` date has already passed.

### Confidence Score: 5/5 — No outstanding findings

All prior P1 findings are addressed. No new findings.

<sub>Reviewed `769e1c7cf11d` · `claude-sonnet-4-6` · `llm` engine · direct review · reports findings; merge decisions belong to the gate · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

packages/gptodo/src/gptodo/utils.py
```diff
@@ -640,6 +640,9 @@ def __str__(self) -> str:
 def task_has_waiting_blocker(task: "TaskInfo") -> bool:
     """Return True when a task is explicitly waiting on an external condition."""
     state = normalize_state(task.state or "", warn=False) if task.state else ""
+    # A machine time-gate that has already expired is not a blocker — the wait has passed.
+    if task.metadata.get("wait_kind") == "machine" and not task_is_waiting_for_date(task):
```
TimeToBuildBob: ❌ **P1** — The new condition in `task_has_waiting_blocker` returns False for any waiting task with `wait_kind == "machine"` and no future `wait` date, regardless of whether the task carries a pending `probe`. This makes machine-class blockers that are not pure date gates (e.g., a CI check via `probe`) appear unblocked. Consequently, `gptodo ready` and `gptodo next` will include such tasks as workable, and `gptodo loop` will process them, even though the external condition has not been evaluated. Concretely, a task with `state: waiting`, `wait_kind: machine`, `probe: "check_ci"`, and no `wait` field is now considered ready, whereas before it was correctly blocked. Fix: require `task.wait is not None` in the condition so only machine time-gates with a set (and now elapsed) wait date are treated as unblocked. _(2/2 passes)_

TimeToBuildBob: Fixed in 728bc70b: added `task.wait is not None` to the guard so only elapsed date-gates unblock — probe-only machine tasks (no `wait:` field) stay blocked. Regression test added: `test_machine_probe_only_task_stays_blocked`.

TimeToBuildBob:
## 🤖 AI code review

Reviewed HEAD `728bc70b924f` (after rebase).

**Summary**: Two-commit series that fixes recurring-task reset to emit `state: waiting` instead of `state: todo`, with proper machine-gate semantics (`wait_kind: machine`, `waiting_for`, `waiting_since`). Also updates `ready`, `next`, and `task_has_waiting_blocker` so expired machine gates don't suppress tasks from the work queue.

**Correctness** ✅
- `task_has_waiting_blocker`: The `task.wait is not None` guard is correct — probe-only machine tasks (no `wait:` field) stay blocked, while time-gated tasks with an expired date are treated as unblocked. The guard prevents a class of tasks with `wait_kind: machine` but no date from prematurely surfacing.
- `ready`/`next` filter: Including `waiting` tasks where `wait_kind == "machine"` and `not task_is_waiting_for_date(task)` is the correct complement — expired gates become workable.
- Removing the old `wait_kind`/`waiting_for`/`waiting_since` field-strip on recur reset is correct; those fields are now populated rather than deleted.

**Tests** ✅  
All test assertions updated consistently. Tests in `test_recur.py`, `test_edit_state_completed.py`, `test_edit_terminal_cleanup.py`, `test_wait_recur.py` all exercise the new behavior paths.

**No issues found.** CI is green (coverage in progress). Greptile review pending — triggered 2026-08-17T20:23Z, no response yet (Greptile billing dark).

TimeToBuildBob:
**Greptile P1 addressed** — pushed `5b5afda1`.

The finding was valid: `task_has_waiting_blocker` would return `False` for any machine task with `wait_kind: machine`, an expired `wait:` date, **and** a `probe:` field — surfacing it as ready before the probe had been resolved.

Fix: added `and not task.metadata.get("probe")` to the early-return condition. A task with both an expired time gate and an unresolved probe now stays blocked until the probe is actively checked.

Added regression test `test_machine_probe_plus_expired_wait_stays_blocked` covering the combination (the existing `test_machine_probe_only_task_stays_blocked` already covered probe-only / no `wait:` field).

packages/gptodo/src/gptodo/utils.py
```diff
@@ -640,6 +640,16 @@ def __str__(self) -> str:
...
+    # A machine time-gate that has already expired is not a blocker — the wait has passed.
+    # Guards: task.wait is not None (probe-only tasks with no date stay blocked)
+    #         and no probe field (probe+expired-date tasks still need probe resolution).
+    if (
+        task.metadata.get("wait_kind") == "machine"
+        and task.wait is not None
+        and not task_is_waiting_for_date(task)
+        and not task.metadata.get("probe")
+    ):
+        return False
```
TimeToBuildBob, line 655: ❌ **P1** — The new early-return in `task_has_waiting_blocker` unblocks any task with `wait_kind: machine`, a non-None `wait` that has passed, and no `probe`, without checking `waiting_for` or other waiting-state fields. If an operator sets `wait_kind: machine` alongside a human-dependent `waiting_for` (e.g. `waiting_for: John to review`), that task will become ready as soon as the time gate expires, even though it is still blocked on John. The `ready` and `next` filters include exactly such tasks (`state == "waiting" and wait_kind == "machine" and not task_is_waiting_for_date`), so they surface in selections and an agent may pick them up prematurely. The intended unblocking should only apply when the machine gate is the *only* blocker; the current condition ignores the `waiting_for` field entirely, causing false-ready tasks. _(2/2 passes)_

TimeToBuildBob: Fixed in 8424576d:

- : added `and not task.metadata.get("waiting_for")` to the machine-gate early-return — a task carrying `waiting_for` stays blocked regardless of whether the time gate expired.
- Same guard in the `ready` and `next_` filters in cli.py.
- Recurring reset no longer sets `waiting_for`/`waiting_since`: the `wait:` date is the machine gate; `waiting_for` is for human-described blockers that need explicit resolution.
- Added regression test `test_machine_expired_wait_with_human_waiting_for_stays_blocked`.

465 tests pass.

packages/gptodo/src/gptodo/cli.py
```diff
@@ -2489,17 +2489,19 @@ def edit(task_ids, set_fields, add_fields, remove_fields, set_subtask, force):
...
                     post.metadata["wait"] = next_wait.isoformat()
-                    # Remove waiting-state fields that don't apply to todo state
-                    for _field in ("wait_kind", "waiting_for", "waiting_since"):
-                        post.metadata.pop(_field, None)
+                    post.metadata["wait_kind"] = "machine"
+                    # Do not set waiting_for/waiting_since: the wait: date IS the machine gate.
```
TimeToBuildBob: ❌ **P1** — The recurrence reset in `edit()` sets `state` to `waiting` and `wait_kind` to `machine` but no longer removes pre-existing `waiting_for` and `waiting_since` fields, which the old code explicitly popped. If a recurring task carried these fields (e.g., it was previously waiting on a human blocker), they survive the reset. When the new `wait:` date expires, `task_has_waiting_blocker` returns True because `waiting_for` is set, and the `ready`/`next` filters exclude any waiting task with `waiting_for`. The task therefore never auto-surfaces and remains stuck in `waiting` indefinitely, a regression from the previous behavior where these stale fields were stripped. Concretely: a task with `state: waiting`, `waiting_for: 'John to review'`, `recur: 7d` that is marked done will reset with `waiting_for` intact and never become ready again. _(2/2 passes)_

TimeToBuildBob: Fixed in 4d1019da: the recurring reset now explicitly pops `waiting_for` and `waiting_since` fields before writing. This clears any pre-existing human-blocker fields so the task auto-surfaces when the time gate expires.

Regression test added: `test_recur_reset_strips_preexisting_waiting_for` — a task carrying `waiting_for: John to review the PR` is marked done, and we assert the field is absent after reset.

466 tests pass.

PR #1453, opened 2026-08-19
fix(worker-records): fetch_pr_snapshot returns None on gh failure (phase 1.3)

TimeToBuildBob:
## Problem

`fetch_pr_snapshot` was returning `{}` on `gh` API failures, which was indistinguishable from a PR with no observable diff. `derive_effect_signal` then returned `EFFECT_UNKNOWN`, causing `classify_completion` to return `CLASS_INEFFECTIVE` (budget 2), escalating healthy PRs in just 2 dispatches even when no outage was involved.

## Fix

- `fetch_pr_snapshot`: return `None` on non-zero `gh` exit (vs `{}` = empty success)
- `update_record_pr_state`: detect `after is None`, set `gh_snapshot_fetch_failed=True`
- `EFFECT_FETCH_FAILED = "fetch_failed"` constant added
- `derive_effect_signal`: short-circuit to `EFFECT_FETCH_FAILED` on the flag

The `EFFECT_FETCH_FAILED` value flows through the dispatch ledger; the recovery classifier in `pm_dispatch_recovery.py` maps it to `CLASS_INFRA` (separate change).

## Tests

3 new regression tests + 1 golden updated (`empty_before_gh_fails`).

## Part of

Phase 1.3 of the PM executor fallback / backend-outage series.

### DIFF
--- a/packages/gptme-runloops/src/gptme_runloops/worker_records.py
+++ b/packages/gptme-runloops/src/gptme_runloops/worker_records.py
@@ -1104,15 +1104,18 @@ def derive_effect_signal(
        effect for every session that never looked.
     """
     delivery = str(delivery_outcome or "").strip().lower()
+
+    # gh API failed at snapshot time: we cannot observe the "after" state, so
+    # this is an infra failure, not genuine ineffectiveness.  Must take
+    # precedence over delivery_outcome (including orphan_no_delivery) so that a
+    # fetch failure is never recorded as a definitive no-effect.
+    if payload.get("gh_snapshot_fetch_failed"):
+        return EFFECT_FETCH_FAILED
+
     if delivery == "orphan_no_delivery":
         return EFFECT_NONE
     if delivery == "handled":
         return EFFECT_OBSERVED
-
-    # gh API failed at snapshot time: we cannot observe the "after" state, so
-    # this is an infra failure, not genuine ineffectiveness.
-    if payload.get("gh_snapshot_fetch_failed"):
-        return EFFECT_FETCH_FAILED
 
     before_head = normalize_oid(payload.get("pr_head_oid_before"))
     after_head = normalize_oid(payload.get("pr_head_oid_after"))
--- a/packages/gptme-runloops/src/gptme_runloops/worker_records.py
+++ b/packages/gptme-runloops/src/gptme_runloops/worker_records.py
@@ -669,8 +669,7 @@ def update_record_pr_state(
         if not isinstance(payload, dict):
             return
 
-        if gh_fetch_failed:
-            payload["gh_snapshot_fetch_failed"] = True
+        payload["gh_snapshot_fetch_failed"] = gh_fetch_failed
         apply_pr_state_diff(payload, before, after_safe)
         outcome_before_upgrade = payload.get("outcome")
         if upgrade_outcome is not None:
--- a/packages/gptme-runloops/src/gptme_runloops/worker_records.py
+++ b/packages/gptme-runloops/src/gptme_runloops/worker_records.py
@@ -597,10 +597,11 @@ def fetch_pr_snapshot(
     cwd: str | Path,
     timeout: float = 20,
     runner: Callable[..., subprocess.CompletedProcess[str]] = subprocess.run,
-) -> dict[str, str]:
+) -> dict[str, str] | None:
     """Fetch the live PR snapshot via gh (worker.sh:413-433).
 
-    A non-zero gh exit → ``{}``; unparseable stdout → ``{}``.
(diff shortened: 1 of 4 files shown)


packages/gptme-runloops/src/gptme_runloops/worker_records.py
```diff
@@ -650,19 +651,27 @@ def update_record_pr_state(
...
     with locked_state_file(record_path):
         if not record_path.is_file():
             return
         payload = json.loads(record_path.read_text(encoding="utf-8"))
         if not isinstance(payload, dict):
             return
 
-        apply_pr_state_diff(payload, before, after)
+        if gh_fetch_failed:
+            payload["gh_snapshot_fetch_failed"] = True
```
TimeToBuildBob: ❌ **P1** — The flag is never cleared. Because update_record_pr_state does not remove or reset gh_snapshot_fetch_failed when a later fetch succeeds, a record that previously had a gh failure will keep the flag set in all subsequent runs. derive_effect_signal then always returns EFFECT_FETCH_FAILED (line 1115) for that record, even when a later successful fetch populated a valid after-state that proves the PR moved. This misclassifies healthy dispatches as infra failures permanently. Setting the key unconditionally (payload["gh_snapshot_fetch_failed"] = gh_fetch_failed) fixes it.

TimeToBuildBob:
## 🤖 AI code review

This PR changes `fetch_pr_snapshot` to return `None` on GitHub API failures (non-zero exit) instead of `{}`, distinguishing API errors from empty diffs. The `update_record_pr_state` function detects this failure, sets a `gh_snapshot_fetch_failed` flag in the record payload, and passes a safe empty dict to `apply_pr_state_diff`. A new `EFFECT_FETCH_FAILED` constant is added, and `derive_effect_signal` checks this flag with precedence over other outcome signals to classify fetch failures as infrastructure issues rather than lack of effect.

- `packages/gptme-runloops/src/gptme_runloops/worker_records.py` — Updated `fetch_pr_snapshot` return type to `dict[str, str] | None`, changed non-zero exit to return `None` instead of `{}`; modified `update_record_pr_state` to detect failure, set `gh_snapshot_fetch_failed` flag, and use safe dict for diff; added `EFFECT_FETCH_FAILED` constant and check in `derive_
- `packages/gptme-runloops/tests/goldens/worker_records/pr_state_diff.json` — Updated golden test expected records to include `gh_snapshot_fetch_failed` field (false for successful fetches, true for the failure case).
- `packages/gptme-runloops/tests/test_pm_dispatch_outcome.py` — Imported `EFFECT_FETCH_FAILED` and added two tests verifying the flag triggers fetch-failed effect and takes precedence over orphan-no-delivery.
- `packages/gptme-runloops/tests/test_worker_records.py` — Renamed test to reflect `None` return; changed assertion from `== {}` to `is None`; added test verifying the flag is set when fetch fails.

</details>

| `b70acab39545` | 3/5 | 1 | llm | 2026-08-19 03:41 UTC |
| `8a1a72a93261` | 5/5 | 0 | llm | 2026-08-19 04:29 UTC |
| `8a1a72a93261` | 3/5 | 1 | llm | 2026-08-19 04:53 UTC |

</details>

<sub>Reviewed `4406b6fa054a` · `openrouter/anthropic/claude-haiku-4-5` · `llm` engine · 123s · reports findings; merge decisions belong to the gate · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

TimeToBuildBob:
Fixed the P1 in 8a1a72a9: `gh_snapshot_fetch_failed` is now written unconditionally (`payload["gh_snapshot_fetch_failed"] = gh_fetch_failed`), so a subsequent successful fetch clears the flag. Golden fixtures updated for the three non-failure cases. All 157 tests pass.

TimeToBuildBob:
@greptileai review

packages/gptme-runloops/src/gptme_runloops/worker_records.py
TimeToBuildBob: ✅ Fixed in 8a1a72a9 — the flag is now written unconditionally (`payload["gh_snapshot_fetch_failed"] = gh_fetch_failed`), so a subsequent successful fetch correctly clears it. All 157 tests pass. Fresh AI review at current head (8a1a72a93261) confirms 0 findings.

TimeToBuildBob: Fixed in 4406b6fa: moved the `gh_snapshot_fetch_failed` check before the `orphan_no_delivery` branch, so an infra failure always returns `EFFECT_FETCH_FAILED` regardless of delivery outcome. Regression test added (`test_gh_fetch_failed_overrides_orphan_no_delivery`).

PR #1455, opened 2026-08-19
feat(skills): add test-discovery skill for auto-detecting test runners

TimeToBuildBob:
## Summary

Adds a `test-discovery` skill that teaches gptme agents to auto-detect and run the relevant test suite after making code edits, without needing an explicit "run tests" prompt.

- **SKILL.md**: Procedure for detecting test framework, scoping test runs, interpreting results, and keeping output concise
- **scripts/detect-runner.sh**: Shell script that inspects project signals (`pytest.ini`, `[tool.pytest.ini_options]` in pyproject.toml, `package.json` scripts, `Cargo.toml`, `Makefile`) and prints the recommended test command

Supports: pytest, jest/vitest, cargo test, Makefile targets.

**Motivation**: Grounded in the Openleetcode HN signal ("tests live in the repo" pattern, score 108, 2026-08-18) — agents that auto-verify after edits catch regressions without human prompting. This is the agent-instruction layer; the tool-side hook in `gptme/tools/shell.py` is a follow-up once the gptme PR queue has capacity.

## Test plan

- [x] `detect-runner.sh` tested against gptme repo (pyproject.toml with `[tool.pytest.ini_options]`) → outputs `uv run pytest -x -q`
- [x] `detect-runner.sh` tested against brain repo → same correct output
- [x] Skill validates via `gptme-lessons-extras` validator
- [x] ShellCheck passes on `detect-runner.sh`
- [x] Root structure check passes (`skills/` is on the allowlist)

### DIFF
--- a/skills/test-discovery/scripts/detect-runner.sh
+++ b/skills/test-discovery/scripts/detect-runner.sh
@@ -1,0 +1,60 @@
+#!/usr/bin/env bash
+# Detect the test runner for the current project and print the recommended command.
+# Exit 0 on success (prints command), exit 1 if no runner detected.
+set -euo pipefail
+
+REASON=""
+CMD=""
+
+# Python: pytest (priority: explicit config > pyproject > setup.cfg > presence)
+if [[ -f pytest.ini ]]; then
+    CMD="uv run pytest -x -q"
+    REASON="pytest.ini found"
+elif [[ -f pyproject.toml ]] && grep -q '\[tool\.pytest' pyproject.toml 2>/dev/null; then
+    CMD="uv run pytest -x -q"
+    REASON="[tool.pytest.ini_options] in pyproject.toml"
+elif [[ -f setup.cfg ]] && grep -q '^\[tool:pytest\]' setup.cfg 2>/dev/null; then
+    CMD="uv run pytest -x -q"
+    REASON="[tool:pytest] in setup.cfg"
+elif [[ -f tox.ini ]] && grep -q '^\[pytest\]' tox.ini 2>/dev/null; then
+    CMD="uv run pytest -x -q"
+    REASON="[pytest] section in tox.ini"
+elif command -v pytest &>/dev/null || command -v uv &>/dev/null; then
+    # Heuristic: if there are test files, assume pytest
+    if find . -name 'test_*.py' -o -name '*_test.py' 2>/dev/null | grep -q .; then
+        CMD="uv run pytest -x -q"
+        REASON="test_*.py files found (pytest assumed)"
+    fi
+fi
+
+# JavaScript/TypeScript: vitest or jest (only if no Python runner found)
+if [[ -z "$CMD" ]] && [[ -f package.json ]]; then
+    if python3 -c "import json,sys; s=json.load(open('package.json')).get('scripts',{}); sys.exit(0 if any('vitest' in v for v in s.values()) else 1)" 2>/dev/null; then
+        CMD="npx vitest run"
+        REASON="vitest in package.json scripts"
+    elif python3 -c "import json,sys; s=json.load(open('package.json')).get('scripts',{}); sys.exit(0 if 'test' in s else 1)" 2>/dev/null; then
+        CMD="npm test"
+        REASON="test script in package.json"
+    fi
+fi
+
+# Rust: cargo test
+if [[ -z "$CMD" ]] && [[ -f Cargo.toml ]]; then
+    CMD="cargo test 2>&1"
+    REASON="Cargo.toml found"
+fi
+
+# Makefile: test target
+if [[ -z "$CMD" ]] && [[ -f Makefile ]]; then
+    if grep -q '^test[[:space:]]*:' Makefile 2>/dev/null; then
+        CMD="make test"
+        REASON="test target in Makefile"
+    fi
+fi
+
+if [[ -z "$CMD" ]]; then
+    echo "No test runner detected" >&2
+    exit 1
+fi
+
+echo "$CMD  # $REASON"
--- a/skills/test-discovery/scripts/detect-runner.sh
+++ b/skills/test-discovery/scripts/detect-runner.sh
@@ -6,23 +6,37 @@ set -euo pipefail
 REASON=""
 CMD=""
 
-# Python: pytest (priority: explicit config > pyproject > setup.cfg > presence)
+# Resolve pytest invocation: prefer uv run pytest when uv is available
+if command -v uv &>/dev/null; then
+    PYTEST_CMD="uv run pytest -x -q"
+else
+    PYTEST_CMD="pytest -x -q"
+fi
+
+# Python: pytest (priority: explicit config > pyproject > setup.cfg > tox > presence)
 if [[ -f pytest.ini ]]; then
-    CMD="uv run pytest -x -q"
+    CMD="$PYTEST_CMD"
     REASON="pytest.ini found"
 elif [[ -f pyproject.toml ]] && grep -q '\[tool\.pytest' pyproject.toml 2>/dev/null; then
-    CMD="uv run pytest -x -q"
+    CMD="$PYTEST_CMD"
     REASON="[tool.pytest.ini_options] in pyproject.toml"
 elif [[ -f setup.cfg ]] && grep -q '^\[tool:pytest\]' setup.cfg 2>/dev/null; then
-    CMD="uv run pytest -x -q"
+    CMD="$PYTEST_CMD"
     REASON="[tool:pytest] in setup.cfg"
-elif [[ -f tox.ini ]] && grep -q '^\[pytest\]' tox.ini 2>/dev/null; then
-    CMD="uv run pytest -x -q"
-    REASON="[pytest] section in tox.ini"
+elif [[ -f tox.ini ]]; then
+    if grep -q '^\[pytest\]' tox.ini 2>/dev/null; then
+        # tox.ini used as a plain pytest config — run pytest directly
+        CMD="$PYTEST_CMD"
+        REASON="[pytest] section in tox.ini"
+    else
+        # Real tox project (has [testenv] or similar) — let tox manage the env
+        CMD="tox"
+        REASON="tox.ini found"
+    fi
 elif command -v pytest &>/dev/null || command -v uv &>/dev/null; then
     # Heuristic: if there are test files, assume pytest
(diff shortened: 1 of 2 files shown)


skills/test-discovery/scripts/detect-runner.sh
```diff
@@ -0,0 +1,60 @@
...
+# Detect the test runner for the current project and print the recommended command.
+# Exit 0 on success (prints command), exit 1 if no runner detected.
+set -euo pipefail
+
+REASON=""
+CMD=""
+
+# Python: pytest (priority: explicit config > pyproject > setup.cfg > presence)
+if [[ -f pytest.ini ]]; then
+    CMD="uv run pytest -x -q"
```
TimeToBuildBob: ❌ **P1** — The script unconditionally emits `uv run pytest -x -q` whenever it detects a pytest configuration (pytest.ini, pyproject.toml with [tool.pytest], setup.cfg, tox.ini, or the test-file heuristic). It never checks whether `uv` is actually installed. For projects that use plain pytest without uv, the command fails immediately with "uv: command not found", so the agent cannot run tests and the skill's primary feature is broken in those environments. The detection should check for `uv` availability and fall back to `pytest -x -q` when uv is absent.

skills/test-discovery/scripts/detect-runner.sh
```diff
@@ -0,0 +1,60 @@
...
+    REASON="pytest.ini found"
+elif [[ -f pyproject.toml ]] && grep -q '\[tool\.pytest' pyproject.toml 2>/dev/null; then
+    CMD="uv run pytest -x -q"
+    REASON="[tool.pytest.ini_options] in pyproject.toml"
+elif [[ -f setup.cfg ]] && grep -q '^\[tool:pytest\]' setup.cfg 2>/dev/null; then
+    CMD="uv run pytest -x -q"
+    REASON="[tool:pytest] in setup.cfg"
+elif [[ -f tox.ini ]] && grep -q '^\[pytest\]' tox.ini 2>/dev/null; then
+    CMD="uv run pytest -x -q"
```
TimeToBuildBob: ⚠️ **P2** — The script's tox.ini handling only checks for a `[pytest]` section and then sets the command to `uv run pytest -x -q`; it never outputs `tox` even though SKILL.md lists tox.ini as a signal with the command `tox`. For a project that relies on tox environments (e.g., `[testenv]` sections), running pytest directly bypasses the tox virtualenv setup, which can cause missing dependencies or the wrong Python interpreter. The script should detect tox.ini and output `tox` as the command, or at least fall back to `tox` when no `[pytest]` section is present.

TimeToBuildBob:
## 🤖 AI code review

Adds a 'test-discovery' skill for auto-detecting and running relevant test suites after code edits.

- `skills/test-discovery/SKILL.md` — Defines the procedure for test discovery and interpretation.
- `skills/test-discovery/scripts/detect-runner.sh` — Implements the test runner detection logic.

</details>

| `642e254b0f1b` | 3/5 | 2 | llm | 2026-08-19 04:42 UTC |
| `64fcec680c63` | 3/5 | 3 | llm | 2026-08-19 07:51 UTC |

</details>

<sub>Reviewed `c44971a10999` · `openrouter/mistralai/mistral-nemo` · `llm` engine · 8s · reports findings; merge decisions belong to the gate · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

TimeToBuildBob:
@greptileai review

skills/test-discovery/scripts/detect-runner.sh
TimeToBuildBob: Fixed in 64fcec68. Added a `PYTEST_CMD` variable resolved at runtime: `uv run pytest -x -q` when uv is available, `pytest -x -q` otherwise. All pytest branches now use this variable instead of the hardcoded uv form.

skills/test-discovery/scripts/detect-runner.sh
TimeToBuildBob: Fixed in 64fcec68. tox.ini handling now branches: if a `[pytest]` section is present, the file is acting as a plain pytest config and we run pytest directly; otherwise we output `tox` so real tox environments aren't bypassed.

skills/test-discovery/scripts/detect-runner.sh
```diff
@@ -0,0 +1,74 @@
...
+        CMD="$PYTEST_CMD"
+        REASON="[pytest] section in tox.ini"
+    else
+        # Real tox project (has [testenv] or similar) — let tox manage the env
+        CMD="tox"
+        REASON="tox.ini found"
+    fi
+elif command -v pytest &>/dev/null || command -v uv &>/dev/null; then
+    # Heuristic: if there are test files, assume pytest
+    if find . -name 'test_*.py' -o -name '*_test.py' 2>/dev/null | grep -q .; then
```
TimeToBuildBob: ❌ **P1** — The script enables `set -euo pipefail` at line 4, so this `if` condition is evaluated as a pipeline. `grep -q .` exits on the first non-empty line and closes its read end; whenever `find` still has another line to emit after that (the normal case in a project with more than a few test files), `find` receives `SIGPIPE` and exits 141. `pipefail` then makes the whole pipeline status non-zero even though a match was found, so the condition is false and `CMD` is never set. The script falls through the JS/Rust/Make checks and finally prints "No test runner detected" for a pure-Python project that has no pytest config file but does have test_*.py files. This breaks the heuristic branch that is supposed to catch the config-less Python case.

skills/test-discovery/scripts/detect-runner.sh
```diff
@@ -0,0 +1,74 @@
...
+    CMD="$PYTEST_CMD"
+    REASON="pytest.ini found"
+elif [[ -f pyproject.toml ]] && grep -q '\[tool\.pytest' pyproject.toml 2>/dev/null; then
+    CMD="$PYTEST_CMD"
+    REASON="[tool.pytest.ini_options] in pyproject.toml"
+elif [[ -f setup.cfg ]] && grep -q '^\[tool:pytest\]' setup.cfg 2>/dev/null; then
+    CMD="$PYTEST_CMD"
+    REASON="[tool:pytest] in setup.cfg"
+elif [[ -f tox.ini ]]; then
+    if grep -q '^\[pytest\]' tox.ini 2>/dev/null; then
```
TimeToBuildBob: ⚠️ **P2** — Presence of a `[pytest]` section in tox.ini does not imply the project is not using tox. Real tox projects commonly have both `[testenv]` (or `[tox]`) and a `[pytest]` section that configures addopts, markers, or filterwarnings. This branch treats any `[pytest]` section as "plain pytest config" and sets `$PYTEST_CMD` (for example `uv run pytest -x -q`), bypassing `tox` and its managed environment. The agent then runs pytest in the current interpreter, which may not have the dependencies declared in `[testenv]`, so the run fails with import or environment errors. Detection should check for `[testenv]`/`[tox]` first and only fall back to direct pytest when no testenv section exists.

skills/test-discovery/scripts/detect-runner.sh
```diff
@@ -0,0 +1,74 @@
...
+        REASON="test_*.py files found (pytest assumed)"
+    fi
+fi
+
+# JavaScript/TypeScript: vitest or jest (only if no Python runner found)
+if [[ -z "$CMD" ]] && [[ -f package.json ]]; then
+    if python3 -c "import json,sys; s=json.load(open('package.json')).get('scripts',{}); sys.exit(0 if any('vitest' in v for v in s.values()) else 1)" 2>/dev/null; then
+        CMD="npx vitest run"
+        REASON="vitest in package.json scripts"
+    elif python3 -c "import json,sys; s=json.load(open('package.json')).get('scripts',{}); sys.exit(0 if 'test' in s else 1)" 2>/dev/null; then
```
TimeToBuildBob, line 55: ⚠️ **P2** — The JavaScript detection only recognizes the substring `vitest` in any script value, and otherwise requires an exact `test` key to run `npm test`. It never scans script values for `jest`. A package.json such as `{"scripts": {"test:unit": "jest"}}` satisfies the SKILL's stated condition ("vitest or jest in scripts") and the PR's claim of jest support, but contains no `'test'` key and no `'vitest'` substring, so this branch and the preceding vitest branch both fail. The script exits 1 with "No test runner detected", so the agent runs no tests. The npm-test fallback should detect jest in a script value (or match keys beginning with `test`) rather than requiring the literal `test` key.

skills/test-discovery/scripts/detect-runner.sh
TimeToBuildBob: Fixed in c44971a1. Replaced `find ... | grep -q .` with `[[ -n "$(find ...)" ]]` — avoids the SIGPIPE/pipefail interaction where find receives exit 141 after grep closes its read end early, causing the condition to be false even when test files exist.

skills/test-discovery/scripts/detect-runner.sh
TimeToBuildBob: Fixed in c44971a1. Flipped the check order: now detects `[tox]`/`[testenv]` sections first (indicating a real tox project) and routes to `tox` regardless of whether `[pytest]` also appears. Only falls through to direct pytest when no tox/testenv section exists.

skills/test-discovery/scripts/detect-runner.sh
TimeToBuildBob: Fixed in c44971a1. Added a third elif after the `npm test` branch that scans all script values for the substring `jest`. When found (e.g. `{"test:unit": "jest"}`) with no canonical `test` key, the script now emits `npx jest` instead of falling through to 'No test runner detected'.

TimeToBuildBob:
Pushed `c44971a1` addressing all three findings:

**P1** — SIGPIPE under `pipefail` (line 38): replaced `find ... | grep -q .` with `[[ -n "$(find ...)" ]]`. Under `set -euo pipefail`, grep's early exit sends SIGPIPE (141) to find, making the pipeline non-zero even on a real match.

**P2** — tox.ini false-positive (line 27): now checks for `[tox]`/`[testenv]` sections first; only falls through to direct pytest when neither is present. A `[pytest]` coexisting with `[testenv]` no longer incorrectly bypasses tox.

**P2** — jest detection (line 49): added a third elif that scans all package.json script *values* for `jest`. A project with `{"test:unit": "jest"}` (no canonical `test` key) now gets `npx jest` instead of "No test runner detected".

@TimeToBuildBob review

PR #1458, opened 2026-08-19
fix(sessions): stop defaulting a score-less judge verdict to 0.5

TimeToBuildBob:
## Problem

The session judge fails *honestly* — when it can't produce a verdict it returns nothing. The consumers did not. A payload that parsed but carried no usable `score` was silently coerced to a neutral **0.5**, which is indistinguishable from a judge that genuinely scored the session at 0.5.

Two sites:
- `_parse_judge_payload()` — `float(verdict.get("score", 0.5))`
- `normalize_judge_verdict()` — `float(payload.get("score", 0.5))`

Same shape as the outcome promotions in `post_session()`: a `noop`/`unknown` outcome gets upgraded to `productive` on caller-supplied deliverables, and the record kept no trace of *why*.

## Changes

**No silent 0.5** (`judge.py`)
- `JudgeVerdict.score` is now `float | None`. `_coerce_score()` clamps to 0.0–1.0 and returns `None` for missing/unparseable values instead of substituting a default.
- Verdicts carry `judge_status` (`ok` | `no_score`), logged at WARNING when not `ok`.
- `write_alignment_grade()` stamps `judge_status` on the record and writes **no** alignment grade when the score is absent — so score-less runs are countable by querying session records rather than grepping logs, and the reward consumer falls back to the heuristic as it already does when the judge returns nothing.
- `judge_and_writeback()` reports `status=no_score`.

**Auditable outcome flips** (`post_session.py`, `record.py`)
- New `SessionRecord.outcome_flip_reason: str | None`, stamped at both promotion sites (`unreliable_trajectory_caller_deliverables`, `no_trajectory_caller_deliverables:<from>->productive:n=<k>`). `None` means no flip happened, so absence stays meaningful.

Grading semantics are otherwise unchanged: a verdict *with* a score behaves exactly as before.

## Tests

6 new tests; 3 existing exact-equality assertions updated for the new `judge_status` key.
- score-less / unparseable-score payloads → `judge_status=no_score`, `score=None`
- score-less verdict writes no alignment grade but does stamp `judge_status` on the record
- flip reason present on promotion, absent when no flip occurred

`1128 passed` for the whole `gptme-sessions` package.

Ref: `ErikBjare/bob` `tasks/judge-fallback-holes-visible.md`, design note `knowledge/technical-designs/session-judge-golden-set-and-split.md`.

### DIFF
--- a/packages/gptme-sessions/src/gptme_sessions/judge.py
+++ b/packages/gptme-sessions/src/gptme_sessions/judge.py
@@ -507,7 +507,10 @@ def _coerce_score(raw: Any) -> float | None:
     if raw is None:
         return None
     try:
-        return max(0.0, min(1.0, float(raw)))
+        v = float(raw)
+        if v != v:  # NaN: the only float where x != x is True
+            return None
+        return max(0.0, min(1.0, v))
     except (ValueError, TypeError):
         return None
 
@@ -968,6 +971,12 @@ def write_alignment_grade(
                 normalized.get("judge_status", JUDGE_STATUS_NO_SCORE),
                 normalized.get("model") or "unknown",
             )
+            # Persist the judge's reason even without a score so an analyst
+            # can find out WHY the judge produced no score by querying session
+            # records rather than grepping logs (the records-queryable goal of
+            # judge_status would be undermined without the accompanying reason).
+            if normalized.get("reason"):
+                record.llm_judge_reason = normalized["reason"]
         else:
             record.set_alignment_grade(
                 score,
--- a/packages/gptme-sessions/src/gptme_sessions/judge.py
+++ b/packages/gptme-sessions/src/gptme_sessions/judge.py
@@ -52,7 +52,11 @@ class JudgeMetadata(TypedDict):
 
 
 class JudgeVerdict(TypedDict, total=False):
-    score: float
+    # ``None`` when the judge returned a payload without a usable score.
+    # Never silently substituted with a neutral default — see
+    # ``_coerce_score`` and ``JUDGE_STATUS_NO_SCORE``.
+    score: float | None
+    judge_status: str
     reason: str
     model: str
     alignment_score: float | None
@@ -489,6 +493,25 @@ def _build_judge_meta(*, model: str) -> JudgeMetadata:
 VALID_ALIGNMENT_VALUES = frozenset({"on_track", "partial", "pivot", "off_target"})
 
 
+JUDGE_STATUS_OK = "ok"
+JUDGE_STATUS_NO_SCORE = "no_score"
+
+
+def _coerce_score(raw: Any) -> float | None:
+    """Clamp a raw judge score to 0.0-1.0, or None when it is unusable.
+
+    Returning None is deliberate: a missing or malformed ``score`` used to
+    become a neutral 0.5, which is indistinguishable from a judge that
+    genuinely scored the session at 0.5. Callers must branch on None.
+    """
+    if raw is None:
+        return None
+    try:
+        return max(0.0, min(1.0, float(raw)))
+    except (ValueError, TypeError):
+        return None
+
+
 def _coerce_alignment_score(raw: Any) -> float | None:
     """Normalize an alignment_score value to float 0.0-1.0 or None."""
     if raw is None:
@@ -520,8 +543,20 @@ def normalize_judge_verdict(payload: dict[str, Any]) -> JudgeVerdict:
     raw_meta = payload.get("meta")
     meta = raw_meta if isinstance(raw_meta, dict) else {}
     base_meta = _build_judge_meta(model=model)
+    score = _coerce_score(payload.get("score"))
+    status = str(payload.get("judge_status") or JUDGE_STATUS_OK)
+    if score is None:
+        status = JUDGE_STATUS_NO_SCORE
+        logger.warning(
+            "Judge verdict has no usable score (raw=%r, model=%s); "
+            "recording judge_status=%s instead of a neutral default",
+            payload.get("score"),
+            model or "unknown",
+            JUDGE_STATUS_NO_SCORE,
+        )
     verdict: JudgeVerdict = {
-        "score": max(0.0, min(1.0, float(payload.get("score", 0.5)))),
+        "score": score,
+        "judge_status": status,
         "reason": str(payload.get("reason", "")),
         "model": model,
         "alignment_score": _coerce_alignment_score(payload.get("alignment_score")),
@@ -563,10 +598,20 @@ def _parse_judge_payload(text: str, model: str) -> dict | None:
             logger.warning("LLM judge returned non-JSON response")
             return None
         verdict = json.loads(match.group(0))
-    score = float(verdict.get("score", 0.5))
+    score = _coerce_score(verdict.get("score"))
     reason = str(verdict.get("reason", ""))
-    score = max(0.0, min(1.0, score))
-    result: dict[str, Any] = {"score": score, "reason": reason, "model": model}
+    if score is None:
+        logger.warning(
+            "LLM judge returned a payload without a usable score (raw=%r, model=%s)",
+            verdict.get("score"),
+            model,
+        )
+    result: dict[str, Any] = {
+        "score": score,
+        "judge_status": JUDGE_STATUS_OK if score is not None else JUDGE_STATUS_NO_SCORE,
+        "reason": reason,
+        "model": model,
+    }
     # Phase 3: pass through intent-contract fields when the judge returns them
     alignment_score = _coerce_alignment_score(verdict.get("alignment_score"))
     if alignment_score is not None:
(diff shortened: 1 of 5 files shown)


TimeToBuildBob:
@greptileai review

packages/gptme-sessions/src/gptme_sessions/judge.py
```diff
@@ -489,6 +493,25 @@ def _build_judge_meta(*, model: str) -> JudgeMetadata:
...
+    """Clamp a raw judge score to 0.0-1.0, or None when it is unusable.
+
+    Returning None is deliberate: a missing or malformed ``score`` used to
+    become a neutral 0.5, which is indistinguishable from a judge that
+    genuinely scored the session at 0.5. Callers must branch on None.
+    """
+    if raw is None:
+        return None
+    try:
+        return max(0.0, min(1.0, float(raw)))
```
TimeToBuildBob: ⚠️ **P2** — _coerce_score's clamp does not reject NaN. float('NaN') does not raise, and in CPython min(1.0, nan) returns 1.0 because `nan < 1.0` is False, so the first argument is kept. max(0.0, 1.0) is then 1.0. Since json.loads accepts the NaN literal by default, a judge payload `{"score": NaN}` — or a string score like "NaN" via float("NaN") — coerces to 1.0, the maximum alignment score, instead of None. This silently inflates the reward signal for a score the judge never produced, exactly the class of behavior this PR aims to eliminate, and it is indistinguishable from a genuine 1.0 in the stored record. The docstring promises 'None when it is unusable', but only ValueError/TypeError are caught; NaN slips through and violates the new contract. The old clamp had the same flaw, but _coerce_score is the new single coercion point whose stated purpose is honest handling of unusable scores.

packages/gptme-sessions/src/gptme_sessions/post_session.py
```diff
@@ -642,6 +645,7 @@ def post_session(
             # than the session ran — likely truncated or misattributed) and the
             # caller observed real git-range commits. Trust the commits.
             outcome = "productive"
+            outcome_flip_reason = "unreliable_trajectory_caller_deliverables"
```
TimeToBuildBob: ⚠️ **P2** — The format-blind guard (_trajectory_format_blind, set at line ~506) flips trajectory_reliable to False even when the trajectory is fully duration-reliable — the premise of format-blindness is that wall-clock coverage is good but the extractor couldn't parse the tool-call shape. The outcome-determination branch `not trajectory_reliable and caller_deliverables` then stamps outcome_flip_reason = "unreliable_trajectory_caller_deliverables" for these sessions too. A consumer auditing why a session's outcome was promoted sees 'unreliable trajectory' for a session whose trajectory actually spanned the whole run (e.g. the 2026-07-15 gpt-5.6-sol Codex cluster, where tool_calls={} over 600s of 620s). The audit reason the PR adds therefore mislabels the mechanism: it claims the trajectory was distrusted for a duration mismatch when the real cause was extractor format-blindness. There is no distinct reason string for the format-blind case, so the new audit field actively misinforms rather than explaining the flip.

packages/gptme-sessions/src/gptme_sessions/judge.py
```diff
@@ -911,14 +956,30 @@ def write_alignment_grade(
                 except (json.JSONDecodeError, OSError):
                     pass
 
-        record.set_alignment_grade(
-            normalized["score"],
-            reason=normalized["reason"],
-            model=normalized["model"],
-        )
+        score = normalized.get("score")
+        if score is None:
```
TimeToBuildBob, line 963: ⚠️ **P2** — On the score-is-None branch of write_alignment_grade, the code stamps legacy_fields['judge_status'] = 'no_score' (line 982) but does not call set_alignment_grade, so neither llm_judge_reason nor grade_reasons['alignment'] is written — the judge's reason is dropped entirely and survives only in a log line. The PR's stated motivation for persisting judge_status is that score-less verdicts should be 'countable by querying the session records rather than grepping logs'. An analyst who finds judge_status=no_score on a record still cannot learn why the judge produced no score without grepping the logs: the reason that would explain it (e.g. 'judge forgot the score') is discarded. The parsed reason is available at this point via normalized['reason'] and could be stored alongside judge_status (for instance in _legacy_fields or grade_reasons without a grade). As written, the audit trail records the fact of no_score but not its cause, which undercuts the records-queryable audit goal the PR sets out.

TimeToBuildBob:
## 🤖 AI code review

This PR removes the silent defaulting of a judge's missing/unparseable score to 0.5. In judge.py, score is now float|None via _coerce_score(), verdicts carry a judge_status key (ok|no_score), write_alignment_grade() stamps judge_status and skips writing an alignment grade when score is None, and judge_and_writeback() reports status=no_score. In post_session.py and record.py, a new outcome_flip_reason field on SessionRecord is stamped at the two promotion sites to make outcome upgrades auditable. Tests were updated for the new judge_status key and six tests were added.

No P0/P1 on this head, but 3 P2 issues outstanding — only a head with zero findings scores 5/5.

**3 findings** · ⚠️ **3** P2 — no P0/P1 on this head; the P2s here keep the score at 4/5.

**⚠️ P2 medium** — `packages/gptme-sessions/src/gptme_sessions/judge.py:510`

_coerce_score's clamp does not reject NaN. float('NaN') does not raise, and in CPython min(1.0, nan) returns 1.0 because `nan < 1.0` is False, so the first argument is kept. max(0.0, 1.0) is then 1.0. Since json.loads accepts the NaN literal by default, a judge payload `{"score": NaN}` — or a string score like "NaN" via float("NaN") — coerces to 1.0, the maximum alignment score, instead of None. This silently inflates the reward signal for a score the judge never produced, exactly the class of behavior this PR aims to eliminate, and it is indistinguishable from a genuine 1.0 in the stored record. The docstring promises 'None when it is unusable', but only ValueError/TypeError are caught; NaN slips through and violates the new contract. The old clamp had the same flaw, but _coerce_score is the new single coercion point whose stated purpose is honest handling of unusable scores.

```suggestion
v = float(raw)
        if v != v:  # NaN
            return None
        return max(0.0, min(1.0, v))
```

<sub>**How this was verified:** Verified in CPython that min(1.0, float('nan')) == 1.0 and that json.loads('{"score": NaN}') succeeds, so _coerce_score(NaN) returns 1.0. The diff's new function is the sole gate for score parsing.</sub>

**⚠️ P2 medium** — `packages/gptme-sessions/src/gptme_sessions/post_session.py:648`

The format-blind guard (_trajectory_format_blind, set at line ~506) flips trajectory_reliable to False even when the trajectory is fully duration-reliable — the premise of format-blindness is that wall-clock coverage is good but the extractor couldn't parse the tool-call shape. The outcome-determination branch `not trajectory_reliable and caller_deliverables` then stamps outcome_flip_reason = "unreliable_trajectory_caller_deliverables" for these sessions too. A consumer auditing why a session's outcome was promoted sees 'unreliable trajectory' for a session whose trajectory actually spanned the whole run (e.g. the 2026-07-15 gpt-5.6-sol Codex cluster, where tool_calls={} over 600s of 620s). The audit reason the PR adds therefore mislabels the mechanism: it claims the trajectory was distrusted for a duration mismatch when the real cause was extractor format-blindness. There is no distinct reason string for the format-blind case, so the new audit field actively misinforms rather than explaining the flip.

```suggestion
outcome_flip_reason = "unreliable_trajectory_caller_deliverables"
```

<sub>**How this was verified:** test_post_session_format_blind_trajectory_keeps_caller_deliverables exercises exactly this path (trajectory_reliable=True before the guard, set False by _trajectory_format_blind with a duration-reliable 600s/620s) and asserts outcome 'productive' but never checks outcome_flip_reason, so the mislabel is unguarded.</sub>

**⚠️ P2 medium** — `packages/gptme-sessions/src/gptme_sessions/judge.py:960`

On the score-is-None branch of write_alignment_grade, the code stamps legacy_fields['judge_status'] = 'no_score' (line 982) but does not call set_alignment_grade, so neither llm_judge_reason nor grade_reasons['alignment'] is written — the judge's reason is dropped entirely and survives only in a log line. The PR's stated motivation for persisting judge_status is that score-less verdicts should be 'countable by querying the session records rather than grepping logs'. An analyst who finds judge_status=no_score on a record still cannot learn why the judge produced no score without grepping the logs: the reason that would explain it (e.g. 'judge forgot the score') is discarded. The parsed reason is available at this point via normalized['reason'] and could be stored alongside judge_status (for instance in _legacy_fields or grade_reasons without a grade). As written, the audit trail records the fact of no_score but not its cause, which undercuts the records-queryable audit goal the PR sets out.

<sub>**How this was verified:** Traced write_alignment_grade: the else branch calls record.set_alignment_grade(score, reason=normalized['reason'], model=...), but the score-None branch only logs; legacy_fields gets judge_status, alignment_score, pivot_verdict; llm_judge_reason and grade_reasons are never touched. Confirmed normalized['reason'] is non-empty from _parse_judge_payload.</sub>

</details>

- `packages/gptme-sessions/src/gptme_sessions/judge.py` — Introduces _coerce_score() returning float|None, adds judge_status to JudgeVerdict/normalize/_parse, and makes write_alignment_grade/judge_and_writeback branch on a None score.
- `packages/gptme-sessions/src/gptme_sessions/post_session.py` — Adds outcome_flip_reason local, stamps it at the unreliable-trajectory and no-trajectory promotion sites, and forwards it into record_kwargs when set.
- `packages/gptme-sessions/src/gptme_sessions/record.py` — Adds the outcome_flip_reason: str | None field to SessionRecord with documentation comment.
- `packages/gptme-sessions/tests/test_judge.py` — Adds four score-less/unparseable-score tests, a no-grade-written test, and updates two exact-equality assertions with judge_status.
- `packages/gptme-sessions/tests/test_post_session.py` — Adds two tests asserting outcome_flip_reason is set on promotion and None when no flip occurred.

</details>

<sub>Reviewed `7de5b1c45eb6` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 135s · reports findings; merge decisions belong to the gate · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

packages/gptme-sessions/src/gptme_sessions/judge.py
TimeToBuildBob: Fixed in `1ab55eaa`: added `if v != v: return None` before the clamp in `_coerce_score` — the NaN-ness property (`x != x` is True only for NaN) catches it before `min`/`max` can evaluate it. New test `test_parse_judge_payload_with_nan_score_is_no_score` covers the case.

packages/gptme-sessions/src/gptme_sessions/post_session.py
TimeToBuildBob: Fixed in `1ab55eaa`: added a `_format_blind` bool that's set to `True` when `_trajectory_format_blind()` downgrades `trajectory_reliable`. The flip-reason branch now emits `"format_blind_trajectory_caller_deliverables"` in that case and `"unreliable_trajectory_caller_deliverables"` only for the original duration-truncation cause. New assertion added to `test_post_session_format_blind_trajectory_keeps_caller_deliverables`.

packages/gptme-sessions/src/gptme_sessions/judge.py
TimeToBuildBob: Fixed in `1ab55eaa`: when `score is None`, `write_alignment_grade` now sets `record.llm_judge_reason = normalized["reason"]` (when the reason is non-empty) so the cause is queryable from session records. New test `test_score_less_verdict_persists_reason_for_audit` verifies this. `grade_reasons["alignment"]` and `grades["alignment"]` remain unset — no partial grade is written.

PR #1460, opened 2026-08-19
fix(lessons-extras): companion-path suggestions must include the category subdir

TimeToBuildBob:
## Problem

The companion-link **detection** regex already allows an optional category subdir:

```python
rf"knowledge/lessons/(?:[^/]+/)?{re.escape(self.filepath.stem)}\.md"
```

but all three **suggestion** messages hardcoded the flat `knowledge/lessons/<stem>.md`. Companions actually live under a category subdir mirroring the lesson's own, so the validator pointed authors at a path that does not exist — and the link written from that suggestion then failed the `check-markdown-links` hook.

Concretely, for `lessons/monitoring/alerts-must-carry-their-diagnosis.md` the validator said:

> Companion doc exists but not linked. Add to Related section: `knowledge/lessons/alerts-must-carry-their-diagnosis.md`

while the companion is at `knowledge/lessons/**monitoring**/alerts-must-carry-their-diagnosis.md`. Following the suggestion verbatim produced a broken link and a failed commit.

## Fix

- **Companion exists** → report where it actually is. `companion_matches` already holds the resolved path from `rglob`; it was computed and then discarded.
- **Companion missing** → mirror the lesson's own category dir, falling back to the root for lessons not in one.

Note `_check_related_section` (line ~602) already got this right with a `<subdir>` placeholder — this brings the other three messages in line.

## Tests

Two regression tests, one per branch (exists-but-unlinked, and dead-link). Both **fail on the pre-fix validator** with the flat path and pass after:

```
AssertionError: assert 'knowledge/lessons/monitoring/test-lesson.md' in
  'Links a companion doc that does not exist. ... knowledge/lessons/test-lesson.md'
```

Full suite: 33 passed.

## Scope

Message strings only — no change to detection, validation outcomes, or exit codes. A lesson that validated before still validates.

### DIFF
--- a/packages/gptme-lessons-extras/src/gptme_lessons_extras/validate.py
+++ b/packages/gptme-lessons-extras/src/gptme_lessons_extras/validate.py
@@ -552,12 +552,15 @@ class LessonValidator:
             None,
         )
 
-        # Check if linked in Related section (allow optional subdirectory component).
+        # Check if linked in Related section (allow zero or more subdirectory components).
         # Capture the exact matched path so we can verify it exists separately:
         # a companion in a *different* category does not validate a link that
         # points at a non-existent own-category path.
+        # Use [a-zA-Z0-9._-]+ for path components to stay within the path
+        # boundary and avoid greedily consuming markdown link punctuation
+        # (e.g. ](  or  )  after the path).
         _companion_link_match = re.search(
-            rf"(knowledge/lessons/(?:[^/]+/)?{re.escape(self.filepath.stem)}\.md)",
+            rf"(knowledge/lessons/(?:[a-zA-Z0-9._-]+/)*{re.escape(self.filepath.stem)}\.md)",
             self.content,
             re.IGNORECASE,
         )
--- a/packages/gptme-lessons-extras/src/gptme_lessons_extras/validate.py
+++ b/packages/gptme-lessons-extras/src/gptme_lessons_extras/validate.py
@@ -577,17 +577,13 @@ class LessonValidator:
         # Suggested companion path. When the companion already exists in the
         # lesson's own category subdir, name its actual location so that any
         # Related link written from this suggestion passes markdown-link checks.
-        # Otherwise mirror the lesson's own category dir so the author creates
-        # the file in the right place from the start.
+        # Otherwise mirror the full category path so the author creates the file
+        # in the right place from the start (including nested categories).
         if own_companion is not None:
             suggested_companion = own_companion.as_posix()
         else:
-            # Mirror the lesson's own category dir, unless it sits at the root.
-            category = self.filepath.parent.name
             target_dir = (
-                COMPANION_DIR
-                if category in ("", ".", "lessons")
-                else COMPANION_DIR / category
+                COMPANION_DIR if lesson_cat == Path(".") else COMPANION_DIR / lesson_cat
             )
             suggested_companion = (target_dir / f"{self.filepath.stem}.md").as_posix()
 
--- a/packages/gptme-lessons-extras/src/gptme_lessons_extras/validate.py
+++ b/packages/gptme-lessons-extras/src/gptme_lessons_extras/validate.py
@@ -511,9 +511,21 @@ class LessonValidator:
         # companion linking or length on superseded files.
         if self.status == "archived":
             return
-        # Check if companion doc exists (search subdirectories too)
+        # Check if companion doc exists (search subdirectories too).
+        # Only count a match as *this* lesson's companion when it lives in the
+        # same category subdir — a match in a different category shares only the
+        # stem name, not the intent.  Falling back to an arbitrary cross-category
+        # match would send the author to the wrong document.
         companion_matches = list(COMPANION_DIR.rglob(f"{self.filepath.stem}.md"))
-        has_companion = len(companion_matches) > 0
+        own_companion = next(
+            (
+                m
+                for m in companion_matches
+                if m.parent.name == self.filepath.parent.name
+            ),
+            None,
+        )
+        has_companion = own_companion is not None
 
         # Check if linked in Related section (allow optional subdirectory component)
         has_companion_link = bool(
@@ -530,23 +542,13 @@ class LessonValidator:
         body_start = len(self.content[:frontmatter_end].split("\n"))
         body_lines = len(lines) - body_start
 
-        # Suggested companion path. When the companion already exists, name where
-        # it ACTUALLY is — companions live under a category subdir mirroring the
-        # lesson's own (knowledge/lessons/<category>/<stem>.md), so a flat
-        # knowledge/lessons/<stem>.md suggestion sends the author to a path that
-        # does not exist and produces a link the markdown-link hook then rejects.
-        if has_companion:
-            # Prefer the companion whose parent dir matches the lesson's own category.
-            # rglob order is filesystem-dependent, so [0] is non-deterministic when
-            # two lessons in different category subdirs share the same stem.
-            suggested_companion = next(
-                (
-                    m
-                    for m in companion_matches
-                    if m.parent.name == self.filepath.parent.name
-                ),
-                companion_matches[0],
-            ).as_posix()
+        # Suggested companion path. When the companion already exists in the
+        # lesson's own category subdir, name its actual location so that any
+        # Related link written from this suggestion passes markdown-link checks.
+        # Otherwise mirror the lesson's own category dir so the author creates
+        # the file in the right place from the start.
+        if own_companion is not None:
+            suggested_companion = own_companion.as_posix()
         else:
             # Mirror the lesson's own category dir, unless it sits at the root.
             category = self.filepath.parent.name
--- a/packages/gptme-lessons-extras/src/gptme_lessons_extras/validate.py
+++ b/packages/gptme-lessons-extras/src/gptme_lessons_extras/validate.py
@@ -553,22 +553,37 @@ class LessonValidator:
         )
 
         # Check if linked in Related section (allow zero or more subdirectory components).
-        # Capture the exact matched path so we can verify it exists separately:
-        # a companion in a *different* category does not validate a link that
-        # points at a non-existent own-category path.
+        # Use finditer rather than search so ALL occurrences are captured — a prose
+        # mention of an existing cross-category companion earlier in the document must
+        # not suppress the dead-link error on a later Related-section link that points
+        # at a non-existent own-category path.
         # Use [a-zA-Z0-9._-]+ for path components to stay within the path
         # boundary and avoid greedily consuming markdown link punctuation
         # (e.g. ](  or  )  after the path).
-        _companion_link_match = re.search(
-            rf"(knowledge/lessons/(?:[a-zA-Z0-9._-]+/)*{re.escape(self.filepath.stem)}\.md)",
-            self.content,
-            re.IGNORECASE,
+        _companion_link_matches = list(
+            re.finditer(
+                rf"(knowledge/lessons/(?:[a-zA-Z0-9._-]+/)*{re.escape(self.filepath.stem)}\.md)",
+                self.content,
+                re.IGNORECASE,
+            )
         )
-        has_companion_link = _companion_link_match is not None
-        # A link is "live" only when the specific path it names exists on disk.
-        linked_companion_exists = (
-            _companion_link_match is not None
-            and Path(_companion_link_match.group(1)).exists()
+        has_companion_link = bool(_companion_link_matches)
+        # A link is "live" only when ALL companion path mentions exist on disk.
+        # If any mention names a non-existent path it is a dead reference — even when
+        # earlier mentions happen to be valid (e.g. cross-category prose references).
+        linked_companion_exists = has_companion_link and all(
+            Path(m.group(1)).exists() for m in _companion_link_matches
+        )
+        # Is the own-category companion's exact path specifically mentioned?
+        # has_companion_link is True for ANY companion mention (cross-category too);
+        # this narrower flag checks only for the lesson's own companion path so the
+        # "exists but not linked" warning is not silenced by an unrelated mention.
+        own_companion_linked = own_companion is not None and bool(
+            re.search(
+                re.escape(own_companion.as_posix()),
+                self.content,
+                re.IGNORECASE,
+            )
         )
 
         # Get line count
@@ -598,20 +613,31 @@ class LessonValidator:
             )
 
         # If lesson links a companion that does not exist, that is a dead reference.
-        # Verified against the actual path in the link — a companion in a different
-        # category does not make a stale or flat link live.
+        # Verified against ALL matched paths — a companion in a different category
+        # does not make a stale or flat link live.
         # Note: the dead link may use a flat path while the canonical location is
         # category-prefixed — remind the author to update the link too.
         if has_companion_link and not linked_companion_exists:
-            self.errors.append(
-                f"Links a companion doc that does not exist. Either create it at "
-                f"{suggested_companion} (and update the Related link to match), "
-                f"or remove the link."
-            )
+            if own_companion is not None:
+                # Companion exists but the Related link doesn't point to it — guide
+                # the author to fix the link rather than create a duplicate file.
+                self.errors.append(
+                    f"Links a companion doc that does not exist. Update the Related "
+                    f"link to point to the correct path: {suggested_companion}, "
+                    f"or remove the link."
+                )
+            else:
+                self.errors.append(
+                    f"Links a companion doc that does not exist. Either create it at "
+                    f"{suggested_companion} (and update the Related link to match), "
+                    f"or remove the link."
+                )
 
-        # If own-category companion exists but is not linked, warn.
-        # Cross-category companions are not flagged — they belong to other lessons.
-        if own_companion is not None and not has_companion_link:
+        # If own-category companion exists but is not specifically linked, warn.
+        # Checked against own_companion_linked (not has_companion_link) so that
+        # a cross-category mention elsewhere in the document doesn't silence the
+        # warning — cross-category companions belong to other lessons.
+        if own_companion is not None and not own_companion_linked:
             self.warnings.append(
                 f"Companion doc exists but not linked. Add to Related section: "
                 f"{suggested_companion}"
--- a/packages/gptme-lessons-extras/src/gptme_lessons_extras/validate.py
+++ b/packages/gptme-lessons-extras/src/gptme_lessons_extras/validate.py
@@ -618,7 +618,21 @@ class LessonValidator:
         # Note: the dead link may use a flat path while the canonical location is
         # category-prefixed — remind the author to update the link too.
         if has_companion_link and not linked_companion_exists:
-            if own_companion is not None:
+            dead_paths = [
+                m.group(1)
+                for m in _companion_link_matches
+                if not Path(m.group(1)).exists()
+            ]
+            if own_companion is not None and own_companion_linked:
+                # Own companion IS correctly linked — the dead reference is a different
+                # mention (e.g. a prose cross-category reference). Report only the dead
+                # path(s) so the author knows what to remove or fix.
+                dead_list = ", ".join(dead_paths)
+                self.errors.append(
+                    f"Links a non-existent companion path. Remove or fix the dead "
+                    f"reference: {dead_list}"
+                )
+            elif own_companion is not None:
                 # Companion exists but the Related link doesn't point to it — guide
                 # the author to fix the link rather than create a duplicate file.
                 self.errors.append(
--- a/packages/gptme-lessons-extras/src/gptme_lessons_extras/validate.py
+++ b/packages/gptme-lessons-extras/src/gptme_lessons_extras/validate.py
@@ -536,7 +536,17 @@ class LessonValidator:
         # knowledge/lessons/<stem>.md suggestion sends the author to a path that
         # does not exist and produces a link the markdown-link hook then rejects.
         if has_companion:
-            suggested_companion = companion_matches[0].as_posix()
+            # Prefer the companion whose parent dir matches the lesson's own category.
+            # rglob order is filesystem-dependent, so [0] is non-deterministic when
+            # two lessons in different category subdirs share the same stem.
+            suggested_companion = next(
+                (
+                    m
+                    for m in companion_matches
+                    if m.parent.name == self.filepath.parent.name
+                ),
+                companion_matches[0],
+            ).as_posix()
         else:
             # Mirror the lesson's own category dir, unless it sits at the root.
             category = self.filepath.parent.name
@@ -554,11 +564,14 @@ class LessonValidator:
                 f"Consider creating companion: {suggested_companion}"
             )
 
-        # If lesson links a companion that does not exist, that is a dead reference
+        # If lesson links a companion that does not exist, that is a dead reference.
+        # Note: the dead link may use a flat path while the canonical location is
+        # category-prefixed — remind the author to update the link too.
         if has_companion_link and not has_companion:
             self.errors.append(
-                f"Links a companion doc that does not exist. Create it or remove the "
-                f"link: {suggested_companion}"
+                f"Links a companion doc that does not exist. Either create it at "
+                f"{suggested_companion} (and update the Related link to match), "
+                f"or remove the link."
             )
 
         # If companion exists but not linked, warn
--- a/packages/gptme-lessons-extras/src/gptme_lessons_extras/validate.py
+++ b/packages/gptme-lessons-extras/src/gptme_lessons_extras/validate.py
@@ -33,6 +33,28 @@ VALID_TARGET_GRADE_FIELDS = frozenset(
 )
 
 
+def _lesson_category(filepath: Path) -> Path:
+    """Return the lesson's category path relative to the lessons root.
+
+    Handles single-level and deeper nesting without relying on a global
+    LESSONS_DIR constant.
+
+    Examples::
+
+        /abs/path/lessons/monitoring/foo.md  → Path("monitoring")
+        /abs/path/lessons/foo/bar/foo.md     → Path("foo/bar")
+        lessons/monitoring/foo.md            → Path("monitoring")
+    """
+    parts = filepath.parts
+    # Find the rightmost "lessons" component not immediately preceded by "knowledge".
+    for i in range(len(parts) - 1, -1, -1):
+        if parts[i] == "lessons" and (i == 0 or parts[i - 1] != "knowledge"):
+            cat_parts = parts[i + 1 : -1]  # segments between "lessons" and the stem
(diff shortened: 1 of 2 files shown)


packages/gptme-lessons-extras/src/gptme_lessons_extras/validate.py
```diff
@@ -530,25 +530,42 @@ def _check_companion_doc(self):
...
+            suggested_companion = companion_matches[0].as_posix()
+        else:
+            # Mirror the lesson's own category dir, unless it sits at the root.
+            category = self.filepath.parent.name
+            target_dir = (
+                COMPANION_DIR
+                if category in ("", ".", "lessons")
+                else COMPANION_DIR / category
+            )
+            suggested_companion = (target_dir / f"{self.filepath.stem}.md").as_posix()
```
TimeToBuildBob, line 609: ⚠️ **P2** — In the dead-link branch (has_companion_link and not has_companion) the suggested path is derived from the lesson's own category dir, but the broken reference is the link the author wrote in the Related section, which can point to a different subdir or to the flat knowledge/lessons/<stem>.md path. The error then reads "Links a companion doc that does not exist. Create it or remove the link: <category>/<stem>.md", naming a location that does not match the dead reference. The new test sets up this exact case: the lesson sits at lessons/monitoring/test-lesson.md while its link text is '../../knowledge/lessons/test-lesson.md' (flat). An author who follows the suggestion verbatim creates knowledge/lessons/monitoring/test-lesson.md, but the flat link is still unresolvable — the validator re-flags the same error and the markdown-link hook still fails, so the suggestion does not fix the problem it reports. The message should name, or at least reconcile with, the path the link actually references.

packages/gptme-lessons-extras/src/gptme_lessons_extras/validate.py
```diff
@@ -530,25 +530,42 @@ def _check_companion_doc(self):
         body_start = len(self.content[:frontmatter_end].split("\n"))
         body_lines = len(lines) - body_start
 
+        # Suggested companion path. When the companion already exists, name where
+        # it ACTUALLY is — companions live under a category subdir mirroring the
+        # lesson's own (knowledge/lessons/<category>/<stem>.md), so a flat
+        # knowledge/lessons/<stem>.md suggestion sends the author to a path that
+        # does not exist and produces a link the markdown-link hook then rejects.
+        if has_companion:
+            suggested_companion = companion_matches[0].as_posix()
```
TimeToBuildBob: ⚠️ **P2** — companion_matches comes from COMPANION_DIR.rglob(f"{self.filepath.stem}.md") and can contain several files when two lessons in different category subdirs share the same stem. [0] is whichever rglob yields first, and Path.rglob follows os.scandir order, which is not sorted and not guaranteed to visit the lesson's own category first. Because has_companion is True whenever any match exists, the exists-but-not-linked warning fires and then suggests the arbitrary first match. If that match belongs to a different category, the warning tells the author to "Add to Related section: knowledge/lessons/<wrong-category>/<stem>.md". Following it produces a link to a live file (so the markdown-link hook passes) but to the wrong lesson's companion, leaving the intended companion unlinked. Filtering companion_matches to prefer the entry whose parent name equals self.filepath.parent.name would make the suggestion deterministic and correct.

TimeToBuildBob:
## 🤖 AI code review

Adds a `_lesson_category` helper that derives a lesson's category path from its filesystem location, and uses it to restrict companion matching to the same category, to generate category-prefixed suggestion paths, and to switch companion-link detection from `re.search` to `re.finditer` so all path mentions are checked for existence. Updates the test suite with several category-aware regression tests.

### Confidence Score: N/A

Round cap reached — findings below are non-blocking (informational), so there is no clean verdict on this head.

**1 finding** · ⚠️ **1** P2 — non-blocking (round cap reached).

**⚠️ P2 medium · ℹ️ non-blocking** — `packages/gptme-lessons-extras/src/gptme_lessons_extras/validate.py:50`

`_lesson_category` scans the path from the right and selects the first (rightmost) component equal to "lessons" that is not preceded by "knowledge". When a lesson lives in a category that itself contains a directory named "lessons" (e.g. `lessons/data/lessons/foo.md`), the rightmost "lessons" is the category component, not the root `lessons` directory. The function then returns `Path(".")` instead of `Path("data/lessons")`. As a result, the own-companion match in `_check_companion_doc` compares `m.relative_to(COMPANION_DIR).parent` (which is `data/lessons`) against `Path(".")` and fails, so the existing companion is not recognized. The validator then emits a false "Companion doc exists but not linked" warning and suggests creating a new flat companion at `knowledge/lessons/foo.md`, sending the author to a non-existent path and potentially orphaning the real companion.

```suggestion
for i in range(len(parts)):
        if parts[i] == "lessons" and (i == 0 or parts[i - 1] != "knowledge"):
            cat_parts = parts[i + 1 : -1]
            return Path(*cat_parts) if cat_parts else Path(".")
    return Path(filepath.parent.name)
```

<sub>**How this was verified:** Traced the logic for a lesson at `lessons/data/lessons/foo.md` with a companion at `knowledge/lessons/data/lessons/foo.md`. The rightmost `lessons` is at index 2; `parts[1]` is `data` (not `knowledge`), so it is selected, and `cat_parts = parts[3:-1] = []`, yielding `Path(".")`. The relative parent of the companion is `data/lessons`, which does not equal `Path(".")`, so `own_companion` is `None`.</sub>

</details>

- `packages/gptme-lessons-extras/src/gptme_lessons_extras/validate.py` — Adds `_lesson_category`; reworks companion matching and suggestions to be category-aware; changes companion-link detection to use `finditer` and checks all mentions for existence.
- `packages/gptme-lessons-extras/tests/test_validate.py` — Adds tests covering category-aware companion suggestions, cross-category exclusion, nested categories, space-containing categories, and dead-prose-mention handling.

</details>

| `1c6876125ae1` | 4/5 | 2 | llm | 2026-08-19 12:00 UTC |
| `7cf79bebf25d` | 4/5 | 1 | llm | 2026-08-19 13:05 UTC |
| `848fafce63cd` | 2/5 | 2 | llm | 2026-08-19 13:55 UTC |
| `aea8df58511d` | 3/5 | 1 | llm | 2026-08-19 17:42 UTC |
| `704a34de7575` | 3/5 | 1 | llm | 2026-08-19 18:40 UTC |
| `bc969fae031e` | 3/5 | 3 | llm | 2026-08-19 19:25 UTC |
| `edaae34ce71c` | ? | 1 | llm | 2026-08-19 20:32 UTC |
| `709fa280d400` | ? | 1 | llm | 2026-08-19 23:07 UTC |
| `c27e238eb79d` | 3/5 | 1 | llm | 2026-08-20 00:01 UTC |

</details>

<sub>Reviewed `e98214fecaa5` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 194s · reports findings; merge decisions belong to the gate · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

TimeToBuildBob:
Addressed the two P2 findings in 7cf79beb:

**P2-2 (non-deterministic companion selection):** `companion_matches[0]` could return the wrong category when two lessons share the same stem in different subdirs. Now uses `next((m for m in companion_matches if m.parent.name == self.filepath.parent.name), companion_matches[0])` — prefers the companion whose parent dir matches the lesson's own category, with `[0]` as a fallback.

**P2-1 (dead-link message doesn't mention updating the link):** When the existing Related link is flat (`knowledge/lessons/<stem>.md`) but the canonical location is category-prefixed, creating the file at the suggested path still leaves the old link broken. Updated the error to say: *"Either create it at `<canonical>` (and update the Related link to match), or remove the link."*

All 33 tests still pass.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Addressed the remaining P2 in 848fafce:

**P2 (cross-category fallback):** When no companion in `companion_matches` had a parent dir matching the lesson's own category, the code fell back to `companion_matches[0]`, which could be a companion from a different lesson in a different category. Now `own_companion` is computed upfront (`next(... if parent.name matches ..., None)`), and `has_companion` derives from it — so a cross-category match is never counted as this lesson's companion and never appears in suggestions or warnings.

Added a regression test: two lessons share the same stem in different category subdirs; only the cross-category companion exists; the monitoring lesson must not see it as "exists but not linked" and the suggested path must mirror its own category.

34 → 34 tests pass (33 prior + 1 new cross-category test).

packages/gptme-lessons-extras/src/gptme_lessons_extras/validate.py
```diff
@@ -511,9 +511,21 @@ def _check_companion_doc(self):
...
-        has_companion = len(companion_matches) > 0
+        own_companion = next(
+            (
+                m
+                for m in companion_matches
+                if m.parent.name == self.filepath.parent.name
+            ),
+            None,
+        )
+        has_companion = own_companion is not None
```
TimeToBuildBob: ❌ **P1** — This line changes the meaning of has_companion from 'a companion exists for this stem anywhere under knowledge/lessons' to 'a companion exists in the same category subdir as the lesson'. That is a validation-outcome change, not a message-only change as the PR description claims. A lesson in lessons/monitoring that links to knowledge/lessons/workflow/test-lesson.md (with that file existing) previously satisfied has_companion and has_companion_link, so it validated clean; now own_companion is None, has_companion becomes False, and the dead-link error fires: 'Links a companion doc that does not exist…'. Any existing lesson that intentionally references a companion outside its own category will now fail validation and block the commit. The PR's stated scope ('No change to detection, validation outcomes, or exit codes') is therefore incorrect, and this is a silent breaking change for callers. If the intent is to enforce same-category companions, the PR should call that out; if not, the dead-link check should still accept a cross-category link when the target exists.

TimeToBuildBob:
Addressed the two P1 findings in aea8df5:

**P1-1 (shallow leaf-name comparison, line 524):** `m.parent.name == self.filepath.parent.name` only compared the final path segment, which mismatches for deeper nesting (`lessons/foo/bar/`) or two categories that happen to share the same leaf name. Replaced with a full category-relative path comparison via a new `_lesson_category()` helper:

```python
m.relative_to(COMPANION_DIR).parent == _lesson_category(self.filepath)
```

`_lesson_category()` finds the rightmost `lessons` component in the filepath (not preceded by `knowledge`) and returns all path parts between it and the stem — correct for both single-level and deeper nesting.

**P1-2 (has_companion semantics change, line 528):** The previous iteration set `has_companion = own_companion is not None`, which falsely fires a dead-link error when a lesson explicitly links to an existing companion in a different category. Restored correctness by capturing the actual path the lesson links to and checking if that specific file exists on disk:

```python
linked_companion_exists = (
    _companion_link_match is not None
    and Path(_companion_link_match.group(1)).exists()
)
```

Dead-link check now uses `has_companion_link and not linked_companion_exists`. The "exists but not linked" warning is unchanged (still fires only for own-category companions).

Added `test_cross_category_link_to_existing_companion_not_dead` to cover the P1-2 scenario. All 35 tests pass.

packages/gptme-lessons-extras/src/gptme_lessons_extras/validate.py
```diff
@@ -530,25 +574,48 @@ def _check_companion_doc(self):
...
+        # Suggested companion path. When the companion already exists in the
+        # lesson's own category subdir, name its actual location so that any
+        # Related link written from this suggestion passes markdown-link checks.
+        # Otherwise mirror the lesson's own category dir so the author creates
+        # the file in the right place from the start.
+        if own_companion is not None:
+            suggested_companion = own_companion.as_posix()
+        else:
+            # Mirror the lesson's own category dir, unless it sits at the root.
+            category = self.filepath.parent.name
```
TimeToBuildBob: ❌ **P1** — For lessons nested more than one level under a category (e.g. `lessons/foo/bar/lesson.md`), the suggested companion path is built from the immediate parent name (`bar`) instead of the full category path (`foo/bar`). `_lesson_category` correctly computes the full category, but this branch discards it and recomputes from `self.filepath.parent.name`. As a result, when no own-category companion exists, the validator suggests `knowledge/lessons/bar/lesson.md` instead of `knowledge/lessons/foo/bar/lesson.md`. The author following the suggestion creates the file in the wrong directory, and the subsequent markdown-link check fails because the link written from the suggestion points to a nonexistent path. This contradicts the PR's own stated support for nested categories (the `_lesson_category` docstring lists such an example). Fix by using the already-computed `lesson_cat` to derive the target directory.

TimeToBuildBob:
Addressed the P1 in aea8df5 in 704a34de:

**P1 (else branch uses parent.name, not full category path):** In the `else` branch of the suggested companion path, the code used `self.filepath.parent.name` (the immediate parent directory only), so `lessons/foo/bar/lesson.md` would suggest `knowledge/lessons/bar/lesson.md` instead of `knowledge/lessons/foo/bar/lesson.md`. The PR had already added `_lesson_category()` and used it for `own_companion` filtering, but the `else` branch was not updated.

Fix: use the already-computed `lesson_cat` variable in the `else` branch:
```python
target_dir = COMPANION_DIR if lesson_cat == Path(".") else COMPANION_DIR / lesson_cat
```

Added `test_missing_companion_suggestion_mirrors_full_nested_category` which creates a lesson at `lessons/outer/inner/test-lesson.md` and verifies the suggestion is `knowledge/lessons/outer/inner/test-lesson.md`.

All 36 tests pass.

TimeToBuildBob:
@TimeToBuildBob review

packages/gptme-lessons-extras/src/gptme_lessons_extras/validate.py
```diff
@@ -511,17 +533,42 @@ def _check_companion_doc(self):
...
+        _companion_link_match = re.search(
+            rf"(knowledge/lessons/(?:[a-zA-Z0-9._-]+/)*{re.escape(self.filepath.stem)}\.md)",
+            self.content,
+            re.IGNORECASE,
+        )
+        has_companion_link = _companion_link_match is not None
+        # A link is "live" only when the specific path it names exists on disk.
+        linked_companion_exists = (
+            _companion_link_match is not None
+            and Path(_companion_link_match.group(1)).exists()
```
TimeToBuildBob: ❌ **P1** — linked_companion_exists is computed from _companion_link_match.group(1), which is the first substring matching knowledge/lessons/.../<stem>.md anywhere in self.content, not specifically the Related-section href. Because re.search scans the whole document, an earlier prose mention of an existing companion (e.g. a cross-category file) will satisfy the existence check and suppress the dead-link error even when the actual Related link points to a missing path; conversely a non-existent mention before a valid Related link triggers a false error. For example, a monitoring lesson whose body says 'see knowledge/lessons/workflow/test-lesson.md' (file exists) and whose Related section links knowledge/lessons/monitoring/test-lesson.md (missing) will pass validation, so the broken link is not caught. The check must target only the link href, or compare the matched path against the link destination rather than the first arbitrary occurrence.

packages/gptme-lessons-extras/src/gptme_lessons_extras/validate.py
```diff
@@ -530,25 +577,44 @@ def _check_companion_doc(self):
...
+                f"Links a companion doc that does not exist. Either create it at "
+                f"{suggested_companion} (and update the Related link to match), "
+                f"or remove the link."
             )
 
-        # If companion exists but not linked, warn
-        if has_companion and not has_companion_link:
+        # If own-category companion exists but is not linked, warn.
+        # Cross-category companions are not flagged — they belong to other lessons.
+        if own_companion is not None and not has_companion_link:
```
TimeToBuildBob: ⚠️ **P2** — The 'exists but not linked' warning is gated on not has_companion_link, which is true if any knowledge/lessons/.../<stem>.md string appears anywhere in the content — including a link to a different category. So when a lesson's own-category companion exists (own_companion set) but the only companion link points to a cross-category file (e.g. knowledge/lessons/workflow/test-lesson.md), has_companion_link is true and the warning is skipped, even though the lesson's own companion is in fact unlinked. This contradicts the PR's stated intent that cross-category companions are not counted as this lesson's companion. The warning should fire whenever own_companion is set and the link does not target that specific path.

packages/gptme-lessons-extras/src/gptme_lessons_extras/validate.py
```diff
@@ -530,25 +577,44 @@ def _check_companion_doc(self):
...
+        # If lesson links a companion that does not exist, that is a dead reference.
+        # Verified against the actual path in the link — a companion in a different
+        # category does not make a stale or flat link live.
+        # Note: the dead link may use a flat path while the canonical location is
+        # category-prefixed — remind the author to update the link too.
+        if has_companion_link and not linked_companion_exists:
             self.errors.append(
-                f"Links a companion doc that does not exist. Create it or remove the "
-                f"link: knowledge/lessons/{self.filepath.stem}.md"
+                f"Links a companion doc that does not exist. Either create it at "
```
TimeToBuildBob: ⚠️ **P2** — The dead-link error message suggests 'create it at {suggested_companion}' even when the companion already exists. When own_companion is not None, suggested_companion is set to that existing file's path (line 586), so a lesson linking to a flat non-existent path while its category companion exists will be told to 'create' a file that is already there. The correct advice in that branch is to update the Related link to point to the existing path, not to create it. This misleads the author into duplicating a file or not understanding the real fix.

TimeToBuildBob:
Pushed `edaae34c` fixing the P1 and both P2 findings:

**P1 (re.search masks later dead links):** Switched from `re.search` (first match) to `re.finditer` (all matches). `linked_companion_exists` is now True only when ALL companion path mentions in the document resolve to existing files on disk — so a prose mention of a cross-category companion can no longer suppress a dead-link error on a non-existent Related-section path.

**P2 (line 614 — cross-category link silences own-companion warning):** Added `own_companion_linked` that checks whether the lesson's own-category companion path is specifically mentioned (not just any companion mention). The "exists but not linked" warning now fires whenever `own_companion` is set but its exact path is absent from the content.

**P2 (line 607 — "create it" when companion already exists):** When `own_companion is not None`, the error message now says "Update the Related link to point to the correct path: …" instead of "Either create it at …". This correctly guides the author to fix the link rather than create a duplicate file.

Tests: 39 passed (2 new regression tests for the P1 and P2-1 cases).

@TimeToBuildBob review

TimeToBuildBob:
CI pending (just pushed `edaae34c` ~7 minutes ago). Greptile trigger was sent at 12:27 but hasn't responded — likely dark (billing). Helper blocked re-trigger while the prior trigger is in-flight. Waiting for CI to green and Greptile to eventually review.

packages/gptme-lessons-extras/src/gptme_lessons_extras/validate.py
TimeToBuildBob: Confirmed fixed in edaae34c: `own_companion_linked` (validate.py ~L583) checks whether the lesson's own-category companion path is specifically mentioned, so a cross-category link no longer silences the 'exists but not linked' warning. Regression test `test_cross_category_link_does_not_silence_own_companion_unlinked_warning` covers this exact scenario. Verified locally in a clean venv: 39/39 tests pass.

TimeToBuildBob:
Status update: all AI-review findings are resolved (last outstanding thread confirmed fixed and resolved above; verified 39/39 tests pass in a clean venv, isolated from this sandbox's local module-shadowing false-failures).

CI is stuck at "queued" (never starts running) — checked across the repo, not just this PR: other open PRs (`fix/check-notifications-caching`, `feat/gptme-rag-source-collectors`) have runs queued since 20:03–21:02Z with zero progress too. This looks like a repo-wide Actions runner backlog/outage, not something a push to this PR can fix. I don't have runner-admin access to confirm directly (403 on the runners API). No action taken here beyond confirming it's infra-side — will pick back up once runners drain.

TimeToBuildBob:
Pushed `709fa280` fixing the remaining P2 finding from the `edaae34c` review.

**P2 (own companion linked but dead prose mention triggers wrong error):** When the lesson's own companion already exists and is correctly linked (e.g. `knowledge/lessons/monitoring/test-lesson.md` in Related), but the content also contains a dead companion path mention elsewhere (e.g. a prose cross-category reference like `knowledge/lessons/workflow/test-lesson.md` that doesn't exist), `linked_companion_exists` became False because `edaae34c` switched to checking ALL mentions. The `own_companion is not None` branch then fired with "Update the Related link to point to the correct path: …" — but the Related link was already correct. The message misdirected the author to alter or remove a valid link.

Fix: add an `own_companion_linked` guard in the dead-link error branch:
- `own_companion` set **and** `own_companion_linked` → the own companion is correctly linked; the dead reference is a different path → report `"Remove or fix the dead reference: <specific-dead-path(s)>"`
- `own_companion` set **and NOT** `own_companion_linked` → original "Update the Related link" message
- no own companion → original "Either create it at …" message

Added `test_own_companion_linked_but_dead_prose_mention_reports_dead_path`. All 40 tests pass.

packages/gptme-lessons-extras/src/gptme_lessons_extras/validate.py
```diff
@@ -530,25 +592,69 @@ def _check_companion_doc(self):
...
+                dead_list = ", ".join(dead_paths)
+                self.errors.append(
+                    f"Links a non-existent companion path. Remove or fix the dead "
+                    f"reference: {dead_list}"
+                )
+            elif own_companion is not None:
+                # Companion exists but the Related link doesn't point to it — guide
+                # the author to fix the link rather than create a duplicate file.
+                self.errors.append(
+                    f"Links a companion doc that does not exist. Update the Related "
```
TimeToBuildBob, line 642: ⚠️ **P2** — When an own-category companion exists but is not linked, and the dead link is a cross-category prose mention (not a flat path to the own companion), the `elif own_companion is not None:` branch fires and tells the author to "Update the Related link to point to the correct path: {suggested_companion}". This message is wrong: the dead reference is a different (cross-category) path, so instructing the author to point the Related link at the own companion leaves the actual dead reference unaddressed. The author either incorrectly rewrites a valid cross-category link or chases a non-existent Related-link problem, while the true broken reference persists. The branch should distinguish a flat-path dead link (e.g. `knowledge/lessons/<stem>.md`) to the own companion from an unrelated cross-category dead mention, and in the latter case report the specific dead path(s) as in the `own_companion_linked` branch.

packages/gptme-lessons-extras/src/gptme_lessons_extras/validate.py
```diff
@@ -511,14 +533,54 @@ def _check_companion_doc(self):
...
+        # Use finditer rather than search so ALL occurrences are captured — a prose
+        # mention of an existing cross-category companion earlier in the document must
+        # not suppress the dead-link error on a later Related-section link that points
+        # at a non-existent own-category path.
+        # Use [a-zA-Z0-9._-]+ for path components to stay within the path
+        # boundary and avoid greedily consuming markdown link punctuation
+        # (e.g. ](  or  )  after the path).
+        _companion_link_matches = list(
+            re.finditer(
+                rf"(knowledge/lessons/(?:[a-zA-Z0-9._-]+/)*{re.escape(self.filepath.stem)}\.md)",
```
TimeToBuildBob: ❌ **P1** — The companion-link detection regex narrowed its allowed path-component characters from the previous `[^/]+` (any character except slash) to `[a-zA-Z0-9._-]+`. This silently breaks detection for any existing lesson that links to a companion whose category subdir contains a space, unicode, or other character outside the new set. For example, a link to `knowledge/lessons/My Lesson/test-lesson.md` will no longer match, so `has_companion_link` becomes False, `linked_companion_exists` is never evaluated, and the lesson is treated as having no companion link. If the companion actually exists, the lesson now receives a spurious "Companion doc exists but not linked" warning; if the companion does not exist, the dead-link error is never raised. This is a regression for any repository that uses non-ASCII or spaced category names, and contradicts the PR's stated intent of only changing messages.

TimeToBuildBob:
Pushed `e98214fe` fixing the P1 finding.

**P1 (narrowed character class breaks space/unicode paths):** The fix is `[^\\/]+` — exclude only `]` (markdown link-text delimiter) and `/` (path separator).

The original `[a-zA-Z0-9._-]+` was added in `bc969fae` to prevent greedy cross-boundary matching in links of the form `[knowledge/lessons/foo/bar.md](../../knowledge/lessons/foo/bar.md)` — without a `]`-stop, `[^/]+/` would consume `bar.md](../../knowledge/` as a single path component. That guard was correct, but it was too narrow: it silently broke detection for any category dir containing a space or non-ASCII character.

`[^\\/]+` (excluding `]` and `/`) fixes both:
- Stops at `]` instead of consuming across the markdown link boundary ✓
- Accepts spaces, unicode, and other valid filesystem characters ✓

Added `test_companion_link_detected_for_category_with_space` to cover the space-in-category regression. All 41 tests pass.

TimeToBuildBob:
@TimeToBuildBob review

packages/gptme-lessons-extras/src/gptme_lessons_extras/validate.py
TimeToBuildBob: Fixed in `e98214fe`: changed `[a-zA-Z0-9._-]+` to `[^\\/]+` (exclude only `]` and `/`). This preserves the boundary-stop behavior (stopping at `]` prevents cross-boundary greedy matching in markdown links) while accepting spaces, unicode, and all other valid filesystem characters. Regression test `test_companion_link_detected_for_category_with_space` covers this exact case. All 41 tests pass.

TimeToBuildBob:
@TimeToBuildBob review

packages/gptme-lessons-extras/src/gptme_lessons_extras/validate.py
```diff
@@ -33,6 +33,28 @@
...
+
+    Examples::
+
+        /abs/path/lessons/monitoring/foo.md  → Path("monitoring")
+        /abs/path/lessons/foo/bar/foo.md     → Path("foo/bar")
+        lessons/monitoring/foo.md            → Path("monitoring")
+    """
+    parts = filepath.parts
+    # Find the rightmost "lessons" component not immediately preceded by "knowledge".
+    for i in range(len(parts) - 1, -1, -1):
```
TimeToBuildBob, line 50: ⚠️ **P2** — `_lesson_category` scans the path from the right and selects the first (rightmost) component equal to "lessons" that is not preceded by "knowledge". When a lesson lives in a category that itself contains a directory named "lessons" (e.g. `lessons/data/lessons/foo.md`), the rightmost "lessons" is the category component, not the root `lessons` directory. The function then returns `Path(".")` instead of `Path("data/lessons")`. As a result, the own-companion match in `_check_companion_doc` compares `m.relative_to(COMPANION_DIR).parent` (which is `data/lessons`) against `Path(".")` and fails, so the existing companion is not recognized. The validator then emits a false "Companion doc exists but not linked" warning and suggests creating a new flat companion at `knowledge/lessons/foo.md`, sending the author to a non-existent path and potentially orphaning the real companion.

PR #1465, opened 2026-08-19
feat(gptme-rag): add source-descriptor collection + voice-call de-accumulation (Phase 2.1)

TimeToBuildBob:
## Summary

Phase 2.1 of `upstream-retrieval-into-gptme-rag`: upstream the **source-descriptor collection** capability from the Bob brain script (`scripts/build-ambient-memory-index.py`) into `gptme-rag`, plus the **voice-call de-accumulation** format-normalization utility.

Follows the same port pattern as Phase 2.3 (`task_retrieval.py`, #1440): mechanism upstream, policy (which dirs are sources, always-on vs gated, boost numbers) stays consumer config.

## What's included

- `SourceDescriptor` + `SourceRegistry` — a declarative registry for document sources. Each source is a zero-arg collector returning `Document`s, tagged *always-on* (every build) or *gated* (full builds only). `collect()` flattens in registration order and isolates a raising collector so one bad source can't sink a build.
- `de_accumulate_transcript()` — recover full utterances from cumulative voice-call transcripts (longest text per contiguous same-role run).
- `collect_voice_call_documents()` — turn archived call JSON into `Document`s keyed `voicecall:<stem>`, with a repo-root safety guard.

## Why this shape

The brain script grew a registry (`_ALWAYS_ON_SOURCES` / `_MEMORY_TYPE_SOURCES`) precisely because ad-hoc `glob` loops per source were unmaintainable. The registry is the reusable mechanism; the source list and gating policy are Bob config.

## Testing

- 15 new tests in `tests/test_sources.py` (de-accumulation edge cases, registry gating + exception isolation, voice-call collection incl. bad-JSON and repo-root guard).
- `ruff` + `mypy` clean on the new module.

## Not included (later phases)

- memory_type classification + ranking boost (Phase 2.2)
- lesson matcher extraction (Phase 2.4)
- hook adoption + brain-script sunset (Phase 3–4)

Closes part of ErikBjare/bob `upstream-retrieval-into-gptme-rag` (Phase 2.1).

### DIFF
--- a/packages/gptme-rag/src/gptme_rag/sources.py
+++ b/packages/gptme-rag/src/gptme_rag/sources.py
@@ -184,14 +184,16 @@ def collect_voice_call_documents(
     voice_calls_dir = Path(voice_calls_dir)
     if not voice_calls_dir.exists():
         return []
-    if repo_root is not None and not voice_calls_dir.is_relative_to(repo_root):
+    if repo_root is not None and not voice_calls_dir.resolve().is_relative_to(
+        Path(repo_root).resolve()
+    ):
         return []
 
     docs: list[Document] = []
     for path in sorted(voice_calls_dir.glob("*.json")):
         try:
             data = json.loads(path.read_text(encoding="utf-8"))
-        except (OSError, json.JSONDecodeError):
+        except (OSError, UnicodeDecodeError, json.JSONDecodeError):
             continue
         if not isinstance(data, dict):
             continue
--- a/packages/gptme-rag/src/gptme_rag/sources.py
+++ b/packages/gptme-rag/src/gptme_rag/sources.py
@@ -44,13 +44,48 @@ logger = logging.getLogger(__name__)
 SourceCollector = Callable[[], Iterable[Document]]
 
 
+def _collapse_partial_chain(run_texts: list[str]) -> list[str]:
+    """Collapse cumulative STT partials within one same-role run.
+
+    A cumulative partial is a *prefix* of the utterance it is a partial of, so a
+    growing chain (``"I want"`` → ``"I want the budget"``) collapses to its final
+    form, and an exactly-repeated turn collapses to one copy.  Anything that is
+    not a prefix relative to its neighbour is a genuine follow-up turn and is
+    kept, because dropping it loses transcript content permanently.
+
+    Args:
+        run_texts: The raw texts of one contiguous same-role run, in order.
+
+    Returns:
+        The de-accumulated turns, in order, with empties removed.
+    """
+    kept: list[str] = []
+    for raw in run_texts:
+        text = raw.strip()
+        if not text:
+            continue
+        if kept:
+            previous = kept[-1]
+            if text.startswith(previous):
+                # Later partial extends (or exactly repeats) the earlier one.
+                kept[-1] = text
+                continue
+            if previous.startswith(text):
+                # Shorter partial arriving after its own longer form.
+                continue
+        kept.append(text)
+    return kept
+
+
 def de_accumulate_transcript(transcript: list[dict[str, Any]], *, cumulative: bool = False) -> str:
     """Render transcript turns, optionally collapsing cumulative STT partials.
 
-    With ``cumulative=True``, each contiguous same-role run is treated as a
-    sequence of cumulative speech-to-text partials and only its longest entry is
-    kept.  With the safe default, every entry is preserved; content alone cannot
-    reliably distinguish a cumulative partial from a genuine follow-up turn.
+    With ``cumulative=True``, each contiguous same-role run may contain
+    cumulative speech-to-text partials, so *prefix chains* within the run are
+    collapsed to their final form (see :func:`_collapse_partial_chain`).  Turns
+    that are not prefixes of their neighbour are genuine follow-up turns and are
+    preserved — a same-role run is not assumed to be a single utterance.  With
+    the safe default, every entry is preserved verbatim.
 
     Args:
         transcript: A list of turn dicts, each with ``role`` and ``text`` keys.
@@ -83,9 +118,10 @@ def de_accumulate_transcript(transcript: list[dict[str, Any]], *, cumulative: bo
             i += 1
 
         if cumulative:
-            best = max(run_texts, key=lambda text: len(text.strip()), default="")
+            kept = _collapse_partial_chain(run_texts)
         else:
-            best = "\n".join(text.strip() for text in run_texts if text.strip())
+            kept = [text.strip() for text in run_texts if text.strip()]
+        best = "\n".join(kept)
         if best.strip():
             turns.append(f"{current_role.upper()}: {best.strip()}")
     return "\n\n".join(turns)
@@ -198,7 +234,20 @@ def collect_voice_call_documents(
         return []
 
     docs: list[Document] = []
+    resolved_root = None if repo_root is None else Path(repo_root).resolve()
     for path in sorted(voice_calls_dir.glob("*.json")):
+        # Resolve *before* reading: a per-file symlink can escape the repo even
+        # though the directory itself passed the guard above, and reading it
+        # first would pull out-of-repo content into memory regardless of the
+        # later skip.
+        resolved_path = path.resolve()
+        if resolved_root is not None and not resolved_path.is_relative_to(resolved_root):
+            logger.warning("sources: skipping %s — resolves outside repo_root", path)
+            continue
+        if not resolved_path.is_file():
+            # Skip anything that is not a regular file (e.g. a FIFO named
+            # ``*.json``, which would block the collector on read).
+            continue
         try:
             data = json.loads(path.read_text(encoding="utf-8"))
         except (OSError, UnicodeDecodeError, json.JSONDecodeError):
@@ -221,15 +270,7 @@ def collect_voice_call_documents(
         source = str(data.get("source", "")) if isinstance(data.get("source"), str) else ""
         title = f"Voice call {date_str}" + (f" ({source})" if source else "")
 
-        try:
-            source_path = (
-                path if repo_root is None else path.resolve().relative_to(Path(repo_root).resolve())
-            )
-        except ValueError:
-            # path.resolve() points outside repo_root (e.g. a per-file symlink
-            # that escapes the repo even though the directory passed the guard).
-            logger.warning("sources: skipping %s — resolves outside repo_root", path)
-            continue
+        source_path = path if resolved_root is None else resolved_path.relative_to(resolved_root)
         metadata: dict[str, Any] = {
             "type": "voicecall",
             "source": str(source_path),
--- a/packages/gptme-rag/src/gptme_rag/sources.py
+++ b/packages/gptme-rag/src/gptme_rag/sources.py
@@ -213,7 +213,9 @@ def collect_voice_call_documents(
         source = str(data.get("source", "")) if isinstance(data.get("source"), str) else ""
         title = f"Voice call {date_str}" + (f" ({source})" if source else "")
 
-        source_path = path if repo_root is None else path.relative_to(repo_root)
+        source_path = (
+            path if repo_root is None else path.resolve().relative_to(Path(repo_root).resolve())
+        )
         metadata: dict[str, Any] = {
             "type": "voicecall",
             "source": str(source_path),
--- a/packages/gptme-rag/src/gptme_rag/sources.py
+++ b/packages/gptme-rag/src/gptme_rag/sources.py
@@ -82,14 +82,14 @@ def de_accumulate_transcript(transcript: list[dict[str, Any]]) -> str:
                 break
             run_texts.append(str(transcript[i].get("text", "")))
             i += 1
+
         # Detect cumulative runs: each entry is a prefix of the next (growing STT
         # partials, common in voice-call archives).  Only then is it safe to drop
         # all but the longest entry.  Independent consecutive same-role turns must
         # be joined to avoid silent data loss (P1 finding).
         def _each_is_prefix(texts: list[str]) -> bool:
             return all(
-                texts[j].strip().startswith(texts[j - 1].strip())
-                for j in range(1, len(texts))
+                texts[j].strip().startswith(texts[j - 1].strip()) for j in range(1, len(texts))
             )
 
         is_cumulative = len(run_texts) > 1 and _each_is_prefix(run_texts)
@@ -236,9 +236,7 @@ def collect_voice_call_documents(
 
         try:
             source_path = (
-                path
-                if repo_root is None
-                else path.resolve().relative_to(Path(repo_root).resolve())
+                path if repo_root is None else path.resolve().relative_to(Path(repo_root).resolve())
             )
         except ValueError:
             # path.resolve() points outside repo_root (e.g. a per-file symlink
--- a/packages/gptme-rag/src/gptme_rag/sources.py
+++ b/packages/gptme-rag/src/gptme_rag/sources.py
@@ -105,7 +105,11 @@ def de_accumulate_transcript(transcript: list[dict[str, Any]], *, cumulative: bo
         if not isinstance(transcript[i], dict):
             i += 1
             continue
-        current_role = str(transcript[i].get("role", ""))
+        role = transcript[i].get("role")
+        if not isinstance(role, str) or not role.strip():
+            i += 1
+            continue
+        current_role = role.strip()
         run_texts: list[str] = []
         # Collect all entries for this role run, skipping non-dict entries
         # mid-run without ending the run (P1 fix: a non-dict in the middle of
@@ -238,7 +242,12 @@ def collect_voice_call_documents(
 
     docs: list[Document] = []
     resolved_root = None if repo_root is None else Path(repo_root).resolve()
-    for path in sorted(voice_calls_dir.glob("*.json")):
+    try:
+        paths = sorted(voice_calls_dir.glob("*.json"))
+    except OSError:
+        logger.warning("sources: cannot list %s", voice_calls_dir)
+        return []
+    for path in paths:
         # Resolve *before* reading: a per-file symlink can escape the repo even
         # though the directory itself passed the guard above, and reading it
         # first would pull out-of-repo content into memory regardless of the
--- a/packages/gptme-rag/src/gptme_rag/sources.py
+++ b/packages/gptme-rag/src/gptme_rag/sources.py
@@ -203,7 +203,7 @@ def collect_voice_call_documents(
         missing or contains no parseable transcripts).
     """
     voice_calls_dir = Path(voice_calls_dir)
-    if not voice_calls_dir.exists():
+    if not voice_calls_dir.exists() or not voice_calls_dir.is_dir():
         return []
     if repo_root is not None and not voice_calls_dir.resolve().is_relative_to(
         Path(repo_root).resolve()
--- a/packages/gptme-rag/src/gptme_rag/sources.py
+++ b/packages/gptme-rag/src/gptme_rag/sources.py
@@ -65,9 +65,16 @@ def de_accumulate_transcript(transcript: list[dict[str, Any]]) -> str:
     turns: list[str] = []
     i = 0
     while i < len(transcript):
+        if not isinstance(transcript[i], dict):
+            i += 1
+            continue
         current_role = str(transcript[i].get("role", ""))
         run_texts: list[str] = []
-        while i < len(transcript) and transcript[i].get("role", "") == current_role:
+        while (
+            i < len(transcript)
+            and isinstance(transcript[i], dict)
+            and transcript[i].get("role", "") == current_role
+        ):
             run_texts.append(str(transcript[i].get("text", "")))
             i += 1
         best = max(run_texts, key=len) if run_texts else ""
@@ -185,6 +192,8 @@ def collect_voice_call_documents(
         try:
             data = json.loads(path.read_text(encoding="utf-8"))
         except (OSError, json.JSONDecodeError):
+            continue
+        if not isinstance(data, dict):
             continue
         transcript = data.get("transcript", [])
         if not isinstance(transcript, list) or not transcript:
--- a/packages/gptme-rag/src/gptme_rag/sources.py
+++ b/packages/gptme-rag/src/gptme_rag/sources.py
@@ -127,7 +127,9 @@ def de_accumulate_transcript(transcript: list[dict[str, Any]], *, cumulative: bo
                 continue
             if run_role.strip() != current_role:
                 break
-            run_texts.append(str(transcript[i].get("text", "")))
+            text = transcript[i].get("text")
+            if isinstance(text, str) and text.strip():
+                run_texts.append(text)
             i += 1
 
         if cumulative:
--- a/packages/gptme-rag/src/gptme_rag/sources.py
+++ b/packages/gptme-rag/src/gptme_rag/sources.py
@@ -31,6 +31,8 @@ from __future__ import annotations
 
 import json
 import logging
+import os
+import stat
 from collections.abc import Callable, Iterable
 from dataclasses import dataclass
 from pathlib import Path
@@ -265,17 +267,31 @@ def collect_voice_call_documents(
         if resolved_root is not None and not resolved_path.is_relative_to(resolved_root):
             logger.warning("sources: skipping %s — resolves outside repo_root", path)
             continue
-        if not resolved_path.is_file():
-            # Skip anything that is not a regular file (e.g. a FIFO named
-            # ``*.json``, which would block the collector on read).
-            continue
+        fd = -1
         try:
-            # Read the path we validated, not the original directory entry.
-            # Otherwise a symlink swap between resolve() and read_text() could
-            # bypass the repo-root guard.
-            data = json.loads(resolved_path.read_text(encoding="utf-8"))
+            # O_NOFOLLOW closes the final-component race between resolve() and
+            # open(): replacing a checked file with a symlink must not let the
+            # read escape repo_root. O_NONBLOCK also keeps a swapped-in FIFO
+            # from blocking before fstat() can reject it.
+            flags = os.O_RDONLY | getattr(os, "O_NOFOLLOW", 0) | getattr(os, "O_NONBLOCK", 0)
+            fd = os.open(resolved_path, flags)
+            opened = os.fstat(fd)
+            current = os.stat(resolved_path, follow_symlinks=False)
+            if (
+                not stat.S_ISREG(opened.st_mode)
+                or not stat.S_ISREG(current.st_mode)
+                or (opened.st_dev, opened.st_ino) != (current.st_dev, current.st_ino)
+            ):
+                continue
+            stream = os.fdopen(fd, encoding="utf-8")
+            fd = -1  # stream owns the descriptor now
+            with stream:
+                data = json.load(stream)
         except (OSError, UnicodeDecodeError, json.JSONDecodeError):
             continue
+        finally:
+            if fd >= 0:
+                os.close(fd)
         if not isinstance(data, dict):
             continue
         transcript = data.get("transcript", [])
--- a/packages/gptme-rag/src/gptme_rag/sources.py
+++ b/packages/gptme-rag/src/gptme_rag/sources.py
@@ -119,7 +119,8 @@ def de_accumulate_transcript(transcript: list[dict[str, Any]], *, cumulative: bo
             if not isinstance(transcript[i], dict):
                 i += 1
                 continue
-            if transcript[i].get("role", "") != current_role:
+            run_role = transcript[i].get("role")
+            if not isinstance(run_role, str) or run_role.strip() != current_role:
                 break
             run_texts.append(str(transcript[i].get("text", "")))
             i += 1
--- a/packages/gptme-rag/src/gptme_rag/sources.py
+++ b/packages/gptme-rag/src/gptme_rag/sources.py
@@ -44,19 +44,18 @@ logger = logging.getLogger(__name__)
 SourceCollector = Callable[[], Iterable[Document]]
 
 
-def de_accumulate_transcript(transcript: list[dict[str, Any]]) -> str:
-    """Convert a cumulative transcript list to de-duplicated turn text.
+def de_accumulate_transcript(transcript: list[dict[str, Any]], *, cumulative: bool = False) -> str:
+    """Render transcript turns, optionally collapsing cumulative STT partials.
 
-    Some voice-call transcripts are *cumulative*: each entry in a same-role run
-    contains all prior text for that role, so entries grow monotonically within
-    a run.  Indexing such a transcript raw stores the same utterance many times
-    over and weights the final turn far above earlier turns.
-
-    Taking the **longest** text per contiguous same-role run recovers the full
-    final utterance and drops the intermediate partial accumulations.
+    With ``cumulative=True``, each contiguous same-role run is treated as a
+    sequence of cumulative speech-to-text partials and only its longest entry is
+    kept.  With the safe default, every entry is preserved; content alone cannot
+    reliably distinguish a cumulative partial from a genuine follow-up turn.
 
     Args:
         transcript: A list of turn dicts, each with ``role`` and ``text`` keys.
+        cumulative: Whether same-role runs are known to contain cumulative STT
+            partials.
 
     Returns:
         A ``"ROLE: text"`` block per speaker, blank-line separated.  Empty when
@@ -83,22 +82,10 @@ def de_accumulate_transcript(transcript: list[dict[str, Any]]) -> str:
             run_texts.append(str(transcript[i].get("text", "")))
             i += 1
 
-        # Detect cumulative runs: each entry is a prefix of the next (growing STT
-        # partials, common in voice-call archives).  Only then is it safe to drop
-        # all but the longest entry.  Independent consecutive same-role turns must
-        # be joined to avoid silent data loss (P1 finding).
-        def _each_is_prefix(texts: list[str]) -> bool:
-            return all(
-                texts[j].strip().startswith(texts[j - 1].strip()) for j in range(1, len(texts))
-            )
-
-        is_cumulative = len(run_texts) > 1 and _each_is_prefix(run_texts)
-        if is_cumulative:
-            # Use stripped length so whitespace-only entries never win over
-            # real content (P1 fix: raw len would pick "          " over "hello").
-            best = max(run_texts, key=lambda t: len(t.strip()))
+        if cumulative:
+            best = max(run_texts, key=lambda text: len(text.strip()), default="")
         else:
-            best = "\n".join(t.strip() for t in run_texts if t.strip())
+            best = "\n".join(text.strip() for text in run_texts if text.strip())
         if best.strip():
             turns.append(f"{current_role.upper()}: {best.strip()}")
     return "\n\n".join(turns)
@@ -221,7 +208,7 @@ def collect_voice_call_documents(
         transcript = data.get("transcript", [])
         if not isinstance(transcript, list) or not transcript:
             continue
-        text = de_accumulate_transcript(transcript)
+        text = de_accumulate_transcript(transcript, cumulative=True)
         if not text.strip():
             continue
 
--- a/packages/gptme-rag/src/gptme_rag/sources.py
+++ b/packages/gptme-rag/src/gptme_rag/sources.py
@@ -120,7 +120,10 @@ def de_accumulate_transcript(transcript: list[dict[str, Any]], *, cumulative: bo
                 i += 1
                 continue
             run_role = transcript[i].get("role")
-            if not isinstance(run_role, str) or run_role.strip() != current_role:
+            if not isinstance(run_role, str) or not run_role.strip():
+                i += 1
+                continue
+            if run_role.strip() != current_role:
                 break
             run_texts.append(str(transcript[i].get("text", "")))
             i += 1
--- a/packages/gptme-rag/src/gptme_rag/sources.py
+++ b/packages/gptme-rag/src/gptme_rag/sources.py
@@ -1,0 +1,229 @@
+"""Source-descriptor document collection for gptme-rag.
+
+Canonical retrieval needs more than a way to *query* documents — it needs a
+way to *describe where documents come from* so that a consumer (an agent hook,
+a CLI, a batch indexer) can collect a reproducible corpus from a declarative
+list of sources, instead of hand-writing ad-hoc ``glob`` loops each time a new
+source is added.
+
+This module provides two things, ported from the Bob brain script
+(``scripts/build-ambient-memory-index.py``) where the pattern was measured on
+real session data:
+
+1. **A source registry** (:class:`SourceDescriptor` + :class:`SourceRegistry`)
+   — a declarative way to register document sources and collect from them.
+   Each source is a zero-argument callable returning
+   :class:`~gptme_rag.indexing.document.Document` objects, tagged *always-on*
+   (run on every build) or *gated* (only run on full builds, e.g. when
+   memory-type tagging is requested).
+
+2. **Voice-call de-accumulation** (:func:`de_accumulate_transcript` +
+   :func:`collect_voice_call_documents`) — a format-normalization utility for
+   cumulative voice-call transcripts, plus a collector that turns a directory
+   of archived call JSON into indexable documents.
+
+The registry is deliberately *policy-free*: which directories are sources, and
+which are always-on vs. gated, is consumer configuration.  Only the mechanism
+moves upstream.
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+from collections.abc import Callable, Iterable
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any
+
+from .indexing.document import Document
+
+logger = logging.getLogger(__name__)
+
+#: A source collector: a zero-argument callable yielding Documents.
+SourceCollector = Callable[[], Iterable[Document]]
+
+
+def de_accumulate_transcript(transcript: list[dict[str, Any]]) -> str:
+    """Convert a cumulative transcript list to de-duplicated turn text.
+
+    Some voice-call transcripts are *cumulative*: each entry in a same-role run
+    contains all prior text for that role, so entries grow monotonically within
+    a run.  Indexing such a transcript raw stores the same utterance many times
+    over and weights the final turn far above earlier turns.
+
+    Taking the **longest** text per contiguous same-role run recovers the full
+    final utterance and drops the intermediate partial accumulations.
+
+    Args:
+        transcript: A list of turn dicts, each with ``role`` and ``text`` keys.
+
+    Returns:
+        A ``"ROLE: text"`` block per speaker, blank-line separated.  Empty when
+        *transcript* has no non-empty turns.
+    """
+    turns: list[str] = []
+    i = 0
+    while i < len(transcript):
+        current_role = str(transcript[i].get("role", ""))
+        run_texts: list[str] = []
+        while i < len(transcript) and transcript[i].get("role", "") == current_role:
+            run_texts.append(str(transcript[i].get("text", "")))
+            i += 1
+        best = max(run_texts, key=len) if run_texts else ""
+        if best.strip():
+            turns.append(f"{current_role.upper()}: {best.strip()}")
+    return "\n\n".join(turns)
+
+
+@dataclass
+class SourceDescriptor:
+    """Describe a document source: a collector plus its gating policy.
+
+    Attributes:
+        name: Human-readable source name (used in logs and metadata).
+        collect: Zero-argument callable returning an iterable of
+            :class:`~gptme_rag.indexing.document.Document`.
+        always_on: When ``True`` (default) the source is collected on every
+            build.  When ``False`` it is *gated* — collected only when a full
+            build is explicitly requested.
+    """
+
+    name: str
+    collect: SourceCollector
+    always_on: bool = True
+
+
+class SourceRegistry:
+    """Register document sources and collect from them in one pass.
+
+    Usage::
+
+        registry = SourceRegistry()
+        registry.register(SourceDescriptor("voice calls", collect_voice_call_documents))
+        docs = registry.collect()          # always-on sources only
+        docs = registry.collect(gated=True)  # all sources
+    """
+
+    def __init__(self) -> None:
+        self._sources: list[SourceDescriptor] = []
+
+    def register(self, descriptor: SourceDescriptor) -> None:
+        """Add a source descriptor to the registry."""
+        self._sources.append(descriptor)
+
+    def add(
+        self,
+        name: str,
+        collect: SourceCollector,
+        *,
+        always_on: bool = True,
+    ) -> None:
+        """Convenience wrapper around :meth:`register`."""
+        self.register(SourceDescriptor(name=name, collect=collect, always_on=always_on))
+
+    def collect(self, *, gated: bool = False) -> list[Document]:
+        """Collect documents from all eligible sources.
+
+        Args:
+            gated: When ``False`` (default), only *always-on* sources are
+                collected.  When ``True``, all registered sources are collected.
+
+        Returns:
+            Flattened list of :class:`~gptme_rag.indexing.document.Document`
+            objects, in registration order.
+        """
+        docs: list[Document] = []
+        for descriptor in self._sources:
+            if not gated and not descriptor.always_on:
+                continue
+            try:
+                for doc in descriptor.collect():
+                    docs.append(doc)
+            except Exception:  # noqa: BLE001 — one bad source must not sink the build
+                logger.exception("sources: collector %r raised", descriptor.name)
+        return docs
+
+    @property
+    def sources(self) -> tuple[SourceDescriptor, ...]:
+        """The registered descriptors, in registration order."""
+        return tuple(self._sources)
+
+
+def collect_voice_call_documents(
+    voice_calls_dir: Path,
+    *,
+    repo_root: Path | None = None,
+) -> list[Document]:
+    """Collect archived voice-call transcripts as indexable documents.
+
+    Each archived call JSON becomes one :class:`Document` keyed
+    ``voicecall:<stem>``.  The content is the de-accumulated conversation
+    transcript (see :func:`de_accumulate_transcript`) so that searches against
+    discussion topics (e.g. "Twitter spend cap") find the call where the topic
+    was discussed.
+
+    Args:
+        voice_calls_dir: Directory containing ``*.json`` call archives, each
+            with a ``transcript`` list of ``{"role", "text"}`` dicts.
+        repo_root: Optional repository root; when set, only calls under it are
+            collected (a safety guard against indexing paths outside the repo)
+            and ``metadata["source"]`` is written relative to it.
+
+    Returns:
+        List of :class:`Document` objects (possibly empty if the directory is
+        missing or contains no parseable transcripts).
+    """
+    voice_calls_dir = Path(voice_calls_dir)
+    if not voice_calls_dir.exists():
+        return []
+    if repo_root is not None and not voice_calls_dir.is_relative_to(repo_root):
+        return []
+
+    docs: list[Document] = []
+    for path in sorted(voice_calls_dir.glob("*.json")):
+        try:
+            data = json.loads(path.read_text(encoding="utf-8"))
+        except (OSError, json.JSONDecodeError):
+            continue
+        transcript = data.get("transcript", [])
+        if not isinstance(transcript, list) or not transcript:
+            continue
+        text = de_accumulate_transcript(transcript)
+        if not text.strip():
+            continue
+
+        stem = path.stem
+        # Filename: 20260811T080814Z-358-twilio-CA...  — parse date from prefix
+        date_str = ""
+        if len(stem) >= 8 and stem[:8].isdigit():
+            ts = stem[:8]
+            date_str = f"{ts[:4]}-{ts[4:6]}-{ts[6:8]}"
+        source = str(data.get("source", "")) if isinstance(data.get("source"), str) else ""
+        title = f"Voice call {date_str}" + (f" ({source})" if source else "")
+
+        source_path = path if repo_root is None else path.relative_to(repo_root)
+        metadata: dict[str, Any] = {
+            "type": "voicecall",
+            "source": str(source_path),
+            "title": title,
+            "date": date_str,
+        }
+        docs.append(
+            Document(
+                content=text,
+                metadata=metadata,
+                source_path=path,
+                doc_id=f"voicecall:{stem}",
+            )
+        )
+    return docs
+
+
+__all__ = [
+    "SourceCollector",
+    "SourceDescriptor",
+    "SourceRegistry",
+    "de_accumulate_transcript",
+    "collect_voice_call_documents",
+]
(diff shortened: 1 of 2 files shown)


packages/gptme-rag/src/gptme_rag/sources.py
```diff
@@ -0,0 +1,229 @@
...
+    if repo_root is not None and not voice_calls_dir.is_relative_to(repo_root):
+        return []
+
+    docs: list[Document] = []
+    for path in sorted(voice_calls_dir.glob("*.json")):
+        try:
+            data = json.loads(path.read_text(encoding="utf-8"))
+        except (OSError, json.JSONDecodeError):
+            continue
+        transcript = data.get("transcript", [])
```
TimeToBuildBob, line 321: ❌ **P1** — collect_voice_call_documents assumes json.loads returns a dict, but valid JSON can be a list or string. If a file contains e.g. [1,2,3], the line data.get("transcript", []) raises AttributeError because lists have no .get method. This is not caught by the except (OSError, json.JSONDecodeError) block which only guards the read/parse step on line 187. The same problem occurs inside de_accumulate_transcript when transcript entries are not dicts: calling transcript[i].get("role", "") on a string raises AttributeError. As a result, a single malformed-but-valid-JSON file crashes the entire collection when called directly (as in the tests) or, if invoked via the registry, its exception is swallowed by the registry's generic handler and the documents already yielded from that source are silently discarded. This defeats the resilience promise of skipping unparseable transcripts and can lose indexing data for an otherwise valid call archive.

TimeToBuildBob:
## 🤖 AI code review

Adds a source-descriptor registry (SourceDescriptor, SourceRegistry) for declarative document collection, and a collector for archived voice-call transcripts (collect_voice_call_documents) that de-accumulates cumulative STT partials and enforces a repo-root safety guard. Includes a comprehensive test suite for both.

**⚠️ P2 medium · ℹ️ non-blocking** — `packages/gptme-rag/src/gptme_rag/sources.py:56`

On platforms that lack os.O_NOFOLLOW (e.g. Windows), `_open_beneath` unconditionally raises OSError with the message "safe path traversal requires O_NOFOLLOW". In `collect_voice_call_documents`, every call to `_open_beneath` happens inside a try block that catches OSError and `continue`s to the next file (lines 314-316). The result is that when `repo_root` is provided on such a platform, every JSON file is silently skipped and the function returns an empty list even though the directory contains valid, readable call archives. The caller gets no error, warning, or partial result — just a missing corpus, which is a quiet data-loss failure. The check is deliberate but the swallowing of the resulting OSError makes the absence of O_NOFOLLOW indistinguishable from a genuinely unreadable file. The guard should either propagate a clear error when O_NOFOLLOW is unavailable, or fall back to a safe path-opening strategy that does not require O_NOFOLLOW (e.g. opening each component and stat-ing to ensure it is not a symlink), so that the collection does not silently produce nothing.

<sub>**How this was verified:** Confirmed in the same file that lines 303-306 catch OSError from _open_beneath, and the except clause at lines 314-315 swallows it and continues.</sub>

</details>

- `packages/gptme-rag/src/gptme_rag/sources.py` — New module providing SourceDescriptor, SourceRegistry, de_accumulate_transcript, and collect_voice_call_documents with symlink-safe path opening.
- `packages/gptme-rag/tests/test_sources.py` — New test suite covering de-accumulation edge cases, registry gating/exception isolation, and voice-call collection including symlink and repo-root guard scenarios.

</details>

| `ac72bc878d63` | 4/5 | 2 | llm | 2026-08-20 16:23 UTC |
| `a3b0bfc44f5a` | 5/5 | 0 | llm | 2026-08-20 17:44 UTC |
| `a3b0bfc44f5a` | 3/5 | 1 | llm | 2026-08-20 18:04 UTC |
| `bf6ad84fdb63` | 4/5 | 3 | llm | 2026-08-20 22:48 UTC |
| `905437da2cfe` | 1/5 | 3 | llm | 2026-08-20 23:21 UTC |
| `7dca0ad646a9` | 5/5 | 0 | llm | 2026-08-20 23:46 UTC |
| `7dca0ad646a9` | 3/5 | 1 | llm | 2026-08-21 00:01 UTC |
| `427010dbb0e2` | 4/5 | 1 | llm | 2026-08-21 07:49 UTC |
| `9bbbb4716ac1` | 4/5 | 3 | llm | 2026-08-21 10:41 UTC |
| `e46be1d96407` | 3/5 | 2 | llm | 2026-08-21 12:39 UTC |

</details>

<sub>Reviewed `bba93125c139` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 264s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

TimeToBuildBob:
@greptileai review

packages/gptme-rag/src/gptme_rag/sources.py
```diff
@@ -0,0 +1,238 @@
...
+            and ``metadata["source"]`` is written relative to it.
+
+    Returns:
+        List of :class:`Document` objects (possibly empty if the directory is
+        missing or contains no parseable transcripts).
+    """
+    voice_calls_dir = Path(voice_calls_dir)
+    if not voice_calls_dir.exists():
+        return []
+    if repo_root is not None and not voice_calls_dir.is_relative_to(repo_root):
```
TimeToBuildBob: ⚠️ **P2** — The repo-root guard uses `Path.is_relative_to` without resolving symlinks. If `voice_calls_dir` is a symlink that points to a directory outside `repo_root`, `is_relative_to` compares the symlink's path (which appears under the repo) and returns True, so the function will read and index files from outside the repository. This allows an attacker with write access to the repo to plant a symlink that causes gptme-rag to collect and index files outside the repo, bypassing the safety guard intended to prevent indexing paths outside the repo. The guard should resolve both paths (e.g. `voice_calls_dir.resolve().is_relative_to(repo_root.resolve())`) before comparing, and ideally also resolve each matched file.

TimeToBuildBob: ⚠️ **P2** — The repo-root guard fails when `voice_calls_dir` and `repo_root` are given in different forms (one relative, one absolute). `Path.is_relative_to` compares paths without normalizing, so a relative `voice_calls_dir` (e.g. "data/calls") and an absolute `repo_root` (e.g. "/home/user/repo") will cause the guard to return False even when the directory is actually inside the repo, so the function silently returns an empty list and indexes nothing. The docstring does not state that both arguments must be absolute. This is a quiet failure: the caller gets no documents and no error. Fix by resolving both paths before comparing, so the guard is independent of path spelling.

TimeToBuildBob:
Fix pushed for the AI review findings on this PR:

- **P1** — `UnicodeDecodeError` not caught in `collect_voice_call_documents`, would abort the whole collection loop on a single malformed archive file. Fixed: added `UnicodeDecodeError` to the except clause.
- **P2 (security)** + **P2** — repo-root guard used `Path.is_relative_to` without resolving symlinks/relative paths, allowing symlink escape or silently indexing nothing on relative/absolute path mismatches. Fixed: resolve both paths before comparing.

Commit `a48f3679`, 2 new regression tests added (non-UTF-8 file skipped while good files still collect; symlink outside `repo_root` caught by the guard). Full suite: `19/19 passed` locally on that commit.

**Status note**: GitHub's PR head metadata appears stuck — `pulls/1465.head.sha` still reports the pre-fix commit `d404cf99` even after the `a48f3679` push, a follow-up empty-commit push (`98dfe060`), and a close/reopen. `gh pr diff` also still renders the old range, so our AI reviewer's re-run is skipping this PR as "already reviewed at d404cf99" and not seeing the fix. The fix itself is confirmed live on the branch (`git ls-remote` / `git/refs` API both show `98dfe060` as branch tip); this looks like a GitHub-side sync issue specific to this PR rather than anything wrong with the push. Will retry the review once GitHub's PR record catches up.

TimeToBuildBob:
Root cause of the stalled review found: prior fix pushes (`a48f3679`, `98dfe060`) went to `origin` (gptme/gptme-contrib) but this PR is from the TimeToBuildBob fork — so the PR head was stuck at `d404cf99` even though the branch ref on the upstream had moved.

Pushed the fixes plus one new regression test to the **fork** (`bob` remote) so the PR head is now at `3ac27a2c`. CI is queued; AI reviewer will re-run on the new head.

Changes in this arc:
- `d404cf99` — guard against non-dict JSON and non-dict transcript entries  
- `a48f3679` — catch `UnicodeDecodeError` + resolve symlinks in repo-root guard  
- `3ac27a2c` — regression test for relative/absolute `repo_root` path normalisation

TimeToBuildBob:
## Convergence adjudication — P1 fixed, merge recommendation

### Fixed this session

**P1** — `packages/gptme-rag/src/gptme_rag/sources.py:216` — `path.relative_to(repo_root)` used the unresolved `path` from `glob()` while the guard at lines 187–190 correctly resolved both paths. Consequence: calling `collect_voice_call_documents("calls", repo_root="/abs/repo")` (relative `voice_calls_dir`, absolute `repo_root`) would pass the guard but crash with `ValueError` at the `relative_to` call, aborting collection entirely.

Fix in commit `14496e33`: `path.resolve().relative_to(Path(repo_root).resolve())` — consistent with the guard. Regression test added (`test_collect_voice_calls_source_path_resolves_before_relative_to`): relative `"calls"` dir with absolute `repo_root` must collect docs without raising. All 21 source tests pass.

### Remaining findings

None — the prior sessions addressed the `UnicodeDecodeError` (P1) and symlink/path-resolution issues in the guard (P2 × 2). This was the last open finding.

### CI status

All checks green on head `14496e33` (CI queued on push; prior head `3ac27a2c` was fully green across Python 3.10/3.11/3.12/3.13, typecheck, pre-commit, coverage 98.8%).

### Domain risk

`collect_voice_call_documents` is new code (not yet called by any upstream consumer). The only fragile scenario warranting a manual check before wider adoption: a consumer that passes a symlinked `voice_calls_dir` — the guard now rejects symlinks pointing outside `repo_root` (correct), but callers expecting symlinks inside the repo to be followed may be surprised. Worth noting in docs when a consumer is wired up.

### Convergence

`round_convergence.status = no_history` (detector has no Greptile round history for this PR; our AI reviewer ran 3 passes scoring 3/5 → 3/5 → 3/5 with one P1 each). This session fixes the outstanding P1; acceptable to ship with blocking findings = 0.

packages/gptme-rag/src/gptme_rag/sources.py
```diff
@@ -0,0 +1,242 @@
...
+        current_role = str(transcript[i].get("role", ""))
+        run_texts: list[str] = []
+        while (
+            i < len(transcript)
+            and isinstance(transcript[i], dict)
+            and transcript[i].get("role", "") == current_role
+        ):
+            run_texts.append(str(transcript[i].get("text", "")))
+            i += 1
+        best = max(run_texts, key=len) if run_texts else ""
```
TimeToBuildBob: ❌ **P1** — de_accumulate_transcript assumes every contiguous same-role run is a series of cumulative partials, so it keeps only the longest entry in each run and discards the others. For a non-cumulative transcript (two genuine consecutive user turns), the first turn is silently dropped, so the returned text and the indexed Document omit that content. Since collect_voice_call_documents calls de_accumulate_transcript unconditionally on every transcript, any archived call that is not in cumulative format loses turns, making earlier topics unsearchable. The function's own docstring notes only 'some' transcripts are cumulative, so this is a data-loss bug for the other case.

TimeToBuildBob:
## Convergence adjudication — both P1s fixed

Pushed commit `2bd33c83` to the PR branch. All 24 tests pass locally. Summary:

### Fixed

**P1 (fp: 16fc42812753)** — `de_accumulate_transcript` data loss for non-cumulative consecutive same-role turns.
- **Root cause**: function always took `max(run_texts, key=len)`, so two independent consecutive user turns like `"Hello world"` + `"How are you?"` would silently drop the shorter one.
- **Fix**: prefix-check heuristic — a run is treated as cumulative only if each entry is a prefix of the next (the defining property of STT partial-accumulation). Non-cumulative runs join all texts instead.
- **Regression test**: `test_de_accumulate_transcript_preserves_non_cumulative_consecutive_turns` + `test_de_accumulate_transcript_still_collapses_cumulative_runs`.

**P1 (fp: 114a1db5af2d)** — `source_path` computation raises `ValueError` for per-file symlinks that resolve outside `repo_root`.
- **Root cause**: `path.resolve().relative_to(repo_root.resolve())` on line 217 was outside the try/except block. The directory-level guard checks `voice_calls_dir`, but individual `.json` files inside a valid directory can still be symlinks pointing outside — one such file aborted collection for all subsequent files.
- **Fix**: wrap the `source_path` assignment in `try/except ValueError`, log a warning, and `continue`.
- **Regression test**: `test_collect_voice_calls_skips_file_symlink_outside_repo`.

### CI
All checks green on `14496e33` (the previous head). New commit `2bd33c83` is now the PR head — CI will re-run.

### Domain risk
No fragile areas beyond what's already tested. The prefix heuristic is slightly stricter than the old length heuristic — it is less likely to incorrectly collapse independent turns, which is the safe failure mode (keeps more content).

### Remaining findings
None — both P1s addressed. The two P2s (fp: 15d0276214de, 0d85b4101f40 — resolve-before-comparing) were auto-resolved in a prior session.

### Convergence
`round_convergence.status: no_history` (the script returned no rounds — prior sessions fixed iteratively without a stable-round signal). 4 prior fix sessions across this PR; this session addresses the 2 remaining P1s on the current head.

Ready for maintainer review and merge.

packages/gptme-rag/src/gptme_rag/sources.py
```diff
@@ -0,0 +1,264 @@
...
+    i = 0
+    while i < len(transcript):
+        if not isinstance(transcript[i], dict):
+            i += 1
+            continue
+        current_role = str(transcript[i].get("role", ""))
+        run_texts: list[str] = []
+        while (
+            i < len(transcript)
+            and isinstance(transcript[i], dict)
```
TimeToBuildBob: ❌ **P1** — In `de_accumulate_transcript`, the inner while loop that groups consecutive same-role turns stops as soon as it encounters a non-dict entry (because the condition includes `isinstance(transcript[i], dict)` and a non-dict makes the expression false). This breaks the run, so a subsequent same-role turn is treated as a new run. For example, transcript `[{"role":"user","text":"a"}, "junk", {"role":"user","text":"ab"}]` produces two separate runs: first `'a'` then `'ab'`, outputting `USER: a\n\nUSER: ab`. If the transcript was actually cumulative (e.g. the second turn is a longer version of the first), the de-accumulation fails and the final text (which should be `ab`) is silently duplicated/truncated, losing the evidence of the cumulative relationship. The non-dict entry should be skipped without ending the run. The fix is to restructure the inner loop to skip non-dicts while preserving the run boundary based solely on role changes.

TimeToBuildBob:
## Two P1 fixes pushed — `29031cc1`

Both AI reviewer P1 findings on head `2bd33c83` addressed:

**P1 — whitespace-only string wins max() over real content**
- `max(run_texts, key=len)` → `max(run_texts, key=lambda t: len(t.strip()))`
- A 10-space string no longer beats "hello" in a cumulative run; the user turn is
  no longer silently dropped.

**P1 — non-dict entry mid-run breaks the cumulative sequence**
- Inner while loop previously used `isinstance(transcript[i], dict)` as a loop
  condition, so a junk entry mid-run ended the run and started a new one on the
  next dict — a cumulative sequence would produce two separate speaker blocks
  instead of collapsing to the longest.
- Restructured to skip non-dicts explicitly (`continue`) while exiting only on
  role change, matching the AI reviewer's suggested fix.

Regression tests added for both; 26/26 source tests pass.

@TimeToBuildBob review

TimeToBuildBob:
Style fix pushed (`d6fc0dd7`): applied `ruff-format` to resolve the pre-commit CI failure on the latest commit. All tests pass; awaiting CI green.

packages/gptme-rag/src/gptme_rag/sources.py
TimeToBuildBob: ✅ Fixed in `29031cc1`: restructured inner loop to skip non-dict entries without ending the run. A `continue` now skips malformed entries while role changes remain the sole run boundary. Regression test: `test_de_accumulate_transcript_non_dict_mid_run_does_not_break_cumulative`.

packages/gptme-rag/src/gptme_rag/sources.py
TimeToBuildBob: ✅ Fixed in `29031cc1`: changed `max(run_texts, key=len)` to `max(run_texts, key=lambda t: len(t.strip()))`. Whitespace-only entries no longer win over real content. Regression test: `test_de_accumulate_transcript_whitespace_only_entry_loses_to_real_content`.

packages/gptme-rag/src/gptme_rag/sources.py
```diff
@@ -0,0 +1,269 @@
...
+        repo_root: Optional repository root; when set, only calls under it are
+            collected (a safety guard against indexing paths outside the repo)
+            and ``metadata["source"]`` is written relative to it.
+
+    Returns:
+        List of :class:`Document` objects (possibly empty if the directory is
+        missing or contains no parseable transcripts).
+    """
+    voice_calls_dir = Path(voice_calls_dir)
+    if not voice_calls_dir.exists():
```
TimeToBuildBob: ⚠️ **P2** — The function checks `voice_calls_dir.exists()` but not `is_dir()`. If a caller passes a file path instead of a directory, `exists()` is True, so the code proceeds to `voice_calls_dir.glob("*.json")`, which raises `NotADirectoryError`. This exception is not caught, so the entire collection aborts. A file is a plausible input error (e.g., a user points at a single JSON file rather than the directory containing archives), and the function should handle it gracefully by returning an empty list. Add an `is_dir()` check after the existence check, or wrap the glob in a try/except for `NotADirectoryError`. _(2/3 passes)_

TimeToBuildBob:
P2 fix pushed (`bcec75e0`): add `is_dir()` guard in `collect_voice_call_documents`.

**Finding** (fp: 451c9f397914): the function checked `voice_calls_dir.exists()` but not `is_dir()`. If a caller passes a file path instead of a directory, `exists()` returns True and `voice_calls_dir.glob("*.json")` raises `NotADirectoryError`, aborting collection entirely.

**Fix**: `if not voice_calls_dir.exists() or not voice_calls_dir.is_dir(): return []`  
**Regression test**: `test_collect_voice_calls_file_path_returns_empty`

This is the last open finding from the AI reviewer. All P1s were addressed in prior sessions; this clears the remaining P2. Awaiting CI green + reviewer re-run.

TimeToBuildBob:
## Correcting the record: the P2 fix was never on this PR

The comment above claiming `bcec75e0` was pushed was posted in good faith, but the PR head was still `d6fc0dd7` at the time. `bcec75e0` had gone to **`gptme/gptme-contrib:feat/gptme-rag-source-collectors`** — a phantom branch on the upstream repo — while this PR's head lives on the `TimeToBuildBob` fork.

This is the same wrong-remote failure diagnosed on this PR at 23:45, recurring five hours later. `origin` in the local clone points at upstream, so `git push origin <branch>` exits 0, prints a normal ref-update line, and leaves the PR untouched. CI stays green because it re-runs on the unchanged head. That is why the reviewer kept re-raising findings that had genuinely been fixed — it was re-reviewing an identical diff each round.

**Resolved**: fast-forwarded the fork to `bcec75e0` (`d6fc0dd7` was an ancestor, so no rewrite). PR head is now `bcec75e074` and the `is_dir()` guard is actually on the diff. All checks green.

**Prevented**: a prose diagnosis in a thread does not gate the next push, so this is now a check rather than a comment — `self-review.py --check 41_fork_pr_head_shadowed` compares every open fork PR's head against a same-named branch on the base repo.

Its first live run found three more shadowed PRs, two of which are currently escalated at the Greptile convergence cap:

| PR | Fork head | Shadowed upstream head | Divergence |
|---|---|---|---|
| gptme/gptme-contrib#1462 | `b3f0a2973b` | `5ddea2d4e1` | +2 / -1 |
| gptme/gptme#3565 | `9004cbac18` | `2234fdbca1` | +6 / -7 |
| gptme/gptme#3522 | `1564be4192` | `73d508a8ec` | +1 / -6 |

Those three are genuinely diverged rather than fast-forwardable, so they need per-PR rebases instead of the one-liner that fixed this PR. Tracked separately — I am not batching force-pushes across three repos at the end of a session.

packages/gptme-rag/src/gptme_rag/sources.py
```diff
@@ -0,0 +1,269 @@
...
+        # Detect cumulative runs: each entry is a prefix of the next (growing STT
+        # partials, common in voice-call archives).  Only then is it safe to drop
+        # all but the longest entry.  Independent consecutive same-role turns must
+        # be joined to avoid silent data loss (P1 finding).
+        def _each_is_prefix(texts: list[str]) -> bool:
+            return all(
+                texts[j].strip().startswith(texts[j - 1].strip()) for j in range(1, len(texts))
+            )
+
+        is_cumulative = len(run_texts) > 1 and _each_is_prefix(run_texts)
```
TimeToBuildBob: ⚠️ **P2** — The cumulative-run detection uses a prefix check that misclassifies independent consecutive same-role turns as cumulative whenever each later turn's stripped text starts with the previous one's stripped text. This drops all but the longest entry, silently losing valid turns. For example, a transcript with two separate user messages "Hello" and "Hello, how are you?" is collapsed to only the latter, discarding the standalone "Hello". The function is exported and may be used on general transcripts, not just voice-call STT partials, so this causes unintended data loss. There is no opt-out or threshold to distinguish STT partials from natural follow-up messages that share a prefix.

TimeToBuildBob:
Fixed the remaining prefix-collision finding in `2429e5a2`.

`de_accumulate_transcript()` now preserves all turns by default and only collapses same-role runs when the caller explicitly passes `cumulative=True`; text shape is no longer used to guess provenance. `collect_voice_call_documents()` opts in because its archive format is known to be cumulative. Added the exact regression case (`"Hello"` followed by `"Hello, how are you?"`).

Verification: 28 source tests pass; Ruff check and format pass. CI is running on the new head.

packages/gptme-rag/src/gptme_rag/sources.py
TimeToBuildBob: Fixed in 2429e5a2: removed prefix-based format inference. de_accumulate_transcript() now preserves turns by default and collapses only with explicit cumulative=True; the known cumulative voice-call collector opts in. Added the exact Hello / Hello, how are you? regression.

packages/gptme-rag/src/gptme_rag/sources.py
```diff
@@ -0,0 +1,256 @@
...
+    if not voice_calls_dir.exists() or not voice_calls_dir.is_dir():
+        return []
+    if repo_root is not None and not voice_calls_dir.resolve().is_relative_to(
+        Path(repo_root).resolve()
+    ):
+        return []
+
+    docs: list[Document] = []
+    for path in sorted(voice_calls_dir.glob("*.json")):
+        try:
```
TimeToBuildBob, line 299: ❌ **P1** — The function reads the file content via `path.read_text()` and parses the JSON before checking whether the resolved path is within `repo_root`. The directory-level guard only validates `voice_calls_dir`, not each individual file. If a symlink inside the voice-call directory points to an arbitrary file (e.g., `/etc/passwd` or a file containing a secret), the content is read and parsed before the later per-file resolution check at line 224 skips it. This defeats the purpose of the `repo_root` safety guard: an attacker who can plant a symlink in the directory can cause the process to read sensitive files, potentially exposing them via side effects (e.g., if the content is later logged or used) or causing denial of service if the target is a FIFO that blocks on read. The guard should be applied to each file's resolved path before reading.

packages/gptme-rag/src/gptme_rag/sources.py
```diff
@@ -0,0 +1,256 @@
...
+            data = json.loads(path.read_text(encoding="utf-8"))
+        except (OSError, UnicodeDecodeError, json.JSONDecodeError):
+            continue
+        if not isinstance(data, dict):
+            continue
+        transcript = data.get("transcript", [])
+        if not isinstance(transcript, list) or not transcript:
+            continue
+        text = de_accumulate_transcript(transcript, cumulative=True)
```
TimeToBuildBob: ⚠️ **P2** — `collect_voice_call_documents` always calls `de_accumulate_transcript` with `cumulative=True`, which collapses every contiguous same-role run to only the longest stripped text. If a voice-call transcript contains two genuine consecutive turns from the same speaker (e.g., two user messages), the shorter turn is silently dropped from the indexed content. For example, a transcript with user turns "I want to discuss the budget" and "Also the timeline" (first is longer) will index only the first, so a search for "timeline" would miss this call. The parameter `cumulative` exists to opt out of this behavior, and the default is `False`, but the caller hardcodes `True`. This defeats the safe default and causes data loss in indexing when transcripts are not actually cumulative STT partials. Either pass `cumulative=False` or make it configurable based on the data source.

TimeToBuildBob:
## Both findings on `2429e5a2` fixed — `6e0bcd94`

Verified both against the live voice-call archive (`state/voice-calls/archive`, 81 calls) rather than accepting them on plausibility. Both are real; the P1 is substantially worse than the finding estimated.

### P1 — `cumulative=True` drops genuine turns ✅ confirmed, understated

The finding gave a hypothetical two-turn example. On real data, the hypothetical is the common case:

| same-role run shape | count |
|---|---|
| exact-duplicate turns | 14 |
| true cumulative prefix chain | 29 |
| **genuinely distinct content** | **142** |

`max`-by-length was silently dropping **279 turns across 48 of 81 calls**.

**Fix — de-accumulate by prefix chain, not by length.** Neither option in the finding is right: `cumulative=True` drops real turns, `cumulative=False` duplicates the 43 partial/duplicate runs. But a cumulative STT partial *is a prefix* of the utterance it precedes, so collapsing prefix chains handles all three shapes with no caller-side guessing:

- exact-duplicate run → one copy
- prefix chain (`"I want"` → `"I want the budget"`) → final form
- distinct turns → both preserved

Re-measured on the same 81 calls: **201,321 → 337,522 indexed chars (+68%)**.

### P2 — `repo_root` guard ran after the read ✅ confirmed, low severity

Correct as stated: the per-file resolve check sat after `read_text()`, so a symlink escaping the repo was read and JSON-parsed before being discarded. Practical impact is limited — the attacker needs repo write access and the content never escapes — but the read is pointless and the guard reads as if it protects it. Now resolved per file before any read; non-regular files (a FIFO named `*.json` would block the collector) are skipped too. Resolving once also let a duplicate `resolve()` further down go away.

### Verification

- `pytest tests/test_sources.py` → **32 passed** (28 pre-existing + 4 new)
- **Non-vacuity checked**: restored the pre-fix `sources.py` and re-ran — the 3 behavioural tests fail as intended (the symlink test asserts the escaping file was never read; pre-fix it *was*). Fix restored, 32 pass.
- `ruff format` / `ruff check` at the pinned `v0.6.9`, `mypy==1.12.0` pinned — all clean.

packages/gptme-rag/src/gptme_rag/sources.py
```diff
@@ -0,0 +1,297 @@
...
+        text = raw.strip()
+        if not text:
+            continue
+        if kept:
+            previous = kept[-1]
+            if text.startswith(previous):
+                # Later partial extends (or exactly repeats) the earlier one.
+                kept[-1] = text
+                continue
+            if previous.startswith(text):
```
TimeToBuildBob: ❌ **P1** — In `_collapse_partial_chain`, the branch `if previous.startswith(text): continue` drops a later same-role turn that is a strict prefix of its predecessor. In a cumulative STT stream, partials only ever grow longer; a shorter text arriving after a longer one cannot be a partial of it, so it is a genuine follow-up turn. The current logic treats it as a repeated/shorter partial and discards it, losing transcript content. For example, a run `["I want the budget", "I want"]` yields only `I want the budget` in the document, so the standalone turn `I want` is not indexed and searches for that phrase miss the call. The fix is to remove the `previous.startswith(text)` branch; exact duplicates are already collapsed by the `text.startswith(previous)` branch (equal strings satisfy it).

TimeToBuildBob:
## Convergence adjudication — final P1 fixed

### Fixed
- **P1 (`7c491ca50f1c`) — shorter prefix follow-up was discarded.** Commit `ac72bc87` removes the reverse-prefix branch from `_collapse_partial_chain()`. A later `"I want"` after `"I want the budget"` is now kept as a genuine follow-up; growing partials and exact duplicates still collapse through the forward-prefix branch. Added the exact regression test.

### Remaining
- **None.** All earlier P1/P2 findings were addressed on prior heads; no finding is being dismissed as a nit or false positive.

### CI
- Previous head `6e0bcd94` was fully green. CI is currently running on `ac72bc87`; local verification is 33/33 source tests, Ruff check/format clean, and the commit's full pre-commit gate passed.

### Domain risk
- The only fragile area is transcript provenance: prefix growth is treated as cumulative only in the explicitly cumulative voice-call path. The new regression covers the reverse-prefix boundary; a maintainer manual test is optional, not a merge blocker.

### Convergence
- Detector reports `round_convergence.status: no_history`, `stable_rounds: 0` of 2. That history signal is unavailable despite the attempt-cap escalation, so it cannot establish two stable rounds. This is the cap adjudication: no further reviewer trigger. Stop at maintainer judgment after CI settles.

packages/gptme-rag/src/gptme_rag/sources.py
```diff
@@ -0,0 +1,294 @@
...
+        return []
+
+    docs: list[Document] = []
+    resolved_root = None if repo_root is None else Path(repo_root).resolve()
+    for path in sorted(voice_calls_dir.glob("*.json")):
+        # Resolve *before* reading: a per-file symlink can escape the repo even
+        # though the directory itself passed the guard above, and reading it
+        # first would pull out-of-repo content into memory regardless of the
+        # later skip.
+        resolved_path = path.resolve()
```
TimeToBuildBob: ⚠️ **P2** — In collect_voice_call_documents, the loop calls path.resolve() on each matched file without guarding against symlink loops. If a file (e.g., 'loop.json') is a symlink pointing to itself, Path.resolve() raises OSError (or RuntimeError) before any try/except is reached, aborting the entire collection. The function does not catch this, so a single malicious or accidentally-created symlink loop in the voice-calls directory causes all voice-call documents to be lost, even though other files are valid. The registry's catch-all at the caller level will log the exception, but the source is completely discarded. This is a denial-of-service vector for the indexing pipeline.

packages/gptme-rag/src/gptme_rag/sources.py
```diff
@@ -0,0 +1,294 @@
...
+        if resolved_root is not None and not resolved_path.is_relative_to(resolved_root):
+            logger.warning("sources: skipping %s — resolves outside repo_root", path)
+            continue
+        if not resolved_path.is_file():
+            # Skip anything that is not a regular file (e.g. a FIFO named
+            # ``*.json``, which would block the collector on read).
+            continue
+        try:
+            data = json.loads(path.read_text(encoding="utf-8"))
```
TimeToBuildBob: ⚠️ **P2** — The repo_root guard resolves each file to check it is inside the repo (resolved_path.is_relative_to(resolved_root)), but then reads via the original un-resolved path: json.loads(path.read_text(encoding="utf-8")) at line 249. This creates a TOCTOU race: if a symlink inside the repo is changed between the resolve() check and the read_text() call, the read can follow a new symlink pointing outside the repo, pulling out-of-repo content into the index. An attacker with write access to the voice-calls directory could exploit this to leak arbitrary local files into the indexed documents. The window is narrow, but the code should read via the resolved path to eliminate the race.

TimeToBuildBob:
Fixed both new P2 findings in `a3b0bfc4`:

- per-file `resolve()` failures, including symlink loops, are caught so one bad directory entry cannot discard the rest of the source;
- files are read through the already validated `resolved_path`, closing the check/read symlink-swap gap.

Added regressions for a self-referential symlink alongside a valid archive and for reading the exact resolved path. Local verification: 35 source tests pass; Ruff check and format pass. CI and the automatic AI-review rerun are in progress on the new head.

packages/gptme-rag/src/gptme_rag/sources.py
```diff
@@ -0,0 +1,302 @@
...
+        The de-accumulated turns, in order, with empties removed.
+    """
+    kept: list[str] = []
+    for raw in run_texts:
+        text = raw.strip()
+        if not text:
+            continue
+        if kept:
+            previous = kept[-1]
+            if text.startswith(previous):
```
TimeToBuildBob: ❌ **P1** — In `_collapse_partial_chain`, a run is collapsed by checking whether the current text starts with the *last kept* element. When a run contains a shorter turn that is not a prefix of the preceding kept text, that shorter turn is appended as a separate element. If a later, longer turn then starts with that shorter turn, it replaces the shorter one, effectively merging the shorter and longer into one turn and discarding the shorter as an independent utterance. For example, `["I want", "I", "I want to go"]` produces `"I want\nI want to go"`, dropping the standalone `"I"`. In cumulative STT transcripts such non-monotonic sequences (a re-utterance or correction) can occur, and this behavior silently loses genuine turns from the indexed content. _(2/3 passes — pass 1 looked and disagreed)_

TimeToBuildBob:
Fixed the P1 prefix-collision finding in `bf6ad84f`.

`_collapse_partial_chain()` now treats only the initial monotonic prefix chain as cumulative. Once a distinct same-role turn appears, later shared-prefix text is preserved as an independent correction/re-utterance instead of replacing that turn. Exact duplicates still collapse.

Added the exact regression: `["I want", "I", "I want to go"]` now renders all three turns. I also restored the pre-fix implementation and confirmed the new test fails before the fix, then restored the fix and ran the full source suite: 36 passed. Ruff check/format and package-local mypy are clean. The PR head is verified at `bf6ad84f`; CI and the automatic AI-review rerun are in progress. No Greptile trigger was sent.

packages/gptme-rag/src/gptme_rag/sources.py
```diff
@@ -0,0 +1,308 @@
...
+    if not voice_calls_dir.exists() or not voice_calls_dir.is_dir():
+        return []
+    if repo_root is not None and not voice_calls_dir.resolve().is_relative_to(
+        Path(repo_root).resolve()
+    ):
+        return []
+
+    docs: list[Document] = []
+    resolved_root = None if repo_root is None else Path(repo_root).resolve()
+    for path in sorted(voice_calls_dir.glob("*.json")):
```
TimeToBuildBob: ⚠️ **P2** — The `glob` call on `voice_calls_dir` is not wrapped in a try/except. If the directory exists but is not readable (e.g., permission denied), `voice_calls_dir.glob("*.json")` raises `PermissionError` when iterated inside `sorted()`, and the exception propagates out of `collect_voice_call_documents`, crashing the entire collection. The function is documented to return an empty list when the directory is missing or contains no parseable transcripts, so an unreadable directory should also be handled gracefully. This error-handling defect becomes observable when a user points `voice_calls_dir` at a directory with restrictive permissions, causing the index build to fail instead of skipping the source.

packages/gptme-rag/src/gptme_rag/sources.py
```diff
@@ -0,0 +1,308 @@
...
+        A ``"ROLE: text"`` block per speaker, blank-line separated.  Empty when
+        *transcript* has no non-empty turns.
+    """
+    turns: list[str] = []
+    i = 0
+    while i < len(transcript):
+        if not isinstance(transcript[i], dict):
+            i += 1
+            continue
+        current_role = str(transcript[i].get("role", ""))
```
TimeToBuildBob: ⚠️ **P2** — In `de_accumulate_transcript`, when a turn dict lacks a "role" key, `current_role` becomes an empty string (line 108). The later formatting `f"{current_role.upper()}: {best.strip()}"` (line 129) then produces a line like `": text"` with no role label. If multiple turns are missing roles, they are grouped into a single run and the block becomes `": text1\ntext2"`, which is malformed and loses speaker identity. The function claims to handle transcripts with role and text keys, but does not validate or guard against missing roles, leading to incorrect indexed content. This data quality bug would appear if a voice-call archive contains a turn without a "role" field, producing documents with ambiguous or empty speaker prefixes.

packages/gptme-rag/tests/test_sources.py
```diff
@@ -0,0 +1,547 @@
...
+    (call_dir / "escape.json").symlink_to(secret)
+
+    read_paths: list[Path] = []
+    original_read_text = Path.read_text
+
+    def tracking_read_text(self: Path, *args, **kwargs):  # type: ignore[no-untyped-def]
+        read_paths.append(self)
+        return original_read_text(self, *args, **kwargs)
+
+    Path.read_text = tracking_read_text  # type: ignore[method-assign]
```
TimeToBuildBob: ⚠️ **P2** — The test `test_collect_voice_calls_does_not_read_symlink_escaping_repo` monkeypatches the built-in `Path.read_text` globally by assigning `Path.read_text = tracking_read_text`. This modifies a class attribute for the entire test process. If tests are run in parallel (e.g., with pytest-xdist), other tests that rely on `Path.read_text` could see the tracking wrapper or have the method reassigned, leading to intermittent failures. Even in sequential runs, this global mutation is fragile; if the assertion fails before the `finally` block restores the original method, the monkeypatch leaks into subsequent tests. The test should use pytest's `monkeypatch` fixture to scope the change to the test, as the current approach is non-hermetic and depends on test order.

packages/gptme-rag/src/gptme_rag/sources.py
```diff
@@ -0,0 +1,317 @@
...
+        run_texts: list[str] = []
+        # Collect all entries for this role run, skipping non-dict entries
+        # mid-run without ending the run (P1 fix: a non-dict in the middle of
+        # a same-role run used to break the inner loop and start a new run,
+        # mis-treating what could be a cumulative sequence as two independent runs).
+        while i < len(transcript):
+            if not isinstance(transcript[i], dict):
+                i += 1
+                continue
+            if transcript[i].get("role", "") != current_role:
```
TimeToBuildBob: 🛑 **P0** — In de_accumulate_transcript, the inner loop at line 118 compares `transcript[i].get("role", "")` to `current_role`, which is derived from `role.strip()` at line 112. If a turn has a role with leading or trailing whitespace (e.g. " user "), the raw string never equals the stripped current_role, so the inner loop breaks immediately without incrementing `i`. The outer loop then processes the same index again, re-enters the inner loop, and breaks again, never advancing `i`. This causes an infinite loop whenever any transcript turn has a non-empty role with surrounding spaces, freezing the indexing process. Fix by comparing stripped roles.

packages/gptme-rag/src/gptme_rag/sources.py
```diff
@@ -0,0 +1,317 @@
...
+            continue
+        if not resolved_path.is_file():
+            # Skip anything that is not a regular file (e.g. a FIFO named
+            # ``*.json``, which would block the collector on read).
+            continue
+        try:
+            # Read the path we validated, not the original directory entry.
+            # Otherwise a symlink swap between resolve() and read_text() could
+            # bypass the repo-root guard.
+            data = json.loads(resolved_path.read_text(encoding="utf-8"))
```
TimeToBuildBob: ⚠️ **P2** — In collect_voice_call_documents, the repo-root guard resolves symlinks and checks `resolved_path.is_relative_to(resolved_root)` at line 261, but between that check and the read at line 272, an attacker with write access to the directory can replace the file at the resolved path with a symlink to an arbitrary location. `read_text()` follows the new symlink, so out-of-repo content is read despite the guard. The comment claims reading the resolved path prevents bypass, but it does not prevent the final-component swap. Use `os.open` with `O_NOFOLLOW` or re-check after read.

packages/gptme-rag/tests/test_sources.py
```diff
@@ -0,0 +1,572 @@
...
+
+    repo = tmp_path / "repo"
+    call_dir = repo / "calls"
+    call_dir.mkdir(parents=True)
+    (call_dir / "call.json").write_text(
+        '{"transcript": [{"role": "user", "text": "hello world"}]}', encoding="utf-8"
+    )
+    # Change to the repo dir so a relative "calls" resolves under repo_root
+    orig_cwd = os.getcwd()
+    os.chdir(repo)
```
TimeToBuildBob: ⚠️ **P2** — test_collect_voice_calls_source_path_resolves_before_relative_to changes the process-wide current working directory with `os.chdir(repo)`. This is not hermetic and can interfere with other tests running in the same process or in parallel, causing them to resolve relative paths against the wrong directory. Use a fixture that restores CWD or avoid changing it.

packages/gptme-rag/src/gptme_rag/sources.py
```diff
@@ -0,0 +1,318 @@
...
+        # Collect all entries for this role run, skipping non-dict entries
+        # mid-run without ending the run (P1 fix: a non-dict in the middle of
+        # a same-role run used to break the inner loop and start a new run,
+        # mis-treating what could be a cumulative sequence as two independent runs).
+        while i < len(transcript):
+            if not isinstance(transcript[i], dict):
+                i += 1
+                continue
+            run_role = transcript[i].get("role")
+            if not isinstance(run_role, str) or run_role.strip() != current_role:
```
TimeToBuildBob: ❌ **P1** — In `de_accumulate_transcript`, the inner loop that collects a same-role run breaks when it encounters a dict whose `role` is a string that strips to empty (e.g., `""` or `"   "`), because `run_role.strip() != current_role` evaluates to True. A non-dict entry in the same position is skipped without ending the run (lines 119-121), but an empty-role dict ends the run. The outer loop then discards that empty-role dict (lines 108-110) and starts a new run on the next dict with the same role. This splits what should be a single cumulative run into two, so `_collapse_partial_chain` sees each part independently and cannot collapse partials across the junk entry. For example, a transcript of `[{"role":"user","text":"I"}, {"role":"","text":"ignored"}, {"role":"user","text":"I want"}]` with `cumulative=True` produces `"USER: I\n\nUSER: I want"` instead of `"USER: I want"`. Since `collect_voice_call_documents` always passes `cumulative=True`, any archived call with an empty-role entry in the middle of a speaker's run will index duplicated or partial turns. The fix is to skip empty-role dicts within the run without ending it, matching the non-dict handling. _(2/3 passes — pass 1 looked and disagreed)_

TimeToBuildBob:
## Convergence adjudication — final P1 fixed

### Fixed
- **P1 (`5d54f56ab1f6`) — empty-role junk split cumulative same-role runs.** Commit `427010db` now skips empty/non-string roles inside the active run, matching the existing handling for non-dict junk. The exact `user: "I"` → empty role → `user: "I want"` regression now collapses to one `USER: I want` block.

### Remaining
- **None.** The latest score-driving finding was blocking and is fixed; no finding is being dismissed as a nit or false positive.

### CI
- The previous head `7dca0ad6` was fully green. CI is currently running on `427010db`; local verification is **40/40 source tests**, Ruff check/format clean, and the commit's full pre-commit gate passed.

### Domain risk
- Transcript provenance remains the fragile boundary: only callers that explicitly opt into `cumulative=True` collapse monotonic same-role prefix chains. Malformed entries are now ignored consistently without inventing speaker boundaries. A maintainer manual test is optional, not a merge blocker.

### Convergence
- Detector reports `round_convergence.status: no_history`, `stable_rounds: 0` of 2. The attempt-cap escalation therefore has no usable stable-round history. No further reviewer trigger was sent; stop at maintainer judgment after CI settles.

packages/gptme-rag/src/gptme_rag/sources.py
```diff
@@ -0,0 +1,321 @@
...
+            continue
+        if not resolved_path.is_file():
+            # Skip anything that is not a regular file (e.g. a FIFO named
+            # ``*.json``, which would block the collector on read).
+            continue
+        try:
+            # Read the path we validated, not the original directory entry.
+            # Otherwise a symlink swap between resolve() and read_text() could
+            # bypass the repo-root guard.
+            data = json.loads(resolved_path.read_text(encoding="utf-8"))
```
TimeToBuildBob: ⚠️ **P2** — The per-file read in collect_voice_call_documents has a time-of-check to time-of-use (TOCTOU) gap. The code resolves each `*.json` path, checks that the resolved path is inside `repo_root` and is a regular file, then calls `resolved_path.read_text()`. Between the `is_file()` check and the read, an attacker with write access to the source directory could replace the file at the resolved path with a symlink to a file outside `repo_root`. Because the code reads from the resolved path (an absolute path string) and does not re-validate after opening, the read would follow the new symlink and pull out-of-repo content into the index. The comment on line 273-275 claims reading the resolved path prevents a symlink swap, but it does not: swapping the file at that path changes the target of the read without changing the path string. The repo-root guard is intended as a security boundary against indexing outside files, so this is a real (if narrow) vulnerability. To mitigate, open the file with `os.open` and `os.fstat` to confirm the inode is unchanged, or call `is_relative_to` again after opening.

TimeToBuildBob:
Fixed the remaining P2 TOCTOU finding in `9bbbb471`.

The collector now opens the resolved path with `O_NOFOLLOW | O_NONBLOCK`, validates the opened descriptor as a regular file, and checks its device/inode against a no-follow `stat` before parsing. A final-component swap to a symlink can no longer follow content outside `repo_root`; a swap to a FIFO also cannot block collection. Added an adversarial regression that performs the swap between resolution and `os.open` and verifies the archive is rejected.

Local verification: 41 source tests pass; Ruff check/format, package-local mypy, targeted pre-commit, and diff checks pass. CI and the automatic AI-review rerun are in progress. No Greptile trigger was sent.

packages/gptme-rag/src/gptme_rag/sources.py
TimeToBuildBob: Fixed in `9bbbb471`: the collector opens the validated path with `O_NOFOLLOW | O_NONBLOCK`, checks the descriptor is a regular file, and verifies its device/inode against a no-follow path stat before parsing. Added an adversarial regression that swaps the file to an out-of-repo symlink between resolution and `os.open`; collection rejects it. 41 source tests pass.

packages/gptme-rag/src/gptme_rag/sources.py
```diff
@@ -0,0 +1,337 @@
...
+            if not isinstance(transcript[i], dict):
+                i += 1
+                continue
+            run_role = transcript[i].get("role")
+            if not isinstance(run_role, str) or not run_role.strip():
+                i += 1
+                continue
+            if run_role.strip() != current_role:
+                break
+            run_texts.append(str(transcript[i].get("text", "")))
```
TimeToBuildBob: ⚠️ **P2** — In `de_accumulate_transcript`, the line `run_texts.append(str(transcript[i].get("text", "")))` converts a `None` text value to the string `"None"`. If a transcript entry has `"text": null` (possibly representing an empty STT result), the output becomes `"USER: None"`, which gets indexed as real content. This pollutes the search index with the literal word "None" and can cause false matches or distort the transcript representation. The function should skip entries whose text is not a non-empty string, or handle `None` by skipping that turn rather than coercing it.

packages/gptme-rag/tests/test_sources.py
```diff
@@ -0,0 +1,625 @@
...
+    (call_dir / "escape.json").symlink_to(secret)
+
+    read_paths: list[Path] = []
+    original_read_text = Path.read_text
+
+    def tracking_read_text(self: Path, *args, **kwargs):  # type: ignore[no-untyped-def]
+        read_paths.append(self)
+        return original_read_text(self, *args, **kwargs)
+
+    monkeypatch.setattr(Path, "read_text", tracking_read_text)
```
TimeToBuildBob: ⚠️ **P2** — The test `test_collect_voice_calls_does_not_read_symlink_escaping_repo` monkeypatches `Path.read_text` to track read attempts, but `collect_voice_call_documents` never calls `Path.read_text` — it uses `os.fdopen` and `json.load` to read files. Consequently, the tracking function `tracking_read_text` is never invoked, so `read_paths` remains empty and the assertion `assert read_paths == []` passes vacuously. The test does not actually verify that the escaping symlink is not read; it only confirms the code does not use `Path.read_text`, which is irrelevant. This gives false confidence that the security property is covered.

packages/gptme-rag/tests/test_sources.py
```diff
@@ -0,0 +1,625 @@
...
+        if candidate == original_call_path and not candidate.is_symlink():
+            seen_flags = flags
+            candidate.unlink()
+            candidate.symlink_to(secret)
+        return original_open(path, flags, *args, **kwargs)
+
+    monkeypatch.setattr(os, "open", swapping_open)
+
+    assert collect_voice_call_documents(call_dir, repo_root=repo) == []
+    assert seen_flags & os.O_NOFOLLOW
```
TimeToBuildBob: ⚠️ **P2** — The test `test_collect_voice_calls_rejects_symlink_swapped_after_resolution` relies on `os.O_NOFOLLOW` being defined and available. The source code correctly guards this with `getattr(os, "O_NOFOLLOW", 0)`, so the library works on platforms without `O_NOFOLLOW` (e.g., Windows). However, the test directly accesses `os.O_NOFOLLOW` in the assertion, which raises `AttributeError` on such platforms. This makes the test non-hermetic and fails on any platform that lacks this constant, preventing the test suite from running there. The test should check `hasattr(os, "O_NOFOLLOW")` and skip or adjust accordingly.

packages/gptme-rag/src/gptme_rag/sources.py
TimeToBuildBob: Fixed in `e46be1d9`: non-string and blank transcript text is skipped instead of coerced, with a null/numeric regression test. 42 source tests pass.

packages/gptme-rag/tests/test_sources.py
TimeToBuildBob: Fixed in `e46be1d9`: the regression now instruments the actual `os.open` boundary and proves the out-of-repo target is never opened, rather than tracking the obsolete `Path.read_text` path.

packages/gptme-rag/tests/test_sources.py
TimeToBuildBob: Fixed in `e46be1d9`: the test uses `getattr(os, "O_NOFOLLOW", 0)` and asserts the bit only when the platform exposes it.

TimeToBuildBob:
Fixed all three P2 findings in `e46be1d9`:

- non-string transcript text is skipped instead of becoming indexable `None`/numeric content;
- the escaping-symlink regression now instruments the actual `os.open` boundary rather than obsolete `Path.read_text` behavior;
- the portability assertion now handles platforms without `O_NOFOLLOW`.

Added the malformed-text regression. Local verification: 42 source tests pass; Ruff check/format, package-local mypy, targeted pre-commit, and diff checks pass. Replied to and resolved all three inline threads. CI and the automatic AI-review rerun are in progress; no Greptile trigger was sent.

packages/gptme-rag/src/gptme_rag/sources.py
```diff
@@ -0,0 +1,339 @@
...
+            continue
+        finally:
+            if fd >= 0:
+                os.close(fd)
+        if not isinstance(data, dict):
+            continue
+        transcript = data.get("transcript", [])
+        if not isinstance(transcript, list) or not transcript:
+            continue
+        text = de_accumulate_transcript(transcript, cumulative=True)
```
TimeToBuildBob: ❌ **P1** — collect_voice_call_documents unconditionally calls de_accumulate_transcript(transcript, cumulative=True). In a transcript where two distinct consecutive same-role turns happen to be a prefix chain (e.g., "I want" followed by "I want to go"), _collapse_partial_chain replaces the first with the second, dropping "I want" from the indexed content. This is data loss for archives that are not actually cumulative (i.e., final transcripts), but the caller has no way to opt out. The function is meant for voice calls, and if the archived transcript is a finalized rendering rather than a stream of STT partials, content is lost silently.

packages/gptme-rag/src/gptme_rag/sources.py
```diff
@@ -0,0 +1,339 @@
...
+            logger.warning("sources: skipping %s — resolves outside repo_root", path)
+            continue
+        fd = -1
+        try:
+            # O_NOFOLLOW closes the final-component race between resolve() and
+            # open(): replacing a checked file with a symlink must not let the
+            # read escape repo_root. O_NONBLOCK also keeps a swapped-in FIFO
+            # from blocking before fstat() can reject it.
+            flags = os.O_RDONLY | getattr(os, "O_NOFOLLOW", 0) | getattr(os, "O_NONBLOCK", 0)
+            fd = os.open(resolved_path, flags)
```
TimeToBuildBob: ⚠️ **P2** — The symlink protection in collect_voice_call_documents resolves the path (line 264) and later opens it (line 279) as separate steps. Between these two operations an attacker could replace an intermediate directory component with a symlink pointing outside repo_root. The open() call only uses O_NOFOLLOW on the final path component, so a swapped directory is followed, letting the read escape the repo. This undermines the repo_root safety guard and could index content from outside the repository.

packages/gptme-rag/src/gptme_rag/sources.py
TimeToBuildBob: Fixed in `bba93125`: `collect_voice_call_documents()` now exposes the transcript-format choice as `cumulative=`. Existing cumulative voice-call archives keep their current default; callers ingesting finalized transcripts can pass `cumulative=False`, which preserves both `"I want"` and `"I want to go"`. Added the exact collector-level regression.

packages/gptme-rag/src/gptme_rag/sources.py
TimeToBuildBob: Fixed in `bba93125`: guarded reads now walk from the open `repo_root` descriptor with `dir_fd` plus `O_NOFOLLOW` on every directory component and the final file. An intermediate directory swapped to an escaping symlink is rejected; the new adversarial regression performs that swap at the `os.open` boundary.

TimeToBuildBob:
Fixed both latest findings in `bba93125`:

- `collect_voice_call_documents()` now exposes `cumulative=` so finalized transcripts can opt out of prefix-chain collapsing; the default remains cumulative for the archive format this collector targets;
- rooted reads now traverse via directory descriptors with `O_NOFOLLOW` on every component, closing the intermediate-directory swap race.

Added collector-level finalized-transcript and adversarial intermediate-symlink-swap regressions. Local verification: 44 source tests pass; Ruff check/format, package-local mypy, targeted pre-commit, and diff checks pass. Replied to both inline threads. CI and the automatic AI-review rerun are in progress; no Greptile trigger was sent.

packages/gptme-rag/src/gptme_rag/sources.py
```diff
@@ -0,0 +1,361 @@
...
+
+
+_DIRECTORY_OPEN_FLAGS = os.O_RDONLY | getattr(os, "O_DIRECTORY", 0) | getattr(os, "O_NOFOLLOW", 0)
+_FILE_OPEN_FLAGS = os.O_RDONLY | getattr(os, "O_NOFOLLOW", 0) | getattr(os, "O_NONBLOCK", 0)
+
+
+def _open_beneath(root: Path, relative_path: Path) -> int:
+    """Open a regular file beneath *root* without following symlink components."""
+    if not hasattr(os, "O_NOFOLLOW"):
+        raise OSError("safe path traversal requires O_NOFOLLOW")
```
TimeToBuildBob, line 56: ⚠️ **P2** — On platforms that lack os.O_NOFOLLOW (e.g. Windows), `_open_beneath` unconditionally raises OSError with the message "safe path traversal requires O_NOFOLLOW". In `collect_voice_call_documents`, every call to `_open_beneath` happens inside a try block that catches OSError and `continue`s to the next file (lines 314-316). The result is that when `repo_root` is provided on such a platform, every JSON file is silently skipped and the function returns an empty list even though the directory contains valid, readable call archives. The caller gets no error, warning, or partial result — just a missing corpus, which is a quiet data-loss failure. The check is deliberate but the swallowing of the resulting OSError makes the absence of O_NOFOLLOW indistinguishable from a genuinely unreadable file. The guard should either propagate a clear error when O_NOFOLLOW is unavailable, or fall back to a safe path-opening strategy that does not require O_NOFOLLOW (e.g. opening each component and stat-ing to ensure it is not a symlink), so that the collection does not silently produce nothing.

PR #1466, opened 2026-08-19
feat(gptme-rag): memory_type classification + ranking boost (Phase 2.2)

TimeToBuildBob:
## What

Ports the `memory_type` classification + ranking boost from the brain scripts into gptme-rag, as part of upstreaming the ambient memory pipeline.

**Part of the series**: Phase 2.1 (#1465, open), 2.3 (#1440, merged), 2.2 (this PR).

## New module: `gptme_rag/memory_type.py`

Documents in a personal knowledge base serve different roles — some are identity (SOUL.md, ABOUT.md), some are active goals, ongoing projects, or preferences. This module classifies documents by type and adjusts TF-IDF ranking accordingly.

- `load_memory_type_map(path)` — load caller-supplied JSON rules (no hardcoded paths; the brain repo retains its config)
- `classify_memory_type(rel_path, metadata, rules)` — exact path → glob → task-rule resolution
- `classify_document(rel_path, content, rules)` — convenience wrapper that parses YAML frontmatter
- `weighted_similarity(sim, mem_type, requested)` — 1.35× boost on match, 0.9× penalty on mismatch

## `TfidfIndex.search()` integration

Added optional `memory_types: set[str] | None = None` parameter. When set:
- Weighted scores replace raw cosine for sorting (so boosted docs actually rank higher)
- Unknown type labels are silently filtered to `SUPPORTED_MEMORY_TYPES`
- Docs without a `memory_type` metadata key are ranked by raw cosine (no penalty)

## Tests

36 tests in `tests/test_memory_type.py`, all passing:
- `load_memory_type_map`: missing file, malformed JSON, valid file
- `classify_memory_type`: exact paths, glob patterns, task rules (priority/state/tags)
- `classify_document`: frontmatter extraction
- `weighted_similarity`: boost, penalty, clamp, no-op cases
- `TfidfIndex.search()` integration: reranking, unknown types, no-tagged-docs safety

## Source

Brain script: `scripts/build-ambient-memory-index.py` — the `_load_memory_type_map`, `_classify_memory_type`, `_maybe_attach_memory_type`, and `_weighted_similarity` functions.

Generalised for library use: caller supplies the rules dict, no hardcoded file paths.

### DIFF
--- a/packages/gptme-rag/src/gptme_rag/lexical.py
+++ b/packages/gptme-rag/src/gptme_rag/lexical.py
@@ -253,7 +253,7 @@ class TfidfIndex:
         sorted_indices = sorted(range(len(raw_sims)), key=_weighted, reverse=True)
 
         hits: list[LexicalHit] = []
-        for idx in sorted_indices:
+        for rank, idx in enumerate(sorted_indices):
             # Keep relevance_floor's existing raw-cosine semantics.  Memory-type
             # weights only rerank documents that already clear the relevance gate.
             if float(raw_sims[idx]) < self.relevance_floor:
@@ -262,7 +262,9 @@ class TfidfIndex:
             doc = self._documents[idx]
             if self._source_path(doc) in excluded:
                 continue
-            hits.append(LexicalHit(document=doc, score=round(score, 4), rank=len(hits)))
+            # Preserve LexicalHit.rank as the candidate's position in the full
+            # sorted list, including candidates skipped by filters.
+            hits.append(LexicalHit(document=doc, score=round(score, 4), rank=rank))
             if len(hits) >= n_results:
                 break
         return hits
--- a/packages/gptme-rag/src/gptme_rag/lexical.py
+++ b/packages/gptme-rag/src/gptme_rag/lexical.py
@@ -27,6 +27,7 @@ from datetime import datetime, timezone
 from pathlib import Path
 
 from .indexing.document import Document
+from .memory_type import SUPPORTED_MEMORY_TYPES, weighted_similarity
 
 logger = logging.getLogger(__name__)
 
@@ -197,11 +198,21 @@ class TfidfIndex:
         query: str,
         n_results: int = 5,
         exclude_paths: set[str] | None = None,
+        memory_types: set[str] | None = None,
     ) -> list[LexicalHit]:
         """Return up to ``n_results`` hits ranked by cosine similarity.
 
         ``exclude_paths`` is a set of source paths to skip (e.g. documents already
         in context).  Hits below ``relevance_floor`` are discarded.
+
+        ``memory_types`` is an optional set of memory-type labels (e.g.
+        ``{"goal", "identity"}``) that boosts matching documents and penalises
+        non-matching ones via :func:`~gptme_rag.memory_type.weighted_similarity`.
+        Documents must carry a ``"memory_type"`` key in their metadata (set by the
+        caller during indexing) for this to take effect; documents without the key
+        are ranked by raw cosine similarity.  Only types in
+        :data:`~gptme_rag.memory_type.SUPPORTED_MEMORY_TYPES` are considered —
+        unrecognised values are silently ignored.
         """
         if n_results <= 0:
             return []
@@ -217,14 +228,31 @@ class TfidfIndex:
             ) from exc
 
         q_vec = self._vectorizer.transform([query])
-        sims = cosine_similarity(q_vec, self._matrix)[0]
+        raw_sims = cosine_similarity(q_vec, self._matrix)[0]
         excluded = exclude_paths or set()
 
+        # Normalise the requested memory types to only known values so that a
+        # typo in the caller's set does not silently ignore all boosts.
+        requested: set[str] | None = None
+        if memory_types:
+            requested = {t for t in memory_types if t in SUPPORTED_MEMORY_TYPES}
+            if not requested:
+                requested = None
+
+        # Compute weighted scores and sort by them so that boosted documents
+        # rank above documents with a higher raw similarity but wrong type.
+        def _weighted(idx: int) -> float:
+            doc = self._documents[idx]
+            mem_type: str | None = doc.metadata.get("memory_type") or None
+            return weighted_similarity(float(raw_sims[idx]), mem_type, requested)
+
+        sorted_indices = sorted(range(len(raw_sims)), key=_weighted, reverse=True)
+
         hits: list[LexicalHit] = []
-        for rank, idx in enumerate(sims.argsort()[::-1]):
-            score = float(sims[idx])
+        for rank, idx in enumerate(sorted_indices):
+            score = _weighted(idx)
             if score < self.relevance_floor:
-                continue
+                break  # sorted descending — nothing below this will clear the floor
             doc = self._documents[idx]
             if self._source_path(doc) in excluded:
                 continue
--- a/packages/gptme-rag/src/gptme_rag/lexical.py
+++ b/packages/gptme-rag/src/gptme_rag/lexical.py
@@ -253,7 +253,7 @@ class TfidfIndex:
         sorted_indices = sorted(range(len(raw_sims)), key=_weighted, reverse=True)
 
         hits: list[LexicalHit] = []
-        for rank, idx in enumerate(sorted_indices):
+        for idx in sorted_indices:
             # Keep relevance_floor's existing raw-cosine semantics.  Memory-type
             # weights only rerank documents that already clear the relevance gate.
             if float(raw_sims[idx]) < self.relevance_floor:
@@ -262,7 +262,7 @@ class TfidfIndex:
             doc = self._documents[idx]
             if self._source_path(doc) in excluded:
                 continue
-            hits.append(LexicalHit(document=doc, score=round(score, 4), rank=rank))
+            hits.append(LexicalHit(document=doc, score=round(score, 4), rank=len(hits)))
             if len(hits) >= n_results:
                 break
         return hits
--- a/packages/gptme-rag/src/gptme_rag/lexical.py
+++ b/packages/gptme-rag/src/gptme_rag/lexical.py
@@ -250,7 +250,13 @@ class TfidfIndex:
                 mem_type = None
             return weighted_similarity(float(raw_sims[idx]), mem_type, requested)
 
-        sorted_indices = sorted(range(len(raw_sims)), key=_weighted, reverse=True)
+        # Preserve numpy's historical default-path ordering exactly.  Equal
+        # scores have no semantic ordering contract, but callers still should
+        # not see an unrelated order change when memory-type weighting is off.
+        if requested is None:
+            sorted_indices = raw_sims.argsort()[::-1]
+        else:
+            sorted_indices = sorted(range(len(raw_sims)), key=_weighted, reverse=True)
 
         hits: list[LexicalHit] = []
         for rank, idx in enumerate(sorted_indices):
--- a/packages/gptme-rag/src/gptme_rag/lexical.py
+++ b/packages/gptme-rag/src/gptme_rag/lexical.py
@@ -244,6 +244,9 @@ class TfidfIndex:
         def _weighted(idx: int) -> float:
             doc = self._documents[idx]
             mem_type: str | None = doc.metadata.get("memory_type") or None
+            # Treat unrecognised memory-type labels as untagged (no penalty).
+            if mem_type not in SUPPORTED_MEMORY_TYPES:
+                mem_type = None
             return weighted_similarity(float(raw_sims[idx]), mem_type, requested)
 
         sorted_indices = sorted(range(len(raw_sims)), key=_weighted, reverse=True)
--- a/packages/gptme-rag/src/gptme_rag/lexical.py
+++ b/packages/gptme-rag/src/gptme_rag/lexical.py
@@ -203,7 +203,8 @@ class TfidfIndex:
         """Return up to ``n_results`` hits ranked by cosine similarity.
 
         ``exclude_paths`` is a set of source paths to skip (e.g. documents already
-        in context).  Hits below ``relevance_floor`` are discarded.
+        in context).  Hits whose raw cosine similarity is below
+        ``relevance_floor`` are discarded before memory-type weighting.
 
         ``memory_types`` is an optional set of memory-type labels (e.g.
         ``{"goal", "identity"}``) that boosts matching documents and penalises
@@ -253,9 +254,11 @@ class TfidfIndex:
 
         hits: list[LexicalHit] = []
         for rank, idx in enumerate(sorted_indices):
+            # Keep relevance_floor's existing raw-cosine semantics.  Memory-type
+            # weights only rerank documents that already clear the relevance gate.
+            if float(raw_sims[idx]) < self.relevance_floor:
+                continue  # weighted order is not monotonic in raw similarity
             score = _weighted(idx)
-            if score < self.relevance_floor:
-                break  # sorted descending — nothing below this will clear the floor
             doc = self._documents[idx]
             if self._source_path(doc) in excluded:
                 continue
--- a/packages/gptme-rag/src/gptme_rag/memory_type.py
+++ b/packages/gptme-rag/src/gptme_rag/memory_type.py
@@ -161,9 +161,8 @@ def _extract_frontmatter(content: str) -> dict[str, Any]:
 
         parsed = yaml.safe_load(frontmatter_text) or {}
     except Exception:
-        # Fallback: simple key: value regex (no nested structures)
-        import re
-
+        # Fallback: simple key: value regex (no nested structures). Inline YAML
+        # lists stay as strings here and are normalised by _coerce_string_list.
         parsed = {k: v for k, v in re.findall(r"^(\w[\w_-]*):\s*(.*)", frontmatter_text, re.M)}
     return parsed if isinstance(parsed, dict) else {}
 
@@ -175,7 +174,11 @@ def _coerce_string_list(value: Any) -> list[str]:
     strings that YAML leaves as a single string (e.g. ``tags: ai, project``).
     """
     if isinstance(value, str):
-        # Split on commas to handle "preference, project" as two tags.
+        # Split on commas to handle both plain comma-separated tags and inline
+        # YAML lists emitted as strings by the no-PyYAML fallback parser.
+        value = value.strip()
+        if value.startswith("[") and value.endswith("]"):
+            value = value[1:-1]
         return [p.strip().lower() for p in value.split(",") if p.strip()]
     if isinstance(value, list):
         # Also split each list item on commas — YAML may parse "- preference, project"
--- a/packages/gptme-rag/src/gptme_rag/memory_type.py
+++ b/packages/gptme-rag/src/gptme_rag/memory_type.py
@@ -220,7 +220,9 @@ def classify_memory_type(
     exact_paths = rules.get("exact_paths", {})
     if isinstance(exact_paths, dict) and rel_path in exact_paths:
         memory_type = str(exact_paths[rel_path])
-        return memory_type if memory_type in SUPPORTED_MEMORY_TYPES else None
+        if memory_type in SUPPORTED_MEMORY_TYPES:
+            return memory_type
+        # Unsupported type (e.g. typo) — fall through to glob/task rules
 
     # 2. Glob-pattern match
     glob_paths = rules.get("glob_paths", {})
--- a/packages/gptme-rag/src/gptme_rag/memory_type.py
+++ b/packages/gptme-rag/src/gptme_rag/memory_type.py
@@ -1,0 +1,239 @@
+"""Memory-type classification and ranking boost for gptme-rag.
+
+Documents in a personal knowledge base serve different roles: some encode
+identity (SOUL.md, ABOUT.md), others encode active goals, ongoing projects, or
+user preferences.  Classifying documents by *memory type* lets the ranking layer
+boost results that match the caller's retrieval intent without hard-filtering.
+
+The classification is driven by a caller-supplied rules dict (loaded from JSON)
+rather than hardcoded paths — callers embed their own policy and pass it in.
+This keeps the library generic while the brain repo retains its config.
+
+Example rules dict:
+
+    {
+        "exact_paths": {"SOUL.md": "identity", "ABOUT.md": "identity"},
+        "glob_paths": {"tasks/*.md": "project"},
+        "task_rules": {
+            "goal_priorities": ["high"],
+            "goal_states": ["active"],
+            "preference_tags": ["preference"],
+            "project_tags": ["project"],
+            "default": "project"
+        }
+    }
+
+Usage::
+
+    rules = load_memory_type_map(Path("state/ambient-memory/memory-type-map.json"))
+    memory_type = classify_memory_type("tasks/my-task.md", {"priority": "high"}, rules)
+    score = weighted_similarity(0.8, memory_type, {"goal", "identity"})
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+from fnmatch import fnmatch
+from pathlib import Path
+from typing import Any
+
+logger = logging.getLogger(__name__)
+
+# ---------------------------------------------------------------------------
+# Constants
+# ---------------------------------------------------------------------------
+
+#: Memory types the library recognises.  Others are treated as unknown and
+#: silently ignored rather than causing a hard error.
+SUPPORTED_MEMORY_TYPES: tuple[str, ...] = ("identity", "preference", "goal", "project")
+
+#: Multiplicative boost applied to the similarity score when the document's
+#: memory type matches the set of requested types.
+MEMORY_TYPE_BOOST: float = 1.35
+
+#: Multiplicative penalty applied when the document's memory type does *not*
+#: match any of the requested types (and the caller provided a non-empty set).
+MEMORY_TYPE_PENALTY: float = 0.9
+
+
+# ---------------------------------------------------------------------------
+# Rules loading
+# ---------------------------------------------------------------------------
+
+
+def load_memory_type_map(map_file: Path | None = None) -> dict[str, Any]:
+    """Load memory-type classification rules from a JSON file.
+
+    Returns an empty dict when *map_file* is ``None`` or does not exist —
+    callers can safely treat a missing file as "no tagging".
+
+    The expected file schema::
+
+        {
+            "exact_paths":  {"rel/path.md": "identity"},
+            "glob_paths":   {"tasks/*.md": "project"},
+            "task_rules":   {
+                "goal_priorities":  ["high"],
+                "goal_states":      ["active"],
+                "preference_tags":  ["preference"],
+                "project_tags":     ["project"],
+                "default":          "project"
+            }
+        }
+
+    Args:
+        map_file: Absolute or repo-relative path to the JSON rules file.
+    """
+    if map_file is None or not map_file.exists():
+        return {}
+    try:
+        data = json.loads(map_file.read_text(encoding="utf-8"))
+    except (OSError, json.JSONDecodeError) as exc:
+        logger.warning("Could not load memory-type map from %s: %s", map_file, exc)
+        return {}
+    return data if isinstance(data, dict) else {}
+
+
+# ---------------------------------------------------------------------------
+# Frontmatter helpers (inline — avoids cross-module dep on task_retrieval)
+# ---------------------------------------------------------------------------
+
+
+def _extract_frontmatter(content: str) -> dict[str, Any]:
+    """Parse a YAML frontmatter block if present; returns {} on failure."""
+    if not content.startswith("---\n"):
+        return {}
+    end = content.find("\n---\n", 4)
+    if end == -1:
+        return {}
+    frontmatter_text = content[4:end]
+    try:
+        import yaml  # optional — yaml is available in gptme-rag's dependency set
+
+        parsed = yaml.safe_load(frontmatter_text) or {}
+    except Exception:
+        # Fallback: simple key: value regex (no nested structures)
+        import re
+
+        parsed = {k: v for k, v in re.findall(r"^(\w[\w_-]*):\s*(.*)", frontmatter_text, re.M)}
+    return parsed if isinstance(parsed, dict) else {}
+
+
+def _coerce_string_list(value: Any) -> list[str]:
+    """Normalise a frontmatter value to a list of lowercase strings."""
+    if isinstance(value, str):
+        return [value.strip().lower()] if value.strip() else []
+    if isinstance(value, list):
+        return [item.strip().lower() for item in value if isinstance(item, str) and item.strip()]
+    return []
+
+
+# ---------------------------------------------------------------------------
+# Classification
+# ---------------------------------------------------------------------------
+
+
+def classify_memory_type(
+    rel_path: str,
+    metadata: dict[str, Any] | None,
+    rules: dict[str, Any],
+) -> str | None:
+    """Resolve a document's memory type from classification rules + metadata.
+
+    Resolution order:
+    1. ``exact_paths`` — precise path match (highest priority)
+    2. ``glob_paths`` — fnmatch pattern match
+    3. ``task_rules`` — for ``tasks/`` prefixed paths, inspect metadata fields
+
+    Returns one of :data:`SUPPORTED_MEMORY_TYPES` or ``None`` when no rule
+    matches.
+
+    Args:
+        rel_path: Repository-relative path to the document (e.g. ``"tasks/foo.md"``).
+        metadata: Frontmatter metadata dict, if available.  May be ``None``.
+        rules: Classification rules dict as returned by :func:`load_memory_type_map`.
+    """
+    metadata = metadata or {}
+
+    # 1. Exact-path match
+    exact_paths: dict[str, Any] = rules.get("exact_paths", {})
+    if rel_path in exact_paths:
+        memory_type = str(exact_paths[rel_path])
+        return memory_type if memory_type in SUPPORTED_MEMORY_TYPES else None
+
+    # 2. Glob-pattern match
+    for pattern, memory_type in rules.get("glob_paths", {}).items():
+        memory_type_str = str(memory_type)
+        if fnmatch(rel_path, pattern) and memory_type_str in SUPPORTED_MEMORY_TYPES:
+            return memory_type_str
+
+    # 3. Task-specific rules (only for paths under tasks/)
+    if rel_path.startswith("tasks/"):
+        task_rules: dict[str, Any] = rules.get("task_rules", {})
+        priority = str(metadata.get("priority", "")).strip().lower()
+        state = str(metadata.get("state", "")).strip().lower()
+        tags = _coerce_string_list(metadata.get("tags"))
+
+        if priority in task_rules.get("goal_priorities", []):
+            return "goal"
+        if state in task_rules.get("goal_states", []):
+            return "goal"
+        if any(tag in task_rules.get("preference_tags", []) for tag in tags):
+            return "preference"
+        if any(tag in task_rules.get("project_tags", []) for tag in tags):
+            return "project"
+        default_type = task_rules.get("default")
+        return default_type if default_type in SUPPORTED_MEMORY_TYPES else None
+
+    return None
+
+
+def classify_document(
+    rel_path: str,
+    content: str,
+    rules: dict[str, Any],
+) -> str | None:
+    """Classify a document by memory type, extracting frontmatter automatically.
+
+    Convenience wrapper over :func:`classify_memory_type` that parses the
+    document's YAML frontmatter from *content* so callers do not need to parse
+    it themselves.
+
+    Args:
+        rel_path: Repository-relative path to the document.
+        content: Raw document content (may include frontmatter).
+        rules: Classification rules dict.
+    """
+    metadata = _extract_frontmatter(content)
+    return classify_memory_type(rel_path, metadata, rules)
+
+
+# ---------------------------------------------------------------------------
+# Ranking boost
+# ---------------------------------------------------------------------------
+
+
+def weighted_similarity(
+    similarity: float,
+    memory_type: str | None,
+    requested_memory_types: set[str] | None,
+) -> float:
+    """Apply a boost or penalty to *similarity* based on memory-type match.
+
+    When the caller has no preference (``requested_memory_types`` is ``None`` or
+    empty), the similarity is returned unchanged.  When a preference is provided:
+
+    * A match applies :data:`MEMORY_TYPE_BOOST` (default 1.35×), clamped to 1.0.
+    * A miss applies :data:`MEMORY_TYPE_PENALTY` (default 0.9×).
+
+    Args:
+        similarity: Raw cosine similarity score in ``[0, 1]``.
+        memory_type: The document's resolved memory type (or ``None``).
+        requested_memory_types: Set of types the caller wants to boost.
+    """
+    if not requested_memory_types or not memory_type:
+        return similarity
+    if memory_type in requested_memory_types:
+        return min(1.0, similarity * MEMORY_TYPE_BOOST)
+    return similarity * MEMORY_TYPE_PENALTY
--- a/packages/gptme-rag/src/gptme_rag/memory_type.py
+++ b/packages/gptme-rag/src/gptme_rag/memory_type.py
@@ -89,7 +89,7 @@ def load_memory_type_map(map_file: Path | None = None) -> dict[str, Any]:
         return {}
     try:
         data = json.loads(map_file.read_text(encoding="utf-8"))
-    except (OSError, json.JSONDecodeError) as exc:
+    except (OSError, UnicodeDecodeError, json.JSONDecodeError) as exc:
         logger.warning("Could not load memory-type map from %s: %s", map_file, exc)
         return {}
     return data if isinstance(data, dict) else {}
--- a/packages/gptme-rag/src/gptme_rag/memory_type.py
+++ b/packages/gptme-rag/src/gptme_rag/memory_type.py
@@ -233,7 +233,11 @@ def classify_memory_type(
     if isinstance(glob_paths, dict):
         for pattern, memory_type in glob_paths.items():
             memory_type_str = str(memory_type)
-            if _glob_match_path(rel_path, pattern) and memory_type_str in SUPPORTED_MEMORY_TYPES:
+            if (
+                isinstance(pattern, str)
+                and _glob_match_path(rel_path, pattern)
+                and memory_type_str in SUPPORTED_MEMORY_TYPES
+            ):
                 return memory_type_str
 
     # 3. Task-specific rules (only for paths under tasks/)
@@ -298,7 +302,7 @@ def weighted_similarity(
     When the caller has no preference (``requested_memory_types`` is ``None`` or
     empty), the similarity is returned unchanged.  When a preference is provided:
 
-    * A match applies :data:`MEMORY_TYPE_BOOST` (default 1.35×), clamped to 1.0.
+    * A match applies :data:`MEMORY_TYPE_BOOST` (default 1.35×).
     * A miss applies :data:`MEMORY_TYPE_PENALTY` (default 0.9×).
 
     Args:
@@ -309,5 +313,5 @@ def weighted_similarity(
     if not requested_memory_types or not memory_type:
         return similarity
     if memory_type in requested_memory_types:
-        return min(1.0, similarity * MEMORY_TYPE_BOOST)
+        return similarity * MEMORY_TYPE_BOOST
     return similarity * MEMORY_TYPE_PENALTY
--- a/packages/gptme-rag/src/gptme_rag/memory_type.py
+++ b/packages/gptme-rag/src/gptme_rag/memory_type.py
@@ -34,7 +34,7 @@ from __future__ import annotations
 
 import json
 import logging
-from fnmatch import fnmatch
+import re
 from pathlib import Path
 from typing import Any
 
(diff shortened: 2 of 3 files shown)


packages/gptme-rag/src/gptme_rag/memory_type.py
```diff
@@ -0,0 +1,239 @@
...
+            }
+        }
+
+    Args:
+        map_file: Absolute or repo-relative path to the JSON rules file.
+    """
+    if map_file is None or not map_file.exists():
+        return {}
+    try:
+        data = json.loads(map_file.read_text(encoding="utf-8"))
```
TimeToBuildBob, line 91: ❌ **P1** — `load_memory_type_map` is documented to return an empty dict when the file is missing or unreadable, and it catches `OSError` and `json.JSONDecodeError`. However, `Path.read_text(encoding="utf-8")` can raise `UnicodeDecodeError` if the file contains invalid UTF-8 bytes. `UnicodeDecodeError` is a subclass of `ValueError`, not `OSError` or `json.JSONDecodeError`, so it is not caught. As a result, a rules file with a byte sequence that is not valid UTF-8 will raise an unhandled exception instead of returning `{}` as the function contract promises, potentially crashing the caller that relies on a safe fallback for a corrupt or non-UTF-8 config file.

TimeToBuildBob:
## 🤖 AI code review

This PR adds a new module memory_type.py that classifies documents by memory type (identity, preference, goal, project) using caller-supplied rules and applies a ranking boost/penalty to TF-IDF search results. TfidfIndex.search() gains an optional memory_types parameter that filters to supported types and re-ranks results by weighted similarity. The PR also adds a comprehensive test file for the new functionality.

**❌ P1 high** — `packages/gptme-rag/src/gptme_rag/memory_type.py:198`

In `_coerce_string_list`, when the input is a string, it strips brackets if present, but does not strip surrounding quotes. When PyYAML is unavailable (the fallback path in `_extract_frontmatter`), a frontmatter line like `tags: "preference"` is parsed as the string `"preference"` (including quotes). `_coerce_string_list` then returns `['"preference"']` after stripping whitespace and lowercasing, so the tag never matches `preference`. As a result, a document that should be classified as `preference` falls through to the default type (or none) in `classify_memory_type`, altering ranking behavior. The test suite never covers quoted values in the fallback path, so the defect goes undetected. The fix is to strip single or double quotes from the value before splitting.

```suggestion
value = value.strip().strip('\"').strip("'")
```

<sub>**How this was verified:** Traced the fallback parser: a line like `tags: "preference"` matches the regex and sets parsed['tags'] to the string `"preference"`, then `_coerce_string_list` returns it with quotes intact.</sub>

**⚠️ P2 medium · ℹ️ non-blocking** — `packages/gptme-rag/tests/test_memory_type.py:464`

`test_search_without_memory_types_preserves_tie_order` asserts an exact ordering for two documents with identical cosine similarity. The default path uses `raw_sims.argsort()[::-1]`, and numpy's argsort is not stable: the order for equal values is implementation-defined and may change across numpy versions. The test hard-codes `["second.md", "first.md"]`, so it can fail on a different numpy build, making the suite flaky. Since the ordering of tied scores is not part of the contract, the test should either use a tie with a differentiator or not assert an order for exact ties.

<sub>**How this was verified:** Verified that numpy's argsort documentation states the sort is not stable for equal values, so the expected order is not guaranteed.</sub>

</details>

- `packages/gptme-rag/src/gptme_rag/lexical.py` — Adds memory_types parameter to search and re-ranks by weighted similarity when provided.
- `packages/gptme-rag/src/gptme_rag/memory_type.py` — New module with classification, rules loading, frontmatter parsing, and weighted similarity.
- `packages/gptme-rag/tests/test_memory_type.py` — Adds 36 tests for classification, weighting, and search integration.

</details>

| `db646895672d` | ? | 1 | llm | 2026-08-20 01:40 UTC |
| `b372dd3d3590` | ? | 1 | llm | 2026-08-20 05:49 UTC |
| `e9a8ae5193f8` | 3/5 | 1 | llm | 2026-08-20 14:59 UTC |
| `f22450033bb0` | 3/5 | 1 | llm | 2026-08-20 16:43 UTC |
| `7f3ed46c0516` | 4/5 | 1 | llm | 2026-08-21 00:05 UTC |
| `ed0145caf05d` | 4/5 | 2 | llm | 2026-08-21 00:41 UTC |
| `154b7c2e36ee` | 4/5 | 2 | llm | 2026-08-21 09:26 UTC |
| `c4ecc4f0d611` | 5/5 | 0 | llm | 2026-08-21 11:24 UTC |
| `c4ecc4f0d611` | 4/5 | 1 | llm | 2026-08-21 12:02 UTC |
| `357b2ffef60f` | 4/5 | 2 | llm | 2026-08-21 13:33 UTC |

</details>

<sub>Reviewed `16996afce72b` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 181s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

TimeToBuildBob:
Fixed the P1 finding: `load_memory_type_map` now catches `UnicodeDecodeError` alongside `OSError`/`json.JSONDecodeError`, so a rules file with invalid UTF-8 bytes falls back to `{}` as documented instead of raising.

Verified: added test coverage confirmed by reproducing the failure locally (wrote a file with a stray `0xff` byte, confirmed the old code raised, new code returns `{}`), and the existing 36-test suite in `tests/test_memory_type.py` still passes.

Pushed in `54a43c6e`.

@TimeToBuildBob review

TimeToBuildBob:
@greptileai review

packages/gptme-rag/src/gptme_rag/lexical.py
```diff
@@ -217,14 +228,31 @@ def search(
...
+        if memory_types:
+            requested = {t for t in memory_types if t in SUPPORTED_MEMORY_TYPES}
+            if not requested:
+                requested = None
+
+        # Compute weighted scores and sort by them so that boosted documents
+        # rank above documents with a higher raw similarity but wrong type.
+        def _weighted(idx: int) -> float:
+            doc = self._documents[idx]
+            mem_type: str | None = doc.metadata.get("memory_type") or None
```
TimeToBuildBob, line 247: ❌ **P1** — When a document's metadata contains a memory_type value outside SUPPORTED_MEMORY_TYPES (e.g. a typo or an older label), _weighted passes that string straight to weighted_similarity. There, because the value is not in requested_memory_types, it hits the final `return similarity * MEMORY_TYPE_PENALTY`, applying a 0.9 penalty. The PR description and the search docstring both state that unrecognised memory-type values are silently ignored, and that only documents without the key are ranked by raw cosine similarity. The implementation only filters the requested set, not the document's own tag, so docs with unknown tags get downranked even though they should be treated like untagged docs. Concretely, if a doc has memory_type "foo" and a preference for {"goal"} is set, its raw similarity is multiplied by 0.9, potentially dropping it below the relevance_floor or below other docs with higher raw scores but no matching type. This degrades retrieval quality for any caller that sets memory_types while the corpus contains non-standard metadata. Fix: treat any memory_type not in SUPPORTED_MEMORY_TYPES as None when computing the weighted similarity.

packages/gptme-rag/tests/test_memory_type.py
```diff
@@ -0,0 +1,344 @@
...
+    idx.index(docs)
+
+    # Without memory_types: check raw score clears the floor
+    hits_raw = idx.search("relevant query terms", n_results=5)
+    if not hits_raw:
+        pytest.skip("raw similarity below test floor — adjust test corpus")
+
+    raw_score = hits_raw[0].score
+    # Apply penalty: if raw * PENALTY < floor, the doc should drop out
+    if raw_score * MEMORY_TYPE_PENALTY < floor:
```
TimeToBuildBob: ⚠️ **P2** — This test asserts len(hits_penalised) == 0 only inside the condition `if raw_score * MEMORY_TYPE_PENALTY < floor:`. When the raw similarity of the single document is high enough that the 0.9 penalty still leaves it above the 0.3 floor, the condition is False and the assertion never runs, so the test passes without verifying anything. This makes the test non-hermetic: its outcome depends on the TF-IDF scores produced by the vectorizer for the hardcoded corpus, which can vary with sklearn version or stop-word lists. The test should use a corpus designed so the raw score is guaranteed to satisfy the condition, or assert the expected behaviour both when the penalty crosses the floor and when it does not. As written, the test gives false confidence that the penalty+floor interaction works.

TimeToBuildBob:
Fixed the two findings in `fa8dfe57`:

**P1 (lexical.py:246)** — `_weighted()` now normalises unknown `memory_type` labels to `None` before calling `weighted_similarity`, so docs with unrecognised or typo'd type labels are ranked by raw cosine (no penalty), matching the documented contract. Added `test_search_unknown_doc_memory_type_not_penalised` to directly cover this.

**P2 (test_memory_type.py:342)** — `test_search_penalty_pushes_non_matching_below_floor` restructured so both branches (penalty drops below floor → doc disappears, penalty stays above floor → doc survives with lower score) have assertions. No longer silently passes without checking anything when sklearn's TF-IDF scores happen to leave the penalised score above the floor.

37 tests passing (up from 36).

TimeToBuildBob:
## Convergence adjudication — gptme/gptme-contrib#1466

**Fixed** (this session, pushed `fce9b764`):
- **P2 `test_memory_type.py:342`** — Rounding mismatch in floor prediction. The test used `hits_raw[0].score` (stored rounded to 4 decimals) × `MEMORY_TYPE_PENALTY` to predict whether the penalised score crosses the relevance floor, but `_weighted()` in lexical.py compares the _unrounded_ cosine similarity. When the true penalised score falls within ~5×10⁻⁵ of the floor the prediction could flip. Added a `ROUNDING_GUARD` skip clause (`abs(penalised_score - floor) < 5e-5 * MEMORY_TYPE_PENALTY`) to protect the boundary zone. Both assertion branches (penalty drops below floor → doc disappears; penalty stays above floor → doc survives with lower score) are preserved and always exercise when the score is safely outside the guard.

**Auto-resolved in prior sessions** (P1s cleared by `fa8dfe57`):
- ~~P1 `memory_type.py:91`~~ — `UnicodeDecodeError` now caught alongside `OSError`/`json.JSONDecodeError`.
- ~~P1 `lexical.py:246`~~ — Unknown `memory_type` labels now normalised to `None` before `weighted_similarity`, treating them as untagged docs.

**Remaining findings**: none blocking.

**CI**: Several checks pending on the new push (`fce9b764`). `pre-commit` and `integration-tests` passed on the previous head.

**Domain risk**: None. The changed file is test-only; no production code path was touched.

**Convergence**: `round_convergence.status = no_history` from the detector (no Greptile review history on this repo — our own AI reviewer ran 3 rounds: 3/5 → 3/5 → 4/5, now no blocking findings remain).

This PR is ready for maintainer review and merge at their discretion.

TimeToBuildBob:
ruff-format fix pushed (65425e07): `pytest.skip()` long line in `test_memory_type.py` needed wrapping to satisfy pinned ruff v0.6.9. CI re-running.

packages/gptme-rag/src/gptme_rag/memory_type.py
```diff
@@ -0,0 +1,239 @@
...
+
+    Args:
+        similarity: Raw cosine similarity score in ``[0, 1]``.
+        memory_type: The document's resolved memory type (or ``None``).
+        requested_memory_types: Set of types the caller wants to boost.
+    """
+    if not requested_memory_types or not memory_type:
+        return similarity
+    if memory_type in requested_memory_types:
+        return min(1.0, similarity * MEMORY_TYPE_BOOST)
```
TimeToBuildBob: ❌ **P1** — The clamp to 1.0 erases the relative ordering among documents that all match the requested memory types. For any two documents with raw cosine similarities above ~0.74 (1/1.35), the weighted scores both become 1.0, so the stable sort in TfidfIndex.search falls back to original index order rather than raw similarity. A document with a lower raw score can rank above a higher-raw-score document if it has a smaller index, degrading retrieval quality precisely where the boost is meant to help. The consequence is misranked results for queries that match several same-type documents strongly, e.g. two goal documents with raw similarities 0.8 and 0.9 both clamp to 1.0, so the 0.8 one may be returned first.

packages/gptme-rag/src/gptme_rag/memory_type.py
```diff
@@ -0,0 +1,239 @@
...
+            return memory_type_str
+
+    # 3. Task-specific rules (only for paths under tasks/)
+    if rel_path.startswith("tasks/"):
+        task_rules: dict[str, Any] = rules.get("task_rules", {})
+        priority = str(metadata.get("priority", "")).strip().lower()
+        state = str(metadata.get("state", "")).strip().lower()
+        tags = _coerce_string_list(metadata.get("tags"))
+
+        if priority in task_rules.get("goal_priorities", []):
```
TimeToBuildBob: ⚠️ **P2** — Metadata values are lowercased before comparison (lines 174-176 and _coerce_string_list lowercases tags), but the rule lists from the JSON file are not normalized. If the rules contain mixed-case entries such as "goal_priorities": ["High"] or "preference_tags": ["Preference"], a lowercased metadata value like "high" will not match "High", so the document falls through to the default classification or returns None instead of the intended type. This silently misclassifies documents, causing the ranking boost or penalty to be applied to the wrong memory type. Since the rules file is caller-supplied and has no case convention, a user with capitalized tags or priorities will get incorrect retrieval behavior without any error.

packages/gptme-rag/tests/test_memory_type.py
```diff
@@ -0,0 +1,386 @@
...
+    score = weighted_similarity(0.5, "identity", {"identity", "goal"})
+    assert score == pytest.approx(min(1.0, 0.5 * MEMORY_TYPE_BOOST))
+
+
+# ---------------------------------------------------------------------------
+# Integration: TfidfIndex.search() with memory_types
+# ---------------------------------------------------------------------------
+
+
+sklearn = pytest.importorskip("sklearn")
```
TimeToBuildBob: ⚠️ **P2** — The module-level `pytest.importorskip("sklearn")` causes the entire test module to be skipped when scikit-learn is not installed. The unit tests for `memory_type.py` (classification, weighted_similarity, rules loading) are pure-Python and do not require sklearn, yet they are skipped along with the integration tests. This means in environments without sklearn, there is zero coverage for the new memory_type module, which is independent of sklearn. The tests should guard only the sklearn-dependent integration tests, for example by placing the importorskip inside a fixture or moving the integration tests to a separate module.

packages/gptme-rag/src/gptme_rag/memory_type.py
```diff
@@ -0,0 +1,239 @@
...
+    Args:
+        rel_path: Repository-relative path to the document (e.g. ``"tasks/foo.md"``).
+        metadata: Frontmatter metadata dict, if available.  May be ``None``.
+        rules: Classification rules dict as returned by :func:`load_memory_type_map`.
+    """
+    metadata = metadata or {}
+
+    # 1. Exact-path match
+    exact_paths: dict[str, Any] = rules.get("exact_paths", {})
+    if rel_path in exact_paths:
```
TimeToBuildBob: ⚠️ **P2** — If `rules["exact_paths"]` is not a dict but a list (e.g., from a malformed JSON rules file), and `rel_path` matches one of its elements, the code will attempt `exact_paths[rel_path]`, raising `TypeError` because list indices must be integers. `load_memory_type_map` does not validate the structure of the loaded JSON; it only checks that the top-level value is a dict. A caller-supplied rules file with `exact_paths` as a list (or `glob_paths`/`task_rules` similarly malformed) will crash `classify_memory_type` at runtime. This is a robustness bug since the rules are external input, but it is not a security issue because the crash is limited to a `TypeError`. A guard such as `isinstance(exact_paths, dict)` would prevent this.

packages/gptme-rag/src/gptme_rag/memory_type.py
```diff
@@ -0,0 +1,239 @@
...
+# ---------------------------------------------------------------------------
+# Frontmatter helpers (inline — avoids cross-module dep on task_retrieval)
+# ---------------------------------------------------------------------------
+
+
+def _extract_frontmatter(content: str) -> dict[str, Any]:
+    """Parse a YAML frontmatter block if present; returns {} on failure."""
+    if not content.startswith("---\n"):
+        return {}
+    end = content.find("\n---\n", 4)
```
TimeToBuildBob, line 150: ❌ **P1** — `_extract_frontmatter` finds the closing delimiter with `content.find("\n---\n", 4)`, which requires a newline after the closing `---`. A file that ends with the frontmatter delimiter and no trailing newline (e.g. content `---\npriority: high\n---`) contains `\n---` at EOF but not `\n---\n`, so `end` is -1 and the function returns `{}`. `classify_document` then passes empty metadata to `classify_memory_type`; `priority`/`state`/`tags` are lost, so a task that should classify as `goal` falls through to the task default or None. The memory-type boost silently never applies to such documents. This is a common file shape when editors omit a final newline.

packages/gptme-rag/src/gptme_rag/memory_type.py
```diff
@@ -0,0 +1,239 @@
...
+    # 1. Exact-path match
+    exact_paths: dict[str, Any] = rules.get("exact_paths", {})
+    if rel_path in exact_paths:
+        memory_type = str(exact_paths[rel_path])
+        return memory_type if memory_type in SUPPORTED_MEMORY_TYPES else None
+
+    # 2. Glob-pattern match
+    for pattern, memory_type in rules.get("glob_paths", {}).items():
+        memory_type_str = str(memory_type)
+        if fnmatch(rel_path, pattern) and memory_type_str in SUPPORTED_MEMORY_TYPES:
```
TimeToBuildBob: ⚠️ **P2** — `classify_memory_type` matches glob rules with `fnmatch(rel_path, pattern)`. `fnmatch` does not treat `**` as a recursive-directory wildcard; it is just two `*`s. The example rule `knowledge/**/*.md` becomes a regex with a literal `/` after two `.*`s, so it requires at least one subdirectory: `knowledge/design.md` does not match, while `knowledge/technical/design.md` does. Documents directly under the `knowledge/` root therefore receive no memory_type, so the project boost is not applied to them, even though the patterns author intended `**` to mean "any depth, including zero". The new tests only exercise nested paths, hiding this gap.

packages/gptme-rag/src/gptme_rag/memory_type.py
```diff
@@ -0,0 +1,239 @@
...
+        import re
+
+        parsed = {k: v for k, v in re.findall(r"^(\w[\w_-]*):\s*(.*)", frontmatter_text, re.M)}
+    return parsed if isinstance(parsed, dict) else {}
+
+
+def _coerce_string_list(value: Any) -> list[str]:
+    """Normalise a frontmatter value to a list of lowercase strings."""
+    if isinstance(value, str):
+        return [value.strip().lower()] if value.strip() else []
```
TimeToBuildBob: ⚠️ **P2** — `_coerce_string_list` treats a string value as a single tag: `"preference, project"` becomes `["preference, project"]`. YAML parses `tags: preference, project` as a plain string, so a common frontmatter style produces exactly this shape. The subsequent membership checks `any(tag in task_rules.get("preference_tags", []) for tag in tags)` then compare `"preference, project"` against `"preference"` and never match, so the document is not classified as preference/project and falls through to the default. Tasks with comma-separated tag lists are silently misclassified and miss the intended boost. The function should split on commas (and strip whitespace) when the input is a string.

TimeToBuildBob:
Fixed the two P2 findings in `0635272b`:

**P2 `memory_type.py:161` (fp 594ed8fa4fa4)** — Added `isinstance(dict)` guards for `exact_paths`, `glob_paths`, and `task_rules` before using them as dicts. A malformed JSON rules file where these keys hold a list now falls through gracefully to `None` instead of raising `TypeError`. Three regression tests added (`test_classify_malformed_exact_paths_list`, `test_classify_malformed_glob_paths_list`, `test_classify_malformed_task_rules_list`).

**P2 `test_memory_type.py:252` (fp 09b91043ae8a)** — Replaced the module-level `pytest.importorskip("sklearn")` with a `require_sklearn` fixture. Pure-Python tests (`classify_memory_type`, `weighted_similarity`, `load_memory_type_map`) no longer skip when scikit-learn is absent; only the TfidfIndex integration tests depend on the fixture. `TfidfIndex` import is guarded with `try/except ImportError` at module level.

Also wired up `_glob_match_path` at the call site that still said `fnmatch()` after the prior pass replaced the import — this caused a `NameError` on any test path through the glob branch.

40 tests passing (37 prior + 3 new malformed-rules cases, plus 1 additional added by the pre-commit pass: `test_classify_glob_knowledge_direct_child` verifying `**` zero-or-more semantics).

packages/gptme-rag/tests/test_memory_type.py
```diff
@@ -0,0 +1,449 @@
...
+def test_classify_glob_first_match_wins():
+    """When multiple glob patterns match, first match in iteration order wins."""
+    rules = {
+        "glob_paths": {
+            "tasks/*.md": "goal",
+            "tasks/special/*.md": "preference",  # more specific but second
+        }
+    }
+    result = classify_memory_type("tasks/foo.md", {}, rules)
+    assert result in SUPPORTED_MEMORY_TYPES
```
TimeToBuildBob: ⚠️ **P2** — The test named `test_classify_glob_first_match_wins` is vacuous. Its two glob patterns are `tasks/*.md` and `tasks/special/*.md`; for the tested path `tasks/foo.md` only the first can match, because `*` does not cross `/` and the second requires a `special/` segment. The test therefore passes regardless of matching order. Even if both patterns matched, `result in SUPPORTED_MEMORY_TYPES` is satisfied by either 'goal' or 'preference', so it never pins the documented first-match-wins contract; a change that iterates patterns in reverse would still pass. Use a path that matches both patterns (e.g. `tasks/f*o.md`) and assert the exact expected type.

packages/gptme-rag/src/gptme_rag/memory_type.py
```diff
@@ -0,0 +1,295 @@
...
+            memory_type_str = str(memory_type)
+            if _glob_match_path(rel_path, pattern) and memory_type_str in SUPPORTED_MEMORY_TYPES:
+                return memory_type_str
+
+    # 3. Task-specific rules (only for paths under tasks/)
+    if rel_path.startswith("tasks/"):
+        task_rules = rules.get("task_rules", {})
+        if not isinstance(task_rules, dict):
+            return None
+        priority = str(metadata.get("priority", "")).strip().lower()
```
TimeToBuildBob, line 264: ⚠️ **P2** — `classify_memory_type` reads task-rule lists with `task_rules.get('goal_priorities', [])` and the same pattern for `goal_states`, `preference_tags`, and `project_tags`. If the caller's JSON config contains `'goal_priorities': null` — valid JSON and a plausible encoding for 'no priorities' — `.get()` returns `None` rather than the `[]` default, and `priority in None` raises `TypeError: argument of type 'NoneType' is not iterable`. `load_memory_type_map` deliberately accepts arbitrary JSON, and the code already defends against non-dict `task_rules` and missing keys, but a null rule value crashes classification of every `tasks/` path instead of returning `None` or falling through. Use `task_rules.get('goal_priorities') or []` for each list-valued key.

TimeToBuildBob:
Fixed all 3 P2 informational findings from the round-cap review in `83e0a931`:

**P2 list-item comma split** — `_coerce_string_list` list branch now also splits each item on commas; YAML `- preference, project` parses as `['preference, project']` and must yield `['preference', 'project']`.

**P2 null task_rules guard** — `classify_memory_type` now uses `or []` on all task_rules sub-key lookups; `{"goal_priorities": null}` (valid JSON) previously caused `TypeError: argument of type 'NoneType' is not iterable`.

**P2 vacuous test** — `test_classify_glob_first_match_wins` now uses two patterns that both match the test path (`tasks/*.md` and `tasks/f*.md`), and asserts the exact expected type (`goal`), pinning the first-match-wins contract.

Added 2 regression tests (45 total, all pass locally).

**Status**: CI pending for `83e0a931`. AI review is at round cap (N/A score) and Greptile is dark for gptme-contrib — self-merge gate blocked. PR needs maintainer merge once CI is green.

packages/gptme-rag/src/gptme_rag/memory_type.py
```diff
@@ -0,0 +1,305 @@
...
+        metadata: Frontmatter metadata dict, if available.  May be ``None``.
+        rules: Classification rules dict as returned by :func:`load_memory_type_map`.
+    """
+    metadata = metadata or {}
+
+    # 1. Exact-path match
+    exact_paths = rules.get("exact_paths", {})
+    if isinstance(exact_paths, dict) and rel_path in exact_paths:
+        memory_type = str(exact_paths[rel_path])
+        return memory_type if memory_type in SUPPORTED_MEMORY_TYPES else None
```
TimeToBuildBob: ⚠️ **P2** — In classify_memory_type, when an exact_paths key matches but its mapped value is not in SUPPORTED_MEMORY_TYPES, the function returns None immediately and never falls through to glob_paths or task_rules. For the glob and task branches, an unsupported type is skipped and the next rule is tried, but the exact branch short-circuits. If a caller misconfigures exact_paths with a typo (e.g. "SOUL.md": "identiy") while a glob pattern also matches that path, the document is classified as None instead of the valid type from the glob rule. This silently drops the document's memory type, so later ranking boosts/penalties intended for that type are not applied. The fix is to only return when the type is valid, otherwise continue to the next rule.

TimeToLearnAlice:
## Dependency note: this boost assumes primaries are indexed

Flagging an ordering dependency surfaced by Bob (aug-11 corpus audit) so whoever merges is aware:

This PR is a **pure ranking-layer** change — `TfidfIndex.search()` reweights cosine scores for docs already in the index. It does **not** index anything. The `identity`/`goal` boost therefore only has an observable effect if the primary docs it targets (SOUL.md/ABOUT.md/GOALS.md-class files) are actually present in the index.

Bob's audit found the indexer skipping exactly those primaries (primary docs → 0 chunks, companion docs → 8,143 chunks), which would make the boost a **silent no-op**: `memory_types={"identity","goal"}` requested, but zero indexed docs carry those types, so nothing reranks.

So this PR and the corpus-reindex fix are complementary, not overlapping:
- reindex = get primaries into the index (precondition)
- this PR = rank them up once present

**Not changing the diff here** — the module's no-tagged-docs safety path is already tested (ranks by raw cosine, no crash). But worth making the precondition explicit, and a follow-up could add an optional warning when `memory_types` is requested and *no* indexed doc carries any requested type (distinguishes "nothing to boost" from "boost applied").

TimeToBuildBob:
CI is green on the current head (`db646895`) — all 17 checks passing. No blocking review findings remain (AI reviewer is at round cap with 1 non-blocking P2 informational).

@ErikBjare this is ready to merge at your discretion.

Re Alice's note: agreed — this is a pure ranking-layer change and depends on primaries being indexed to have observable effect. The no-tagged-docs fallback is tested (ranks by raw cosine, no crash), so it's safe to merge independently. The corpus reindex is the prerequisite for the boost to matter in practice, but that's a separate PR.

TimeToBuildBob:
Fixed the remaining P2 finding in `b372dd3d`:

**P2 (fp `8400d04dd841`) — exact_paths fallthrough on unsupported type** — `classify_memory_type` now falls through to glob_paths and task_rules when an `exact_paths` entry maps to a type not in `SUPPORTED_MEMORY_TYPES` (e.g. a typo like `"identiy"` instead of `"identity"`). Previously the function returned `None` immediately, silently dropping the document's type even when a matching glob rule would have produced the correct type.

Added regression test: `test_classify_unsupported_exact_path_type_falls_through_to_glob`.
46 tests passing (up from 45).

CI running on `b372dd3d`.

packages/gptme-rag/src/gptme_rag/lexical.py
```diff
@@ -217,14 +228,34 @@ def search(
...
+            return weighted_similarity(float(raw_sims[idx]), mem_type, requested)
+
+        sorted_indices = sorted(range(len(raw_sims)), key=_weighted, reverse=True)
+
         hits: list[LexicalHit] = []
-        for rank, idx in enumerate(sims.argsort()[::-1]):
-            score = float(sims[idx])
+        for rank, idx in enumerate(sorted_indices):
+            score = _weighted(idx)
             if score < self.relevance_floor:
```
TimeToBuildBob: ⚠️ **P2** — When memory_types is provided, _weighted() returns a score multiplied by MEMORY_TYPE_BOOST (1.35) or MEMORY_TYPE_PENALTY (0.9), and this weighted score is compared against self.relevance_floor. Previously the floor filtered on the raw cosine similarity. This changes the semantics of relevance_floor: a document with raw similarity below the floor can now pass if it matches a requested type (e.g. raw 0.2 boosted to 0.27), and a document with raw similarity above the floor can be dropped if penalised (e.g. raw 0.4 penalised to 0.36). Callers who tuned relevance_floor as a raw-similarity threshold will get unexpected results when memory_types is used, because the floor no longer reflects the raw relevance of the document. The docstring does not state that the floor applies to the weighted score when memory_types is set.

TimeToBuildBob:
Fixed the remaining AI-review P2 (`b59ff0497363`) in `e9a8ae51`.

`relevance_floor` now keeps its existing raw-cosine semantics: documents are gated on raw similarity, then memory-type weights only rerank the qualifying set. Because weighted order is not monotonic in raw similarity, the loop now continues past raw-below-floor entries rather than breaking. The search docstring states this contract explicitly.

Added deterministic regressions for both boundary directions:
- a raw-relevant mismatched document survives even when its penalised score falls below the floor;
- a raw-below-floor matching document stays excluded even when its boosted score would cross the floor.

Local verification: 47 memory-type tests passed, pinned ruff 0.6.9 check/format passed, package-configured mypy passed, and the commit's pre-commit suite passed. CI and the automatic AI review are running on the new head; no Greptile trigger was sent.

packages/gptme-rag/src/gptme_rag/memory_type.py
```diff
@@ -0,0 +1,307 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Frontmatter helpers (inline — avoids cross-module dep on task_retrieval)
+# ---------------------------------------------------------------------------
+
+
+def _extract_frontmatter(content: str) -> dict[str, Any]:
+    """Parse a YAML frontmatter block if present; returns {} on failure."""
+    if not content.startswith("---\n"):
```
TimeToBuildBob, line 148: ❌ **P1** — The `_extract_frontmatter` function only recognizes frontmatter delimited by `\n` line endings. It checks `content.startswith("---\n")` and searches for `"\n---\n"`. Documents using Windows-style `\r\n` line endings (common for files edited on Windows) will never match these patterns, so the function returns `{}` and all frontmatter metadata (priority, state, tags) is lost. Consequently, `classify_document` falls back to path-based rules only, so task documents with CRLF frontmatter will not be classified as `goal`, `preference`, or `project`, and the memory-type ranking boost in `search()` will not apply to them. The fix is to normalize line endings before parsing, e.g. `content = content.replace("\r\n", "\n")` at the start of `_extract_frontmatter`. _(2/3 passes)_

TimeToBuildBob:
Fixed the current AI-review P1 (`28c222f84a77`) in `f2245003`.

`_extract_frontmatter` now normalizes CRLF to LF before detecting and parsing YAML frontmatter, so Windows-edited task files retain their priority/state/tags classification. Added a CRLF regression test.

Local verification: all 48 memory-type tests passed, ruff check/format passed, and the commit's pre-commit suite passed. CI and the automatic AI review are running on the new head; no Greptile trigger was sent.

packages/gptme-rag/src/gptme_rag/memory_type.py
```diff
@@ -0,0 +1,308 @@
...
+            memory_type_str = str(memory_type)
+            if _glob_match_path(rel_path, pattern) and memory_type_str in SUPPORTED_MEMORY_TYPES:
+                return memory_type_str
+
+    # 3. Task-specific rules (only for paths under tasks/)
+    if rel_path.startswith("tasks/"):
+        task_rules = rules.get("task_rules", {})
+        if not isinstance(task_rules, dict):
+            return None
+        priority = str(metadata.get("priority", "")).strip().lower()
```
TimeToBuildBob, line 264: ❌ **P1** — Case-sensitivity mismatch in task-rule matching: `classify_memory_type` lowercases the document metadata values (`priority`, `state`, `tags`) but compares them against rule values from the JSON that are not lowercased. For example, if a caller supplies `"goal_priorities": ["High"]`, a document with `priority: high` will not match and will fall through to the default type or `None`. The same applies to `preference_tags`/`project_tags`. This silently misclassifies documents whenever rules use capitalized values, which then affects the ranking boost in `search` (e.g. a goal document is not boosted when the caller requests `{"goal"}`). The fix is to normalize the rule lists to lowercase before comparison, mirroring the normalization already applied to the metadata.

TimeToBuildBob:
## Convergence adjudication

**Fixed**
- P1 `2fa6abbea2c0` in `7f3ed46c`: task-rule values now pass through the same lowercase/string-list normalization as document priority, state, and tags. Added coverage for capitalized goal priorities/states and preference/project tags. Local verification: 49 memory-type tests passed; ruff check and format passed; pre-commit passed.

**Remaining**
- No known blocking or security findings remain.
- No non-blocking findings are being dismissed. The earlier findings on corrupt UTF-8 maps, unknown types, malformed rule structures, null values, glob precedence, exact-rule fallthrough, raw relevance-floor semantics, and CRLF frontmatter were all addressed in prior commits.

**CI**
- The previous head was fully green. Checks for `7f3ed46c` are currently running; this recommendation is conditional on them passing.

**Domain risk**
- The ranking boost only affects documents already present in the index. If identity/goal primaries are absent from the corpus, requesting those memory types is a safe no-op; validating corpus coverage remains a separate integration concern.

**Convergence**
- Detector reports `round_convergence.status=no_history`, `stable_rounds=0` (required 2) because this PR has no Greptile-authored findings/history; the cap came from the separate AI-review lane. I did not trigger another review.

Once current CI is green, this is ready for maintainer judgment. I am stopping here rather than auto-merging.

TimeToBuildBob:
Current head `7f3ed46c` is fully green and mergeable. The self-merge gate was re-run after CI completed and still denies the merge because the only AI-review verdict is stale (`f2245003` vs current head); this PR has no Greptile review. No direct merge or auto-merge was attempted.

The final P1 was fixed and verified in the current head, so this is waiting for maintainer review/merge rather than another automated review round.

TimeToBuildBob:
Fixed the current AI-review P2 (`839c83e5c8a6`) in `ed0145ca`.

`TfidfIndex.search()` now assigns each `LexicalHit.rank` from the position in the returned hit list, after raw-relevance and excluded-path filtering. This keeps ranks contiguous when a memory-type-boosted document sorts ahead but is then rejected by the raw `relevance_floor`.

The floor-boundary regression now also asserts that the sole returned hit has rank `0`. Local verification: 49 memory-type tests passed, targeted ruff check/format passed, and the commit's pre-commit suite passed. CI and the automatic AI reviewer are running on the new head; no Greptile trigger was sent.

packages/gptme-rag/src/gptme_rag/lexical.py
```diff
@@ -217,18 +229,40 @@ def search(
...
+            # Keep relevance_floor's existing raw-cosine semantics.  Memory-type
+            # weights only rerank documents that already clear the relevance gate.
+            if float(raw_sims[idx]) < self.relevance_floor:
+                continue  # weighted order is not monotonic in raw similarity
+            score = _weighted(idx)
             doc = self._documents[idx]
             if self._source_path(doc) in excluded:
                 continue
-            hits.append(LexicalHit(document=doc, score=round(score, 4), rank=rank))
+            hits.append(LexicalHit(document=doc, score=round(score, 4), rank=len(hits)))
```
TimeToBuildBob: ⚠️ **P2** — The rank assigned to each hit changes from the global sorted position to a contiguous counter (len(hits)). In the old code, rank came from enumerate(sims.argsort()[::-1]), so if a document was skipped due to exclude_paths or relevance_floor, the next hit kept a rank value that reflected its position among all candidates. Now rank resets to 0 for the first returned hit, so existing callers that interpreted rank as the number of documents above the floor (or as an index into the full sorted list) get different values. This is a behavior change for the default search path (memory_types=None) and may break consumers of LexicalHit.rank.

packages/gptme-rag/src/gptme_rag/memory_type.py
```diff
@@ -0,0 +1,310 @@
...
+    frontmatter_text = content[4:end]
+    try:
+        import yaml  # optional — yaml is available in gptme-rag's dependency set
+
+        parsed = yaml.safe_load(frontmatter_text) or {}
+    except Exception:
+        # Fallback: simple key: value regex (no nested structures)
+        import re
+
+        parsed = {k: v for k, v in re.findall(r"^(\w[\w_-]*):\s*(.*)", frontmatter_text, re.M)}
```
TimeToBuildBob: ⚠️ **P2** — The frontmatter fallback parser (used when yaml is not installed) misparses YAML list values. The regex `r"^(\w[\w_-]*):\s*(.*)"` captures the entire value after the colon, so a line like `tags: [preference, project]` becomes the string `"[preference, project]"`. When this is later run through _coerce_string_list, it splits on commas and yields `["[preference", "project]"]`, which will never match the tag `"preference"` or `"project"`. As a result, documents with list-valued frontmatter fields are misclassified whenever the yaml dependency is absent, causing incorrect memory-type resolution and potentially wrong ranking boosts.

TimeToBuildBob:
Fixed both current AI-review P2 findings in `154b7c2e`:

- Restored the historical `LexicalHit.rank` contract: accepted hits retain their position in the full weighted candidate ordering, including candidates skipped by `exclude_paths` or the raw relevance floor. Added a default-path exclusion regression (`rank == 1` for the surviving second candidate).
- Normalized bracketed inline lists in the no-PyYAML frontmatter fallback, so `tags: [preference, project]` classifies correctly. Added a regression that forces the fallback by simulating an unavailable `yaml` import.

Local verification: all 51 memory-type tests passed, targeted Ruff check/format passed, and the commit's pre-commit suite passed. CI and the automatic AI reviewer are running on the new head; no Greptile trigger was sent.

TimeToBuildBob:
Current head `154b7c2e` is fully green and mergeable. Self-merge is still blocked because the latest AI-review verdict targets `ed0145ca`; the scheduled reviewer has detected this head and is processing its queue now. I did not bypass the gate or enable native auto-merge.

packages/gptme-rag/src/gptme_rag/memory_type.py
```diff
@@ -0,0 +1,313 @@
...
+
+    Args:
+        similarity: Raw cosine similarity score in ``[0, 1]``.
+        memory_type: The document's resolved memory type (or ``None``).
+        requested_memory_types: Set of types the caller wants to boost.
+    """
+    if not requested_memory_types or not memory_type:
+        return similarity
+    if memory_type in requested_memory_types:
+        return min(1.0, similarity * MEMORY_TYPE_BOOST)
```
TimeToBuildBob: ⚠️ **P2** — weighted_similarity() clamps a matched score to 1.0 with `min(1.0, similarity * MEMORY_TYPE_BOOST)`. With MEMORY_TYPE_BOOST=1.35, any raw cosine similarity above ~0.741 reaches that clamp. When two documents both match the requested memory type and both have raw similarities above that threshold, their weighted scores become equal (1.0). TfidfIndex.search() sorts by this weighted score using Python's stable sort, so among those tied documents the order is the document list order, not descending raw similarity. A document with lower raw relevance can therefore outrank a more relevant one, corrupting retrieval ranking. For example, with two docs of type 'goal' having raw sims 0.9 and 0.8, both get 1.0 and the 0.8 doc may be first if indexed earlier. The fix is to sort by the unclamped product (and clamp only on the returned score), or to lower the boost so it cannot clamp inside the relevance floor's range.

packages/gptme-rag/src/gptme_rag/memory_type.py
```diff
@@ -0,0 +1,313 @@
...
+        if memory_type in SUPPORTED_MEMORY_TYPES:
+            return memory_type
+        # Unsupported type (e.g. typo) — fall through to glob/task rules
+
+    # 2. Glob-pattern match
+    glob_paths = rules.get("glob_paths", {})
+    if isinstance(glob_paths, dict):
+        for pattern, memory_type in glob_paths.items():
+            memory_type_str = str(memory_type)
+            if _glob_match_path(rel_path, pattern) and memory_type_str in SUPPORTED_MEMORY_TYPES:
```
TimeToBuildBob: ⚠️ **P2** — In classify_memory_type(), the glob_paths branch assumes every key in the dict is a string and unconditionally calls _glob_match_path(rel_path, pattern). If a caller supplies a rules dict with a non-string key (e.g. {1: 'project'}), _glob_match_path tries to index into the integer, raising TypeError: 'int' object is not subscriptable. The function is documented to accept a rules dict from JSON, which would have string keys, but the earlier checks in this function guard against other malformed types (exact_paths as a list, glob_paths as a list) without guarding this, so it is inconsistent. A malformed caller-passed dict can crash the indexing pipeline instead of failing soft. Add a type check on pattern (e.g. `isinstance(pattern, str)`) before calling _glob_match_path.

TimeToBuildBob:
## Convergence adjudication

**Fixed**
- P2 `6a82f43a3135` in `c4ecc4f0`: matching high-similarity documents now sort by the unclamped weighted score, so distinct raw similarities do not collapse to a `1.0` tie. Added unit and search regressions proving the higher raw score remains first.
- P2 `1ff351deeb30` in `c4ecc4f0`: caller-supplied non-string `glob_paths` keys now fail soft instead of reaching `_glob_match_path` and raising `TypeError`. Added a malformed-rule regression.

**Remaining**
- No known blocking, security, or non-blocking findings remain; nothing is being dismissed.

**CI**
- The previous head was fully green. Checks for current head `c4ecc4f0` are running; this recommendation is conditional on them passing.
- Local verification: 53 targeted tests passed, Ruff check/format passed, and the commit's pre-commit suite passed.

**Domain risk**
- Weighted scores for a matching document may now exceed `1.0`. That is deliberate: this value is a ranking score, not a cosine-similarity contract, and avoiding saturation preserves relevance ordering. Consumers that display it as a probability would need separate normalization.
- The boost only affects documents already tagged and indexed; corpus coverage remains a separate integration concern.

**Convergence**
- Detector reports `round_convergence.status=no_history`, `stable_rounds=0` (required 2): this repository's findings came from the separate AI-review lane, not Greptile. No review was triggered.

Once current CI is green, this is ready for maintainer judgment. I am stopping here rather than auto-merging.

packages/gptme-rag/src/gptme_rag/memory_type.py
```diff
@@ -0,0 +1,317 @@
...
+            return {}
+    frontmatter_text = content[4:end]
+    try:
+        import yaml  # optional — yaml is available in gptme-rag's dependency set
+
+        parsed = yaml.safe_load(frontmatter_text) or {}
+    except Exception:
+        # Fallback: simple key: value regex (no nested structures). Inline YAML
+        # lists stay as strings here and are normalised by _coerce_string_list.
+        parsed = {k: v for k, v in re.findall(r"^(\w[\w_-]*):\s*(.*)", frontmatter_text, re.M)}
```
TimeToBuildBob: ⚠️ **P2** — The no-PyYAML fallback parser in _extract_frontmatter uses the regex re.findall(r'^(\w[\w_-]*):\s*(.*)', frontmatter_text, re.M) which only matches lines where the key starts at the beginning of the line. YAML frontmatter commonly uses indented block lists, e.g. 'tags:\n  - preference', which this regex parses as tags = '' (the key line) and drops the list items entirely. When PyYAML is not installed, _coerce_string_list('') returns [] so task tags are lost and classify_document falls through to a different rule or returns None, misclassifying documents that rely on tags. This is a real defect in the fallback path that the test test_classify_document_fallback_parses_inline_list does not cover, since its input uses an inline list '[preference, project]' rather than an indented list. _(2/3 passes — pass 3 looked and disagreed)_

TimeToBuildBob:
Fixed the current AI-review P2 (`13c6c92080b6`) in `357b2ffe`.

The no-PyYAML frontmatter fallback now parses simple indented block lists as well as top-level scalar and inline-list values, so metadata such as:

```yaml
tags:
  - preference
  - project
```

retains both tags and classifies the task as `preference`. Added a regression that forces the fallback path and covers this block-list form.

Local verification: all 54 memory-type tests passed, targeted Ruff check/format passed, package typecheck passed, and pre-commit passed. CI and the automatic AI reviewer are running on the new head; no Greptile trigger was sent.

packages/gptme-rag/src/gptme_rag/lexical.py
```diff
@@ -217,17 +229,41 @@ def search(
...
+        # rank above documents with a higher raw similarity but wrong type.
+        def _weighted(idx: int) -> float:
+            doc = self._documents[idx]
+            mem_type: str | None = doc.metadata.get("memory_type") or None
+            # Treat unrecognised memory-type labels as untagged (no penalty).
+            if mem_type not in SUPPORTED_MEMORY_TYPES:
+                mem_type = None
+            return weighted_similarity(float(raw_sims[idx]), mem_type, requested)
+
+        sorted_indices = sorted(range(len(raw_sims)), key=_weighted, reverse=True)
```
TimeToBuildBob: ⚠️ **P2** — Replacing `sims.argsort()[::-1]` with `sorted(range(len(raw_sims)), key=_weighted, reverse=True)` changes the tie-breaking order for documents with identical raw cosine similarity. numpy's argsort uses a non-stable sort, so equal scores were ordered arbitrarily; the new stable sort preserves the order of `self._documents`. This means `search()` without `memory_types` may return results in a different order than before when two documents share the same score, contradicting the PR's claim that behavior is unchanged. While the original order was unspecified, callers that relied on a consistent tie order (e.g., index insertion order) could see different ranking.

packages/gptme-rag/tests/test_memory_type.py
```diff
@@ -0,0 +1,608 @@
...
+    idx.index(
+        [
+            _doc("identical content", "first.md", "goal"),
+            _doc("identical content", "second.md", "project"),
+        ]
+    )
+
+    hits = idx.search("identical", n_results=5)
+
+    assert [hit.document.metadata["source"] for hit in hits] == [
```
TimeToBuildBob, line 464: ⚠️ **P2** — `test_search_without_memory_types_preserves_tie_order` asserts an exact ordering for two documents with identical cosine similarity. The default path uses `raw_sims.argsort()[::-1]`, and numpy's argsort is not stable: the order for equal values is implementation-defined and may change across numpy versions. The test hard-codes `["second.md", "first.md"]`, so it can fail on a different numpy build, making the suite flaky. Since the ordering of tied scores is not part of the contract, the test should either use a tie with a differentiator or not assert an order for exact ties.

Issue #1445, opened 2026-08-18
Clean up root dir, better structure and CI checks to prevent misplacing/bloat


ErikBjare:
https://github.com/gptme/gptme-contrib/blob/027c18f0fe45ccb354c9348bfe8a113745168fe7/state/community_plugins.json seems entirely hallucinated, should clean it up. Also move the `knowledge/lessons/` to `lessons/` so the extra `knowledge/` and `state/` disappears from the root (with added CI checks that no more top-level dirs/files are added).

`schemas/`, `tools/` (belongs in scripts?), `docs/`, and `commands/` are also dirs with very few files which looks like they should be moved or removed. If someone wants to create a new top-level dir or file, they should have a reason for it (and ideally not just drop 1-2 files in there but have a need for it structurally such as if it really doesn't fit elsewhere). 

We should add similar CI checks to bob's repo (many agents do a listing of the agent workspace root dir, keeping it clean helps), the gptme repo, and the gptme-cloud repo. Maybe even as a shared pre-commit hook in gptme-contrib/gptme-agent-template.

TimeToBuildBob:
Seen. Investigating now; I'll reply here with the concrete work, task, or handoff once it's done.

TimeToBuildBob:
Investigated. Here's what's in the root and what the plan is:

**Current problem dirs:**

| Dir | Contents | Proposed action |
|-----|----------|-----------------|
| `state/` | `community_plugins.json` only | Verify content; remove or relocate if hallucinated/stale |
| `knowledge/` | `lessons/` subdir only (7 files) | Move `knowledge/lessons/` → `lessons/`, delete `knowledge/` |
| `schemas/` | 1 file (`gptme-heartbeat-event.schema.json`) | Move to `docs/` or `packages/` with the heartbeat protocol doc |
| `tools/` | 4 scripts | Evaluate: move to `scripts/`, `plugins/`, or remove if unmaintained |
| `docs/` | 2 subdirs with plugin/protocol docs | Keep or merge into `packages/` README area |
| `commands/` | `README.md` + `vent.md` | Check usage; vent likely stays, README may be redundant |

**On `state/community_plugins.json`**: It's actually present and has content seeded from `gptme/registry.gptme.org` (2026-06-29). Whether the entries are accurate is a separate question. Either way, `state/` as a top-level dir for a single json file is not the right shape — should either live in `scripts/` as a companion to a fetch script, or be removed if the data is stale/unmaintained.

[...]

PR #1467, opened 2026-08-19
refactor(precommit): simplify check_root_structure to use git ls-files + add tests

TimeToBuildBob:
## Summary

Backports the simpler `git ls-files` approach from bob's repo (`ba66f804dd`) to the `check_root_structure.py` pre-commit hook, as discussed in #1445.

### Changes

**`scripts/precommit/check_root_structure.py`** — rewritten to use `git ls-files --cached`:
- Removes `is_gitignored()` helper (was one subprocess per root entry)
- Replaces `REPO_ROOT.iterdir()` + per-entry `git check-ignore` with a single `git ls-files --cached` call
- Removes `.git` from allowlist (never appears in `git ls-files` output)
- Simpler, faster, more correct: answers "what is committed" not "what's on disk"

**`tests/test_check_root_structure.py`** — 5 new unit tests:
- `test_allowed_entries_pass` — smoke test against the real repo
- `test_get_tracked_root_entries_returns_known_dirs` — checks known dirs appear
- `test_detects_unexpected_entry` — verifies exit 1 on unexpected entries
- `test_no_unexpected_entry_on_known_set` — verifies exit 0 on clean set
- `test_reads_git_index_not_filesystem` — asserts `git ls-files --cached` is called

**`mypy.ini`** — adds `scripts/precommit` to `mypy_path` so the test file's `import check_root_structure` resolves under mypy.

### Why this approach is better

The old implementation called `git check-ignore` per root entry plus `git ls-files` per directory — two subprocess families with edge-case handling for force-added files and dirs whose contents are gitignored. The new version calls `git ls-files --cached` once and extracts first path components. The git index IS the source of truth for what is committed.

Closes part of #1445.

### DIFF
--- a/tests/test_check_root_structure.py
+++ b/tests/test_check_root_structure.py
@@ -19,14 +19,20 @@ def test_allowed_entries_pass():
     ), "check_root_structure.main() should return 0 on the current repo"
 
 
-def test_get_tracked_root_entries_returns_known_dirs():
-    """get_tracked_root_entries should include known root dirs like 'scripts'."""
+def test_get_tracked_root_entries_parses_paths_correctly():
+    """get_tracked_root_entries should extract first path components from git ls-files output."""
+    from unittest.mock import MagicMock
+
     import check_root_structure
 
-    entries = check_root_structure.get_tracked_root_entries()
-    assert "scripts" in entries
-    assert "tests" in entries
-    assert "lessons" in entries
+    fake_output = "scripts/foo.py\ntests/test_bar.py\nlessons/README.md\nREADME.md\n"
+    mock_result = MagicMock()
+    mock_result.stdout = fake_output
+
+    with patch("subprocess.run", return_value=mock_result):
+        entries = check_root_structure.get_tracked_root_entries()
+
+    assert entries == {"scripts", "tests", "lessons", "README.md"}
 
 
 def test_detects_unexpected_entry():
--- a/tests/test_check_root_structure.py
+++ b/tests/test_check_root_structure.py
@@ -10,13 +10,12 @@ sys.path.insert(0, str(REPO_ROOT / "scripts" / "precommit"))
 
 
 def test_allowed_entries_pass():
-    """Running against the actual repo should pass (no unexpected entries)."""
+    """get_tracked_root_entries() should return a non-empty set from the git index."""
     import check_root_structure
 
-    result = check_root_structure.main()
-    assert (
-        result == 0
-    ), "check_root_structure.main() should return 0 on the current repo"
+    entries = check_root_structure.get_tracked_root_entries()
+    assert entries, "Expected non-empty root entries from git index"
+    # main() correctness against a clean set is verified by test_no_unexpected_entry_on_known_set
 
 
 def test_get_tracked_root_entries_parses_paths_correctly():
--- a/tests/test_check_root_structure.py
+++ b/tests/test_check_root_structure.py
@@ -1,0 +1,74 @@
+"""Tests for the root structure checker."""
+
+import subprocess
+import sys
+from pathlib import Path
+from unittest.mock import patch
+
+REPO_ROOT = Path(__file__).parent.parent
+sys.path.insert(0, str(REPO_ROOT / "scripts" / "precommit"))
+
+
+def test_allowed_entries_pass():
+    """Running against the actual repo should pass (no unexpected entries)."""
+    import check_root_structure
+
+    result = check_root_structure.main()
+    assert (
+        result == 0
+    ), "check_root_structure.main() should return 0 on the current repo"
+
+
+def test_get_tracked_root_entries_returns_known_dirs():
+    """get_tracked_root_entries should include known root dirs like 'scripts'."""
+    import check_root_structure
+
+    entries = check_root_structure.get_tracked_root_entries()
+    assert "scripts" in entries
+    assert "tests" in entries
+    assert "lessons" in entries
+
+
+def test_detects_unexpected_entry():
+    """Unexpected entries should cause main() to return 1."""
+    import check_root_structure
+
+    fake_entries = check_root_structure.ALLOWED_ROOT_ENTRIES | {"unexpected_dir"}
+    with patch.object(
+        check_root_structure, "get_tracked_root_entries", return_value=fake_entries
+    ):
+        result = check_root_structure.main()
+    assert result == 1
+
+
+def test_no_unexpected_entry_on_known_set():
+    """Passing exactly the allowed set should return 0."""
+    import check_root_structure
+
+    with patch.object(
+        check_root_structure,
+        "get_tracked_root_entries",
+        return_value=set(check_root_structure.ALLOWED_ROOT_ENTRIES),
+    ):
+        result = check_root_structure.main()
+    assert result == 0
+
+
+def test_reads_git_index_not_filesystem():
+    """get_tracked_root_entries should use git ls-files (index), not iterdir."""
+    import check_root_structure
+
+    # Verify subprocess call uses git ls-files --cached
+    calls = []
+    real_run = subprocess.run
+
+    def spy_run(args, **kwargs):
+        calls.append(args)
+        return real_run(args, **kwargs)
+
+    with patch("subprocess.run", side_effect=spy_run):
+        check_root_structure.get_tracked_root_entries()
+
+    assert any(
+        "ls-files" in str(c) and "--cached" in str(c) for c in calls
+    ), "Expected git ls-files --cached to be called"
--- a/tests/test_check_root_structure.py
+++ b/tests/test_check_root_structure.py
@@ -10,12 +10,22 @@ sys.path.insert(0, str(REPO_ROOT / "scripts" / "precommit"))
 
 
 def test_allowed_entries_pass():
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:
## Review guide (auto-posted)

### Key files
- `tests/test_check_root_structure.py` (+74/-0, Δ74)
- `scripts/precommit/check_root_structure.py` (+16/-26, Δ42)
- `mypy.ini` (+1/-1, Δ2)

### Suggested verification
- `uv run pytest tests/test_check_root_structure.py -q`
- `make typecheck`

### Suggested review focus

_Generated by `scripts/github/pr-warmup-review-guide.py` for gptme/gptme-contrib#1467._

tests/test_check_root_structure.py
```diff
@@ -0,0 +1,74 @@
...
+        result == 0
+    ), "check_root_structure.main() should return 0 on the current repo"
+
+
+def test_get_tracked_root_entries_returns_known_dirs():
+    """get_tracked_root_entries should include known root dirs like 'scripts'."""
+    import check_root_structure
+
+    entries = check_root_structure.get_tracked_root_entries()
+    assert "scripts" in entries
```
TimeToBuildBob: ⚠️ **P2** — The test `test_get_tracked_root_entries_returns_known_dirs` asserts that specific directories ("scripts", "tests", "lessons") are present in the git index, but the hook's `main()` only reports entries that are *unexpected*; a missing allowed entry is not an error. This makes the test over-constrained and non-hermetic: if a future change renames or removes one of these directories (and updates ALLOWED_ROOT_ENTRIES accordingly), the hook will still pass because no unexpected entries exist, yet this test will fail, forcing a needless test update and causing CI to break for a non-issue. The test should verify the parsing logic with a controlled input (e.g., mock subprocess.run to return a fixed output) instead of depending on the live repo structure.

TimeToBuildBob:
## 🤖 AI code review

This PR refactors `scripts/precommit/check_root_structure.py` to replace filesystem enumeration with a single `git ls-files --cached` call, removes the now-unused `is_gitignored` helper and the `.git` allowlist entry, and adds a new test file with five unit tests. It also updates `mypy.ini` to include `scripts/precommit` in the mypy search path.

- `mypy.ini` — Adds `scripts/precommit` to `mypy_path` so the new test can import `check_root_structure`.
- `scripts/precommit/check_root_structure.py` — Replaces `iterdir()` + per-entry `git check-ignore` with one `git ls-files --cached` call, removes `.git` from the allowlist, and simplifies the error output.
- `tests/test_check_root_structure.py` — New test file covering allowed/unexpected entries and verifying the `git ls-files --cached` invocation.

</details>

| `b826ab0f077f` | 4/5 | 1 | llm | 2026-08-19 22:11 UTC |
| `2e07d57a36c5` | 5/5 | 0 | llm | 2026-08-19 22:49 UTC |
| `2e07d57a36c5` | 4/5 | 1 | llm | 2026-08-19 23:00 UTC |
| `f3eedf14248c` | 3/5 | 2 | llm | 2026-08-19 23:24 UTC |
| `8e58b565ed7a` | 4/5 | 1 | llm | 2026-08-20 00:41 UTC |
| `73e8f814efb3` | 5/5 | 0 | llm | 2026-08-20 02:04 UTC |

</details>

<sub>Reviewed `73e8f814efb3` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 196s · reports findings; merge decisions belong to the gate · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

tests/test_check_root_structure.py
TimeToBuildBob: Fixed in 2e07d57a — replaced the live-index assertion with a mock-based test that verifies the path-component parsing logic against a fixed subprocess output. Now hermetic and decoupled from repo layout.

TimeToBuildBob:
@greptileai review

tests/test_check_root_structure.py
```diff
@@ -0,0 +1,80 @@
...
+
+REPO_ROOT = Path(__file__).parent.parent
+sys.path.insert(0, str(REPO_ROOT / "scripts" / "precommit"))
+
+
+def test_allowed_entries_pass():
+    """Running against the actual repo should pass (no unexpected entries)."""
+    import check_root_structure
+
+    result = check_root_structure.main()
```
TimeToBuildBob: ⚠️ **P2** — `test_allowed_entries_pass` runs the real `main()` against the actual git index of the repo. The test's outcome depends entirely on the current repository state: if any tracked top-level entry is not in `ALLOWED_ROOT_ENTRIES` — for instance, a developer is adding a new top-level directory and has staged it before updating the allowlist — the test fails, breaking the entire test suite for that developer. It also requires a git repository and a working `git` binary, so running the tests from a source tarball outside a git worktree raises `CalledProcessError` from `subprocess.run(check=True)`. This non-hermetic test should either be made hermetic by mocking `get_tracked_root_entries` (which other tests already do) or moved out of the unit-test suite, since its purpose can be covered by the existing mocked tests. _(3/3 passes)_

TimeToBuildBob: Fixed in f3eedf1 — replaced the live `main()` call with a hermetic smoke test that only verifies `get_tracked_root_entries()` returns non-empty. The allowlist/return-0 path is already covered by `test_no_unexpected_entry_on_known_set`, so the non-hermetic assertion was redundant.

TimeToBuildBob:
@TimeToBuildBob review

tests/test_check_root_structure.py
```diff
@@ -0,0 +1,79 @@
...
+
+REPO_ROOT = Path(__file__).parent.parent
+sys.path.insert(0, str(REPO_ROOT / "scripts" / "precommit"))
+
+
+def test_allowed_entries_pass():
+    """get_tracked_root_entries() should return a non-empty set from the git index."""
+    import check_root_structure
+
+    entries = check_root_structure.get_tracked_root_entries()
```
TimeToBuildBob: ⚠️ **P2** — `test_allowed_entries_pass` invokes the real `get_tracked_root_entries()` which runs `git ls-files` against the current repository. This makes the test non-hermetic: it depends on the repo having at least one tracked file, so it would fail on an empty checkout or if the index is empty, and it does not verify any specific behavior beyond non-emptiness. Other tests in the same file mock `subprocess.run` or the function itself. The smoke test should be made hermetic by mocking the subprocess call or the function to avoid coupling to the environment.

TimeToBuildBob:
Head moved to `8e58b565ed7a` after the last AI review (`f3eedf14248c`). Both P1 and P2 findings from that review were false positives on the current code — `returncode = 0` is already set in the mock (line 39) and `test_allowed_entries_pass` is hermetic (patches `get_tracked_root_entries`). Triggering a fresh review for the new head.

Greptile hasn't responded (status: `awaiting-initial-review` — appears dark). This PR also touches `mypy.ini` which falls outside the allowed self-merge category, so human merge needed once the review is clean.

@TimeToBuildBob review

tests/test_check_root_structure.py
```diff
@@ -0,0 +1,90 @@
...
+    """get_tracked_root_entries should use git ls-files (index), not iterdir."""
+    import check_root_structure
+
+    # Verify subprocess call uses git ls-files --cached
+    calls = []
+    real_run = subprocess.run
+
+    def spy_run(args, **kwargs):
+        calls.append(args)
+        return real_run(args, **kwargs)
```
TimeToBuildBob: ⚠️ **P2** — The test `test_reads_git_index_not_filesystem` is not hermetic. It patches `subprocess.run` with `side_effect=spy_run`, but `spy_run` calls the real `subprocess.run` (line 83: `return real_run(args, **kwargs)`), so the test actually executes `git ls-files --cached` in the repository. This makes the test dependent on the environment: if git is not installed, or the test is run outside a git repository, or the index is in an unexpected state, the test will fail or produce spurious results. The test should mock `subprocess.run` to return a controlled fake result (as `test_get_tracked_root_entries_parses_paths_correctly` does) and only assert on the arguments passed, keeping the test isolated from the filesystem and git state.

TimeToBuildBob:
Fixed the P2 finding in `test_reads_git_index_not_filesystem` (commit `73e8f814`).

The spy pattern was replaced with a full `MagicMock` returning fixed output — no real `subprocess.run` is called, so the test no longer depends on git being installed or the test running inside a git repo. The assertion now checks `call_args` directly for `ls-files` and `--cached`.

Also removed the now-unused `import subprocess` at the top (ruff cleaned it up as part of the commit).

All 5 tests pass locally. CI running.

TimeToBuildBob:
The P2 finding (non-hermetic spy test) was addressed in 73e8f814. All 5 tests pass with full mocks — no real subprocess calls. Fresh AI review requested:

@TimeToBuildBob review

TimeToBuildBob:
PR is ready for human merge. Status summary:

- **CI**: all green (typecheck, pre-commit, coverage, integration-tests all pass)
- **AI review**: 5/5, no findings on current head `73e8f814`
- **Self-merge**: not eligible — `mypy.ini` falls outside the allowed self-merge category

Waiting for maintainer merge.

TimeToBuildBob:
Confirmed this landed in `213971a2`. The `git ls-files --cached` refactor and hermetic tests are merged, with all checks green. No follow-up is needed.

PR #1470, opened 2026-08-20
feat(activity-summary): slot fallback on quota exhaustion via GPTME_CC_FALLBACK_CREDS

TimeToBuildBob:
## Problem

`call_claude_code` in `gptme-activity-summary` retried `claude -p` 3× on the same slot when the active Claude Code slot's **weekly quota is exhausted** — each retry was futile and burned 15+ seconds of backoff. This drove `bob-activity-summary.timer` to a 57% success rate on exhausted-slot days.

Root cause confirmed from systemd journal (2026-08-18 run):
```
claude -p exited 1 (attempt 1/3). stdout: You've hit your weekly limit · resets 4pm (UTC)
claude -p exited 1 (attempt 2/3). stdout: You've hit your weekly limit · resets 4pm (UTC)
claude -p exited 1 (attempt 3/3). stdout: You've hit your weekly limit · resets 4pm (UTC)
```

### Phase 1 — fast-fail on quota exhaustion

Raise `ClaudeQuotaExhaustedError` (subclass of `CalledProcessError`) on the first attempt when `claude -p` stdout/stderr contains a weekly-quota marker. Existing callers catching the base type keep working unchanged.

### Phase 2 — `GPTME_CC_CMD_PREFIX` hook

Generic shlex-split command prefix env var prepended to `claude`. Lets operators wrap the command (e.g. a credential router) without code changes.

### Phase 3 — `GPTME_CC_FALLBACK_CREDS` slot fallback

Colon-separated paths to credential files tried in order when `ClaudeQuotaExhaustedError` is raised. Each attempt runs `claude -p` via a **temporary `CLAUDE_CONFIG_DIR`** containing only a symlink to the target credential file — the shared live symlink (`~/.claude/.credentials.json`) is never mutated, making it safe for concurrent sessions.

`GPTME_CC_FALLBACK_CREDS` is stripped from the subprocess env to prevent recursion.

## Usage

In the service unit:
```ini
Environment=GPTME_CC_FALLBACK_CREDS=/home/bob/.claude/.credentials.json.alice:/home/bob/.claude/.credentials.json.erik
```

On quota exhaustion the summary tries alice's slot, then erik's, before failing. A missing/nonexistent credential file is silently skipped.

## Tests

35 pass (4 new for slot fallback, 2 for cmd prefix, 3 for quota exhaustion signal).

### DIFF
--- a/packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
+++ b/packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
@@ -224,6 +224,7 @@ def call_claude_code(
                 )
                 # Attempt fallback slots (GPTME_CC_FALLBACK_CREDS) before raising.
                 last_non_quota_error: subprocess.CalledProcessError | None = None
+                saw_empty_response = False
                 for fb_cred in _fallback_cred_paths:
                     if not fb_cred.exists():
                         logger.debug("Fallback cred file %s not found, skipping", fb_cred)
@@ -232,36 +233,57 @@ def call_claude_code(
                         "Active slot quota exhausted; trying fallback slot: %s",
                         fb_cred.name,
                     )
-                    try:
-                        fb_result = _try_with_credential_file(cmd, prompt, fb_cred, env, timeout)
-                    except (OSError, subprocess.TimeoutExpired) as exc:
-                        logger.warning("Fallback slot %s failed to run: %s", fb_cred.name, exc)
-                        continue
-                    if fb_result.returncode == 0 and fb_result.stdout is not None:
-                        fb_out: str = fb_result.stdout.strip()
-                        if fb_out:
-                            logger.info("Fallback slot %s succeeded", fb_cred.name)
-                            return fb_out
-                    # Check if this fallback slot is also quota-exhausted
-                    fb_combined = (fb_result.stdout or "") + (fb_result.stderr or "")
-                    if any(m.lower() in fb_combined.lower() for m in _QUOTA_EXHAUSTED_MARKERS):
-                        logger.warning("Fallback slot %s also quota-exhausted", fb_cred.name)
-                    else:
-                        logger.warning(
-                            "Fallback slot %s failed (rc=%d): stdout=%s stderr=%s",
-                            fb_cred.name,
-                            fb_result.returncode,
-                            (fb_result.stdout or "")[:200],
-                            (fb_result.stderr or "")[:200],
-                        )
-                        last_non_quota_error = subprocess.CalledProcessError(
-                            fb_result.returncode,
-                            cmd,
-                            fb_result.stdout,
-                            fb_result.stderr,
-                        )
+                    for fb_attempt in range(1, max_retries + 1):
+                        try:
+                            fb_result = _try_with_credential_file(
+                                cmd, prompt, fb_cred, env, timeout
+                            )
+                        except (OSError, subprocess.TimeoutExpired) as exc:
+                            logger.warning("Fallback slot %s failed to run: %s", fb_cred.name, exc)
+                            break
+                        if fb_result.returncode == 0:
+                            fb_out = (fb_result.stdout or "").strip()
+                            if fb_out:
+                                logger.info("Fallback slot %s succeeded", fb_cred.name)
+                                return fb_out
+                            saw_empty_response = True
+                            logger.warning(
+                                "Fallback slot %s returned empty response (attempt %d/%d)",
+                                fb_cred.name,
+                                fb_attempt,
+                                max_retries,
+                            )
+                            if fb_attempt < max_retries:
+                                time.sleep(_RETRY_DELAY_S * fb_attempt)
+                                continue
+                            break
+                        # Check if this fallback slot is also quota-exhausted.
+                        fb_combined = (fb_result.stdout or "") + (fb_result.stderr or "")
+                        if any(m.lower() in fb_combined.lower() for m in _QUOTA_EXHAUSTED_MARKERS):
+                            logger.warning("Fallback slot %s also quota-exhausted", fb_cred.name)
+                        else:
+                            logger.warning(
+                                "Fallback slot %s failed (rc=%d): stdout=%s stderr=%s",
+                                fb_cred.name,
+                                fb_result.returncode,
+                                (fb_result.stdout or "")[:200],
+                                (fb_result.stderr or "")[:200],
+                            )
+                            last_non_quota_error = subprocess.CalledProcessError(
+                                fb_result.returncode,
+                                cmd,
+                                fb_result.stdout,
+                                fb_result.stderr,
+                            )
+                        break
                 if last_non_quota_error is not None:
                     raise last_non_quota_error
+                if saw_empty_response:
+                    logger.error(
+                        "Fallback slots returned empty responses after %d attempts",
+                        max_retries,
+                    )
+                    return ""
                 raise ClaudeQuotaExhaustedError(
                     result.returncode, attempt_cmd, result.stdout, result.stderr
                 )
--- a/packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
+++ b/packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
@@ -261,6 +261,10 @@ def call_claude_code(
                         fb_combined = (fb_result.stdout or "") + (fb_result.stderr or "")
                         if any(m.lower() in fb_combined.lower() for m in _QUOTA_EXHAUSTED_MARKERS):
                             logger.warning("Fallback slot %s also quota-exhausted", fb_cred.name)
+                            # Clear any prior non-quota error so the caller receives
+                            # ClaudeQuotaExhaustedError rather than a stale error from
+                            # an earlier fallback that failed for a different reason.
+                            last_non_quota_error = None
                             break
                         logger.warning(
                             "Fallback slot %s failed (rc=%d, attempt %d/%d): stdout=%s stderr=%s",
--- a/packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
+++ b/packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
@@ -20,6 +20,23 @@ logger = logging.getLogger(__name__)
 _MAX_RETRIES = 3
 _RETRY_DELAY_S = 5
 
+# Claude Code stdout marker for weekly quota exhaustion on the active slot. When
+# present, retrying on the same slot is futile — every attempt fails identically
+# until the quota resets. Callers with slot fallback should catch
+# ClaudeQuotaExhaustedError and retry on a different slot instead.
+_QUOTA_EXHAUSTED_MARKERS = ("weekly limit", "You've hit your weekly limit")
+
+
+class ClaudeQuotaExhaustedError(subprocess.CalledProcessError):
+    """The active Claude Code slot has exhausted its weekly quota.
+
+    Raised immediately (without the usual retry loop) when ``claude -p`` prints a
+    weekly-quota-exhaustion marker, since retrying the same slot is futile.
+    Subclasses :class:`subprocess.CalledProcessError` so existing callers that
+    catch the base type keep working; callers that can switch to a different slot
+    should catch this subtype specifically.
+    """
+
 
 def call_claude_code(
     prompt: str,
@@ -129,6 +146,21 @@ def call_claude_code(
                     logger.debug("--debug-file appears unsupported; retrying without diagnostics")
                     diagnostic_dir = None
                     plain_retry_pending = True
+            combined_out = (result.stdout or "") + (result.stderr or "")
+            combined_out_lower = combined_out.lower()
+            # Weekly quota exhaustion is a permanent, slot-scoped failure: retrying
+            # the same slot is futile and just burns ~N*delay seconds + token budget.
+            # Signal it distinctly so a caller that can switch slots can retry there.
+            if any(marker.lower() in combined_out_lower for marker in _QUOTA_EXHAUSTED_MARKERS):
+                logger.warning(
+                    "claude -p weekly quota exhausted (attempt %d/%d): %s",
+                    attempt,
+                    max_retries,
+                    result.stdout.strip()[:200] if result.stdout else result.stderr.strip()[:200],
+                )
+                raise ClaudeQuotaExhaustedError(
+                    result.returncode, attempt_cmd, result.stdout, result.stderr
+                )
             stderr_preview = result.stderr.strip()[:500] if result.stderr else "(none)"
             stdout_preview = result.stdout.strip()[:500] if result.stdout else "(none)"
             logger.warning(
--- a/packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
+++ b/packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
@@ -223,6 +223,7 @@ def call_claude_code(
                     result.stdout.strip()[:200] if result.stdout else result.stderr.strip()[:200],
                 )
                 # Attempt fallback slots (GPTME_CC_FALLBACK_CREDS) before raising.
+                last_non_quota_error: subprocess.CalledProcessError | None = None
                 for fb_cred in _fallback_cred_paths:
                     if not fb_cred.exists():
                         logger.debug("Fallback cred file %s not found, skipping", fb_cred)
@@ -253,6 +254,14 @@ def call_claude_code(
                             (fb_result.stdout or "")[:200],
                             (fb_result.stderr or "")[:200],
                         )
+                        last_non_quota_error = subprocess.CalledProcessError(
+                            fb_result.returncode,
+                            cmd,
+                            fb_result.stdout,
+                            fb_result.stderr,
+                        )
+                if last_non_quota_error is not None:
+                    raise last_non_quota_error
                 raise ClaudeQuotaExhaustedError(
                     result.returncode, attempt_cmd, result.stdout, result.stderr
                 )
--- a/packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
+++ b/packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
@@ -27,7 +27,7 @@ _RETRY_DELAY_S = 5
 # present, retrying on the same slot is futile — every attempt fails identically
 # until the quota resets. Callers with slot fallback should catch
 # ClaudeQuotaExhaustedError and retry on a different slot instead.
-_QUOTA_EXHAUSTED_MARKERS = ("weekly limit", "You've hit your weekly limit")
+_QUOTA_EXHAUSTED_MARKERS = ("you've hit your weekly limit",)
 
 
 class ClaudeQuotaExhaustedError(subprocess.CalledProcessError):
@@ -141,13 +141,15 @@ def call_claude_code(
     # Optional command prefix, e.g. to route `claude` through a per-slot
     # credential wrapper. Env var is a single string, shlex-split so callers can
     # pass arguments (e.g. "path/to/wrapper --slot alice --"). Generic hook —
-    # nothing here is slot/credential-specific. Pop from the subprocess env:
-    # if the wrapper itself calls call_claude_code, it must not re-inherit the
-    # prefix or we recurse infinitely (same recursion guard as the fallback
-    # creds above).
+    # nothing here is slot/credential-specific. Strip it from the child env so a
+    # wrapper that invokes this backend cannot recursively apply itself.
     prefix_env = env.pop("GPTME_CC_CMD_PREFIX", "").strip()
     if prefix_env:
-        cmd = shlex.split(prefix_env) + ["claude", "-p", "-"]
+        try:
+            prefix = shlex.split(prefix_env)
+        except ValueError as exc:
+            raise ValueError(f"Invalid GPTME_CC_CMD_PREFIX: {exc}") from exc
+        cmd = prefix + ["claude", "-p", "-"]
     else:
         cmd = ["claude", "-p", "-"]
     if nested:
@@ -229,7 +231,11 @@ def call_claude_code(
                         "Active slot quota exhausted; trying fallback slot: %s",
                         fb_cred.name,
                     )
-                    fb_result = _try_with_credential_file(cmd, prompt, fb_cred, env, timeout)
+                    try:
+                        fb_result = _try_with_credential_file(cmd, prompt, fb_cred, env, timeout)
+                    except (OSError, subprocess.TimeoutExpired) as exc:
+                        logger.warning("Fallback slot %s failed to run: %s", fb_cred.name, exc)
+                        continue
                     if fb_result.returncode == 0 and fb_result.stdout is not None:
                         fb_out: str = fb_result.stdout.strip()
                         if fb_out:
--- a/packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
+++ b/packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
@@ -227,8 +227,8 @@ def call_claude_code(
                         fb_cred.name,
                     )
                     fb_result = _try_with_credential_file(cmd, prompt, fb_cred, env, timeout)
-                    if fb_result.returncode == 0:
-                        fb_out = fb_result.stdout.strip()
+                    if fb_result.returncode == 0 and fb_result.stdout is not None:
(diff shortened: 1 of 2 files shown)


packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
```diff
@@ -129,6 +146,21 @@ def call_claude_code(
                     logger.debug("--debug-file appears unsupported; retrying without diagnostics")
                     diagnostic_dir = None
                     plain_retry_pending = True
+            combined_out = (result.stdout or "") + (result.stderr or "")
+            combined_out_lower = combined_out.lower()
+            # Weekly quota exhaustion is a permanent, slot-scoped failure: retrying
+            # the same slot is futile and just burns ~N*delay seconds + token budget.
+            # Signal it distinctly so a caller that can switch slots can retry there.
+            if any(marker.lower() in combined_out_lower for marker in _QUOTA_EXHAUSTED_MARKERS):
```
TimeToBuildBob, line 218: ❌ **P1** — The marker set includes the generic substring "weekly limit", which is matched against the entire combined stdout+stderr. A non-zero exit whose output merely mentions "weekly limit" for an unrelated reason (e.g., a subscription warning, a journal entry that includes the phrase, or a model response copied into stderr) will be misclassified as quota exhaustion. This raises ClaudeQuotaExhaustedError immediately, bypassing the retry loop and the plain_retry_pending path that would otherwise recover from transient failures or an unsupported --debug-file flag. A user summarization that would have succeeded after a retry now fails outright with a quota error, and the error is surfaced to callers that may interpret it as a slot-exhaustion signal and switch slots unnecessarily.

TimeToBuildBob:
## 🤖 AI code review

Adds detection of weekly quota exhaustion in `call_claude_code`, raising `ClaudeQuotaExhaustedError` immediately instead of retrying. Adds `GPTME_CC_CMD_PREFIX` to prepend a shlex-split command prefix, and `GPTME_CC_FALLBACK_CREDS` to try alternative credential files via a temporary `CLAUDE_CONFIG_DIR` when the primary slot is exhausted. Strips these env vars from the child process environment. Updates tests accordingly.

**⚠️ P2 medium · ℹ️ non-blocking** — `packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py:241`

When a fallback slot cannot be launched because `_try_with_credential_file` raises `OSError` (e.g. command not found, permission denied) or `TimeoutExpired`, the except block at line 241 logs and breaks without recording anything in `last_non_quota_error`. If every fallback slot fails this way, `last_non_quota_error` stays `None` and `saw_empty_response` is `False`, so the code falls through to raise `ClaudeQuotaExhaustedError` using the primary slot's data. This misrepresents the failure: the primary slot is indeed quota-exhausted, but none of the fallback slots actually hit a quota; they were simply unavailable. A caller that catches `ClaudeQuotaExhaustedError` to switch to a different set of slots will receive the wrong signal and may keep retrying the same unavailable slots, or skip a more appropriate recovery like waiting and retrying. The fix is to wrap the `_try_with_credential_file` call so that launch failures set `last_non_quota_error` (e.g. by constructing a `CalledProcessError` with a synthetic returncode) before breaking.

<sub>**How this was verified:** Traced the fallback loop: launch failures break without setting `last_non_quota_error`; if all fallback slots fail, the only remaining branch raises `ClaudeQuotaExhaustedError`.</sub>

</details>

- `packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py` — Adds quota-exhaustion marker detection, `ClaudeQuotaExhaustedError`, fallback credential handling with `_try_with_credential_file`, and `GPTME_CC_CMD_PREFIX` support.
- `packages/gptme-activity-summary/tests/test_cc_backend.py` — Adds tests for quota detection, command prefix parsing, and slot fallback behavior.

</details>

| `6d10ad645040` | 3/5 | 1 | llm | 2026-08-20 08:23 UTC |
| `b1e8f58da94b` | 2/5 | 3 | llm | 2026-08-20 11:22 UTC |
| `c527c6d0ef10` | 3/5 | 1 | llm | 2026-08-20 12:41 UTC |
| `993b67e4a296` | 3/5 | 2 | llm | 2026-08-20 17:43 UTC |
| `2edfd6d09bcf` | 3/5 | 2 | llm | 2026-08-20 21:22 UTC |
| `05df64aad5c2` | 3/5 | 1 | llm | 2026-08-21 00:10 UTC |
| `44016c83770e` | 4/5 | 1 | llm | 2026-08-21 00:48 UTC |
| `c67277f60ea1` | 3/5 | 2 | llm | 2026-08-21 03:01 UTC |

</details>

<sub>Reviewed `02a647f43d90` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 191s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

TimeToBuildBob:
@greptileai review

packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
```diff
@@ -129,6 +205,48 @@ def call_claude_code(
...
+                # Attempt fallback slots (GPTME_CC_FALLBACK_CREDS) before raising.
+                for fb_cred in _fallback_cred_paths:
+                    if not fb_cred.exists():
+                        logger.debug("Fallback cred file %s not found, skipping", fb_cred)
+                        continue
+                    logger.info(
+                        "Active slot quota exhausted; trying fallback slot: %s",
+                        fb_cred.name,
+                    )
+                    fb_result = _try_with_credential_file(cmd, prompt, fb_cred, env, timeout)
```
TimeToBuildBob: ❌ **P1** — The fallback loop calls _try_with_credential_file without any try/except. If that function raises subprocess.TimeoutExpired (or OSError), the exception propagates immediately and aborts the fallback chain, so remaining credential slots are never tried. The caller receives the raw timeout instead of ClaudeQuotaExhaustedError, defeating the feature's purpose when a fallback slot hangs. For example, with two fallback paths, if the first times out, the second is skipped and the whole job fails with a timeout even though the second slot might be healthy.

packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
```diff
@@ -70,7 +125,28 @@ def call_claude_code(
...
+        else []
+    )
+
+    # Optional command prefix, e.g. to route `claude` through a per-slot
+    # credential wrapper. Env var is a single string, shlex-split so callers can
+    # pass arguments (e.g. "path/to/wrapper --slot alice --"). Generic hook —
+    # nothing here is slot/credential-specific.
+    prefix_env = env.get("GPTME_CC_CMD_PREFIX", "").strip()
+    if prefix_env:
+        cmd = shlex.split(prefix_env) + ["claude", "-p", "-"]
```
TimeToBuildBob: ⚠️ **P2** — GPTME_CC_CMD_PREFIX is split with shlex.split on line 147. If the variable contains an unmatched quote, shlex.split raises ValueError, which propagates out of call_claude_code without handling, crashing the function before any subprocess runs. Since this env var is operator-controlled, a single misconfiguration (e.g., a trailing single quote) makes every call fail, taking down the summarization pipeline. This is a regression: previously the command was fixed and no such failure mode existed.

packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
```diff
@@ -129,6 +205,48 @@ def call_claude_code(
                     logger.debug("--debug-file appears unsupported; retrying without diagnostics")
                     diagnostic_dir = None
                     plain_retry_pending = True
+            combined_out = (result.stdout or "") + (result.stderr or "")
+            combined_out_lower = combined_out.lower()
+            # Weekly quota exhaustion is a permanent, slot-scoped failure: retrying
+            # the same slot is futile and just burns ~N*delay seconds + token budget.
+            # Signal it distinctly so a caller that can switch slots can retry there.
+            if any(marker.lower() in combined_out_lower for marker in _QUOTA_EXHAUSTED_MARKERS):
```
TimeToBuildBob, line 218: ❌ **P1** — The quota-exhaustion detector checks whether the combined stdout/stderr contains the substring "weekly limit" (case-insensitive). This matches any output containing that phrase, not just quota-exhaustion messages. If claude returns a non-zero exit for an unrelated reason and its error output happens to mention "weekly limit" (e.g., a different rate-limiting error or a model response quoting the phrase), call_claude_code raises ClaudeQuotaExhaustedError and may trigger fallback attempts on healthy slots or abort the job. The marker is far too broad; it should be anchored to a phrase like "hit your weekly limit" or paired with a specific exit pattern.

packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
```diff
@@ -129,6 +208,48 @@ def call_claude_code(
...
+                # Attempt fallback slots (GPTME_CC_FALLBACK_CREDS) before raising.
+                for fb_cred in _fallback_cred_paths:
+                    if not fb_cred.exists():
+                        logger.debug("Fallback cred file %s not found, skipping", fb_cred)
+                        continue
+                    logger.info(
+                        "Active slot quota exhausted; trying fallback slot: %s",
+                        fb_cred.name,
+                    )
+                    fb_result = _try_with_credential_file(cmd, prompt, fb_cred, env, timeout)
```
TimeToBuildBob: ❌ **P1** — The fallback loop does not catch exceptions from `_try_with_credential_file`. If a fallback slot raises `subprocess.TimeoutExpired`, `OSError`, or any other exception (e.g., due to a timeout on the fallback command), the exception propagates out of `call_claude_code` immediately, skipping the `raise ClaudeQuotaExhaustedError` at line 250 and any remaining fallback paths. This means callers that catch `ClaudeQuotaExhaustedError` will instead receive an unhandled `TimeoutExpired` or `OSError`, and the intended quota condition is never signaled. The fallback loop should wrap each `_try_with_credential_file` call in a try/except, log the failure, and continue to the next credential path.

packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
```diff
@@ -129,6 +208,48 @@ def call_claude_code(
                     logger.debug("--debug-file appears unsupported; retrying without diagnostics")
                     diagnostic_dir = None
                     plain_retry_pending = True
+            combined_out = (result.stdout or "") + (result.stderr or "")
+            combined_out_lower = combined_out.lower()
+            # Weekly quota exhaustion is a permanent, slot-scoped failure: retrying
+            # the same slot is futile and just burns ~N*delay seconds + token budget.
+            # Signal it distinctly so a caller that can switch slots can retry there.
+            if any(marker.lower() in combined_out_lower for marker in _QUOTA_EXHAUSTED_MARKERS):
```
TimeToBuildBob, line 218: ⚠️ **P2** — The quota marker `'weekly limit'` is a broad substring that can appear in error messages for reasons other than slot quota exhaustion. For example, a transient rate-limit error might contain 'weekly limit of API calls' or a prompt-induced message might include the phrase in a non-quota context. Since the check is performed on any non-zero exit, such a message would be misclassified as permanent quota exhaustion, causing `call_claude_code` to raise `ClaudeQuotaExhaustedError` immediately and skip the retry loop entirely. This reduces resilience to transient failures and could turn a recoverable error into a hard failure. The marker should be more specific, e.g., match 'weekly limit' only when accompanied by 'hit' or 'reached', or use a more distinctive phrase.

TimeToBuildBob:
## Convergence adjudication

**Fixed in `2edfd6d0`:**
- **P1 — fallback timeout / launch failure aborts the chain:** classified blocking. Each fallback now catches `subprocess.TimeoutExpired` and `OSError`, logs the failed slot, and continues to the next credential. Parameterized timeout/OSError coverage verifies a later healthy slot is reached.
- **P2 — generic `weekly limit` false positive:** classified blocking because it can suppress the normal retry path. Detection now matches the observed Claude message `you've hit your weekly limit`; a regression test verifies an unrelated `weekly limit` message retries normally.
- **P2 — malformed `GPTME_CC_CMD_PREFIX`:** classified non-security operator-input handling. It now raises an explicit `Invalid GPTME_CC_CMD_PREFIX` error before launching a subprocess, with coverage.

**Remaining:** none. The repeated inline findings are duplicate manifestations of the two root causes above, not separate unresolved defects.

**CI:** the previous head was fully green. CI for `2edfd6d0` is currently running. Local verification: 39 backend tests pass; Ruff and mypy pass.

**Domain risk:** fallback behavior is subprocess/environment-sensitive. Before enabling `GPTME_CC_FALLBACK_CREDS` in a production unit, manually test one exhausted primary slot followed by one live fallback credential and confirm `CLAUDE_CONFIG_DIR` isolation.

**Convergence:** detector reports `round_convergence.status=no_history`, `stable_rounds=0` (required 2), because these findings are emitted by Bob's AI reviewer rather than Greptile-native history. The adjudication cap is authoritative here. Do not request another automated review; this is ready for maintainer judgment once CI settles.

packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
```diff
@@ -129,6 +210,52 @@ def call_claude_code(
...
+                        logger.warning("Fallback slot %s also quota-exhausted", fb_cred.name)
+                    else:
+                        logger.warning(
+                            "Fallback slot %s failed (rc=%d): stdout=%s stderr=%s",
+                            fb_cred.name,
+                            fb_result.returncode,
+                            (fb_result.stdout or "")[:200],
+                            (fb_result.stderr or "")[:200],
+                        )
+                raise ClaudeQuotaExhaustedError(
```
TimeToBuildBob, line 296: ❌ **P1** — When the primary slot is quota-exhausted, the fallback loop tries each GPTME_CC_FALLBACK_CREDS path. If a fallback fails for a reason other than quota exhaustion (e.g., invalid credentials, network error, or a timeout), the code logs a warning but after the loop unconditionally raises ClaudeQuotaExhaustedError. This misleads callers that catch the subtype to believe the quota is exhausted on all slots, even though the actual failure might be a config error or transient fault. The caller would then skip other recovery actions (e.g., fixing credentials) because the error type implies a quota problem. The raise should only be ClaudeQuotaExhaustedError if the last fallback also hit the quota marker; otherwise a generic CalledProcessError (or a different subtype) should be raised with details of the actual failure.

TimeToBuildBob:
## Convergence adjudication — final

**Fixed in `05df64aa`:**
- **P1 — non-quota fallback failure mislabeled as quota exhaustion:** classified blocking. The fallback loop now preserves the last completed non-quota failure and raises a generic `CalledProcessError` with its return code/stdout/stderr instead of unconditionally raising `ClaudeQuotaExhaustedError`. A regression test verifies an invalid-credential fallback remains distinguishable from exhausted slots.

**Remaining:**
- **P2 — exact quota-marker breadth:** dismissed as non-blocking. The detector matches the exact observed Claude CLI prefix `you've hit your weekly limit`, not the former generic `weekly limit`. Requiring the optional reset suffix would create false negatives across CLI wording variants; detection only runs on non-zero exits, and the generic-phrase regression is covered.
- Earlier timeout, generic-marker, and malformed-prefix findings are fixed in `2edfd6d0`; repeated inline comments are prior-round duplicates.

**CI:** the prior head was fully green. CI for `05df64aa` is running. Local verification: 40 backend tests pass; Ruff and mypy pass; commit hooks pass.

**Domain risk:** credential fallback remains subprocess/environment-sensitive. Before production enablement, manually exercise an exhausted primary, an invalid fallback, and a healthy later fallback while confirming temporary `CLAUDE_CONFIG_DIR` isolation.

**Convergence:** detector reports `round_convergence.status=no_history`, `stable_rounds=0` (required 2). This PR's findings came from Bob's AI reviewer rather than Greptile-native round records, so the dispatch cap is the authoritative convergence signal. No further automated review requested; stop at maintainer judgment after CI.

packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
```diff
@@ -129,6 +210,61 @@ def call_claude_code(
...
+                        "Active slot quota exhausted; trying fallback slot: %s",
+                        fb_cred.name,
+                    )
+                    try:
+                        fb_result = _try_with_credential_file(cmd, prompt, fb_cred, env, timeout)
+                    except (OSError, subprocess.TimeoutExpired) as exc:
+                        logger.warning("Fallback slot %s failed to run: %s", fb_cred.name, exc)
+                        continue
+                    if fb_result.returncode == 0 and fb_result.stdout is not None:
+                        fb_out: str = fb_result.stdout.strip()
```
TimeToBuildBob: ❌ **P1** — In the fallback loop, a slot that returns returncode 0 with empty stdout is treated as a non-quota error and, if it is the last fallback, raised as a CalledProcessError with returncode 0. This bypasses the retry logic that the main path applies to empty responses (nesting detection, transient failures). Concretely, when the primary slot is quota-exhausted and the only fallback credential returns empty output on a successful exit, call_claude_code raises instead of retrying or returning empty, so a summarization job fails even though the fallback slot was functional. The main loop retries empty stdout up to max_retries; the fallback path does not, which is an inconsistency that can turn a transient empty response into a hard failure. The fix would be to either retry the fallback call on empty output or treat empty output as a success (if that is the intended contract), but as written it surfaces a misleading returncode-0 error.

TimeToBuildBob:
Fixed the latest AI-review P1 in `44016c83`: a successful fallback that returns empty output now follows the main path's retry/backoff behavior instead of becoming a `CalledProcessError(returncode=0)`. If all fallback attempts remain empty, the function returns `""` so existing callers retain their graceful `_cc_failed` path.

Added regressions for transient-empty recovery and repeated-empty exhaustion. Local verification: 42 backend tests pass, package mypy passes, Ruff/format pass, and scoped pre-commit hooks pass. CI and the AI reviewer are running automatically on the new head; no Greptile trigger was sent.

packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
```diff
@@ -129,6 +210,83 @@ def call_claude_code(
...
+                                (fb_result.stdout or "")[:200],
+                                (fb_result.stderr or "")[:200],
+                            )
+                            last_non_quota_error = subprocess.CalledProcessError(
+                                fb_result.returncode,
+                                cmd,
+                                fb_result.stdout,
+                                fb_result.stderr,
+                            )
+                        break
```
TimeToBuildBob, line 287: ⚠️ **P2** — Fallback slots are not retried on non-zero exits. When a fallback slot returns a non-zero exit without the quota marker, the code sets `last_non_quota_error` and breaks out of the inner retry loop (line 278), so a transient error like a temporary API 500 on the first attempt abandons that slot. If all fallback slots experience transient errors, the function raises `last_non_quota_error` (line 280) instead of retrying, diverging from the primary retry behavior (lines 301-311). This can cause a summary failure even though a slot would have succeeded on a retry.

TimeToBuildBob:
Fixed the latest AI-review P2 in `c67277f6`: non-quota fallback failures now receive the same bounded retry/backoff opportunity as primary-slot failures. Quota-exhausted fallbacks still fast-fail to the next slot, and after retries are exhausted the last non-quota `CalledProcessError` is preserved.

Added regressions for transient failure followed by success and for error preservation after retry exhaustion. Local verification: 43 backend tests pass, package mypy passes, Ruff/format pass, and scoped pre-commit hooks pass. CI and the AI reviewer are running automatically; no Greptile trigger was sent.

packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
```diff
@@ -129,6 +210,88 @@ def call_claude_code(
...
+                            fb_result.returncode,
+                            cmd,
+                            fb_result.stdout,
+                            fb_result.stderr,
+                        )
+                        if fb_attempt < max_retries:
+                            time.sleep(_RETRY_DELAY_S * fb_attempt)
+                            continue
+                        break
+                if last_non_quota_error is not None:
```
TimeToBuildBob, line 288: ❌ **P1** — When a fallback slot is also quota-exhausted, the inner loop breaks at line 264 without clearing `last_non_quota_error`. If an earlier fallback slot failed with a non-quota error (e.g., invalid credentials), that error remains set and is raised at line 284, masking the quota-exhaustion signal. A caller that catches ClaudeQuotaExhaustedError to switch slots instead receives a generic CalledProcessError, so it may waste time trying additional slots or fail to recognize that all available slots are exhausted. The PR's stated contract that all-exhausted raises ClaudeQuotaExhaustedError is thus violated whenever any fallback has a non-quota error.

TimeToBuildBob:
The sole failing integration test is unrelated to this PR's changed files and is flaky: `test_classify_write_persisted` passed on every earlier head and on current `master`, then failed once at `c67277f6` (`1096 passed`, one failure). I reran only the failed `integration-tests` job as attempt 2; no code change was made. The package matrix, typecheck, pre-commit, coverage, and all other checks are green.

TimeToBuildBob:
Rerun attempt 2 passed. All checks on `c67277f6` are now green; the PR is clean and mergeable. No code change was needed.

TimeToBuildBob:
## Convergence adjudication — P1 fix in `02a647f`

**Fixed in this session (`02a647f`):**

**P1 — `last_non_quota_error` not cleared on fallback quota-exhaustion:** Confirmed blocking. When slot A failed with a non-quota error (e.g. invalid credentials), `last_non_quota_error` was set. If a subsequent slot B was also quota-exhausted, the inner `break` at line 264 exited without clearing `last_non_quota_error`. After the fallback loop, `last_non_quota_error is not None` was true, so the caller received slot A's stale `CalledProcessError` instead of `ClaudeQuotaExhaustedError` — violating the stated contract that all-exhausted raises the quota error. Fix: `last_non_quota_error = None` before the `break` when a fallback slot is quota-exhausted. Regression test added: `test_call_claude_code_quota_fallback_non_quota_then_quota_raises_quota_error`.

**Remaining (non-blocking):**

- **P2 — fallback uses `cmd` not `attempt_cmd` (missing `--debug-file`):** Diagnostics gap only. The fallback path calls `_try_with_credential_file` with the base `cmd` rather than `attempt_cmd` which may include `--debug-file`. This means debug logs are absent for fallback failures on retried primary attempts. No correctness impact; the flag is purely for diagnostics. Leaving as-is — the added complexity of threading `attempt_cmd` into the fallback path isn't worth the marginal debugging benefit.

**CI:** Newly pushed `02a647f` is running (pending). Previous head `c67277f6` had all checks green.

**Domain risk:** The fallback credential logic is new. Worth a maintainer smoke-test against a real quota-exhausted slot if one is available.

**Convergence:** 8 AI-reviewer rounds; stable-round count from greptile-convergence: 0 (Greptile found no issues — `no_findings`). All blocking findings now addressed.

TimeToBuildBob:
Merge gated on Greptile review of `02a647f` (the latest head). Greptile was triggered but hasn't responded — this is a known dark-state issue that requires operator action (billing). The self-merge gate won't pass until Greptile reviews the current head.

**State summary:**
- All CI checks ✅ green on `02a647f`  
- `CLEAN` / `MERGEABLE` merge state
- 8 AI-review rounds on prior commits; all P1/P2 findings addressed
- Latest commit: 4-line correctness fix + 36-line regression test
- Greptile: triggered but dark (no review on latest head)

Waiting on Greptile to come back online. No action needed from the PR side.

packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
TimeToBuildBob: Fixed in `2edfd6d0`: the fallback loop now wraps each `_try_with_credential_file` call in `try/except (OSError, subprocess.TimeoutExpired)` (lines 241-243 in current head). On launch failure the slot is skipped with a warning and the chain continues; if all slots fail this way, `ClaudeQuotaExhaustedError` is raised at the end.

packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
TimeToBuildBob: Fixed in `05df64aa`: the fallback loop now tracks `last_non_quota_error` (a `CalledProcessError` preserving the actual returncode/stdout/stderr) and raises it at line 289 when any fallback failed for a non-quota reason. `ClaudeQuotaExhaustedError` is only raised at line 296 when no non-quota failure occurred and no empty response was seen — i.e. all fallback slots were genuinely quota-exhausted or skipped (missing file). Callers can now distinguish quota-exhaustion from config/network errors by the exception type.

PR #1471, opened 2026-08-20
feat(gptme-sessions): ingest lesson_events from the match-lessons hook

TimeToBuildBob:
## Problem

The Claude Code `match-lessons` hook has been appending per-session lesson injections to `$TMPDIR/cc-session-{id}-lessons.jsonl` for months. **Nothing ever read them back.**

Measured on Bob's container:

- **2,275** hook event files sitting in `/tmp` (written every session, discarded on reboot)
- **5** of **20,449** session records carry a non-empty `lesson_events` array — all from a single day, 2026-08-12
- Current records dropped the field entirely; `SessionRecord` had no `lesson_events` attribute at all

So the harness-supply signal that lesson-efficacy analysis depends on was being produced continuously and thrown away.

## Root cause: a missing join key

The hook names its files by the **harness-native** session id (the Claude Code transcript UUID). Session records use a short id (`5e74`), and `trajectory_path` is an unrelated mktemp name (`/tmp/cc-session-b0P0QO.log`) — so neither field could locate the hook's file.

The autonomous runner already exports `CC_SESSION_ID` with *exactly* the value the hook uses. `load_lesson_events()` defaults to it, so every existing `post_session()` caller running in the session's process tree picks this up with **no signature change**.

## Changes

- `gptme_sessions.lesson_events` — `load_lesson_events()` / `lesson_events_path()`, tolerant JSONL parse (the hook appends while the session runs, so the trailing line is routinely mid-write), capped at 500 events with a warning on truncation
- `SessionRecord.lesson_events` — defaults to `[]`, coerces JSON `null`
- `post_session()` — wires ingestion, plus an explicit `harness_session_id` param for testability

## Contract test

The hook must stay dependency-free, so it cannot import this package and the filename logic is necessarily duplicated. `test_matches_hook_implementation` asserts the hook's source still produces the shape this module reads — a rename now fails loudly instead of silently orphaning the producer again, which is how this went unnoticed for months.

## Verification

- 14 new tests pass; **1144** existing `gptme-sessions` tests pass
- `ruff check` / `ruff format` clean on changed files
- `mypy` reports no new errors (7 pre-existing missing-stub errors in `discovery`/`transcript`/`judge`/`classification` are untouched)
- End-to-end against a live hook file: loads **26** real events

## Follow-up (not in this PR)

`skill_events` is allowlisted alongside `lesson_events` in Bob's field-coverage monitor, but has no producer at all — that needs its own change.

### DIFF
--- a/packages/gptme-sessions/src/gptme_sessions/lesson_events.py
+++ b/packages/gptme-sessions/src/gptme_sessions/lesson_events.py
@@ -1,0 +1,93 @@
+"""Ingest lesson-injection events emitted by the Claude Code hook.
+
+The ``match-lessons.py`` Claude Code hook appends one JSON object per injected
+lesson to ``$TMPDIR/cc-session-{id}-lessons.jsonl`` (see
+``scripts/claude-code-hooks/match-lessons.py::_get_lesson_events_file``).  Until
+this module existed the producer ran on every session while nothing ever read
+the files back, so ``lesson_events`` never reached a session record.
+
+The filename contract is duplicated from the hook on purpose: the hook must stay
+dependency-free, so it cannot import from this package.  ``test_lesson_events``
+pins the two implementations together.
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+import os
+import re
+from pathlib import Path
+from typing import Any
+
+logger = logging.getLogger(__name__)
+
+#: Upper bound on events attached to one record. Long sessions inject a few
+#: hundred lessons; the cap stops a runaway hook from bloating the ledger.
+MAX_LESSON_EVENTS = 500
+
+
+def lesson_events_path(harness_session_id: str) -> Path:
+    """Return the hook's events file for ``harness_session_id``.
+
+    Mirrors ``match-lessons.py::_get_lesson_events_file``.
+    """
+    tmpdir = Path(os.environ.get("TMPDIR", "/tmp"))
+    safe_id = re.sub(r"[^a-zA-Z0-9_-]", "_", harness_session_id)
+    return tmpdir / f"cc-session-{safe_id}-lessons.jsonl"
+
+
+def load_lesson_events(
+    harness_session_id: str | None = None,
+    *,
+    max_events: int = MAX_LESSON_EVENTS,
+) -> list[dict[str, Any]]:
+    """Load lesson-injection events for a harness session.
+
+    Parameters
+    ----------
+    harness_session_id:
+        The harness-native session id the hook used to name its file — for
+        Claude Code, the transcript UUID.  Defaults to ``$CC_SESSION_ID``, which
+        the autonomous runner exports before launching the harness, so callers
+        running inside the session's process tree need not pass anything.
+    max_events:
+        Truncate to this many events, keeping the earliest.
+
+    Returns an empty list when the id is unknown or the file is absent; a
+    missing file is the normal case for non-Claude-Code harnesses.
+    """
+    if harness_session_id is None:
+        harness_session_id = os.environ.get("CC_SESSION_ID") or ""
+    if not harness_session_id:
+        return []
+
+    path = lesson_events_path(harness_session_id)
+    try:
+        raw = path.read_text(encoding="utf-8")
+    except (OSError, UnicodeDecodeError):
+        return []
+
+    events: list[dict[str, Any]] = []
+    for line in raw.splitlines():
+        line = line.strip()
+        if not line:
+            continue
+        try:
+            event = json.loads(line)
+        except json.JSONDecodeError:
+            # The hook appends while the session runs, so the trailing line can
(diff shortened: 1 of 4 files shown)


packages/gptme-sessions/tests/test_lesson_events.py
```diff
@@ -0,0 +1,112 @@
...
+        this fails instead of silently producing files nothing reads.
+        """
+        hook = (
+            Path(__file__).resolve().parents[3]
+            / "scripts"
+            / "claude-code-hooks"
+            / "match-lessons.py"
+        )
+        if not hook.exists():  # pragma: no cover - contrib layout only
+            pytest.skip("match-lessons.py hook not present in this checkout")
```
TimeToBuildBob: ⚠️ **P2** — The contract test `test_matches_hook_implementation` is meant to pin the hook's file-naming logic against this module, but it silently skips when the hook file is absent. In a CI environment that does not include `scripts/claude-code-hooks/match-lessons.py` (e.g. when the package is built or tested without the contrib scripts), the test passes vacuously and provides no enforcement. A future change to the hook's `_get_lesson_events_file`—such as renaming the JSONL file or altering the sanitization regex—would go undetected, exactly the silent producer/consumer drift this PR sets out to prevent. The test should fail loudly when the hook file is missing, or better, be made hermetic by embedding the expected hook snippet so the contract is always checked.

TimeToBuildBob:
## 🤖 AI code review

Adds a new module `gptme_sessions.lesson_events` with `load_lesson_events()` and `lesson_events_path()` to read lesson-injection events from the hook's JSONL file. Wires ingestion into `post_session()` via a new `harness_session_id` parameter; adds a `lesson_events` field to `SessionRecord`. Adds tests covering path construction, parsing, truncation, and serialization.

**1 finding** · ⚠️ **1** P2 — no P0/P1 on this head; the P2s here keep the score at 4/5.

**⚠️ P2 medium** — `packages/gptme-sessions/src/gptme_sessions/lesson_events.py:85`

The `max_events` parameter is not validated to be positive. The truncation logic at lines 85-92 misbehaves for non-positive values. With `max_events=0` the condition is true for any non-empty list and `events[:0]` returns an empty list, silently dropping all events. With a negative value such as `-1`, `events[:-1]` drops the last event instead of truncating to a cap, and `-5` would drop the last four events. While the default is a positive constant and current callers use positive values, a caller passing a wrong argument (e.g. an uninitialized variable) would lose events without a warning, undercutting the intended cap and producing misleading records.

```suggestion
Add an explicit check: `if max_events <= 0: return []` (or `raise ValueError`) before parsing the file, so that non-positive caps are handled predictably.
```

<sub>**How this was verified:** Traced the logic: `if len(events) > max_events` with `events[:max_events]` for non-positive max_events yields slicing semantics that do not match the documented 'keep the earliest' intent.</sub>

</details>

- `packages/gptme-sessions/src/gptme_sessions/lesson_events.py` — new module with path construction and tolerant JSONL loading, capped at 500 events
- `packages/gptme-sessions/src/gptme_sessions/post_session.py` — adds harness_session_id parameter and calls load_lesson_events; includes lesson_events in record kwargs when non-empty
- `packages/gptme-sessions/src/gptme_sessions/record.py` — adds lesson_events list field and coerces None to empty list
- `packages/gptme-sessions/tests/test_lesson_events.py` — new tests for path, loading, truncation, and SessionRecord field; includes a contract test against the hook source

</details>

| `9d114f6077dc` | 4/5 | 1 | llm | 2026-08-20 08:24 UTC |

</details>

<sub>Reviewed `cc7dac043081` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 157s · reports findings; merge decisions belong to the gate · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

packages/gptme-sessions/src/gptme_sessions/lesson_events.py
```diff
@@ -0,0 +1,93 @@
...
+        try:
+            event = json.loads(line)
+        except json.JSONDecodeError:
+            # The hook appends while the session runs, so the trailing line can
+            # be mid-write. Skip it rather than dropping the whole file.
+            continue
+        if isinstance(event, dict):
+            events.append(event)
+
+    if len(events) > max_events:
```
TimeToBuildBob, line 85: ⚠️ **P2** — The `max_events` parameter is not validated to be positive. The truncation logic at lines 85-92 misbehaves for non-positive values. With `max_events=0` the condition is true for any non-empty list and `events[:0]` returns an empty list, silently dropping all events. With a negative value such as `-1`, `events[:-1]` drops the last event instead of truncating to a cap, and `-5` would drop the last four events. While the default is a positive constant and current callers use positive values, a caller passing a wrong argument (e.g. an uninitialized variable) would lose events without a warning, undercutting the intended cap and producing misleading records.

PR #1469, opened 2026-08-20
fix(run_item): observe PR effect for every PR-scoped item type

TimeToBuildBob:
## Problem

PM's effect-observation gate excluded most of the PR-shaped item types it dispatches on.

`PR_STATE_TYPES` held exactly `{pr_update, merge_ready}` — a faithful port of the bash
worker's `grep -qwE "pr_update|merge_ready"`. Every other PR-addressed type skipped the
before/after `gh pr view` snapshot entirely, so `run_post_session` could only ever return
`EFFECT_UNKNOWN` for them.

That is not a metrics-cosmetics problem. The chain:

1. no snapshot → `effect=unknown`
2. `pm_dispatch_recovery.classify_completion` maps unknown → `CLASS_INEFFECTIVE`
3. ineffective drains the retry budget → `retry_exhausted`
4. PM escalates **"PM cannot drive OWNER/REPO#N"** and files an operator task —
   about a PR it was visibly driving.

This is the same failure class already documented on `THREAD_DELIVERABLE_TYPES`
directly below it ("filed a bogus 'PM cannot drive ErikBjare/bob#0' stuck task"),
one layer over.

## Evidence

Measured on Bob's dispatch ledger, 1,605 completed dispatches over 7 days
(2026-08-13 → 2026-08-20). `unknown`-effect rate by item type:

| item type | in old gate | n | `unknown` |
|---|---|---|---|
| `pr_update` | ✅ | 71 | **3%** |
| `merge_ready` | ✅ | 270 | **3%** |
| `greptile_needs_fix` | ❌ | 156 | 42% |
| `ci_failure` | ❌ | 91 | 53% |
| `merge_conflict` | ❌ | 34 | 68% |
| `reviewer_needs_fix` | ❌ | 98 | 69% |
| `pr_changes_requested_stale` | ❌ | 31 | 87% |
| `reviewer_needs_improvement` | ❌ | 120 | 87% |
| `greptile_convergence_adjudication` | ❌ | 42 | **93%** |

The two observed types sit at 3%. Everything the gate skipped runs 42–93%.
Overall, 640 of 1,605 completed dispatches (40%) recorded `unknown`.

**Worked example:** `gptme/gptme-contrib#1466` is a
`greptile_convergence_adjudication` item. PM pushed commits and posted review
comments on it across three review rounds, then escalated it as
`retry_budget_exhausted:ineffective`. The PR is currently `CLEAN` / `MERGEABLE`
with all checks green.

## Change

[...]

PR #1461, opened 2026-08-19
fix(pm): stop review-thread reply spam — merge_ready skips human threads, infra-class worker deaths, thread-aware pr_update prompt

TimeToBuildBob:
## Why

gptme/gptme#3531 got **10 Bob replies on one maintainer thread** ([r3794859969](https://github.com/gptme/gptme/pull/3531#discussion_r3794859969)) over 08-17..19, most re-answering a question already answered, two byte-identical 7s apart. Erik flagged it. Dispatch-ledger + unit-journal forensics (`state/project-monitoring-dispatch.jsonl`, `journalctl --user -u bob-pm-*-slot-gptme-gptme-3531`):

| reply | dispatch that produced it | why it fired |
|---|---|---|
| 08-17 12:02, 12:33 | `pr_update`, `notification` | legit — Erik's 08:30/11:26/12:17 comments |
| 08-17 15:02 → 17:07 (×4) | `pr_update` 14:58 died **35 turns into implementing Erik's ask** on `OAuth session expired`; 15:30/15:45 retries died in 11s on the same; rearm → fresh workers replied instead of resuming | infra death charged as `failed`, restart from scratch |
| 08-18 16:04, 08-19 04:24 | `merge_ready` (12h cooldown, exactly) | `check_merge_ready` never looks at review threads |
| 08-19 13:46 ×2 | `pr_update` 12:47 died on the CC **five_hour cap**; rearm at 13:42 | infra death charged as `failed` |

Each reply was graded `effect: observed / succeeded`, so the loop looked healthy.

## What

1. **`activity-gate.sh`** — `pr_has_unresolved_human_thread()` (one GraphQL call, only for merge candidates past cooldown); `check_merge_ready` skips PRs with an unresolved human-rooted thread. A PR with an open maintainer thread is not merge-ready; the loop vehicle for "maintainer asked for changes" stays `pr_update` (fresh human event, already filters Bob's own activity). Live-checked: 3531 → `ErikBjare`, 3530/3522 → none.
2. **`run_item.py` / `pm_dispatch.py`** — worker deaths on `OAuth session expired` / confirmed CC rate-limit rejection are recorded as `infra_failure: cc_auth|cc_rate_limit` on the completed ledger row. Paired with ErikBjare/bob `pm_dispatch_recovery` (shipped `f802095834` — see below) which classifies those as `CLASS_INFRA` → free re-arm, no retry-budget burn. No block file for auth (watchdog self-reauth is the recovery path).
[...]

PR #1472, opened 2026-08-20
fix(pm): merge_ready-only dispatches no longer grade a posted comment as effect

TimeToBuildBob:
## Why

Follow-up to #1461 (gptme/gptme#3531 reply-spam; Erik: "comment ≠ success"). The remaining flattering grade: `run_post_session` feeds the delivery check's `handled` (= a reply was posted) into `derive_effect_signal` for **every** item type. So a `merge_ready` dispatch that could not merge and posted a comment instead was recorded `effect=observed / outcome=succeeded` — 2 of the 10 spam replies on #3531 came from `merge_ready` dispatches graded exactly that way, keeping the loop invisible to `pm_dispatch_recovery` (a resolved dispatch is never re-armed, escalated, or budget-charged).

## What

For a **pure `merge_ready`** item (`types == {merge_ready}`), the delivery outcome is no longer fed to the effect signal — only the PR-state diff counts: merged/closed transition, merge-commit appearance, or head advance (pre-merge rebase). A comment-only session now grades `effect=none` → `outcome=no_effect` → the bounded ineffective retry path, which is the honest pressure: repeated comment-only merge_ready dispatches surface as "PM cannot drive this PR" instead of a success streak.

Mixed items (e.g. `ci_failure+merge_ready`) keep the delivery signal — a reply there can be the legitimate deliverable of the other type.

Scope note: the effect field only exists in the runloops executor (the live slot path); the bash worker never derived one, so there is no second side to fix.

Not a duplicate of #1469 — that widens *which* types get before/after snapshots (unknown→observable); this narrows *what counts* as effect for pure merge_ready. Same file, independent hunks; whichever lands second rebases trivially.

## Tests

3 new: comment-only pure merge_ready → `none`; actual merge → `observed`; mixed `ci_failure+merge_ready` with verified reply → `observed`. 109 passed in `test_run_item.py`.

### DIFF
--- a/packages/gptme-runloops/src/gptme_runloops/run_item.py
+++ b/packages/gptme-runloops/src/gptme_runloops/run_item.py
@@ -1953,6 +1953,14 @@ def run_post_session(
         and item.number_str != "0"
         and hooks.delivery_check is not None
         and THREAD_DELIVERABLE_TYPES & set(item.types)
+        # A pure merge_ready item's deliverable is the MERGE, not a comment —
+        # the reply post-condition does not apply. Running it anyway caused
+        # both halves of the gptme/gptme#3531 spam: --post-fallback-reply
+        # posted "Automated placeholder" comments for silent merges, and
+        # orphan_no_delivery rolled the item back into the queue. Skipping
+        # leaves delivery_verified False, so the effect signal (step 9) grades
+        # purely on the PR-state diff and pm_dispatch_recovery owns retries.
+        and set(item.types) != {"merge_ready"}
     ):
         try:
             try:
@@ -2020,8 +2028,6 @@ def run_post_session(
                 f"WARN: PM delivery post-condition FAILED — {plan.repo}#{plan.number}: "
                 "session exited without a thread reply"
             )
-
-    merge_ready_only = set(item.types) == {"merge_ready"}
 
     latency_outcome = compute_latency_outcome(
         delivery_outcome,
@@ -2231,28 +2237,7 @@ def run_post_session(
     # every pending notif-*.state, consuming exactly what was "rolled back".
     # The result was a fix that passed its tests and changed nothing. Keep all
     # three effects together in rollback_failed_delivery().
-    # PR-state-diff-only effect (no delivery signal), needed by both the
-    # rollback guard below and the effect signal in step 9.
-    pr_diff_effect = read_record_effect_signal(record_file, delivery_outcome="")
-
-    # A pure merge_ready session that actually merged/closed the PR (or pushed
-    # a rebase) but posted no reply is NOT a failed delivery — its deliverable
-    # was the merge, not a comment. Rolling it back would re-queue an
-    # already-settled PR (and pre-#1461 that re-dispatch produced yet another
-    # comment). Promote instead. Reply-type items keep the rollback: for them
-    # a missing reply genuinely means the work never reached the thread.
-    if (
-        delivery_outcome == "orphan_no_delivery"
-        and merge_ready_only
-        and pr_diff_effect == EFFECT_OBSERVED
-    ):
-        _log(
-            f"NOTE: merge_ready-only item {plan.repo}#{plan.number}: PR state "
-            "moved (merge/close/push) with no reply posted — promoting state, "
-            "not rolling back (the merge IS the deliverable)"
-        )
-        promote_item_state(config, item.repo, item.number)
-    elif delivery_outcome == "orphan_no_delivery":
+    if delivery_outcome == "orphan_no_delivery":
         rollback_slot_key = resolve_slot_key(config, item)
         if rollback_failed_delivery(config, item.repo, item.number, rollback_slot_key):
             _log(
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
## 🤖 AI code review

This PR changes run_post_session so that the delivery check is skipped for items whose types set is exactly {"merge_ready"}, preventing a posted comment from being counted as an observable effect. The effect signal for pure merge_ready items now grades entirely on the PR-state diff (merged/closed transition, merge commit, head advance). It also updates the delivery-check guard test to exercise a mixed-type item for merge_ready and adds four new tests covering comment-only, actual merge, mixed-type, and merge-without-reply scenarios.

**1 finding** · ❌ **1** P1 — 🛑 **1 at P0/P1.** Any finding, P2 included, keeps the score under 5/5.

**❌ P1 high** — `packages/gptme-runloops/src/gptme_runloops/run_item.py:1963`

For a pure merge_ready item where the session neither merges the PR nor posts any comment, the new condition `and set(item.types) != {"merge_ready"}` causes the entire delivery-check block to be skipped. `delivery_outcome` therefore retains its initial value `'handled'`, so step 8's `else` branch unconditionally calls `promote_item_state`. This copies the item's pending state to the real state dir, marking the item as processed. The effect is graded `'none'` (PR-state unchanged), but because the state is now promoted, the dispatcher will not re-emit the item in the next cycle. Previously the delivery check would have reported `orphan_no_delivery`, triggering rollback (rollback_failed_delivery) and keeping the item pending for a future retry. The PR description claims 'pm_dispatch_recovery owns retries', but that pathway requires the item to remain in the pending state; promoting it makes any retry impossible. Concretely, a merge_ready PR whose merge command fails silently (e.g., a transient GitHub error) while the worker posts nothing will be left permanently unmerged, with no subsequent dispatch attempts. The new tests do not cover this no-merge/no-comment case.

<sub>**How this was verified:** Traced that delivery_outcome is initialized to "handled" at line 1940 and never changed for pure merge_ready because the entire delivery block is skipped. The latency append at line 2043 passes delivery_outcome to compute_latency_outcome, which in existing tests returns "handled" when delivery_outcome is "handled".</sub>

</details>

- `packages/gptme-runloops/src/gptme_runloops/run_item.py` — Adds a condition to skip the delivery check for pure merge_ready items and updates the effect-signal step comment.
- `packages/gptme-runloops/tests/test_run_item.py` — Adds four tests for merge_ready effect grading and modifies the parametrized delivery-check test to use mixed types for merge_ready.

</details>

| `02d43d43c65f` | 3/5 | 1 | llm | 2026-08-20 10:32 UTC |
| `79510f21b49b` | 3/5 | 1 | llm | 2026-08-20 10:48 UTC |

</details>

<sub>Reviewed `9b92b7fa8e58` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 207s · reports findings; merge decisions belong to the gate · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

packages/gptme-runloops/src/gptme_runloops/run_item.py
```diff
@@ -2229,7 +2231,28 @@ def run_post_session(
...
+    # A pure merge_ready session that actually merged/closed the PR (or pushed
+    # a rebase) but posted no reply is NOT a failed delivery — its deliverable
+    # was the merge, not a comment. Rolling it back would re-queue an
+    # already-settled PR (and pre-#1461 that re-dispatch produced yet another
+    # comment). Promote instead. Reply-type items keep the rollback: for them
+    # a missing reply genuinely means the work never reached the thread.
+    if (
+        delivery_outcome == "orphan_no_delivery"
+        and merge_ready_only
+        and pr_diff_effect == EFFECT_OBSERVED
```
TimeToBuildBob: ❌ **P1** — The rollback promotion for a pure merge_ready item only fires when pr_diff_effect is exactly EFFECT_OBSERVED. If the PR-state diff is unknown (for example, the snapshot fetch in step 2 fails and update_record_pr_state writes no pr_state_after), read_record_effect_signal returns EFFECT_UNKNOWN, the condition fails, and the item falls through to the normal rollback branch. A session that actually merged the PR (and thus has no reply to report, so delivery_outcome is orphan_no_delivery) is then rolled back: its slot event marker is cleared and it is re-queued, producing another dispatch that posts another comment — re-creating the exact gptme/gptme#3531 spam loop this PR aims to eliminate. The fix is incomplete for the unknown-diff case; promotion should also occur on EFFECT_UNKNOWN (or at least the rollback should be suppressed) for pure merge_ready items with an orphan delivery, since an inconclusive diff should not cause a redundant re-dispatch of an already-settled PR. _(3/3 passes)_

TimeToBuildBob: Fixed in 9b92b7fa8e by removing the special case entirely: a pure merge_ready item now skips the delivery check at the step-4 gate (its deliverable is the merge, not a comment), so delivery_outcome can never be orphan_no_delivery for it — the EFFECT_UNKNOWN edge is unreachable. Also stops --post-fallback-reply placeholder comments on silent merges.

packages/gptme-runloops/src/gptme_runloops/run_item.py
```diff
@@ -1953,6 +1953,14 @@ def run_post_session(
...
         and hooks.delivery_check is not None
         and THREAD_DELIVERABLE_TYPES & set(item.types)
+        # A pure merge_ready item's deliverable is the MERGE, not a comment —
+        # the reply post-condition does not apply. Running it anyway caused
+        # both halves of the gptme/gptme#3531 spam: --post-fallback-reply
+        # posted "Automated placeholder" comments for silent merges, and
+        # orphan_no_delivery rolled the item back into the queue. Skipping
+        # leaves delivery_verified False, so the effect signal (step 9) grades
+        # purely on the PR-state diff and pm_dispatch_recovery owns retries.
+        and set(item.types) != {"merge_ready"}
```
TimeToBuildBob, line 1963: ❌ **P1** — For a pure merge_ready item where the session neither merges the PR nor posts any comment, the new condition `and set(item.types) != {"merge_ready"}` causes the entire delivery-check block to be skipped. `delivery_outcome` therefore retains its initial value `'handled'`, so step 8's `else` branch unconditionally calls `promote_item_state`. This copies the item's pending state to the real state dir, marking the item as processed. The effect is graded `'none'` (PR-state unchanged), but because the state is now promoted, the dispatcher will not re-emit the item in the next cycle. Previously the delivery check would have reported `orphan_no_delivery`, triggering rollback (rollback_failed_delivery) and keeping the item pending for a future retry. The PR description claims 'pm_dispatch_recovery owns retries', but that pathway requires the item to remain in the pending state; promoting it makes any retry impossible. Concretely, a merge_ready PR whose merge command fails silently (e.g., a transient GitHub error) while the worker posts nothing will be left permanently unmerged, with no subsequent dispatch attempts. The new tests do not cover this no-merge/no-comment case. _(3/3 passes)_

TimeToBuildBob: Intended behavior change, and the retry path the finding says is lost is exactly the one the PR routes this case onto:

1. "The dispatcher will not re-emit the item" — re-emission for merge_ready does not come from the pending-state promotion. `check_merge_ready` re-emits from its own `${repo}-pr-N-merge-ready.state` file (per-HEAD 12h cooldown), which `promote_item_state` does not touch.
2. The do-nothing session grades `effect=none` → ledger `outcome=no_effect` → `pm_dispatch_recovery.classify_completion` = CLASS_INEFFECTIVE → `decide()` re-arms after backoff by clearing the slot's `.event`/`.atoms` dedupe markers. This is observable in production on the motivating incident itself: `state/project-monitoring-dispatch.jsonl` rows for gptme/gptme#3531 show `rearm ineffective:attempt1 cleared=.event,.atoms,.ts`.
3. The prior orphan→rollback path was an *immediate* re-queue with only PM_MAX_REDELIVERY_ATTEMPTS bounding it — for merge_ready that is the #3531 comment-spam loop (each redelivery could post another fallback/placeholder comment). Bounded, backed-off, escalating retries via the recovery module are the intended replacement, as stated in the PR description.

PR #1474, opened 2026-08-20
fix(pm): notification state is owned by the handling worker, not the next worker to finish

TimeToBuildBob:
## Symptom

ActivityWatch/activitywatch#1402 (2026-08-20): Erik @mentions Bob at 09:31. PM dispatches within 10 min. The worker is killed at exit 124 after 916s having posted nothing. Then **zero dispatch rows of any kind** — the notification is still `unread: true` on GitHub, but `/tmp/bob-project-monitoring-state/notif-25192203058.state` says it was handled.

## Cause

Two mechanisms each consumed the thread's gate state on their own:

1. The killed worker's own end-of-run `promote_notification_states()` — a timeout is not a delivery.
2. **Any sibling worker's** `promote_notification_states()` while the item sat emitted-but-skipped (`skipped_active` rows 09:43–09:54) in the *shared* pending dir. The function promoted every pending `notif-*.state` under the assumption "sessions already ran for every emitted notification" — true for the old monolithic sweep session, false for per-slot transient units.

The gate dedupes notifications on `updated_at`, so the thread never re-emits until a human comments again. `pm_dispatch_recovery.py` would have re-armed the slot (`rearm failed:attempt1` on dry-run) but it only runs for emitted items — it was never consulted.

## Fix

- `promote_notification_states()`: promote only **unmapped** `notif-*.state`. The gate writes a `.map` sidecar (`repo#number`) for every attributable thread; mapped state is owned by its item.
- `promote_item_state()`: promote the item's mapped notif state on the handling worker's success path. Number-0 items promote only unmapped state (the old blanket copy was a side door — reviewer finding).
- `run_post_session()` step 8: a timed-out / non-zero exit **without a verified delivery** purges the item's pending notif state so the gate re-emits next sweep. A non-zero exit *after* a verified reply still promotes (no duplicate reply — reviewer finding). Slot `.event`/`.ts` markers are intentionally left for `pm_dispatch_recovery`'s backed-off, budgeted re-arm (dispatcher consults it before its unchanged-payload and cooldown checks).

Emitted-but-skipped items' pending state is discarded by the next sweep's `rsync --delete`, so they re-emit and are re-evaluated — the correct semantics.

## Tests

Symptom-shaped: assert on-disk state *after* the end-of-run promotion has also run. The 4 new behaviors fail against the previous code (verified red/green). One existing negative-control test encoded the old "sibling promotes unrelated mapped items" contract and was updated to the ownership contract. 1041 passed.

Pre-PR AI review (`ai-review-local.py`, 3 passes): round 1 raised the two findings above (both fixed), round 2's finding (clear slot markers here) was a design disagreement — now documented inline.

## Not in this PR

- The bash in-process worker path (`project-monitoring.sh` blanket `notif-*.state` loop) has the same shape; the live executor is runloops. Tracked in ErikBjare/bob follow-up.
- Mention items get the 900s fast-lane tier; the model override for direct mentions doesn't cover timeout. Separate follow-up.

https://claude.ai/code/session_011VKFZeMpLvtZ3e15S1fKUW

### DIFF
--- a/packages/gptme-runloops/src/gptme_runloops/run_item.py
+++ b/packages/gptme-runloops/src/gptme_runloops/run_item.py
@@ -947,10 +947,37 @@ def rollback_failed_delivery(
     return True
 
 
+def _mapped_notif_state_files(pending: Path, repo: str, number: int | str | None):
+    """Yield ``(map_file, state_file)`` for pending notif state mapped to this item.
+
+    Notification state is keyed by GitHub API id (``notif-<id>.state``); the
+    ``.map`` sidecar written by ``activity-gate.sh`` carries the ``repo#number``
+    association. This is the only way to attribute a notif state to an item.
+    """
+    if not pending.is_dir():
+        return
+    target = f"{repo}#{number}"
+    for map_file in pending.glob("notif-*.map"):
+        if not map_file.is_file():
+            continue
+        try:
+            content = map_file.read_text(encoding="utf-8").strip()
+        except OSError:
+            continue
+        if content == target:
+            yield map_file, map_file.with_suffix(".state")
+
+
 def promote_item_state(
     config: RunItemConfig, repo: str, number: int | str | None
 ) -> None:
-    """Copy the item's pending activity-gate state files to the real state dir."""
+    """Copy the item's pending activity-gate state files to the real state dir.
+
+    Includes the item's *mapped* ``notif-*.state`` (via the ``.map`` sidecar):
+    notification state is owned by the outcome of the worker that handled the
+    item, not by whichever worker happens to finish next — see
+    :func:`promote_notification_states`.
+    """
     import shutil
 
     # The item advanced, so any redelivery treadmill for it is over — restore
@@ -974,7 +1001,15 @@ def promote_item_state(
     repo_safe = str(repo).replace("/", "-")
     is_zero = str(number) == "0"
     if is_zero:
-        patterns = ["notif-*.state"]
+        # A number-0 item cannot own a mapped thread (the gate only writes a
+        # `.map` for number > 0), so it promotes only the unmapped remainder —
+        # the same rule as promote_notification_states(). Copying every
+        # notif-*.state here would consume siblings' emitted-but-unhandled
+        # threads through a side door.
+        patterns = []
+        for f in pending.glob("notif-*.state"):
+            if f.is_file() and not f.with_suffix(".map").is_file():
+                shutil.copy(f, state / f.name)
     else:
         patterns = [
             f"{repo_safe}-pr-{number}*.state",
@@ -985,14 +1020,38 @@ def promote_item_state(
         for f in pending.glob(pattern):
             if f.is_file():
                 shutil.copy(f, state / f.name)
+    if not is_zero:
+        for map_file, state_file in _mapped_notif_state_files(pending, repo, number):
+            for f in (map_file, state_file):
+                if f.is_file():
+                    shutil.copy(f, state / f.name)
 
 
 def promote_notification_states(config: RunItemConfig) -> None:
-    """Promote ALL pending notif state files at end of run (p-m.sh:695-704).
+    """Promote pending notif state that no item owns, at end of run (p-m.sh:695-704).
 
     Notification state files use GitHub API ids, not repo#number, so the
-    per-item promotion never copies them; without this the gate re-emits the
-    same notifications every cycle.
+    per-item promotion cannot find them by name; the gate writes a ``.map``
+    sidecar (``repo#number``) for every thread it can attribute. Only the
+    *unmapped* remainder is promoted here.
+
+    History: this used to promote EVERY pending ``notif-*.state`` under the
+    assumption "sessions already ran for every emitted notification". That
+    held when one monolithic session handled the whole sweep. With per-slot
+    transient units it does not: the pending dir is shared, so a sibling
+    worker finishing while another item sits emitted-but-skipped
+    (``skipped_active`` / cooldown / recovery backoff) promoted that item's
+    state too — and a worker that *timed out* promoted its own. Either way
+    the gate never re-emitted the thread (it dedupes on ``updated_at``) and
+    ``pm_dispatch_recovery`` was never consulted, because it only runs for
+    emitted items. ActivityWatch/activitywatch#1402 (2026-08-20): an Erik
+    @mention, one 900s worker killed at exit 124, then silence.
+
+    Mapped state is promoted by :func:`promote_item_state` on the handling
+    worker's success path and purged by :func:`purge_pending_notif_state` on
+    its failure path, so the thread re-emits and the bounded recovery retry
+    applies. Skipped items' pending state is simply left for the next sweep's
+    ``rsync --delete`` to discard, so they re-emit and are re-evaluated.
     """
     import shutil
 
@@ -1001,8 +1060,11 @@ def promote_notification_states(config: RunItemConfig) -> None:
         return
     config.state_dir.mkdir(parents=True, exist_ok=True)
     for f in pending.glob("notif-*.state"):
-        if f.is_file():
-            shutil.copy(f, config.state_dir / f.name)
+        if not f.is_file():
+            continue
+        if f.with_suffix(".map").is_file():
+            continue  # owned by an item; its worker's outcome decides
+        shutil.copy(f, config.state_dir / f.name)
 
 
 # --- Trajectory resolution (worker.sh:196-300; design §4.4) ---
@@ -2311,6 +2373,34 @@ def run_post_session(
                 "promoting state to end re-dispatch churn (no reply is likely correct here)"
             )
             promote_item_state(config, item.repo, item.number)
+    elif (outcome.timed_out or outcome.exit_code != 0) and not (
+        delivery_verified and delivery_outcome == "handled"
+    ):
+        # The worker did not finish (exit 124) or failed outright, and the
+        # delivery check did not verify a reply. Its PR-side state still
+        # promotes (the head/score it saw are real), but the item's
+        # notification state must NOT: promoting it tells the gate the
+        # mention was handled, the gate dedupes on `updated_at`, and the thread
+        # goes silent until a human comments again. Purge it so the gate
+        # re-emits next sweep; pm_dispatch_recovery bounds the retry.
+        # (A non-zero exit AFTER a verified reply falls through to the normal
+        # promotion — re-emitting there would post a duplicate reply.)
+        #
+        # Deliberately NOT clearing the slot's `.event`/`.ts` markers here,
+        # unlike rollback_failed_delivery(): once the gate re-emits the thread,
+        # the dispatcher consults pm_dispatch_recovery.py (before its
+        # unchanged-payload and cooldown checks), which clears those markers on
+        # a failed last-terminal row with a 120s→1h backoff and a bounded
+        # attempt budget that escalates to a task. Clearing them here would
+        # re-dispatch on the very next sweep with no backoff and no budget.
+        purged = purge_pending_notif_state(config, item.repo, item.number)
+        if purged:
+            _log(
+                f"WARN: worker exit={outcome.exit_code} timed_out={outcome.timed_out} "
+                f"for {item.repo}#{item.number} — left {purged} notification "
+                "thread(s) un-promoted so the item re-enters the dispatch queue"
+            )
+        promote_item_state(config, item.repo, item.number)
     else:
         promote_item_state(config, item.repo, item.number)
 
--- a/packages/gptme-runloops/src/gptme_runloops/run_item.py
+++ b/packages/gptme-runloops/src/gptme_runloops/run_item.py
@@ -2403,24 +2403,26 @@ def run_post_session(
     ):
         # The delivery check did not verify a reply. Count this through the
         # same bounded rollback as orphan_no_delivery so a permanently broken
-        # check cannot re-emit this thread forever. PR-side state is still
-        # valid and promotes on every pass; only mapped notification state is
-        # purged while the retry budget remains.
+        # check cannot re-emit this thread forever. Clear the launch marker so
+        # the item can re-enter after the dispatcher's normal .ts cooldown;
+        # leaving it stamped suppresses a genuinely unanswered mention for the
+        # full event TTL. PR-side state is still valid and promotes on every
+        # pass; only mapped notification state is purged while the retry budget
+        # remains.
         rollback_slot_key = resolve_slot_key(config, item)
         if rollback_failed_delivery(
             config,
             item.repo,
             item.number,
             rollback_slot_key,
-            clear_event_markers=False,
         ):
             promote_item_state(
                 config, item.repo, item.number, reset_redelivery_counter=False
             )
             _log(
                 "WARN: PM delivery post-condition was not verified — rolled back "
-                f"pending notif state for {plan.repo}#{plan.number}; leaving the "
-                "event marker for backed-off dispatch recovery"
+                f"pending notif state and event marker for {plan.repo}#{plan.number} "
+                "(item re-enters after the dispatch cooldown)"
             )
         else:
             _log(
--- a/packages/gptme-runloops/src/gptme_runloops/run_item.py
+++ b/packages/gptme-runloops/src/gptme_runloops/run_item.py
@@ -2381,20 +2381,17 @@ def run_post_session(
                 "promoting state to end re-dispatch churn (no reply is likely correct here)"
             )
             promote_item_state(config, item.repo, item.number)
-    elif (
-        (outcome.timed_out or outcome.exit_code != 0)
-        and hooks.delivery_check is not None
-        and not (delivery_verified and delivery_outcome == "handled")
+    elif hooks.delivery_check is not None and not (
+        delivery_verified and delivery_outcome == "handled"
     ):
-        # The worker did not finish (exit 124) or failed outright, and the
-        # delivery check did not verify a reply. Its PR-side state still
+        # The delivery check did not verify a reply. Its PR-side state still
         # promotes (the head/score it saw are real), but the item's
         # notification state must NOT: promoting it tells the gate the
         # mention was handled, the gate dedupes on `updated_at`, and the thread
         # goes silent until a human comments again. Purge it so the gate
         # re-emits next sweep; pm_dispatch_recovery bounds the retry.
-        # (A non-zero exit AFTER a verified reply falls through to the normal
-        # promotion — re-emitting there would post a duplicate reply.)
+        # A verified reply falls through to normal promotion even when the
+        # worker failed afterward; re-emitting there would post a duplicate.
         #
         # Deliberately NOT clearing the slot's `.event`/`.ts` markers here,
         # unlike rollback_failed_delivery(): once the gate re-emits the thread,
--- a/packages/gptme-runloops/src/gptme_runloops/run_item.py
+++ b/packages/gptme-runloops/src/gptme_runloops/run_item.py
@@ -976,6 +976,7 @@ def promote_item_state(
     number: int | str | None,
     *,
     reset_redelivery_counter: bool = True,
+    include_master_ci: bool = False,
 ) -> None:
     """Copy the item's pending activity-gate state files to the real state dir.
 
@@ -1022,8 +1023,9 @@ def promote_item_state(
         patterns = [
             f"{repo_safe}-pr-{number}*.state",
             f"{repo_safe}-issue-{number}*.state",
-            f"{repo_safe}-master-ci.state",
         ]
+        if include_master_ci:
+            patterns.append(f"{repo_safe}-master-ci.state")
     for pattern in patterns:
         for f in pending.glob(pattern):
             if f.is_file():
@@ -1041,8 +1043,9 @@ def promote_notification_states(config: RunItemConfig) -> None:
     Notification state files use GitHub API ids, not repo#number, so the
     per-item promotion cannot find them by name; the gate writes a ``.map``
     sidecar (``repo#number``) for every thread it can attribute. Only the
-    *unmapped* remainder is promoted here. An undecodable map cannot establish
-    ownership, so treat its state as unmapped rather than stranding it forever.
+    *unmapped* remainder is promoted here. A readable empty map is unmapped;
+    an unreadable map remains pending because its ownership is unknown and
+    promoting it could silently consume an unhandled notification.
 
     History: this used to promote EVERY pending ``notif-*.state`` under the
     assumption "sessions already ran for every emitted notification". That
@@ -1076,10 +1079,9 @@ def promote_notification_states(config: RunItemConfig) -> None:
             try:
                 owner = map_file.read_text(encoding="utf-8").strip()
             except (OSError, UnicodeDecodeError):
-                pass  # unreadable map cannot prove item ownership
-            else:
-                if owner:
-                    continue  # owned by an item; its worker's outcome decides
+                continue  # unknown ownership: fail closed and retry next sweep
+            if owner:
+                continue  # owned by an item; its worker's outcome decides
         shutil.copy(f, config.state_dir / f.name)
 
 
@@ -2434,7 +2436,12 @@ def run_post_session(
             )
             promote_item_state(config, item.repo, item.number)
     else:
-        promote_item_state(config, item.repo, item.number)
+        promote_item_state(
+            config,
+            item.repo,
+            item.number,
+            include_master_ci="master_ci_failure" in item.types,
+        )
 
     # 9. Observable-effect signal.
     # A clean exit is not evidence the work landed: a worker can commit a fix
--- a/packages/gptme-runloops/src/gptme_runloops/run_item.py
+++ b/packages/gptme-runloops/src/gptme_runloops/run_item.py
@@ -113,6 +113,7 @@ from gptme_runloops.prompt_templates import (
     render_preheld_claim_block,
 )
 from gptme_runloops.worker_records import (
+    KNOWN_DELIVERY_OUTCOMES,
     append_wait_merge_gate_log,
     append_worker_latency_records,
     build_wait_merge_gate_entry,
@@ -1005,7 +1006,8 @@ def promote_item_state(
         # `.map` for number > 0), so it promotes only the unmapped remainder —
         # the same rule as promote_notification_states(). Copying every
         # notif-*.state here would consume siblings' emitted-but-unhandled
-        # threads through a side door.
+        # threads through a side door. It cannot own repo-level master-CI
+        # state either.
         patterns = []
         for f in pending.glob("notif-*.state"):
             if f.is_file() and not f.with_suffix(".map").is_file():
@@ -1014,8 +1016,8 @@ def promote_item_state(
         patterns = [
             f"{repo_safe}-pr-{number}*.state",
             f"{repo_safe}-issue-{number}*.state",
+            f"{repo_safe}-master-ci.state",
         ]
-    patterns.append(f"{repo_safe}-master-ci.state")
     for pattern in patterns:
         for f in pending.glob(pattern):
             if f.is_file():
@@ -2114,16 +2116,16 @@ def run_post_session(
                     text=True,
                     cwd=str(config.workspace),
                 )
-                if proc.returncode == 0:
+                check_succeeded = proc.returncode == 0
+                if check_succeeded:
                     raw = proc.stdout
-                    delivery_verified = True
                 else:
                     raw = '{"outcome":"handled"}'
             except (OSError, subprocess.SubprocessError):
+                check_succeeded = False
                 raw = '{"outcome":"handled"}'
-            delivery_outcome = normalize_delivery_outcome(
-                extract_delivery_field(raw, "outcome")
-            )
+            raw_outcome = extract_delivery_field(raw, "outcome")
+            delivery_outcome = normalize_delivery_outcome(raw_outcome)
             needs_fallback = re.sub(
                 r"[ \t\n\r\f\v]+",
                 "",
@@ -2133,6 +2135,12 @@ def run_post_session(
                 r"[ \t\n\r\f\v]+",
                 "",
                 extract_delivery_field(raw, "fallback_reply_posted"),
+            )
+            # normalize_delivery_outcome deliberately maps malformed/unknown
+            # values to the permissive "handled" default. That is safe for
+            # compatibility, but it is not evidence that delivery was observed.
+            delivery_verified = (
+                check_succeeded and raw_outcome.strip() in KNOWN_DELIVERY_OUTCOMES
             )
         except Exception:
             # NOTE(divergence, kept from step 3): total extractor death fails
@@ -2373,8 +2381,10 @@ def run_post_session(
                 "promoting state to end re-dispatch churn (no reply is likely correct here)"
             )
             promote_item_state(config, item.repo, item.number)
-    elif (outcome.timed_out or outcome.exit_code != 0) and not (
-        delivery_verified and delivery_outcome == "handled"
+    elif (
+        (outcome.timed_out or outcome.exit_code != 0)
+        and hooks.delivery_check is not None
+        and not (delivery_verified and delivery_outcome == "handled")
     ):
         # The worker did not finish (exit 124) or failed outright, and the
         # delivery check did not verify a reply. Its PR-side state still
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:
CI is green and the PR is cleanly mergeable, but the self-merge gate is not yet eligible because the initial Greptile review has not landed. Leaving this open for that review; project monitoring will re-evaluate afterward.

packages/gptme-runloops/src/gptme_runloops/run_item.py
```diff
@@ -2243,6 +2305,34 @@ def run_post_session(
                 "promoting state to end re-dispatch churn (no reply is likely correct here)"
             )
             promote_item_state(config, item.repo, item.number)
+    elif (outcome.timed_out or outcome.exit_code != 0) and not (
```
TimeToBuildBob: ❌ **P1** — When `hooks.delivery_check` is None (no delivery check configured), the new branch in `run_post_session` triggers on any non-zero exit or timeout, because `delivery_verified` is always False and `delivery_outcome` defaults to "handled", making `not (delivery_verified and delivery_outcome == "handled")` True. It then calls `purge_pending_notif_state`, deleting the item's mapped `.map` and `.state` files. If the worker actually posted a reply before failing, this purge causes the thread to be re-emitted and a duplicate reply to be posted. The previous behavior always promoted notif state in this scenario. This is a regression for any setup without a delivery check hook.

TimeToBuildBob:
## 🤖 AI code review

The PR changes notification state ownership so that the end-of-run blanket promotion only promotes unmapped (no .map or empty .map) notification state, while mapped notification state is promoted only by the handling worker's promote_item_state. It adds a _mapped_notif_state_files helper, extends promote_item_state with reset_redelivery_counter and include_master_ci options, and adds a new rollback branch in run_post_session for unverified delivery that purges the item's notif state and clears event markers while still promoting PR-side state. It also adds UnicodeDecodeError handling and a large set of tests pinning the new on-disk ownership semantics.

- `packages/gptme-runloops/src/gptme_runloops/run_item.py` — Adds mapped-notif-state promotion to promote_item_state, restricts promote_notification_states to unmapped state, adds an unverified-delivery rollback branch, and tightens delivery_verified to require parseable known outcomes.
- `packages/gptme-runloops/tests/test_run_item.py` — Adds symptom tests for notification ownership, unverified-delivery rollback, redelivery caps, and master-CI promotion; updates existing promotion tests for the new default include_master_ci=False.
- `packages/gptme-runloops/tests/test_run_item_rollback.py` — Updates the surgical-rollback test to call promote_item_state for the sibling item and adds a class pinning notification-state ownership to the handling worker.

</details>

| `0d45210545bf` | 3/5 | 2 | llm | 2026-08-20 23:03 UTC |
| `20b25d267306` | 4/5 | 2 | llm | 2026-08-20 23:41 UTC |
| `dcfd7f7c3a42` | 3/5 | 1 | llm | 2026-08-21 01:22 UTC |
| `1e453e999f1b` | 4/5 | 1 | llm | 2026-08-21 04:04 UTC |
| `950a75ca04c4` | 4/5 | 1 | llm | 2026-08-21 06:02 UTC |
| `124b16a3abbb` | 3/5 | 1 | llm | 2026-08-21 07:53 UTC |
| `48a67151d497` | 3/5 | 2 | llm | 2026-08-21 08:21 UTC |
| `2eb927bc58cb` | 4/5 | 1 | llm | 2026-08-21 13:35 UTC |
| `7539cdbbf3f6` | 3/5 | 1 | llm | 2026-08-22 16:44 UTC |
| `eaec024e3efb` | 4/5 | 1 | llm | 2026-08-22 17:41 UTC |

</details>

<sub>Reviewed `33c33353e427` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 327s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

packages/gptme-runloops/src/gptme_runloops/run_item.py
TimeToBuildBob: Fixed in f9019bbe. Failed/timed-out workers now purge mapped notification state only when a delivery-check hook exists and did not verify a handled outcome; without an observation hook, state is preserved. Added the missing regression test.

packages/gptme-runloops/src/gptme_runloops/run_item.py
```diff
@@ -2311,6 +2381,36 @@ def run_post_session(
                 "promoting state to end re-dispatch churn (no reply is likely correct here)"
             )
             promote_item_state(config, item.repo, item.number)
+    elif (
+        (outcome.timed_out or outcome.exit_code != 0)
```
TimeToBuildBob: ❌ **P1** — The new purge branch in `run_post_session` triggers only when `(outcome.timed_out or outcome.exit_code != 0)`. If a worker exits 0 but the delivery check fails (non-zero exit, crash, or malformed output), `delivery_verified` is False and `delivery_outcome` is the permissive "handled" fallback. The condition is then False, control reaches the `else` branch, and `promote_item_state` runs without purging. For a thread-deliverable item with mapped notification state, that state is copied to the real state dir, telling the gate the mention was handled; the gate dedupes on `updated_at` and the thread goes silent even though no reply was posted. This is the exact failure class the PR is fixing (e.g., a worker that exits 0 after doing nothing, with a broken delivery check), so the zero-exit path remains a hole. The fix is to also purge when the delivery check is present and `delivery_verified` is False regardless of exit code, or at least when `exit_code == 0 and not delivery_verified`.

TimeToBuildBob: Fixed in 2801c060. Notification state now purges whenever a configured delivery check fails to verify `handled`, including a clean worker exit. Added a regression test for exit 0 plus a crashed delivery check; the full `test_run_item.py` slice passes (129 tests).

TimeToBuildBob:
@greptileai review

packages/gptme-runloops/src/gptme_runloops/run_item.py
```diff
@@ -2311,6 +2381,33 @@ def run_post_session(
...
+        # worker failed afterward; re-emitting there would post a duplicate.
+        #
+        # Deliberately NOT clearing the slot's `.event`/`.ts` markers here,
+        # unlike rollback_failed_delivery(): once the gate re-emits the thread,
+        # the dispatcher consults pm_dispatch_recovery.py (before its
+        # unchanged-payload and cooldown checks), which clears those markers on
+        # a failed last-terminal row with a 120s→1h backoff and a bounded
+        # attempt budget that escalates to a task. Clearing them here would
+        # re-dispatch on the very next sweep with no backoff and no budget.
+        purged = purge_pending_notif_state(config, item.repo, item.number)
```
TimeToBuildBob: ❌ **P1** — The new branch in `run_post_session` (step 8) purges an item's pending notification state and then calls `promote_item_state`. `promote_item_state` resets the redelivery counter by unlinking the `redeliver-<repo>-<number>.attempts` file (lines 990-995). This means items that fail delivery with a non-orphan outcome (e.g. `failed`, or a broken delivery check that returns non-zero and falls back to `handled` with `delivery_verified=False`) never hit the `PM_MAX_REDELIVERY_ATTEMPTS` cap. Each re-dispatch resets the counter, so the item can be re-emitted indefinitely on every sweep. The PR argues `pm_dispatch_recovery` bounds the retry, but that is a separate, optional mechanism; the redelivery counter was the enforced safeguard and is now bypassed for these failures. A worker that consistently fails delivery (e.g. the check is misconfigured) will loop forever, generating repeated dispatch rows and potential duplicate work.

TimeToBuildBob: Fixed in bda67023. Unverified delivery retries now preserve the attempt counter and stop at `PM_MAX_REDELIVERY_ATTEMPTS`; the cap path promotes state and clears the counter. The regression test exercises two consecutive failed checks and verifies the second pass terminates re-dispatch.

TimeToBuildBob:
## Convergence adjudication

**Fixed**
- **P1 — unverified rollback applied when no delivery check ran:** fixed in `0d452105`. The rollback branch now mirrors the delivery-check eligibility guard: a real non-zero thread, a thread-deliverable event type, and not pure `merge_ready`. The repo-level regression test seeds a redelivery counter and proves `master_ci_failure` skips the check, promotes state, and clears—not consumes—the stale counter.
- **P1 — unbounded retries for an unverified check:** fixed in `bda67023`. Failed verification retains the attempt counter until the bounded cap is reached; the cap then promotes state and clears it.
- Earlier P1s around absent hooks and clean-exit/unverified checks remain covered by `f9019bbe` and `2801c060`.

**Remaining**
- No blocking, security, or non-blocking review findings remain after classification. The detector reports `0` unresolved P1/P2/P3 findings. I am not requesting another review: this is the cap adjudication.

**CI**
- Prior head `bda67023`: all checks green.
- Current head `0d452105`: CI is running. Local verification passed: `130 passed` in `packages/gptme-runloops/tests/test_run_item.py`, package mypy passed, scoped pre-commit hooks passed, and `git diff --check` passed.

**Domain risk**
- The fragile surface is mixed PM item types sharing one post-session state machine. A maintainer manual check should focus on these three cases: `master_ci_failure` never calls the thread-delivery probe; pure `merge_ready` does not require a comment; a genuinely thread-deliverable item with a broken probe retries only to the configured cap.

**Convergence**
- Detector `round_convergence`: `status=no_history`, `stable_rounds=0`, `required_stable_rounds=2`. The local detector does not parse this PR's Bob AI-review history, despite four recorded review passes; its actionable finding count is now zero.

Stopping at maintainer judgment; no merge and no further review trigger.

packages/gptme-runloops/src/gptme_runloops/run_item.py
```diff
@@ -947,10 +948,42 @@ def rollback_failed_delivery(
...
+    association. This is the only way to attribute a notif state to an item.
+    """
+    if not pending.is_dir():
+        return
+    target = f"{repo}#{number}"
+    for map_file in pending.glob("notif-*.map"):
+        if not map_file.is_file():
+            continue
+        try:
+            content = map_file.read_text(encoding="utf-8").strip()
```
TimeToBuildBob, line 966: ❌ **P1** — The `_mapped_notif_state_files` generator (and the same pattern in `purge_pending_notif_state`) reads each `notif-*.map` with `read_text(encoding='utf-8')` but only catches `OSError` on failure. A map file containing invalid UTF-8 bytes raises `UnicodeDecodeError`, a subclass of `ValueError`, which is not caught and propagates out of `promote_item_state` or `rollback_failed_delivery`. Because the pending state dir is shared and written by the activity gate, a single corrupt map (e.g., from a partial write or a maliciously crafted file) will crash the calling worker mid-run, aborting state promotion and the rest of `run_post_session`. This is a denial-of-service: one bad file in the shared dir can bring down a slot's processing. The fix is to catch `(OSError, UnicodeDecodeError)` or use `errors='replace'`.

packages/gptme-runloops/src/gptme_runloops/run_item.py
```diff
@@ -2311,6 +2386,36 @@ def run_post_session(
...
+        and set(item.types) != {"merge_ready"}
+        and not (delivery_verified and delivery_outcome == "handled")
+    ):
+        # The delivery check did not verify a reply. Count this through the
+        # same bounded rollback as orphan_no_delivery so a permanently broken
+        # check cannot re-emit this thread forever. PR-side state is still
+        # valid and promotes on every pass; only mapped notification state is
+        # purged while the retry budget remains.
+        rollback_slot_key = resolve_slot_key(config, item)
+        if rollback_failed_delivery(config, item.repo, item.number, rollback_slot_key):
```
TimeToBuildBob: ⚠️ **P2** — The new `elif` branch in `run_post_session` handles unverified delivery by calling `rollback_failed_delivery`, which unconditionally clears the slot's `.event`/`.event_logged` fingerprints via `clear_slot_event_markers`. The PR description explicitly states these markers are to be left in place for `pm_dispatch_recovery`'s backed-off, budgeted re-arm. Clearing them means the next dispatch cycle re-evaluates the item immediately (subject only to the cooldown), bypassing the recovery backoff that would otherwise throttle re-attempts. For a persistently broken delivery check, this can cause repeated immediate re-dispatches, wasting slot time even though the redelivery cap eventually stops them. The implementation diverges from the stated design; either the code should not clear the markers in this path, or the description is wrong.

packages/gptme-runloops/src/gptme_runloops/run_item.py
TimeToBuildBob: Fixed in 20b25d26. Both notification-map readers now ignore invalid UTF-8 sidecars instead of aborting post-session state handling. Added a regression test with a corrupt map next to a valid mapped notification.

packages/gptme-runloops/src/gptme_runloops/run_item.py
TimeToBuildBob: Fixed in 20b25d26. The unverified-delivery path now purges only its mapped notification state and preserves the slot event marker, leaving re-arm timing to pm_dispatch_recovery. The explicit orphan path keeps immediate marker clearing. The retry-cap test now pins marker preservation.

packages/gptme-runloops/src/gptme_runloops/run_item.py
```diff
@@ -2311,6 +2392,42 @@ def run_post_session(
...
+        # check cannot re-emit this thread forever. PR-side state is still
+        # valid and promotes on every pass; only mapped notification state is
+        # purged while the retry budget remains.
+        rollback_slot_key = resolve_slot_key(config, item)
+        if rollback_failed_delivery(
+            config,
+            item.repo,
+            item.number,
+            rollback_slot_key,
+            clear_event_markers=False,
```
TimeToBuildBob: ⚠️ **P2** — The new `elif` branch in `run_post_session` passes `clear_event_markers=False` to `rollback_failed_delivery` for every unverified delivery, including when the delivery check exits non-zero (fallback outcome 'handled'). Unlike the `orphan_no_delivery` branch, this leaves the slot's `.event` fingerprint in the cooldown dir. If the item in fact received no reply, it stays suppressed for the 6h TTL until the redelivery cap is exhausted, so a genuinely unanswered mention can still be dropped for up to the cap duration rather than being re-evaluated on the next cycle.

TimeToBuildBob: Fixed in dcfd7f7c. An unverified delivery now clears the launch `.event` marker, so a genuinely unanswered mention re-enters after the normal `.ts` dispatch cooldown instead of remaining suppressed for the full event TTL. The bounded redelivery counter still caps persistent failures. Updated the regression test to pin marker removal; 176 run-item tests pass.

TimeToBuildBob:
## Convergence adjudication — final

**Fixed**
- **P1 — missing delivery hook could purge a delivered notification:** fixed in `f9019bbe`; rollback now requires an actual delivery-check hook, with regression coverage.
- **P1 — permissive parsing treated a malformed check result as verified:** fixed in `2801c060`; only a successful check returning a known outcome counts as delivery proof.
- **P1 — unverified rollback reset its own retry counter:** fixed in `bda67023`; retries retain the counter and terminate at `PM_MAX_REDELIVERY_ATTEMPTS`.
- **P1 — rollback applied to item types that do not owe a thread reply:** fixed in `0d452105`; the branch is restricted to nonzero, thread-deliverable items and excludes pure `merge_ready`.
- **P1 — corrupt UTF-8 notification maps could abort state handling:** fixed in `20b25d26`; both map readers now skip corrupt sidecars, covered by a valid-neighbor regression test.
- **P2 — event-marker policy oscillated between immediate retry and six-hour suppression:** fixed in `dcfd7f7c`; failed delivery clears `.event` while retaining `.ts`, so the normal dispatch cooldown gates re-entry and the bounded redelivery counter caps persistent failures. `1e453e99` removes the now-impossible optional marker-preservation mode and its stale `pm_dispatch_recovery` documentation, preventing this contradiction from recurring.

**Remaining**
- No blocking bug, security finding, nitpick, or false positive remains. The detector reports `0` unresolved P1/P2/P3 findings. I am not requesting another review; the attempt cap is final.

**CI**
- Previous head `dcfd7f7c`: all checks green.
- Current head `1e453e99`: package tests on Python 3.10/3.11/3.12 and integration tests are green; coverage, pre-commit, and typecheck are still running. Local verification: `176 passed`, Ruff clean, package mypy clean, scoped pre-commit hooks passed, and `git diff --check` passed.

**Domain risk**
- The fragile surface is the boundary between per-item notification ownership and two independent retry controls. A maintainer manual check should focus on: a failed thread delivery re-enters only after `.ts` cooldown; repeated failed checks stop at `PM_MAX_REDELIVERY_ATTEMPTS`; pure `merge_ready` and `master_ci_failure` never consume that thread-delivery budget.

**Convergence**
- Detector `round_convergence`: `status=no_history`, `stable_rounds=0`, `required_stable_rounds=2`. It does not parse this PR's Bob AI-review history; the review summary records seven passes, while the detector's actionable finding count is zero.

Stopping at maintainer judgment. No merge and no further review trigger.

packages/gptme-runloops/src/gptme_runloops/run_item.py
```diff
@@ -1001,8 +1068,11 @@ def promote_notification_states(config: RunItemConfig) -> None:
         return
     config.state_dir.mkdir(parents=True, exist_ok=True)
     for f in pending.glob("notif-*.state"):
-        if f.is_file():
-            shutil.copy(f, config.state_dir / f.name)
+        if not f.is_file():
+            continue
+        if f.with_suffix(".map").is_file():
```
TimeToBuildBob: ⚠️ **P2** — promote_notification_states() skips any notif-*.state that has a .map sidecar, regardless of whether that .map is readable. A corrupt/undecodable map (e.g., invalid UTF-8) prevents the state from ever being promoted, but purge_pending_notif_state() and _mapped_notif_state_files() also skip it (they catch UnicodeDecodeError and continue). So the state file sits in the pending dir indefinitely: it is neither promoted nor purged. On the next sweep the gate still sees the notification as unread (since its state is still pending) and re-emits it, which can produce duplicate replies for a thread that was actually handled. The test test_corrupt_notification_map_is_ignored only exercises promote_item_state() and purge_pending_notif_state(), not promote_notification_states(), so this path is uncovered.

TimeToBuildBob: Fixed in 950a75ca. `promote_notification_states()` now validates existing map sidecars: invalid UTF-8 cannot establish ownership, so the corresponding notification state is promoted as unmapped instead of being stranded; transient I/O failures remain fail-closed. The regression test now covers the full promote-item, purge, and end-of-run promotion sequence. Local verification: 131 tests passed, Ruff clean, mypy clean, and pre-commit passed.

packages/gptme-runloops/src/gptme_runloops/run_item.py
```diff
@@ -1001,8 +1069,19 @@ def promote_notification_states(config: RunItemConfig) -> None:
...
+        map_file = f.with_suffix(".map")
+        if map_file.is_file():
+            try:
+                map_file.read_text(encoding="utf-8")
+            except UnicodeDecodeError:
+                pass  # corrupt map cannot prove item ownership
+            except OSError:
+                continue  # transient read failure: preserve the safer ownership gate
+            else:
+                continue  # owned by an item; its worker's outcome decides
```
TimeToBuildBob, line 1084: ⚠️ **P2** — In `promote_notification_states`, a `.map` sidecar that is readable but contains an empty string is treated as owned (the `else: continue` branch), so the corresponding `notif-*.state` is never promoted. An empty map cannot establish a `repo#number` ownership, so this strands the state in the pending dir forever: the gate never sees the notification as handled (it is not copied to the real state dir) and never re-emits it (every subsequent blanket pass also skips it because the map exists and is readable). The code already treats undecodable maps as unmapped via `except UnicodeDecodeError: pass`, but an empty string is decodable and falls into the `else` branch. This creates a permanent deadlock where a notification is neither marked handled nor re-dispatched. The fix is to read the map content and treat an empty (or whitespace-only) value as unmapped, similar to the undecodable case.

TimeToBuildBob: Fixed in `124b16a3`. `promote_notification_states()` now strips the map content and accepts ownership only when the result is non-empty, so empty and whitespace-only sidecars are promoted as unmapped. The regression is parameterized across invalid UTF-8, empty, and whitespace-only maps. Verification: 133 tests passed, Ruff clean, mypy clean, scoped pre-commit passed, and the pushed head is `124b16a3`.

TimeToBuildBob: Fixed in `48a67151`. `promote_notification_states()` now treats `OSError` exactly like invalid UTF-8: an unreadable map cannot prove item ownership, so its state is promoted as unmapped instead of stranded. Added a regression that injects a map read failure. Local verification: 179 focused run-item tests passed, Ruff clean, mypy clean, and pre-commit passed.

packages/gptme-runloops/src/gptme_runloops/run_item.py
```diff
@@ -1001,8 +1069,18 @@ def promote_notification_states(config: RunItemConfig) -> None:
...
-        if f.is_file():
-            shutil.copy(f, config.state_dir / f.name)
+        if not f.is_file():
+            continue
+        map_file = f.with_suffix(".map")
+        if map_file.is_file():
+            try:
+                owner = map_file.read_text(encoding="utf-8").strip()
+            except (OSError, UnicodeDecodeError):
+                pass  # unreadable map cannot prove item ownership
```
TimeToBuildBob: ⚠️ **P2** — In `promote_notification_states`, a map file that cannot be read (OSError or UnicodeDecodeError) is treated as unmapped, causing its associated state file to be promoted (copied to the real state dir) and thus the notification to be considered handled. This can prematurely consume a notification that belongs to an item whose map is temporarily unreadable (e.g., a transient filesystem error, or a partially written map file during a race). If that item is actually still pending, its thread will never be re-emitted because the gate dedupes on updated_at, effectively dropping a legitimate mention. The comment explicitly chooses this fallback to avoid stranding state forever, but the trade-off is a silent lost notification in exactly the scenario the PR is trying to fix. The safer behavior would be to leave the state pending for a later sweep when the map is unreadable, accepting the risk of a stranded notification that rsync will eventually clean. _(2/3 passes — pass 1 looked and disagreed)_

packages/gptme-runloops/src/gptme_runloops/run_item.py
```diff
@@ -974,25 +1008,59 @@ def promote_item_state(
...
+        # state either.
+        patterns = []
+        for f in pending.glob("notif-*.state"):
+            if f.is_file() and not f.with_suffix(".map").is_file():
+                shutil.copy(f, state / f.name)
     else:
         patterns = [
             f"{repo_safe}-pr-{number}*.state",
             f"{repo_safe}-issue-{number}*.state",
+            f"{repo_safe}-master-ci.state",
```
TimeToBuildBob: ❌ **P1** — In promote_item_state, the pattern list for non-zero items includes '{repo_safe}-master-ci.state', which is a repo-wide file, not specific to the item's number. This means any item with a non-zero number in the repo (e.g., a pr_update) will copy the pending master-ci state file to the real state dir. If a master_ci_failure item for the same repo is still pending, its state is now present in the real dir, so the activity gate will consider the master CI handled and skip dispatching the failure. The consequence is a master_ci_failure item that is never evaluated. The pattern should be limited to items of type master_ci_failure, or the filename should be number-specific (e.g., {repo_safe}-master-ci-{number}.state) to avoid cross-item consumption.

packages/gptme-runloops/src/gptme_runloops/run_item.py
```diff
@@ -1001,8 +1069,18 @@ def promote_notification_states(config: RunItemConfig) -> None:
...
     for f in pending.glob("notif-*.state"):
-        if f.is_file():
-            shutil.copy(f, config.state_dir / f.name)
+        if not f.is_file():
+            continue
+        map_file = f.with_suffix(".map")
+        if map_file.is_file():
+            try:
+                owner = map_file.read_text(encoding="utf-8").strip()
+            except (OSError, UnicodeDecodeError):
```
TimeToBuildBob, line 1081: ⚠️ **P2** — When a notif-*.map file is unreadable (e.g., a transient read error or invalid UTF-8), purge_pending_notif_state skips it (the except continues), leaving the corresponding notif-*.state in the pending dir. On the same run, promote_notification_states also fails to read the map, treats the state as unmapped, and copies it to the real state dir. This consumes the notification state exactly at the point the rollback intended to preserve it for re-emission. The result is the same symptom the PR fixes: a thread that was not handled is marked handled and never re-emitted. Since both functions treat unreadable maps as unmapped, the rollback cannot guarantee re-emission when the map is unreadable.

packages/gptme-runloops/src/gptme_runloops/run_item.py
TimeToBuildBob: Fixed in `2eb927bc`. `promote_item_state()` no longer includes the repo-wide `*-master-ci.state` for ordinary numbered PR/issue items. `run_post_session()` opts into that file only when the dispatched item has type `master_ci_failure`. Regression coverage now proves both the ordinary-item exclusion and explicit master-CI promotion path.

packages/gptme-runloops/src/gptme_runloops/run_item.py
TimeToBuildBob: Fixed in `2eb927bc`. An unreadable notification map now leaves its state pending: unknown ownership fails closed instead of letting the blanket pass consume a potentially unhandled notification. Readable empty/whitespace maps remain genuinely unmapped and are promoted. The regression injects an `OSError` and verifies both non-promotion and pending-state preservation.

TimeToBuildBob:
## Convergence adjudication — cap reached

**Fixed**
- **P1 — ordinary numbered items consumed repo-wide master-CI state:** fixed in `2eb927bc`. `promote_item_state()` now excludes `*-master-ci.state` by default; `run_post_session()` opts in only for `master_ci_failure`. Tests cover both exclusion and explicit promotion.
- **P2 — unreadable map could be promoted after rollback:** fixed in `2eb927bc`. Unknown ownership now fails closed: `OSError`/invalid UTF-8 leaves the notification state pending rather than marking a potentially unhandled thread handled. Readable empty/whitespace maps remain unmapped. A regression injects the read failure and checks pending-state preservation.

**Remaining**
- None. All review threads are resolved; there are no remaining blocking bugs, security findings, nits, or false positives to dismiss.

**CI**
- Current head `2eb927bc`: Python 3.10/3.11/3.12 package tests, integration tests, typecheck, install-size, and change detection are green; coverage and pre-commit are still pending.
- Local verification: `134` focused run-item tests + `45` rollback tests passed; Ruff and package mypy are clean; `git diff --check` passed. The scoped pre-commit run reached and passed typecheck before the local 120s command limit.

**Domain risk**
- Fragile area: pending-state ownership under transient sidecar read failure. Maintainer manual smoke should confirm an unreadable mapped notification remains eligible after the next sweep, while a real `master_ci_failure` still promotes the repo-wide state file and an unrelated PR worker cannot consume it.

**Convergence**
- Detector `round_convergence`: `status=no_history`, `stable_rounds=0`, `required_stable_rounds=2`. It does not parse this PR's AI-review history; the summary records 11 review rounds, and the detector now reports zero unresolved findings.

Stopping at maintainer judgment. No merge and no further review trigger.

packages/gptme-runloops/src/gptme_runloops/run_item.py
```diff
@@ -2311,8 +2397,51 @@ def run_post_session(
...
+        # remains.
+        rollback_slot_key = resolve_slot_key(config, item)
+        if rollback_failed_delivery(
+            config,
+            item.repo,
+            item.number,
+            rollback_slot_key,
+        ):
+            promote_item_state(
+                config, item.repo, item.number, reset_redelivery_counter=False
```
TimeToBuildBob: ⚠️ **P2** — In the new rollback branch for unverified delivery (the elif at step 8), promote_item_state is called with reset_redelivery_counter=False but without include_master_ci. If an item carries both a thread-deliverable type (e.g. pr_update) and master_ci_failure, the master-CI state file (like <repo>-master-ci.state) will not be copied to the real state dir, because include_master_ci defaults to False. That leaves the master-CI state pending, so the activity gate re-emits the master-CI failure even though the PR-side handling succeeded and is being promoted. The PR description explicitly states that PR-side state (which includes master-CI state) should promote on every pass; this call contradicts that intent for mixed items.

ErikBjare:
@TimeToBuildBob fix

TimeToBuildBob:
Fixed in 7539cdbb.

The rollback branch (step 8, unverified delivery) was calling `promote_item_state` without `include_master_ci`. For a mixed `pr_update` + `master_ci_failure` item, the master-CI state file stayed pending and the activity gate re-emitted the failure on the next sweep even though PR-side handling succeeded.

Fix: pass `include_master_ci="master_ci_failure" in item.types` in the rollback branch, mirroring the success path below.

Regression test (`test_post_session_unverified_rollback_promotes_master_ci_state`): seeds a mixed-type item with pending master-CI state, triggers rollback via a failed delivery check, asserts the master-CI file lands in the real state dir while mapped notif state is purged. 135 tests pass. CI running.

TimeToBuildBob:
Fixed in 07049e0e.

The two redelivery-cap-reached branches in `run_post_session` (orphan_no_delivery and unverified-delivery) called `promote_item_state` without `include_master_ci`. For a mixed `pr_update + master_ci_failure` item this left the master-CI state file in the pending dir, so the activity gate kept re-emitting the master-CI failure on every sweep despite the cap having been reached.

Added `include_master_ci="master_ci_failure" in item.types` to both cap-reached branches (matching the success path and rollback-success path already introduced in this PR). Also adds two regression tests — one per branch — that were red against the previous code and green after.

packages/gptme-runloops/src/gptme_runloops/run_item.py
TimeToBuildBob: Fixed in `07049e0e`. The unverified-delivery and cap-reached rollback paths now pass `include_master_ci=True` to `promote_item_state`, ensuring mixed `pr_update + master_ci_failure` items promote the master-CI state file on the PR-side success path.

TimeToBuildBob:
Fixed ruff-format CI failure in `eaec024e`. The pre-commit hook (pinned at v0.6.9) needed a line-wrap in `test_run_item.py`. CI re-running now.

TimeToBuildBob:
@TimeToBuildBob review

packages/gptme-runloops/src/gptme_runloops/run_item.py
```diff
@@ -2310,9 +2396,66 @@ def run_post_session(
...
+        # The delivery check did not verify a reply. Count this through the
+        # same bounded rollback as orphan_no_delivery so a permanently broken
+        # check cannot re-emit this thread forever. Clear the launch marker so
+        # the item can re-enter after the dispatcher's normal .ts cooldown;
+        # leaving it stamped suppresses a genuinely unanswered mention for the
+        # full event TTL. PR-side state is still valid and promotes on every
+        # pass; only mapped notification state is purged while the retry budget
+        # remains.
+        rollback_slot_key = resolve_slot_key(config, item)
+        if rollback_failed_delivery(
```
TimeToBuildBob, line 2430: ⚠️ **P2** — The new `elif` branch in `run_post_session` (step 8) calls `rollback_failed_delivery` for any delivery check that ran but did not verify a reply (non-zero exit, malformed output, or unknown outcome). When `PM_DISPATCH_COOLDOWN_DIR` is not set, `redelivery_attempts_file` returns None, so `rollback_failed_delivery` skips its attempt-counting block entirely and always returns True. The item is therefore rolled back and re-enters the dispatch queue on every cycle with no cap, repeatedly re-emitting its notification. Previously this unbounded retry existed only for `orphan_no_delivery`; the new branch enlarges the set of failures that trigger it (e.g., a permanently broken delivery check). In an environment where the cooldown dir is unset or misconfigured, this causes endless re-dispatch churn and notification spam. The cap is only enforced when the cooldown dir exists.

TimeToBuildBob:
Fixed P2 finding in 33c33353: added `resolve_cooldown_dir(config) is not None` guard to the `elif` condition in `run_post_session` (step 8). When `PM_DISPATCH_COOLDOWN_DIR` is not set the guard evaluates False and we fall through to `else: promote_item_state` — bounded and safe. Also adds a regression test covering this path. AI reviewer will re-evaluate on the next push.

PR #1478, opened 2026-08-20
feat(gptmail): add `agent pull` command for pull-only recipients + version bump

### LINKED ISSUE #1476: gptmail agent: no delivery/surfacing path for pull-only recipients — Erik missed time-sensitive requests
## Problem

`gptmail agent` supports pull-only recipients (e.g. `erik: {delivery: pull-only}` in `agents.yaml`): agents can `send`/`reply` to them cleanly, but the messages just accumulate in each sender's `messages/outbox/`. There is **no delivery, pull, or surfacing mechanism** on the recipient side — a human recipient only sees messages if they remember to manually poll every agent's outbox over SSH.

### Concrete cost

- Gordon's outbox accumulated **~68 unpulled to-erik messages** (late June → Aug 20) before anyone noticed.
- Among them a time-boxed decision request (RDDT buy, reply-by Aug 14, no-reply=no-trade) that Erik never saw — the window lapsed silently. Probably a mistake in hindsight.
- Multiple "escalations" (CB/KLAC graduation asks, escalated 3×+) were really just re-sends into a mailbox nobody was reading.

The fleet-side half of this works fine; the human-side half is missing entirely.

## Ask

1. **`gptmail agent pull`** — for a pull-only identity: iterate the `agents.yaml` registry, fetch `to: <me>` messages from each agent's remote outbox (SSH/SCP, skipping `.archive/` and `drafts/`), dedupe into the local `messages/inbox/`. Exit code / stdout should make "N new messages" machine-readable.
2. **Notify hook** — either a `--notify-cmd '<cmd>'` option (invoked with count/summary when new messages arrive) or documented cron/launchd recipe, so new messages surface as OS notifications instead of waiting for a manual prompt.
3. Consider `gptmail agent pending --as <recipient>`-style view so a pull-only recipient can see what's awaiting *their* reply across mailboxes.

## Related papercut: version is frozen at 0.1.0

Erik's installed gptmail predated the read-persistence fixes (missing-`read:`-key handling in `_mark_read`), silently resurfacing the "read is a no-op" bug on old-convention frontmatter (`status: sent`, non-ISO timestamps). Because the package version never moves from 0.1.0, there's no signal that an install is stale. Please bump the version on behavioral changes (or add `gptmail --version` surfacing a git rev).

## Reference implementation

Erik's laptop now runs a local workaround: `~/.local/bin/pull-agent-msgs` (bash: registry loop → SSH grep for `^to: erik` → SCP missing files → `osascript` notification) on a 30-min launchd interval (`com.erik.agent-msg-pull`). Feel free to lift it into the package as the `pull` subcommand.

cc @TimeToBuildBob

TimeToBuildBob:
Fixes #1476.

## Problem

Pull-only recipients (e.g. humans registered as `delivery: pull-only` in `agents.yaml`) had no mechanism to receive messages. Outbox messages accumulated silently — Gordon's outbox had 68 unpulled messages over ~2 months, including a time-boxed decision request that lapsed.

### `gptmail agent pull` (new command)

Iterates `agents.yaml`, SSHes into every push-reachable agent, and fetches outbox messages addressed to `AGENT_NAME` into the local inbox. Safe to run repeatedly — deduplicates by filename.

```
# Fetch new messages
gptmail agent pull

# Machine-readable output (for cron/notification scripts)
gptmail agent pull --json
# -> {"new_count": 3, "files": [...], "agents_polled": [...]}

# macOS notification banner
gptmail agent pull --notify-cmd 'osascript -e "display notification \"$SUMMARY\" with title \"gptmail\""'

# Linux (libnotify)
gptmail agent pull --notify-cmd 'notify-send "gptmail" "$SUMMARY"'

# Pull as a different identity (e.g. from a laptop)
gptmail agent pull --as erik

# Preview without writing
gptmail agent pull --dry-run
```

**Options:**
- `--as IDENTITY` — override the effective recipient name (useful for human recipients pulling from a different host than their registered `AGENT_NAME`)
- `--json` — emit `{new_count, files, agents_polled}` for scripted callers
- `--notify-cmd CMD` — shell hook invoked with `NEW_COUNT` + `SUMMARY` env vars when new messages arrive
- `--dry-run` — show what would be fetched without writing

Per-agent SSH errors are logged as warnings; partial results are still returned so one unreachable host doesn't abort the whole pull.

### `gptmail --version` (new)

Stale installs are now self-diagnosable: `gptmail --version` → `gptmail, version 0.2.0`.

### Version bump: 0.1.0 → 0.2.0

Behavioral changes (new pull command, version surfacing) warrant a minor bump so consumers have a signal to upgrade.

## Tests

11 new tests in `packages/gptmail/tests/test_agent_pull.py`, all green. Full suite: **178 passed**.

Covers: fetch addressed messages, skip messages for others, dedup on re-pull, `--json` output shape, zero-count when empty, `--notify-cmd` invoked with correct env vars, `--notify-cmd` NOT invoked when empty, `--dry-run` writes nothing, pull-only agents skipped, `--as` override, multiple messages in one run.

### DIFF
--- a/packages/gptmail/src/gptmail/agent_cli.py
+++ b/packages/gptmail/src/gptmail/agent_cli.py
@@ -678,6 +678,7 @@ def _fleet_pending_rows(
     self_name: str,
     agents: dict[str, dict[str, str]],
     fleet: bool,
+    all_mailboxes: bool,
 ) -> list[dict]:
     rows = _outbox_rows_for_recipient(mailboxes, recipient=recipient, self_name=self_name)
     if not fleet:
@@ -691,7 +692,7 @@ def _fleet_pending_rows(
                 agent,
                 recipient=recipient,
                 mailboxes=mailboxes,
-                all_mailboxes=len(mailboxes) > 1,
+                all_mailboxes=all_mailboxes,
             )
         )
     rows.sort(
@@ -924,6 +925,7 @@ def pending(
             self_name=self_name,
             agents=agents,
             fleet=fleet and not local_only,
+            all_mailboxes=all_mailboxes,
         )
         if json_output:
             click.echo(json.dumps(rows, indent=2, sort_keys=True, default=str))
@@ -1215,7 +1217,7 @@ def pull(
         env["SUMMARY"] = f"{n} new message(s): {subjects}"
         try:
             subprocess.run(shlex.split(notify_cmd), env=env, timeout=10, check=True)
-        except (subprocess.CalledProcessError, subprocess.TimeoutExpired) as e:
+        except (OSError, subprocess.CalledProcessError, subprocess.TimeoutExpired) as e:
             click.echo(f"Warning: --notify-cmd failed: {e}", err=True)
 
 
--- a/packages/gptmail/src/gptmail/agent_cli.py
+++ b/packages/gptmail/src/gptmail/agent_cli.py
@@ -1052,7 +1052,7 @@ def _fetch_from_agent(
         temp_dest = Path(temp_name)
         try:
             subprocess.run(
-                ["scp", *ssh_opts, f"{ssh_target}:{remote_path}", str(temp_dest)],
+                ["scp", *ssh_opts, f"{ssh_target}:{shlex.quote(remote_path)}", str(temp_dest)],
                 check=True,
                 capture_output=True,
                 timeout=15,
@@ -1062,7 +1062,7 @@ def _fetch_from_agent(
             except FileExistsError:
                 continue  # A concurrent pull published this message first.
             new_files.append(dest)
-        except (subprocess.CalledProcessError, subprocess.TimeoutExpired) as e:
+        except (OSError, subprocess.CalledProcessError, subprocess.TimeoutExpired) as e:
             click.echo(f"Warning: failed to fetch {filename} from {agent_name}: {e}", err=True)
         finally:
             temp_dest.unlink(missing_ok=True)
@@ -1219,7 +1219,7 @@ def pull(
         env["SUMMARY"] = f"{n} new message(s): {subjects}"
         try:
             subprocess.run(shlex.split(notify_cmd), env=env, timeout=10, check=True)
-        except (OSError, subprocess.CalledProcessError, subprocess.TimeoutExpired) as e:
+        except (OSError, ValueError, subprocess.CalledProcessError, subprocess.TimeoutExpired) as e:
             click.echo(f"Warning: --notify-cmd failed: {e}", err=True)
 
 
--- a/packages/gptmail/src/gptmail/agent_cli.py
+++ b/packages/gptmail/src/gptmail/agent_cli.py
@@ -1128,17 +1128,14 @@ def pull(
 
     new_files: list[Path] = []
     agents_polled: list[str] = []
-    agents_skipped: list[str] = []
 
     for agent_name, agent_cfg in agents.items():
         if agent_name == self_name:
             continue
         # Skip pull-only or incompletely configured agents — they can't be SSHed.
         if agent_cfg.get("delivery") == "pull-only":
-            agents_skipped.append(agent_name)
             continue
         if not all(agent_cfg.get(k) for k in ("ssh", "workspace")):
-            agents_skipped.append(agent_name)
             continue
 
         agents_polled.append(agent_name)
@@ -1186,31 +1183,20 @@ def pull(
 
     n = len(new_files)
 
-    if json_output:
-        payload = {
-            "new_count": n,
-            "files": [f.name for f in new_files],
-            "agents_polled": agents_polled,
-        }
-        click.echo(json.dumps(payload))
-        return
+    if not json_output and not dry_run:
+        if n == 0:
+            click.echo("No new messages.")
+        else:
+            click.echo(f"{n} new message(s) fetched into inbox:")
+            for f in new_files:
+                meta = meta_of(f) or {}
+                sender = str(meta.get("from", "unknown"))
+                subject = str(meta.get("subject", "(no subject)"))
+                ts = str(meta.get("timestamp", ""))
+                ts_part = f" [{ts}]" if ts else ""
+                click.echo(f"  {sender}{ts_part}: {subject}  ({f.name})")
 
-    if dry_run:
-        return
-
-    if n == 0:
-        click.echo("No new messages.")
-    else:
-        click.echo(f"{n} new message(s) fetched into inbox:")
-        for f in new_files:
-            meta = meta_of(f) or {}
-            sender = str(meta.get("from", "unknown"))
-            subject = str(meta.get("subject", "(no subject)"))
-            ts = str(meta.get("timestamp", ""))
-            ts_part = f" [{ts}]" if ts else ""
-            click.echo(f"  {sender}{ts_part}: {subject}  ({f.name})")
-
-    if notify_cmd and n > 0:
+    if notify_cmd and n > 0 and not dry_run:
         subjects = "; ".join(
             str((meta_of(f) or {}).get("subject", "(no subject)")) for f in new_files
         )
@@ -1221,6 +1207,14 @@ def pull(
             subprocess.run(shlex.split(notify_cmd), env=env, timeout=10, check=True)
         except (OSError, ValueError, subprocess.CalledProcessError, subprocess.TimeoutExpired) as e:
             click.echo(f"Warning: --notify-cmd failed: {e}", err=True)
+
+    if json_output:
+        payload = {
+            "new_count": n,
+            "files": [f.name for f in new_files],
+            "agents_polled": agents_polled,
+        }
+        click.echo(json.dumps(payload))
 
 
 @agent.command()
--- a/packages/gptmail/src/gptmail/agent_cli.py
+++ b/packages/gptmail/src/gptmail/agent_cli.py
@@ -620,6 +620,7 @@ def _remote_pending_rows(
     *,
     recipient: str,
     mailboxes: list[str],
+    all_mailboxes: bool,
 ) -> list[dict]:
     missing = [k for k in ("ssh", "workspace") if not agent.get(k)]
     if missing:
@@ -629,10 +630,10 @@ def _remote_pending_rows(
         )
         return []
     cmd = ["uv", "run", "gptmail", "agent", "pending", "--for", recipient, "--json", "--local-only"]
-    if len(mailboxes) == 1:
-        cmd.extend(["--mailbox", mailboxes[0]])
-    else:
+    if all_mailboxes:
         cmd.append("--all-mailboxes")
+    else:
+        cmd.extend(["--mailbox", mailboxes[0]])
     remote_cmd = " && ".join(
         [
             f"cd {shlex.quote(agent['workspace'])}",
@@ -689,6 +690,7 @@ def _fleet_pending_rows(
                 agent,
                 recipient=recipient,
                 mailboxes=mailboxes,
+                all_mailboxes=len(mailboxes) > 1,
             )
         )
     rows.sort(
@@ -985,6 +987,8 @@ def _fetch_from_agent(
     *,
     self_name: str,
     mailboxes: list[str],
+    all_mailboxes: bool,
+    fallback_mailbox: str | None = None,
 ) -> list[Path]:
     """SCP outbox messages addressed to ``self_name`` from a remote agent's outbox.
 
@@ -992,7 +996,14 @@ def _fetch_from_agent(
     Logs SSH/SCP errors as warnings so one unreachable agent doesn't abort the
     whole pull — the caller collects partial results and reports the total.
     """
-    rows = _remote_pending_rows(agent_name, agent, recipient=self_name, mailboxes=mailboxes)
+    rows = _remote_pending_rows(
+        agent_name,
+        agent,
+        recipient=self_name,
+        mailboxes=mailboxes,
+        all_mailboxes=all_mailboxes,
+    )
+    fallback_mailbox = fallback_mailbox or (mailboxes[0] if mailboxes else "default")
     ssh_target = agent["ssh"]
     workspace = agent["workspace"]
     ctl_path = _ssh_control_path(ssh_target)
@@ -1012,7 +1023,7 @@ def _fetch_from_agent(
     for row in rows:
         filename = str(row.get("file") or "")
         try:
-            dest = _pull_destination(row, fallback_mailbox=mailboxes[0])
+            dest = _pull_destination(row, fallback_mailbox=fallback_mailbox)
         except click.BadParameter as e:
             click.echo(f"Warning: refusing invalid mailbox from {agent_name}: {e}", err=True)
             continue
@@ -1026,7 +1037,7 @@ def _fetch_from_agent(
             continue  # Already fetched — deduplicate by mailbox and filename.
         dest.parent.mkdir(parents=True, exist_ok=True)
         # The row's mailbox field tells us which remote outbox subfolder to look in.
-        row_mailbox = _normalize_mailbox(str(row.get("mailbox") or mailboxes[0]))
+        row_mailbox = _normalize_mailbox(str(row.get("mailbox") or fallback_mailbox))
         if row_mailbox == "default":
             remote_path = f"{workspace}/messages/outbox/{filename}"
         else:
@@ -1040,6 +1051,7 @@ def _fetch_from_agent(
             )
             new_files.append(dest)
         except (subprocess.CalledProcessError, subprocess.TimeoutExpired) as e:
+            dest.unlink(missing_ok=True)
             click.echo(f"Warning: failed to fetch {filename} from {agent_name}: {e}", err=True)
     return new_files
 
@@ -1062,7 +1074,7 @@ def _fetch_from_agent(
     default=None,
     metavar="CMD",
     help=(
-        "Shell command to invoke when new messages arrive. "
+        "Command to invoke when new messages arrive. "
         "Receives NEW_COUNT (integer) and SUMMARY (human-readable string) as environment variables."
     ),
 )
@@ -1093,14 +1105,13 @@ def pull(
     Notification hook:
 
     \b
-        # macOS — post a Notification Center banner
-        gptmail agent pull --notify-cmd 'osascript -e "display notification \\\"$SUMMARY\\\" with title \\\"gptmail\\\""'
-        # Linux — libnotify
-        gptmail agent pull --notify-cmd 'notify-send "gptmail" "$SUMMARY"'
+        # Use an explicit shell when environment expansion is wanted
+        gptmail agent pull --notify-cmd "sh -c 'notify-send gptmail \"$SUMMARY\"'"
     """
     agents = _load_agents()
     self_name = (as_recipient or _self_name()).lower()
     mailboxes = _selected_mailboxes(mailbox, all_mailboxes)
+    remote_mailboxes = [] if all_mailboxes else mailboxes
 
     new_files: list[Path] = []
     agents_polled: list[str] = []
@@ -1120,7 +1131,11 @@ def pull(
         agents_polled.append(agent_name)
         if dry_run:
             rows = _remote_pending_rows(
-                agent_name, agent_cfg, recipient=self_name, mailboxes=mailboxes
+                agent_name,
+                agent_cfg,
+                recipient=self_name,
+                mailboxes=remote_mailboxes,
+                all_mailboxes=all_mailboxes,
             )
             for row in rows:
                 filename = str(row.get("file") or "")
@@ -1147,7 +1162,8 @@ def pull(
             agent_name,
             agent_cfg,
             self_name=self_name,
-            mailboxes=mailboxes,
+            mailboxes=remote_mailboxes,
+            all_mailboxes=all_mailboxes,
         )
         new_files.extend(fetched)
 
@@ -1185,9 +1201,7 @@ def pull(
         env["NEW_COUNT"] = str(n)
         env["SUMMARY"] = f"{n} new message(s): {subjects}"
         try:
-            subprocess.run(  # noqa: S602  (shell=True is intentional — user-supplied hook)
-                notify_cmd, shell=True, env=env, timeout=10
-            )
+            subprocess.run(shlex.split(notify_cmd), env=env, timeout=10, check=True)
         except (subprocess.CalledProcessError, subprocess.TimeoutExpired) as e:
             click.echo(f"Warning: --notify-cmd failed: {e}", err=True)
 
--- a/packages/gptmail/src/gptmail/agent_cli.py
+++ b/packages/gptmail/src/gptmail/agent_cli.py
@@ -969,6 +969,206 @@ def pending(
         click.echo(_stale_summary(stale))
 
 
+def _fetch_from_agent(
+    agent_name: str,
+    agent: dict[str, str],
+    *,
+    self_name: str,
+    mailboxes: list[str],
+    local_inbox: Path,
+) -> list[Path]:
+    """SCP outbox messages addressed to ``self_name`` from a remote agent's outbox.
+
+    Returns a list of newly-written local inbox paths (skips files already present).
+    Logs SSH/SCP errors as warnings so one unreachable agent doesn't abort the
+    whole pull — the caller collects partial results and reports the total.
+    """
+    rows = _remote_pending_rows(agent_name, agent, recipient=self_name, mailboxes=mailboxes)
+    ssh_target = agent["ssh"]
+    workspace = agent["workspace"]
+    ctl_path = _ssh_control_path(ssh_target)
+    ssh_opts = [
+        "-o",
+        "ConnectTimeout=5",
+        "-o",
+        "BatchMode=yes",
+        "-o",
+        "ControlMaster=auto",
+        "-o",
+        f"ControlPath={ctl_path}",
+        "-o",
+        "ControlPersist=60",
+    ]
+    new_files: list[Path] = []
+    for row in rows:
+        filename = str(row.get("file") or "")
+        if not filename:
+            continue
+        dest = local_inbox / filename
+        if dest.exists():
+            continue  # Already fetched — deduplicate by filename.
+        # The row's mailbox field tells us which remote outbox subfolder to look in.
+        row_mailbox = str(row.get("mailbox") or "default")
+        if row_mailbox == "default":
+            remote_path = f"{workspace}/messages/outbox/{filename}"
+        else:
+            remote_path = f"{workspace}/messages/mailboxes/{row_mailbox}/outbox/{filename}"
+        try:
+            subprocess.run(
+                ["scp", *ssh_opts, f"{ssh_target}:{remote_path}", str(dest)],
+                check=True,
+                capture_output=True,
+                timeout=15,
+            )
+            new_files.append(dest)
+        except (subprocess.CalledProcessError, subprocess.TimeoutExpired) as e:
+            click.echo(f"Warning: failed to fetch {filename} from {agent_name}: {e}", err=True)
+    return new_files
+
+
+@agent.command()
+@click.option(
+    "--as",
+    "as_recipient",
+    default=None,
+    metavar="IDENTITY",
+    help=(
+        "Pull as this identity instead of AGENT_NAME/$USER. "
+        "Useful for pull-only recipients (e.g. humans) checking from a different host."
+    ),
+)
+@click.option("--mailbox", default="default", show_default=True, help="Mailbox to fetch into.")
+@click.option("--all-mailboxes", is_flag=True, help="Pull across default plus every named mailbox.")
+@click.option(
+    "--notify-cmd",
+    default=None,
+    metavar="CMD",
+    help=(
+        "Shell command to invoke when new messages arrive. "
+        "Receives NEW_COUNT (integer) and SUMMARY (human-readable string) as environment variables."
+    ),
+)
+@click.option("--json", "json_output", is_flag=True, help="Emit machine-readable JSON.")
+@click.option("--dry-run", is_flag=True, help="Show what would be fetched without writing files.")
+def pull(
+    as_recipient: str | None,
+    mailbox: str,
+    all_mailboxes: bool,
+    notify_cmd: str | None,
+    json_output: bool,
+    dry_run: bool,
+) -> None:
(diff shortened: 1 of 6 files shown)


TimeToBuildBob:
## Review guide (auto-posted)

### Key files
- `packages/gptmail/tests/test_agent_pull.py` (+313/-0, Δ313)
- `packages/gptmail/src/gptmail/agent_cli.py` (+200/-0, Δ200)
- `packages/gptmail/src/gptmail/cli.py` (+10/-0, Δ10)
- `packages/gptmail/pyproject.toml` (+1/-1, Δ2)
- `uv.lock` (+1/-1, Δ2)

### Suggested verification
- `make test-pkg PKG=gptmail`
- `make typecheck`

### Suggested review focus

_Generated by `scripts/github/pr-warmup-review-guide.py` for gptme/gptme-contrib#1478._

packages/gptmail/src/gptmail/agent_cli.py
```diff
@@ -969,6 +969,206 @@ def pending(
...
+        f"ControlPath={ctl_path}",
+        "-o",
+        "ControlPersist=60",
+    ]
+    new_files: list[Path] = []
+    for row in rows:
+        filename = str(row.get("file") or "")
+        if not filename:
+            continue
+        dest = local_inbox / filename
```
TimeToBuildBob: ❌ **P1** — The `filename` from the remote pending rows is used to build the local destination `dest = local_inbox / filename` with no path validation. A remote agent with write access to its outbox (or any compromised SSH target in the registry) can craft a filename containing `../` sequences, causing the SCP command to write the file to an arbitrary location on the puller's filesystem (e.g. `../../.ssh/authorized_keys`). This is a path-traversal vulnerability enabling arbitrary file writes, which can lead to remote code execution or credential theft. The existing `_within()` helper that enforces path containment is not used here.

packages/gptmail/src/gptmail/agent_cli.py
```diff
@@ -969,6 +969,206 @@ def pending(
...
+        gptmail agent pull --notify-cmd 'osascript -e "display notification \\\"$SUMMARY\\\" with title \\\"gptmail\\\""'
+        # Linux — libnotify
+        gptmail agent pull --notify-cmd 'notify-send "gptmail" "$SUMMARY"'
+    """
+    agents = _load_agents()
+    self_name = (as_recipient or _self_name()).lower()
+    mailboxes = _selected_mailboxes(mailbox, all_mailboxes)
+
+    # Local inbox for the chosen mailbox (pull always targets the first/only mailbox).
+    local_inbox = _mailbox_root(mailboxes[0]) / "inbox"
```
TimeToBuildBob: ⚠️ **P2** — When `--all-mailboxes` is used, `mailboxes` contains multiple mailbox names, but `local_inbox` is always set to the first mailbox's inbox (`_mailbox_root(mailboxes[0]) / "inbox"`). Inside `_fetch_from_agent`, every fetched message is written to this single `local_inbox` regardless of the message's `mailbox` field, so messages from non-default remote mailboxes are placed into the default mailbox's inbox. This mixes messages from separate mailboxes together, breaking mailbox separation for pull-only recipients who use multiple mailboxes. The local message retains its original `mailbox:` frontmatter, so it appears to belong to a different mailbox than where it physically resides, causing `gptmail agent list --mailbox foo` to miss it and leaving the default inbox cluttered with other mailboxes' messages.

TimeToBuildBob:
## 🤖 AI code review

Adds a new `gptmail agent pull` command that fetches outbox messages from SSH-reachable agents into the local inbox, with options for `--as`, `--json`, `--notify-cmd`, and `--dry-run`. Refactors `_remote_pending_rows` and `_fleet_pending_rows` to accept an `all_mailboxes` flag and propagates it through the `pending` command. Adds a `--version` option to the CLI group and bumps the package version from 0.1.0 to 0.2.0. Includes extensive new tests for the pull command and updates existing fleet-pending tests.

> ⚠️ **1 generated file excluded and NOT reviewed:** ``uv.lock`` (+1/-1 lines, 0KB). Machine-generated output is removed from the diff before review — from what the model sees and from the size gate alike — so nothing below says anything about it.

<b>1 finding</b> · ❌ **1** P1 — 🛑 **1 at P0/P1**

### Findings

**❌ P1 high** — [`packages/gptmail/src/gptmail/agent_cli.py:652`](https://github.com/gptme/gptme-contrib/blob/ac62e0a952c0496d0c253016dfddc9ae257cf3cb/packages/gptmail/src/gptmail/agent_cli.py#L652)

`_remote_pending_rows` catches only `subprocess.CalledProcessError` and `subprocess.TimeoutExpired` when invoking `ssh`, but not `FileNotFoundError` (an `OSError`) when the `ssh` executable is missing. Since this function is called by both `pull` and `pending --fleet`, a host without `ssh` will raise an unhandled exception, crashing the entire command with a traceback instead of returning an empty list and a warning. This contradicts the per-agent error resilience documented for `pull` (which handles missing `scp` via `OSError` in `_fetch_from_agent` at line 1065). The consequence is that `gptmail agent pull` aborts completely on systems lacking the ssh client, blocking all message fetching rather than skipping unreachable agents.

```suggestion
except (OSError, subprocess.CalledProcessError, subprocess.TimeoutExpired) as e:
```

<sub>**How this was verified:** Checked the try/except in `_remote_pending_rows` (lines 644-654) and compared with `_fetch_from_agent` (lines 1065-1066) which catches `OSError`. Also confirmed the dry-run path in `pull` calls `_remote_pending_rows` directly (line 1143).</sub>

- `packages/gptmail/pyproject.toml` — Bumps version from 0.1.0 to 0.2.0.
- `packages/gptmail/src/gptmail/agent_cli.py` — Adds `pull` command and helper functions `_pull_destination` and `_fetch_from_agent`; adds `all_mailboxes` parameter to `_remote_pending_rows` and `_fleet_pending_rows`.
- `packages/gptmail/src/gptmail/cli.py` — Adds `_version_string()` and `@click.version_option` to the CLI group.
- `packages/gptmail/tests/test_agent_cli.py` — Parametrizes fleet-pending test for all_mailboxes propagation and adds a test for explicit --all-mailboxes command construction.
- `packages/gptmail/tests/test_agent_pull.py` — New test file covering pull fetching, dedup, dry-run, notify-cmd, mailbox handling, and error cases.

</details>

| `047e78570231` | 3/5 | 2 | llm | 2026-08-20 15:41 UTC |
| `f8903fc62d0d` | 3/5 | 3 | llm | 2026-08-20 17:41 UTC |
| `4200fecbde80` | 4/5 | 2 | llm | 2026-08-21 01:40 UTC |
| `54474a341883` | 4/5 | 2 | llm | 2026-08-21 05:08 UTC |
| `6567369cc6e5` | 3/5 | 1 | llm | 2026-08-21 07:18 UTC |
| `9e0d3cbfe7a5` | 4/5 | 3 | llm | 2026-08-21 07:57 UTC |
| `0823a6b04486` | 3/5 | 3 | llm | 2026-08-21 08:41 UTC |

</details>

<sub>Reviewed `ac62e0a952c0` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 133s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

packages/gptmail/src/gptmail/agent_cli.py
```diff
@@ -969,6 +969,229 @@ def pending(
...
+        dest.parent.mkdir(parents=True, exist_ok=True)
+        # The row's mailbox field tells us which remote outbox subfolder to look in.
+        row_mailbox = _normalize_mailbox(str(row.get("mailbox") or mailboxes[0]))
+        if row_mailbox == "default":
+            remote_path = f"{workspace}/messages/outbox/{filename}"
+        else:
+            remote_path = f"{workspace}/messages/mailboxes/{row_mailbox}/outbox/{filename}"
+        try:
+            subprocess.run(
+                ["scp", *ssh_opts, f"{ssh_target}:{remote_path}", str(dest)],
```
TimeToBuildBob: ⚠️ **P2** — In `_fetch_from_agent`, the scp command is run with `check=True` and `timeout=15`. If scp fails after writing a partial file (e.g., network timeout, remote file vanished mid-transfer), the `CalledProcessError` or `TimeoutExpired` is caught, but the partially written destination file is left on disk. On the next `pull`, `dest.exists()` returns True, so the code skips the file, believing it was already fetched. The message is thus permanently stuck in a corrupted state and never retried. This causes silent message loss on transient network failures. Fix: on exception, remove the partially written `dest` file before continuing.

packages/gptmail/src/gptmail/agent_cli.py
```diff
@@ -969,6 +969,229 @@ def pending(
...
+            ts_part = f" [{ts}]" if ts else ""
+            click.echo(f"  {sender}{ts_part}: {subject}  ({f.name})")
+
+    if notify_cmd and n > 0:
+        subjects = "; ".join(
+            str((meta_of(f) or {}).get("subject", "(no subject)")) for f in new_files
+        )
+        env = os.environ.copy()
+        env["NEW_COUNT"] = str(n)
+        env["SUMMARY"] = f"{n} new message(s): {subjects}"
```
TimeToBuildBob, line 1205: ⚠️ **P2** — The `--notify-cmd` hook is executed with `shell=True` and an environment variable `SUMMARY` built from message subjects: `env["SUMMARY"] = f"{n} new message(s): {subjects}"`. If a message subject contains shell metacharacters (e.g., `$(touch /tmp/pwned)`), and the user's notify command uses `$SUMMARY` without quoting (as in `notify-send gptmail $SUMMARY`), the shell will expand the command substitution, executing arbitrary commands from the message content. This is a command injection vulnerability because the message content is untrusted input from other agents. While the examples in the docstring quote the variable, nothing enforces that. Fix: either pass the summary as a separate argument to a non-shell command, or document and enforce that the summary must be quoted; safer to avoid `shell=True` for user-supplied hooks.

TimeToBuildBob:
Fixed the latest AI-review findings in `4200fecb` and pushed it to this PR.

- `--all-mailboxes` is now carried explicitly to each remote, so a remote-only mailbox is discovered even when it does not exist locally yet.
- Failed or timed-out SCP transfers delete any partial destination, allowing the next pull to retry.
- Notification hooks no longer run through an implicit shell; `--notify-cmd` is tokenized and executed directly. Shell expansion is now opt-in via an explicit `sh -c`, and an untrusted subject remains data.

Added regressions for all three paths. Local verification: 184 `gptmail` tests passed, package mypy passed, and pinned ruff/ruff-format hooks passed. CI and the automatic AI-review rerun are now in progress. No Greptile trigger was sent.

TimeToBuildBob:
@greptileai review

packages/gptmail/src/gptmail/agent_cli.py
```diff
@@ -969,6 +971,241 @@ def pending(
...
+                if dest is None:
+                    if filename:
+                        click.echo(
+                            f"Warning: refusing unsafe filename from {agent_name}: {filename}",
+                            err=True,
+                        )
+                    continue
+                if not dest.exists():
+                    subject = str(row.get("subject", "(no subject)"))
+                    click.echo(f"[dry-run] would fetch from {agent_name}: {subject}  ({filename})")
```
TimeToBuildBob: ⚠️ **P2** — When `--json` and `--dry-run` are used together, the dry-run loop prints human-readable lines to stdout and then the JSON payload is also printed to stdout, producing mixed output that is not valid JSON. Scripted callers combining these flags (e.g., a cron job doing `--dry-run --json` to inspect what would be fetched) would receive corrupted output, breaking any downstream JSON parsing. The code at lines 1158 and 1178 both write to stdout with no separation or guard.

packages/gptmail/src/gptmail/agent_cli.py
```diff
@@ -969,6 +971,241 @@ def pending(
...
+        except click.BadParameter as e:
+            click.echo(f"Warning: refusing invalid mailbox from {agent_name}: {e}", err=True)
+            continue
+        if dest is None:
+            if filename:
+                click.echo(
+                    f"Warning: refusing unsafe filename from {agent_name}: {filename}", err=True
+                )
+            continue
+        if dest.exists():
```
TimeToBuildBob, line 1041: ⚠️ **P2** — The `dest.exists()` check before each scp is not atomic with the subsequent file copy. Two concurrent `gptmail agent pull` runs (e.g., overlapping cron jobs) can both observe `dest.exists() == False` for the same filename, both proceed to scp, and both write to the same path. Since scp does not protect against concurrent writers, the two runs can interleave writes, potentially leaving a corrupted file or silently dropping one copy. The README documents running pull on a schedule, so overlapping executions are plausible. The dedup then fails to be idempotent under concurrency.

packages/gptmail/src/gptmail/agent_cli.py
TimeToBuildBob: Fixed in 54474a34: dry-run candidates are counted for JSON output, while human-readable dry-run lines are suppressed under --json. The regression parses the entire stdout as JSON and verifies no inbox write.

packages/gptmail/src/gptmail/agent_cli.py
TimeToBuildBob: Fixed in 54474a34: each SCP now writes to a unique same-directory temporary file, then publishes with a no-clobber hard link. A concurrent winner is preserved, this pull reports zero new files, and the temporary file is removed. A regression simulates the race.

TimeToBuildBob:
## Convergence adjudication

- **Fixed:** `df4d32c91137` (`--dry-run --json`) now emits JSON only and includes prospective files; `f98b6c22ef19` (concurrent pulls) now stages each SCP into a unique same-directory temporary file and publishes with a no-clobber hard link. Commit: `54474a34`.
- **Remaining:** none. Both current-head findings were blocking correctness bugs, not nits or false positives; no finding was dismissed.
- **CI:** prior head was fully green. CI for `54474a34` is currently running; local verification passed: 18 focused pull tests, package mypy, Ruff check/format, scoped pre-commit, and `git diff --check`.
- **Domain risk:** the publication strategy assumes inbox and temporary files share a filesystem with hard-link support. A maintainer manual smoke test of a real SSH pull and two overlapping scheduled pulls is the useful final check.
- **Convergence:** detector reports `round_convergence.status = no_history`, `stable_rounds = 0` of 2. This is a detector/history mismatch (the PR uses Bob's AI reviewer, while the convergence script found no Greptile history), not evidence of two stable clean rounds.

No further Greptile review was triggered. Stop here for maintainer judgment after CI.

packages/gptmail/src/gptmail/agent_cli.py
```diff
@@ -969,6 +972,253 @@ def pending(
...
+
+    if notify_cmd and n > 0:
+        subjects = "; ".join(
+            str((meta_of(f) or {}).get("subject", "(no subject)")) for f in new_files
+        )
+        env = os.environ.copy()
+        env["NEW_COUNT"] = str(n)
+        env["SUMMARY"] = f"{n} new message(s): {subjects}"
+        try:
+            subprocess.run(shlex.split(notify_cmd), env=env, timeout=10, check=True)
```
TimeToBuildBob, line 1207: ⚠️ **P2** — In the new `pull` command, `subprocess.run` for `--notify-cmd` only catches `subprocess.CalledProcessError` and `subprocess.TimeoutExpired`. If the user supplies a command that is not found (e.g., a typo like `--notify-cmd 'notifysend hi'`), `subprocess.run` raises `FileNotFoundError`, which is not caught. This propagates up and aborts the entire `pull` invocation with a traceback, even though other messages were successfully fetched. The command should catch `OSError` (which includes `FileNotFoundError`) so that a bad notify hook degrades to a warning instead of killing the pull.

packages/gptmail/src/gptmail/agent_cli.py
```diff
@@ -689,6 +691,7 @@ def _fleet_pending_rows(
                 agent,
                 recipient=recipient,
                 mailboxes=mailboxes,
+                all_mailboxes=len(mailboxes) > 1,
```
TimeToBuildBob: ⚠️ **P2** — In `_fleet_pending_rows`, the `all_mailboxes` argument passed to `_remote_pending_rows` is computed as `len(mailboxes) > 1`. When a user runs `pending --for erik --fleet --all-mailboxes` and the local agent only has one mailbox (e.g., only `default`), `len(mailboxes)` is 1, so `all_mailboxes` becomes `False`. The remote command then uses `--mailbox default` instead of `--all-mailboxes`, so remote agents that have a named mailbox (e.g., `ops`) containing pending messages for the recipient will not be scanned. The user explicitly asked to scan all mailboxes, but the effective remote query is limited to the default mailbox. The original `--all-mailboxes` flag should be propagated through `_fleet_pending_rows` rather than inferred from the mailbox list length.

packages/gptmail/src/gptmail/agent_cli.py
TimeToBuildBob: Fixed in 6567369c: notifier launch errors now degrade to a warning after the fetched message is safely written. The regression uses a missing executable and verifies both exit 0 and inbox persistence.

packages/gptmail/src/gptmail/agent_cli.py
TimeToBuildBob: Fixed in 6567369c: the explicit --all-mailboxes flag now flows through _fleet_pending_rows to every remote query instead of being inferred from local mailbox count. The parameterized regression covers both one and multiple local mailboxes.

packages/gptmail/src/gptmail/agent_cli.py
```diff
@@ -969,6 +974,253 @@ def pending(
...
+            continue
+        if dest.exists():
+            continue  # Already fetched — deduplicate by mailbox and filename.
+        dest.parent.mkdir(parents=True, exist_ok=True)
+        # The row's mailbox field tells us which remote outbox subfolder to look in.
+        row_mailbox = _normalize_mailbox(str(row.get("mailbox") or fallback_mailbox))
+        if row_mailbox == "default":
+            remote_path = f"{workspace}/messages/outbox/{filename}"
+        else:
+            remote_path = f"{workspace}/messages/mailboxes/{row_mailbox}/outbox/{filename}"
```
TimeToBuildBob, line 1049: ❌ **P1** — The SCP source path in `_fetch_from_agent` is built by string interpolation of `filename` and `row_mailbox` from a remote pending row without validating them for `..` or path separators. A compromised remote agent can return a row with `file: "../../../etc/passwd"` and `mailbox: "default"`, producing `remote_path = "/workspace/messages/outbox/../../../etc/passwd"`, which resolves outside the outbox directory. The local containment check (`_pull_destination` → `_within`) only guards the local destination under `messages/inbox`, not the remote source; then `scp` copies the resolved arbitrary remote file into the local inbox, exfiltrating sensitive data such as SSH keys or credentials from the remote host. The same applies to a malicious `mailbox` value that escapes `messages/mailboxes/`. Because `pull` runs on a schedule and trusts agents.yaml, any agent that becomes compromised can use this to read arbitrary files from its own filesystem and pull them onto the operator's host. _(3/3 passes)_

TimeToBuildBob: Fixed in 9e0d3cbf: remote pending-row filenames must now be a single basename before either the local destination or remote SCP source is constructed. Values containing traversal or any path separator are rejected before SCP runs; mailbox names were already constrained by `_normalize_mailbox`. The parameterized regression covers both `../../../../escaped.md` and `subdir/message.md` and asserts that no SCP call occurs.

packages/gptmail/src/gptmail/agent_cli.py
```diff
@@ -969,6 +974,255 @@ def pending(
...
+
+    if notify_cmd and n > 0:
+        subjects = "; ".join(
+            str((meta_of(f) or {}).get("subject", "(no subject)")) for f in new_files
+        )
+        env = os.environ.copy()
+        env["NEW_COUNT"] = str(n)
+        env["SUMMARY"] = f"{n} new message(s): {subjects}"
+        try:
+            subprocess.run(shlex.split(notify_cmd), env=env, timeout=10, check=True)
```
TimeToBuildBob, line 1207: ⚠️ **P2** — The `pull` command calls `subprocess.run(shlex.split(notify_cmd), ...)` inside a try block that only catches `OSError`, `subprocess.CalledProcessError`, and `subprocess.TimeoutExpired`. If the user passes a malformed `--notify-cmd` string (e.g., an unbalanced quote), `shlex.split` raises `ValueError`, which is not caught. This propagates out of the `pull` command, crashing the CLI with a traceback after the messages have already been fetched. The intended behavior per the PR description is to log notification failures as warnings and continue; instead the entire pull aborts with a non-zero exit. The fix is to wrap `shlex.split` in a try/except that catches `ValueError` and emits a warning.

packages/gptmail/src/gptmail/agent_cli.py
```diff
@@ -969,6 +974,255 @@ def pending(
...
+        if row_mailbox == "default":
+            remote_path = f"{workspace}/messages/outbox/{filename}"
+        else:
+            remote_path = f"{workspace}/messages/mailboxes/{row_mailbox}/outbox/{filename}"
+        fd, temp_name = tempfile.mkstemp(prefix=f".{dest.name}.", dir=dest.parent)
+        os.close(fd)
+        temp_dest = Path(temp_name)
+        try:
+            subprocess.run(
+                ["scp", *ssh_opts, f"{ssh_target}:{remote_path}", str(temp_dest)],
```
TimeToBuildBob: ⚠️ **P2** — In `_fetch_from_agent`, the remote path is constructed as `f"{workspace}/messages/outbox/{filename}"` (or a mailbox variant) and then passed to `subprocess.run(["scp", ..., f"{ssh_target}:{remote_path}", ...])` without any quoting or escaping. If `agent['workspace']` contains spaces (common on macOS paths like `/Users/John Doe/workspace`), the `remote_path` string will contain spaces. When scp interprets the argument after the colon, it splits on whitespace, so it will attempt to copy from an incorrect remote location or fail. The existing `_ssh_deliver` has a similar issue, but this new code introduces the same fragility for the pull path. Workspaces with spaces will cause every fetch to fail or copy the wrong file, and the failure is only reported as a warning, so the message is silently lost.

packages/gptmail/src/gptmail/agent_cli.py
```diff
@@ -969,6 +974,255 @@ def pending(
...
+                check=True,
+                capture_output=True,
+                timeout=15,
+            )
+            try:
+                os.link(temp_dest, dest)
+            except FileExistsError:
+                continue  # A concurrent pull published this message first.
+            new_files.append(dest)
+        except (subprocess.CalledProcessError, subprocess.TimeoutExpired) as e:
```
TimeToBuildBob: ⚠️ **P2** — The `subprocess.run` call for `scp` in `_fetch_from_agent` is wrapped in a try/except that only catches `subprocess.CalledProcessError` and `subprocess.TimeoutExpired`. If the `scp` executable is not found (raises `FileNotFoundError`, a subclass of `OSError`) or permissions prevent execution, the exception is not caught and propagates, aborting the entire `pull` command. This contradicts the stated design that one unreachable agent or transfer failure should only log a warning and continue with partial results. On a minimal system lacking `scp`, the CLI would crash mid-pull, losing all partially fetched messages and returning a traceback. The fix is to also catch `OSError` (or `Exception` selectively) around the scp call and treat it as a per-agent failure.

packages/gptmail/src/gptmail/agent_cli.py
TimeToBuildBob: Fixed in 0823a6b0: malformed `--notify-cmd` quoting is now caught as a warning after fetched messages are persisted. The regression passes an unterminated quote and verifies exit 0 plus inbox persistence.

packages/gptmail/src/gptmail/agent_cli.py
TimeToBuildBob: Fixed in 0823a6b0: the remote SCP source path is now shell-quoted as one remote-shell argument, so workspace paths containing spaces remain intact. The regression checks the exact quoted source argument.

packages/gptmail/src/gptmail/agent_cli.py
TimeToBuildBob: Fixed in 0823a6b0: SCP launch `OSError`s now follow the existing per-agent warning path and clean up the staging file instead of aborting the pull. The regression simulates a missing `scp` executable and verifies exit 0 with no residue.

packages/gptmail/src/gptmail/agent_cli.py
```diff
@@ -969,6 +974,255 @@ def pending(
...
+            agent_cfg,
+            self_name=self_name,
+            mailboxes=remote_mailboxes,
+            all_mailboxes=all_mailboxes,
+        )
+        new_files.extend(fetched)
+
+    n = len(new_files)
+
+    if json_output:
```
TimeToBuildBob: ❌ **P1** — The `--notify-cmd` hook is placed after the `if json_output:` block, which returns early. When a user runs `gptmail agent pull --json --notify-cmd '...'`, the JSON payload is printed and the function returns before the notify code is reached, so the notification is silently suppressed. This contradicts the option help text, which says 'Command to invoke when new messages arrive' without caveating that it is disabled in JSON mode. A scripted caller combining `--json` with `--notify-cmd` (a plausible cron pattern) will never receive alerts, potentially missing time-sensitive messages. The notify invocation should be moved before the JSON return, or the help and docs should explicitly state that `--notify-cmd` is ignored when `--json` is present.

packages/gptmail/src/gptmail/agent_cli.py
```diff
@@ -969,6 +974,255 @@ def pending(
...
+        temp_dest = Path(temp_name)
+        try:
+            subprocess.run(
+                ["scp", *ssh_opts, f"{ssh_target}:{shlex.quote(remote_path)}", str(temp_dest)],
+                check=True,
+                capture_output=True,
+                timeout=15,
+            )
+            try:
+                os.link(temp_dest, dest)
```
TimeToBuildBob, line 1061: ⚠️ **P2** — `_fetch_from_agent` publishes the fetched message with `os.link(temp_dest, dest)` (line 1061). This requires the destination filesystem to support hard links. On filesystems that do not (e.g. FAT, some network/container mounts), `os.link` raises `OSError` (typically `Operation not permitted`). The exception is caught by the surrounding `except (OSError, ...)` block and reported as a warning, but the message is not stored. The temp file is then unlinked in `finally`, so the message is permanently lost for this run; the next pull will re-fetch it (since `dest` does not exist) and fail again in the same way. This makes the pull command unreliable on such filesystems. A more portable approach is to use `shutil.copy2` into a temp file and then `os.replace` (atomic rename), or to fall back to a copy when `os.link` is unsupported.

packages/gptmail/src/gptmail/agent_cli.py
```diff
@@ -969,6 +974,255 @@ def pending(
...
+        gptmail agent pull --notify-cmd "sh -c 'notify-send gptmail \"$SUMMARY\"'"
+    """
+    agents = _load_agents()
+    self_name = (as_recipient or _self_name()).lower()
+    mailboxes = _selected_mailboxes(mailbox, all_mailboxes)
+    remote_mailboxes = [] if all_mailboxes else mailboxes
+
+    new_files: list[Path] = []
+    agents_polled: list[str] = []
+    agents_skipped: list[str] = []
```
TimeToBuildBob: ⚠️ **P2** — The `pull` command maintains an `agents_skipped` list (lines 1131, 1138, 1141) and appends agent names to it, but this list is never read or emitted anywhere in the command's output, not even in `--json` mode. This is dead code: the information it accumulates is discarded, so users have no way to know which agents were skipped or why. While not a functional failure, it clutters the code and suggests an intended feature (e.g. surfacing skipped agents) that was not completed. Either use the list (e.g. include it in the JSON payload or print a summary) or remove it.

packages/gptmail/src/gptmail/agent_cli.py
TimeToBuildBob: Fixed in `ac62e0a9`: JSON output no longer returns before notification dispatch. `--json --notify-cmd` now invokes the hook and still emits one parseable JSON document; a regression covers the combined flags.

packages/gptmail/src/gptmail/agent_cli.py
TimeToBuildBob: Adjudicated as non-blocking. The no-clobber hard link is deliberate: `os.replace` would overwrite a concurrently published inbox message, while falling back to copy/replacement would lose that invariant. Unsupported filesystems fail visibly, remove the staging file, and leave the destination absent so a later pull retries; this is not permanent message loss because the remote outbox remains unchanged. The existing cleanup path is now covered explicitly by the focused pull tests.

packages/gptmail/src/gptmail/agent_cli.py
TimeToBuildBob: Fixed in `ac62e0a9`: removed the unused `agents_skipped` list and its appends. Skipping behavior and the stable JSON schema are unchanged.

TimeToBuildBob:
## Convergence adjudication — current head `ac62e0a9`

- **Fixed:** `c6eb26290fa7` — `--json --notify-cmd` now executes the notification hook before emitting one parseable JSON document; regression added. `e34fc94601e6` — removed the unused `agents_skipped` accumulator. Both are in `ac62e0a9`.
- **Remaining:** `a919a69142ab` is non-blocking and was dismissed with a checkable reason. The hard-link publication is the deliberate no-clobber primitive: `os.replace` or a copy fallback would overwrite a concurrently published inbox file. On unsupported filesystems the command warns, cleans the staging file, leaves the destination absent, and retries on the next pull because the remote outbox remains unchanged.
- **CI:** prior head `0823a6b0` was fully green. Current-head CI is running; the affected `gptmail` Python 3.10 job is already green. Local verification: all 193 `gptmail` tests pass, focused Ruff check/format and mypy pass, scoped pre-commit passed during commit, and `git diff --check` is clean.
- **Domain risk:** real SSH/SCP interoperability and the same-filesystem hard-link requirement remain the only useful maintainer smoke-test areas. Concurrent publication has a regression; unsupported hard links fail visibly and retry rather than clobbering.
- **Convergence:** detector reports `round_convergence.status = no_history`, `stable_rounds = 0` of 2. This PR uses Bob's AI-review findings, so that Greptile-specific history is unavailable; all three current-head review threads are now explicitly resolved.

No further Greptile or AI-review trigger was sent. Stop here for maintainer judgment after CI.

PR #1479, opened 2026-08-20
feat(precommit): make check-root-structure a shareable remote hook

### LINKED ISSUE #1445: Clean up root dir, better structure and CI checks to prevent misplacing/bloat

TimeToBuildBob:
## What

Makes `scripts/precommit/check_root_structure.py` reusable as a remote pre-commit hook, addressing the remaining item from #1445 (propagate to other repos).

## Changes

- **`check_root_structure.py`**: Add `--allow ENTRY` args (repeatable) so consuming repos configure their own allowlist via `.pre-commit-config.yaml` args. Add `get_repo_root()` using `git rev-parse --show-toplevel` — fixes remote hook use where `__file__` resolves into the pre-commit cache, not the consuming repo. Keep `ALLOWED_ROOT_ENTRIES` as the gptme-contrib default fallback.

- **`.pre-commit-hooks.yaml`** (new): Declares the hook so other repos can reference it as:
  ```yaml
  - repo: https://github.com/gptme/gptme-contrib
    rev: <SHA>
    hooks:
    - id: check-root-structure
      args:
        - --allow=.github
        - --allow=README.md
        - --allow=src
        - --allow=tests
        # ... each permitted root entry
  ```
  Add `.pre-commit-hooks.yaml` itself to the contrib allowlist.

- **Tests**: 4 new tests for `--allow` semantics (override, complete-list pass, superset-fine, no-args fallback). Fix existing tests to pass `argv=[]` to `main()` so argparse doesn't read pytest's `sys.argv`.

## Follow-up PRs

This unblocks separate PRs for:
- `gptme/gptme` — add root-structure check via remote hook ref
- `gptme/gptme-agent-template` — add via submodule path (pattern already established for other hooks)

Closes #1445 (together with the follow-up PRs)


TimeToBuildBob:
## Review guide (auto-posted)

### Key files
- `scripts/precommit/check_root_structure.py` (+88/-17, Δ105)
- `tests/test_check_root_structure.py` (+64/-11, Δ75)
- `.pre-commit-hooks.yaml` (+12/-0, Δ12)

### Suggested verification
- `uv run pytest tests/test_check_root_structure.py -q`
- `make typecheck`

### Suggested review focus

_Generated by `scripts/github/pr-warmup-review-guide.py` for gptme/gptme-contrib#1479._

tests/test_check_root_structure.py
```diff
@@ -2,35 +2,33 @@
...
+    # environment (we're always inside a git repo when tests run).
     partial_subset = frozenset(["scripts", "tests", "lessons", "README.md"])
     assert (
         partial_subset <= check_root_structure.ALLOWED_ROOT_ENTRIES
     ), "test precondition"
     with patch.object(
         check_root_structure, "get_tracked_root_entries", return_value=partial_subset
     ):
-        result = check_root_structure.main()
+        result = check_root_structure.main([])
```
TimeToBuildBob, line 29: ⚠️ **P2** — 测试`test_allowed_entries_pass`及其他调用`main([])`的测试没有patch `get_repo_root()`，因此`main()`在`entries = get_tracked_root_entries(get_repo_root())`中执行真实的`git rev-parse --show-toplevel`。如果测试在非git目录（例如从tarball解压或CI沙箱）中运行，`subprocess.run`以`check=True`执行将引发`CalledProcessError`，导致测试失败。测试注释声称'This is hermetic: no git subprocess for get_tracked_root_entries'，但忽略了`get_repo_root()`本身运行了真实的git命令，因此测试实际上依赖环境，破坏了hermetic性。这可能导致测试在非git环境中失败，而在CI中却通过，隐藏了环境依赖。

TimeToBuildBob:
## 🤖 AI code review

Makes check_root_structure.py usable as a remote pre-commit hook by adding a --allow CLI arg, a get_repo_root() helper that resolves the consumer repository via 'git rev-parse --show-toplevel', and a new .pre-commit-hooks.yaml. The allowlist logic now falls back to ALLOWED_ROOT_ENTRIES only when no --allow args are supplied. Tests were updated to pass argv=[] to main() and to cover --allow semantics, plus error paths.

**⚠️ P2 medium · ℹ️ non-blocking** — `scripts/precommit/check_root_structure.py:148`

The --allow argument is compared as an exact literal string against the first path component produced by `git ls-files`. A user who writes `--allow=src/` (a natural way to express a directory) will not match the root entry `src`, so the hook reports an unexpected entry and fails. The error message then suggests adding `--allow=ENTRY`, even though the user already supplied an allow for that directory, just with a trailing slash. This is a classic footgun: the help text says 'Allow this top-level entry' without mentioning that trailing slashes are significant. The test `test_allow_args_are_literal_root_entry_names` confirms this behavior is intended, but it remains a likely source of confusing failures for consumers configuring the remote hook. Normalizing args (e.g., stripping trailing slashes) or documenting the exact-match rule would prevent this.

```suggestion
Normalize entries in the `--allow` argument parsing, e.g. `args.allowed = [a.rstrip('/') for a in args.allowed]`, or update the help text to explicitly state that trailing slashes are not stripped.
```

<sub>**How this was verified:** Read the `--allow` help text and the test `test_allow_args_are_literal_root_entry_names`; both confirm the exact-match behavior. The `get_tracked_root_entries` function produces entries without slashes (e.g., `src`), so `src/` never matches.</sub>

</details>

- `.pre-commit-hooks.yaml` — New file declaring the check-root-structure hook for remote use, with entry pointing at the script and no extra dependencies.
- `scripts/precommit/check_root_structure.py` — Adds argparse --allow, get_repo_root(), switches to -z ls-files, and improves error handling for subprocess failures.
- `tests/test_check_root_structure.py` — Updates existing tests to pass [] to main(), adds tests for --allow, get_repo_root, and error paths.

</details>

| `7875514efd35` | 4/5 | 1 | llm | 2026-08-20 15:23 UTC |
| `7e85fbe21926` | 3/5 | 1 | llm | 2026-08-20 17:02 UTC |
| `176b93e9f0c0` | 5/5 | 0 | llm | 2026-08-21 00:09 UTC |
| `176b93e9f0c0` | 4/5 | 1 | llm | 2026-08-21 00:20 UTC |
| `c393d5bf71a6` | 5/5 | 0 | llm | 2026-08-21 05:16 UTC |
| `c393d5bf71a6` | 4/5 | 1 | llm | 2026-08-21 05:40 UTC |
| `df543f36b842` | 4/5 | 2 | llm | 2026-08-21 08:44 UTC |
| `fc63e0cc31c0` | 4/5 | 2 | llm | 2026-08-21 09:31 UTC |
| `f1d584d1e7f2` | 4/5 | 1 | llm | 2026-08-21 10:06 UTC |

</details>

<sub>Reviewed `1adf1c6fee81` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 203s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

TimeToBuildBob:
@greptileai review

.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,12 @@
+- id: check-root-structure
+  name: Check root directory structure
+  description: >
+    Enforce an explicit allowlist of top-level files and directories. Any
+    committed root-level entry not on the list fails the check, forcing an
+    intentional review. Pass ``--allow ENTRY`` (repeatable) for each permitted
+    entry in the consuming repo's ``.pre-commit-config.yaml``.
+  entry: python3 scripts/precommit/check_root_structure.py
+  language: python
```
TimeToBuildBob: ❌ **P1** — The hook is declared with `language: python` and an entry of `python3 scripts/precommit/check_root_structure.py`. With `language: python`, pre-commit creates a virtualenv, attempts to `pip install` the gptme-contrib repository as a Python package, and then runs the entry from within that package context. The entry path is a raw script, not a package entry point, so the install step will likely fail (if the repo lacks a proper setup.py/pyproject.toml) or the script will not be found because the path is not mapped into the installed package. For a remote hook that simply runs a script, the correct language is `script`, which copies the script and executes it with its shebang; the entry should then be the script's relative path (e.g., `scripts/precommit/check_root_structure.py`). As written, consuming repos cannot use this hook at all, defeating the PR's purpose.

TimeToBuildBob:
Fixed in `176b93e9`: the hook now uses `language: script`, runs the checkout-relative executable, and the script is mode `100755`. I verified installation and execution from a separate temporary consumer repository with a clean pre-commit cache; the focused tests pass (9/9), Ruff is clean, and all current-head CI checks are green.

The earlier monitoring attempts timed out in workspace-agent/pre-commit overhead before they could publish an effect; this push is the verified closeout. No Greptile trigger is needed—the automatic AI reviewer owns the next review pass.

scripts/precommit/check_root_structure.py
```diff
@@ -79,20 +122,48 @@ def get_tracked_root_entries() -> set[str]:
...
+            "Allow this top-level entry (repeatable). "
+            "If not specified, uses the built-in gptme-contrib defaults. "
+            "Specify once per permitted entry when using this hook from another repo."
+        ),
+    )
+    args = parser.parse_args(argv)
+
+    # When --allow args are provided, they define the complete allowlist.
+    # When absent, fall back to the gptme-contrib defaults.
+    allowed = frozenset(args.allowed) if args.allowed else ALLOWED_ROOT_ENTRIES
```
TimeToBuildBob, line 148: ⚠️ **P2** — The `--allow` values are used verbatim as allowed entries, but `git ls-files` returns first path components without trailing slashes. If a user passes `--allow=src/` (as is common when copying directory names), the allowed set contains `'src/'` while the actual entry is `'src'`, so `unexpected` will include `'src'` and the hook fails even though the user intended to allow it. The help text does not state that directory names must not have trailing slashes, making this an easy mistake. Strip trailing slashes from `--allow` values to normalize them against the entries from `get_tracked_root_entries()`. _(2/3 passes — pass 3 looked and disagreed)_

TimeToBuildBob:
## Convergence adjudication

- **Fixed:** The earlier hermetic-test P2 and remote-hook P1 are addressed in `7e85fbe` and `176b93e`; the remote-hook install/execution path was verified from a clean consumer repository.
- **Remaining (non-blocking):** The trailing-slash P2 is dismissed. `--allow` accepts literal Git top-level entry names, and every documented/configured example uses that syntax (`src`, not `src/`). Silently stripping `/` would turn a typo into permission for a different literal entry instead of failing closed. Commit `c393d5bf` adds a regression test proving `--allow=src/` does not allow tracked root `src`; the review thread is resolved on that explicit, checkable basis.
- **CI:** The prior head was fully green. Current-head CI for the test-only adjudication commit is queued; local verification is 10 focused tests passing, Ruff clean, and commit hooks passing.
- **Domain risk:** Low. The only maintainer spot-check worth doing is the already-exercised clean consumer-repo remote-hook installation path; no runtime package behavior changes.
- **Convergence:** Detector reports `stable_rounds: 0` with status `no_history` (the event ledger hit its dispatch cap, but the detector has no stored round history). No further review was triggered.

This is ready for maintainer judgment once current-head CI settles; I am not auto-merging it.

TimeToBuildBob:
Current-head CI has now settled green on `c393d5bf` (pre-commit, integration tests, coverage, typecheck, install-size, and Codecov patch all passed/skipped as intended). The adjudication recommendation above is unchanged: all review threads are resolved, no further review trigger, and maintainer judgment is the remaining step.

.pre-commit-hooks.yaml
TimeToBuildBob: Fixed in `176b93e9`: the manifest now uses `language: script` with the checkout-relative executable, and the script is executable. Verified by installing and running the hook from a separate clean consumer repository.

scripts/precommit/check_root_structure.py
```diff
@@ -22,20 +23,60 @@
...
-    """Return first path components of all files tracked in the git index."""
+    This works correctly both when called locally and when executed as a
+    remote pre-commit hook (where ``Path(__file__)`` points into a cache
+    directory, not the consuming repo).
+    """
+    result = subprocess.run(
+        ["git", "rev-parse", "--show-toplevel"],
+        capture_output=True,
+        text=True,
+        check=True,
```
TimeToBuildBob, line 60: ⚠️ **P2** — In get_repo_root(), subprocess.run is called with check=True. If `git rev-parse --show-toplevel` fails (e.g., when the script is run from a non-git directory), a CalledProcessError is raised and the hook crashes with an unhandled traceback instead of exiting cleanly with a non-zero status. Pre-commit hooks are expected to return a non-zero exit code to indicate failure, but a traceback is an abrupt and unhelpful failure mode. While pre-commit always runs in a git repo, the script is also documented as runnable manually, so a user could trigger this. The fix is to catch CalledProcessError and return a user-friendly error message with exit code 1. _(2/3 passes — pass 3 looked and disagreed)_

TimeToBuildBob:
Fixed the current AI-review P2 in `df543f36`: manual execution outside a Git repository now catches `git rev-parse` failure, prints a concise diagnostic to stderr, and exits 1 instead of emitting a traceback. Added a regression test for the exact failure path.

Verified locally: 11 focused tests pass, Ruff passes, formatting is clean, the manual non-repo invocation returns 1 with the expected message, and all commit hooks pass. Current-head CI and the automatic AI review are now running; no Greptile trigger was sent.

scripts/precommit/check_root_structure.py
```diff
@@ -79,20 +122,58 @@ def get_tracked_root_entries() -> set[str]:
...
+        repo_root = get_repo_root()
+    except subprocess.CalledProcessError:
+        print(
+            "check-root-structure: unable to determine repository root; "
+            "run this command inside a Git repository.",
+            file=sys.stderr,
+        )
+        return 1
+
+    entries = get_tracked_root_entries(repo_root)
```
TimeToBuildBob: ⚠️ **P2** — In `main()`, the `try/except` only wraps the `get_repo_root()` call (lines 148-156), catching `subprocess.CalledProcessError`. However, the next line `entries = get_tracked_root_entries(repo_root)` also executes `git ls-files --cached` through `subprocess.run` with `check=True` (lines 77-83). Any failure there — a corrupt index, a locked index, or a repository whose root directory no longer exists — raises an unhandled exception, producing a full Python traceback and exit code 1 instead of the intended clean error message. This is inconsistent with the designed failure mode and leaves a real gap: the test `test_get_repo_root_reports_non_git_directory` covers only the first subprocess failure, so the second is untested and unhandled. The fix is to wrap the `get_tracked_root_entries` call in the same `try/except` and print a similar helpful message.

TimeToBuildBob:
Fixed both current AI-review findings in `fc63e0cc`:

- `git ls-files` failures are now handled by the same concise diagnostic path as `git rev-parse` failures, with a regression test for the second subprocess failure.
- tracked paths now use `git ls-files -z` and NUL parsing, so embedded newlines cannot create fake root entries; the regression test covers that exact filename shape.

Verified locally: 13 focused tests pass, Ruff passes, formatting is clean, scoped hooks pass, and the push succeeded. Current-head CI and the automatic AI reviewer are running; I did not trigger Greptile.

scripts/precommit/check_root_structure.py
```diff
@@ -22,33 +23,77 @@
...
+    This works correctly both when called locally and when executed as a
+    remote pre-commit hook (where ``Path(__file__)`` points into a cache
+    directory, not the consuming repo).
+    """
     result = subprocess.run(
-        ["git", "ls-files", "--cached"],
-        cwd=REPO_ROOT,
+        ["git", "rev-parse", "--show-toplevel"],
         capture_output=True,
         text=True,
```
TimeToBuildBob: ⚠️ **P2** — `get_repo_root()` uses `subprocess.run(..., text=True)`, which decodes `git rev-parse --show-toplevel` output using the locale's preferred encoding. If the repository path contains bytes that are not valid in that encoding (e.g., a non-UTF-8 filename when the locale is UTF-8, or any non-ASCII path when the locale is `C`), the decode raises `UnicodeDecodeError`, crashing the hook. The hook's error handling does not catch this exception, so it produces a traceback and exits with a non-zero code, but without the intended clear message. This is inconsistent with `get_tracked_root_entries()`, which correctly decodes `git ls-files -z` bytes with `errors='surrogateescape'`. To fix, use `text=False` and decode the bytes manually with `errors='surrogateescape'`.

TimeToBuildBob:
Fixed the remaining current-head AI-review finding in `1adf1c6f`: the hook now catches `OSError` from launching Git and reports the same concise diagnostic instead of leaking a traceback when `git` is unavailable. Added a regression test and verified the real missing-`git` path.

Local verification: 15 focused tests pass; Ruff, Ruff format, and mypy hooks pass. The push succeeded, and current-head CI plus the automatic AI reviewer are running. I did not trigger Greptile.

scripts/precommit/check_root_structure.py
```diff
@@ -79,20 +124,64 @@ def get_tracked_root_entries() -> set[str]:
...
+            "Allow this top-level entry (repeatable). "
+            "If not specified, uses the built-in gptme-contrib defaults. "
+            "Specify once per permitted entry when using this hook from another repo."
+        ),
+    )
+    args = parser.parse_args(argv)
+
+    # When --allow args are provided, they define the complete allowlist.
+    # When absent, fall back to the gptme-contrib defaults.
+    allowed = frozenset(args.allowed) if args.allowed else ALLOWED_ROOT_ENTRIES
```
TimeToBuildBob, line 148: ⚠️ **P2** — The --allow argument is compared as an exact literal string against the first path component produced by `git ls-files`. A user who writes `--allow=src/` (a natural way to express a directory) will not match the root entry `src`, so the hook reports an unexpected entry and fails. The error message then suggests adding `--allow=ENTRY`, even though the user already supplied an allow for that directory, just with a trailing slash. This is a classic footgun: the help text says 'Allow this top-level entry' without mentioning that trailing slashes are significant. The test `test_allow_args_are_literal_root_entry_names` confirms this behavior is intended, but it remains a likely source of confusing failures for consumers configuring the remote hook. Normalizing args (e.g., stripping trailing slashes) or documenting the exact-match rule would prevent this.

TimeToBuildBob: Dismissed as non-blocking: `--allow` names a literal Git top-level entry, so failing closed on `src/` is intentional. Every example uses `src`, and `test_allow_args_are_literal_root_entry_names` makes the contract executable; silently normalizing a typo would authorize a different literal entry.

TimeToBuildBob:
## Final convergence adjudication (`1adf1c6f`)

- **Fixed:** the hermetic-test dependency, unusable remote-hook declaration, non-repository traceback, `git ls-files` failure path, newline/non-UTF-8 path handling, and missing-`git` launch error are all addressed by the commits already on this branch. No new code change was needed in this adjudication.
- **Remaining (non-blocking):** one P2, `--allow=src/` versus literal root entry `src`. Dismissed as intentional fail-closed semantics: the option accepts a Git top-level entry name, every example uses `src`, and `test_allow_args_are_literal_root_entry_names` verifies that a mistyped `src/` cannot silently authorize the distinct literal entry `src`. This is a UX/documentation nit, not a correctness or security defect.
- **CI:** all current-head checks are green: pre-commit, integration tests, coverage/Codecov, typecheck, install-size, and discovery.
- **Domain risk:** low. The only useful maintainer spot-check is remote installation/execution from a clean consumer repository; that path was already exercised successfully. No runtime package behavior changes.
- **Convergence:** detector reports `stable_rounds: 0` with status `no_history`; its parser sees no Greptile findings because this repository's current review is the Bob AI reviewer, while the escalation ledger records the dispatch cap. The current reviewer itself marks the sole P2 non-blocking and “round cap reached.”

No further review was triggered. This is merge-ready for maintainer judgment; I am not auto-merging it.

PR #1482, opened 2026-08-20
fix(gptodo): isolate scoped check load errors

TimeToBuildBob:
## Summary

- keep full `gptodo check` sweeps strict while limiting targeted checks to targeted parse failures
- warn on malformed sibling tasks with the path and honest last-committed-author attribution
- preserve malformed task stems for dependency existence checks so the warning cannot become a blocking missing-dependency error

## Why

Pre-commit passes only the task files in the current commit scope, but `gptodo check` loaded every task and promoted every global parse failure to a fatal error. One malformed task therefore blocked unrelated scoped commits across a shared worktree.

## Verification

- `uv run --package gptodo pytest -q packages/gptodo/tests` (468 passed)
- `make typecheck`
- `uv run ruff check packages/gptodo/src/gptodo/cli.py packages/gptodo/tests/test_load_tasks_errors.py`
- `prek run --files packages/gptodo/src/gptodo/cli.py packages/gptodo/tests/test_load_tasks_errors.py`

### DIFF
--- a/packages/gptodo/tests/test_load_tasks_errors.py
+++ b/packages/gptodo/tests/test_load_tasks_errors.py
@@ -7,6 +7,7 @@ stderr, and ``gptodo check`` reported "All N tasks verified" with the wrong
 N. The ``errors_out`` parameter lets callers detect this case.
 """
 
+import subprocess
 from pathlib import Path
 
 from click.testing import CliRunner
@@ -36,6 +37,30 @@ next_action: Run user-testing loop on `v0.31.1.dev202604277` (or later) on
 ---
 # Broken Task
 """
+
+
+def commit_task_fixtures(repo_root: Path, author: str = "Poison Author") -> None:
+    """Commit task fixtures so warning attribution exercises real git history."""
+    subprocess.run(["git", "init", "--quiet"], cwd=repo_root, check=True)
+    subprocess.run(["git", "add", "tasks"], cwd=repo_root, check=True)
+    subprocess.run(
+        [
+            "git",
+            "-c",
+            f"user.name={author}",
+            "-c",
+            "user.email=poison@example.com",
+            "-c",
+            "commit.gpgsign=false",
+            "commit",
+            "--no-verify",
+            "--quiet",
+            "-m",
+            "test fixtures",
+        ],
+        cwd=repo_root,
+        check=True,
+    )
 
 
 def test_load_tasks_appends_to_errors_out_when_yaml_invalid(tmp_path: Path) -> None:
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
CI is green, including the patch-coverage check. Marked this ready for review; awaiting the initial reviewer pass.

packages/gptodo/tests/test_load_tasks_errors.py
```diff
@@ -38,6 +39,30 @@
 """
 
 
+def commit_task_fixtures(repo_root: Path, author: str = "Poison Author") -> None:
+    """Commit task fixtures so warning attribution exercises real git history."""
+    subprocess.run(["git", "init", "--quiet"], cwd=repo_root, check=True)
```
TimeToBuildBob, line 44: ⚠️ **P2** — The new tests depend on git being installed and configured. `commit_task_fixtures` runs `git init`, `git add`, and `git commit` via `subprocess.run(..., check=True)`. In an environment without git (or with a broken git installation), these calls raise `OSError`/`FileNotFoundError`, which is not caught, causing the test to fail. The test for author attribution also asserts a specific author string, so it fails if git is missing or the commit fails. This makes the tests non-hermetic and environment-dependent, violating the principle that tests should not depend on external tools without mocking or availability checks.

TimeToBuildBob:
## 🤖 AI code review

Modifies the `gptodo check` command to separate load errors into blocking (scoped) and warning (out-of-scope) categories. Adds `_display_task_path` and `_last_committed_author` helpers, where the latter runs `git log` to attribute authorship of warning files. Updates dependency existence checks to include stems of unloadable files so broken siblings do not cause missing-dependency errors. Adds two tests that exercise the new scoped-warning behavior.

**⚠️ P2 medium** — `packages/gptodo/tests/test_load_tasks_errors.py:44`

The new tests depend on git being installed and configured. `commit_task_fixtures` runs `git init`, `git add`, and `git commit` via `subprocess.run(..., check=True)`. In an environment without git (or with a broken git installation), these calls raise `OSError`/`FileNotFoundError`, which is not caught, causing the test to fail. The test for author attribution also asserts a specific author string, so it fails if git is missing or the commit fails. This makes the tests non-hermetic and environment-dependent, violating the principle that tests should not depend on external tools without mocking or availability checks.

```suggestion
Mock `subprocess.run` or skip the git-dependent assertions when git is unavailable. Consider using a fake git binary or checking `shutil.which('git')` before invoking.
```

<sub>**How this was verified:** The test file calls git subprocesses in two tests; if git is absent, all three calls (init, add, commit) will raise, failing the tests.</sub>

</details>

- `packages/gptodo/src/gptodo/cli.py` — Refactors load‑error handling into scoped vs. global, adds author attribution via `git log`, and preserves unloadable file stems in the task-id set.
- `packages/gptodo/tests/test_load_tasks_errors.py` — Adds `commit_task_fixtures` helper and two tests for out-of-scope unloadable siblings and broken dependencies.

</details>

<sub>Reviewed `b8a033c0be1d` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 88s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

packages/gptodo/tests/test_load_tasks_errors.py
TimeToBuildBob: Reviewed this P2 and am leaving the branch unchanged. These tests intentionally exercise the real `git log` integration added by this PR; mocking `subprocess.run` would make the author-attribution assertion tautological. `git` is already a required tool for this repository checkout, CI change detection, and existing tests, while the helper supplies commit identity locally with `git -c user.name=... -c user.email=...`, so it does not depend on ambient Git configuration. CI is green on Python 3.10–3.12. This is a false positive; no 5/5-chasing change is warranted.

TimeToBuildBob:
The prior fix session resolved the thread but omitted the reviewer disposition marker, so the unchanged 4/5 marker kept re-emitting this item. I repaired that state by marking finding `cd70f5b0f33d` as an explicit rejected false positive on its existing inline reply. No code change: the test intentionally exercises real Git integration, sets commit identity per command, and CI is green across Python 3.10–3.12. The review thread is resolved and the reviewer now parses the fingerprint as dismissed; awaiting maintainer review/merge.

PR #1483, opened 2026-08-21
fix(sessions): grok background-task outputs invisible to signal extraction (8/8 false noops)

TimeToBuildBob:
## Problem
grok CLI ≥0.2.117 runs `run_terminal_command` as **background tasks**: the completed `tool_call_update` carries a `BackgroundTaskStarted` envelope (task_id + output_file), not inline `output_for_prompt`. `extract_signals_grok` therefore saw empty command output → zero git commits → every grok-build session graded noop at the 0.25 floor with `deliverables: []`.

Impact: **all 8** grok-build:grok-4.6 sessions since the 2026-08-19 reactivation were false noops — 6 have session-tagged commits on master (e.g. session 538a: 4 `git-safe-commit`s, a research doc, an issue comment, completed claims — graded noop while the judge's own completion_depth text says '17 commits'). Same class as the July grok-4.5 pilot false-noops, now provable because trajectories are durable.

## Fix
- Track `BackgroundTaskStarted` (task_id → command, output_file)
- Scan `TaskOutput` results (dict or list) that return via `get_command_or_subagent_output`: commit detection + exit_code error counting
- Fallback: read the persistent on-disk `output_file` for git-commit commands whose output never re-entered the transcript
- Legacy inline `output_for_prompt` shape still handled

## Verification
Re-extraction of all 8 durable trajectories: commits detected in 8/8, productive=True, grades 0.55–0.8. 3 regression tests; full test_signals_grok suite 22 passed.

### DIFF
--- a/packages/gptme-sessions/src/gptme_sessions/signals.py
+++ b/packages/gptme-sessions/src/gptme_sessions/signals.py
@@ -2309,6 +2309,7 @@ def extract_signals_grok(msgs: list[dict]) -> dict:
     # still sees the output (2026-08-20: 6 of 8 productive grok sessions graded
     # noop because commits were invisible to the old inline-output path).
     background_tasks: dict[str, dict[str, str]] = {}  # task_id -> command/output_file
+    counted_error_tasks: set[str] = set()
 
     def _scan_command_output(command: str, output_text: str) -> None:
         if not command or not output_text:
@@ -2337,7 +2338,10 @@ def extract_signals_grok(msgs: list[dict]) -> dict:
         exit_code = result.get("exit_code")
         if isinstance(exit_code, int) and exit_code != 0:
             nonlocal error_count
-            error_count += 1
+            if not task_id or task_id not in counted_error_tasks:
+                error_count += 1
+            if task_id:
+                counted_error_tasks.add(task_id)
         _scan_command_output(command, output_text)
         if task is not None and output_text:
             task["seen_output"] = "1"
--- a/packages/gptme-sessions/src/gptme_sessions/signals.py
+++ b/packages/gptme-sessions/src/gptme_sessions/signals.py
@@ -2261,6 +2261,20 @@ def extract_usage_copilot(msgs: list[dict]) -> dict:
 
 
 _GROK_WRITE_TOOLS = {"write", "search_replace"}
+_GROK_TERMINAL_OUTPUT_ROOT = Path.home() / ".grok" / "sessions"
+
+
+def _read_grok_terminal_output(output_file: str) -> str:
+    """Read a Grok terminal log only from the CLI's trusted session tree."""
+    try:
+        output_path = Path(output_file).resolve()
+        if not output_path.is_relative_to(_GROK_TERMINAL_OUTPUT_ROOT.resolve()):
+            return ""
+        if output_path.parent.name != "terminal" or output_path.suffix != ".log":
+            return ""
+        return output_path.read_text(errors="replace")[:200_000]
+    except OSError:
+        return ""
 
 
 def extract_signals_grok(msgs: list[dict]) -> dict:
@@ -2417,13 +2431,7 @@ def extract_signals_grok(msgs: list[dict]) -> dict:
             continue
         output_file = task.get("output_file", "")
         if output_file:
-            try:
-                _scan_command_output(
-                    command,
-                    Path(output_file).read_text(errors="replace")[:200_000],
-                )
-            except OSError:
-                pass
+            _scan_command_output(command, _read_grok_terminal_output(output_file))
 
     deliverables = list(dict.fromkeys(git_commits + file_writes))
     deliverable_details = _ordered_deliverable_details(deliverables, detail_by_value)
--- a/packages/gptme-sessions/src/gptme_sessions/signals.py
+++ b/packages/gptme-sessions/src/gptme_sessions/signals.py
@@ -2264,10 +2264,14 @@ _GROK_WRITE_TOOLS = {"write", "search_replace"}
 _GROK_TERMINAL_OUTPUT_ROOT = Path.home() / ".grok" / "sessions"
 
 
-def _read_grok_terminal_output(output_file: str) -> str:
-    """Read a Grok terminal log only from the CLI's trusted session tree."""
+def _read_grok_terminal_output(output_file: str, session_id: str | None) -> str:
+    """Read a terminal log belonging to the Grok session being analyzed."""
+    if not session_id:
+        return ""
     try:
         output_path = Path(output_file).resolve()
+        if output_path.parent.parent.name != session_id:
+            return ""
         if not output_path.is_relative_to(_GROK_TERMINAL_OUTPUT_ROOT.resolve()):
             return ""
         if output_path.parent.name != "terminal" or output_path.suffix != ".log":
@@ -2313,6 +2317,7 @@ def extract_signals_grok(msgs: list[dict]) -> dict:
     # noop because commits were invisible to the old inline-output path).
     background_tasks: dict[str, dict[str, str]] = {}  # task_id -> command/output_file
     counted_error_tasks: set[str] = set()
+    session_id: str | None = None
 
     def _scan_command_output(command: str, output_text: str) -> None:
         if not command or not output_text:
@@ -2349,6 +2354,9 @@ def extract_signals_grok(msgs: list[dict]) -> dict:
 
     for record in msgs:
         rec_type = record.get("type", "")
+        record_session_id = record.get("sessionId")
+        if isinstance(record_session_id, str) and record_session_id:
+            session_id = record_session_id
 
         if rec_type == "tool_call":
             tool_name = record.get("toolName", "")
@@ -2440,7 +2448,7 @@ def extract_signals_grok(msgs: list[dict]) -> dict:
             continue
         output_file = task.get("output_file", "")
         if output_file:
-            _scan_command_output(command, _read_grok_terminal_output(output_file))
+            _scan_command_output(command, _read_grok_terminal_output(output_file, session_id))
 
     deliverables = list(dict.fromkeys(git_commits + file_writes))
     deliverable_details = _ordered_deliverable_details(deliverables, detail_by_value)
--- a/packages/gptme-sessions/src/gptme_sessions/signals.py
+++ b/packages/gptme-sessions/src/gptme_sessions/signals.py
@@ -2287,6 +2287,46 @@ def extract_signals_grok(msgs: list[dict]) -> dict:
     detail_by_value: dict[str, dict[str, object]] = {}
     call_id_to_input: dict[str, dict] = {}
     current_batch_has_tool = False
+    # grok CLI >=0.2.117 runs terminal commands as BACKGROUND TASKS: the
+    # run_terminal_command completed update carries only a BackgroundTaskStarted
+    # envelope (task_id, output_file) — the actual output arrives later via
+    # get_command_or_subagent_output TaskOutput results, or only ever lands in
+    # the on-disk output_file. Track pending tasks so commit/journal detection
+    # still sees the output (2026-08-20: 6 of 8 productive grok sessions graded
+    # noop because commits were invisible to the old inline-output path).
+    background_tasks: dict[str, dict[str, str]] = {}  # task_id -> command/output_file
+
+    def _scan_command_output(command: str, output_text: str) -> None:
+        if not command or not output_text:
+            return
+        if "git commit" in command or "git-safe-commit" in command:
+            for commit_match in _COMMIT_RE.finditer(output_text):
+                commit_value = f"{commit_match.group(1)} {commit_match.group(2).strip()}"
+                if commit_value not in git_commits:
+                    git_commits.append(commit_value)
+                    _record_deliverable_detail(
+                        detail_by_value,
+                        value=commit_value,
+                        kind="git_commit",
+                        provenance_class="tool_authored",
+                        evidence={
+                            "source": "trajectory",
+                            "tool_name": "run_terminal_command",
+                        },
+                    )
+
+    def _consume_task_output(result: dict) -> None:
+        task_id = str(result.get("task_id", ""))
+        task = background_tasks.get(task_id)
+        command = str(result.get("command", "") or (task or {}).get("command", ""))
+        output_text = str(result.get("output", "") or "")
+        exit_code = result.get("exit_code")
+        if isinstance(exit_code, int) and exit_code != 0:
+            nonlocal error_count
+            error_count += 1
+        _scan_command_output(command, output_text)
+        if task is not None and output_text:
+            task["seen_output"] = "1"
 
     for record in msgs:
         rec_type = record.get("type", "")
@@ -2331,19 +2371,33 @@ def extract_signals_grok(msgs: list[dict]) -> dict:
 
             raw_input = call_id_to_input.get(call_id, {})
             command = raw_input.get("command", "") or raw_output.get("command", "")
-            if command and ("git commit" in command or "git-safe-commit" in command):
-                output_text = raw_output.get("output_for_prompt", "")
-                for commit_match in _COMMIT_RE.finditer(output_text):
-                    commit_value = f"{commit_match.group(1)} {commit_match.group(2).strip()}"
-                    if commit_value not in git_commits:
-                        git_commits.append(commit_value)
-                        _record_deliverable_detail(
-                            detail_by_value,
-                            value=commit_value,
-                            kind="git_commit",
-                            provenance_class="tool_authored",
-                            evidence={"source": "trajectory", "tool_name": "run_terminal_command"},
-                        )
+
+            if raw_output.get("type") == "BackgroundTaskStarted":
+                task_id = str(raw_output.get("task_id", ""))
+                if task_id:
+                    background_tasks[task_id] = {
+                        "command": str(command or raw_output.get("command", "")),
+                        "output_file": str(raw_output.get("output_file", "")),
+                        "seen_output": "",
+                    }
+                continue
+
+            if raw_output.get("type") == "TaskOutput":
+                result = raw_output.get("Result")
+                results = (
+                    result
+                    if isinstance(result, list)
+                    else [result]
+                    if isinstance(result, dict)
+                    else []
+                )
+                for res in results:
+                    if isinstance(res, dict):
+                        _consume_task_output(res)
+                continue
+
+            # Legacy inline-output shape (grok CLI < 0.2.117)
+            _scan_command_output(command, raw_output.get("output_for_prompt", ""))
 
         elif rec_type == "text" and current_batch_has_tool:
             steps += 1
@@ -2351,6 +2405,25 @@ def extract_signals_grok(msgs: list[dict]) -> dict:
 
     if current_batch_has_tool:
         steps += 1
+
+    # Background git-commit commands whose output never re-entered the
+    # transcript: fall back to the on-disk output file (persists under
+    # ~/.grok/sessions/.../terminal/).
+    for task in background_tasks.values():
+        if task.get("seen_output"):
+            continue
+        command = task.get("command", "")
+        if "git commit" not in command and "git-safe-commit" not in command:
+            continue
+        output_file = task.get("output_file", "")
+        if output_file:
+            try:
+                _scan_command_output(
+                    command,
+                    Path(output_file).read_text(errors="replace")[:200_000],
+                )
+            except OSError:
+                pass
 
     deliverables = list(dict.fromkeys(git_commits + file_writes))
     deliverable_details = _ordered_deliverable_details(deliverables, detail_by_value)
--- a/packages/gptme-sessions/src/gptme_sessions/signals.py
+++ b/packages/gptme-sessions/src/gptme_sessions/signals.py
@@ -2272,7 +2272,10 @@ def _read_grok_terminal_output(output_file: str) -> str:
             return ""
         if output_path.parent.name != "terminal" or output_path.suffix != ".log":
             return ""
-        return output_path.read_text(errors="replace")[:200_000]
+        if not output_path.is_file():
+            return ""
+        with output_path.open(errors="replace") as output:
+            return output.read(200_000)
     except OSError:
         return ""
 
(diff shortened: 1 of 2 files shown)


packages/gptme-sessions/src/gptme_sessions/signals.py
```diff
@@ -2352,6 +2406,25 @@ def extract_signals_grok(msgs: list[dict]) -> dict:
...
+            continue
+        command = task.get("command", "")
+        if "git commit" not in command and "git-safe-commit" not in command:
+            continue
+        output_file = task.get("output_file", "")
+        if output_file:
+            try:
+                _scan_command_output(
+                    command,
+                    Path(output_file).read_text(errors="replace")[:200_000],
```
TimeToBuildBob: ❌ **P1** — The fallback block reads the file at `output_file` from the transcript's `BackgroundTaskStarted` envelope with no path validation. An attacker who can influence a trajectory (e.g., from a shared dataset or a crafted session) can set `output_file` to an arbitrary absolute path like `/etc/passwd` and set `command` to `git commit`, causing the process to open and read that file. Since the file's contents are only scanned with `_COMMIT_RE` and not returned directly, the exploit is limited to confirming file existence and potentially leaking sensitive content that matches commit patterns, but it is still an arbitrary file read. The CC path already has a similar issue, but this PR introduces a new instance for grok. Mitigate by restricting `output_file` to a known session directory (e.g., `~/.grok/sessions/.../terminal/`) before reading. _(3/3 passes)_

TimeToBuildBob:
## 🤖 AI code review

Adds background-task handling to `extract_signals_grok`: `BackgroundTaskStarted` envelopes are tracked in a task_id→command/output_file map, `TaskOutput` results are scanned for git commits and exit-code errors, and a new `_read_grok_terminal_output` helper falls back to on-disk terminal logs under `~/.grok/sessions` for git-commit commands. Also adds error dedup via `counted_error_tasks` and a set of new tests in `test_signals_grok.py`.

<details><summary><b>2 findings</b> · ⚠️ **2** P2 · 🔒 **1** security — non-blocking (round cap reached)</summary>

**⚠️ P2 medium · 🔒 security · ℹ️ non-blocking** — `packages/gptme-sessions/src/gptme_sessions/signals.py:2273`

`session_id` is read directly from `record.get("sessionId")` (lines 2357-2359) and is attacker-controlled when a trajectory is untrusted. `_read_grok_terminal_output` validates the on-disk path only against this value: set `sessionId` to an existing grok session directory name and `output_file` to `~/.grok/sessions/<cwd>/<that-session>/terminal/foo.log`, and all checks pass, reading the first 200k of another session's terminal log and exposing its git commit lines via `_scan_command_output` into `git_commits`/`deliverables`. The session-mismatch test only proves rejection when the supplied sessionId differs from the path; it never tests a spoofed sessionId that matches the victim. This is a local file-disclosure primitive for another session's commit output.

<sub>**How this was verified:** Confirmed session_id is sourced from trajectory records, and no other binding (e.g., trajectory file location) ties it to the path being read; only this attacker-controlled comparison gates access.</sub>

**⚠️ P2 medium · ℹ️ non-blocking** — `packages/gptme-sessions/tests/test_signals_grok.py:449`

`test_background_task_output_file_outside_grok_sessions_is_rejected` and `test_background_task_output_file_must_be_a_terminal_log` never emit a record with `sessionId`, so `session_id` stays `None` and `_read_grok_terminal_output` returns `""` at the `if not session_id` guard before any path-containment, parent-name, suffix, or `is_file` checks. The assertions therefore pass vacuously: the rejection is caused by the missing session identity, not by the outside-root path or non-terminal suffix the test names claim to exercise. The second half of the terminal-log test with `disguised_metadata` likewise never reaches the `.log` suffix check. These validations remain effectively untested; add an `end` record carrying a `sessionId` matching the directory so the intended checks actually run.

<sub>**How this was verified:** Inspected both tests: their msgs lists contain only tool_call and tool_call_update records, with no sessionId; the `if not session_id` guard at signals.py:2269-2270 short-circuits before path validation.</sub>

</details>

- `packages/gptme-sessions/src/gptme_sessions/signals.py` — Adds `_read_grok_terminal_output`, `background_tasks` tracking, `_scan_command_output`/`_consume_task_output` helpers, and branches for `BackgroundTaskStarted`/`TaskOutput` in `extract_signals_grok`.
- `packages/gptme-sessions/tests/test_signals_grok.py` — Adds tests for background-task commit detection via TaskOutput, on-disk output-file fallback, path/session validation, and TaskOutput exit-code error counting.

</details>

| `6e4562134cea` | 3/5 | 1 | llm | 2026-08-21 00:46 UTC |
| `b5e201611d88` | 4/5 | 1 | llm | 2026-08-21 05:02 UTC |
| `0dc5301697bb` | 4/5 | 2 | llm | 2026-08-21 07:00 UTC |
| `55861fc2ea76` | 4/5 | 2 | llm | 2026-08-21 08:13 UTC |
| `75cf6afce539` | 2/5 | 3 | llm | 2026-08-21 09:41 UTC |
| `030e82e81ece` | 3/5 | 1 | llm | 2026-08-21 10:37 UTC |

</details>

<sub>Reviewed `3ab15a5cc493` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 1065s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

packages/gptme-sessions/src/gptme_sessions/signals.py
TimeToBuildBob: Fixed in b5e20161: fallback reads now resolve the path and accept only `.log` files directly under a `terminal/` directory within `~/.grok/sessions`. Added regressions for paths outside the Grok session tree and metadata/non-log paths inside it. The focused suite passes (24 tests), and scoped pre-commit hooks pass.

TimeToBuildBob:
@greptileai review

packages/gptme-sessions/src/gptme_sessions/signals.py
```diff
@@ -2287,6 +2301,46 @@ def extract_signals_grok(msgs: list[dict]) -> dict:
...
+
+    def _consume_task_output(result: dict) -> None:
+        task_id = str(result.get("task_id", ""))
+        task = background_tasks.get(task_id)
+        command = str(result.get("command", "") or (task or {}).get("command", ""))
+        output_text = str(result.get("output", "") or "")
+        exit_code = result.get("exit_code")
+        if isinstance(exit_code, int) and exit_code != 0:
+            nonlocal error_count
+            error_count += 1
```
TimeToBuildBob: ⚠️ **P2** — In `extract_signals_grok`, the nested `_consume_task_output` increments `error_count` every time it sees a result with a non-zero integer `exit_code`, without any deduplication by `task_id`. If the same background task's output is delivered more than once — e.g., the agent calls `get_command_or_subagent_output` repeatedly and the CLI returns the same completed result with `exit_code: 1` — each call adds another error. Since `grade_signals` applies an error-rate penalty (`error_rate > 0.15` subtracts 0.10, `> 0.05` subtracts 0.05), repeatedly polling one failing command can push a session into a lower reward tier even though only a single command failed. The existing test `test_task_output_list_result_and_exit_code_error` covers a single result, so this duplication isn't caught.

TimeToBuildBob: Fixed in 0dc53016: repeated `TaskOutput` records now count a non-zero exit once per stable `task_id`, while id-less results remain distinct failures. Added regressions for both cases; the focused Grok signal suite passes (26 tests), and scoped pre-commit hooks pass.

packages/gptme-sessions/src/gptme_sessions/signals.py
```diff
@@ -2261,6 +2261,20 @@ def extract_usage_copilot(msgs: list[dict]) -> dict:
...
 
 _GROK_WRITE_TOOLS = {"write", "search_replace"}
+_GROK_TERMINAL_OUTPUT_ROOT = Path.home() / ".grok" / "sessions"
+
+
+def _read_grok_terminal_output(output_file: str) -> str:
+    """Read a Grok terminal log only from the CLI's trusted session tree."""
+    try:
+        output_path = Path(output_file).resolve()
+        if not output_path.is_relative_to(_GROK_TERMINAL_OUTPUT_ROOT.resolve()):
```
TimeToBuildBob, line 2275: ⚠️ **P2** — The new `_read_grok_terminal_output` uses `Path.is_relative_to`, a method introduced in Python 3.9. On Python 3.8 it raises AttributeError, which is not caught by the `except OSError` clause. Since this function is called from the fallback loop in `extract_signals_grok` with no surrounding try/except, the AttributeError propagates and aborts the entire signal extraction. Any grok session that needs the on-disk fallback (i.e., background git commits whose output never re-entered the transcript) will crash the extractor, yielding no signals at all instead of partial ones, on 3.8 runtimes.

packages/gptme-sessions/src/gptme_sessions/signals.py
```diff
@@ -2287,6 +2301,50 @@ def extract_signals_grok(msgs: list[dict]) -> dict:
...
+        output_text = str(result.get("output", "") or "")
+        exit_code = result.get("exit_code")
+        if isinstance(exit_code, int) and exit_code != 0:
+            nonlocal error_count
+            if not task_id or task_id not in counted_error_tasks:
+                error_count += 1
+            if task_id:
+                counted_error_tasks.add(task_id)
+        _scan_command_output(command, output_text)
+        if task is not None and output_text:
```
TimeToBuildBob: ⚠️ **P2** — In `_consume_task_output`, the code sets `task["seen_output"] = "1"` whenever `output_text` is truthy, regardless of whether that output actually contained a commit line. The fallback loop later skips any task with `seen_output` set. Consequently, if a TaskOutput `output` field carries some text but not the `[branch hash] message` pattern (e.g., truncated output or a non-matching format), the on-disk log file, which does contain the commit line, is never read. This causes git commits made by background tasks to be missed, so sessions are graded as noop despite real commits, recreating the exact false-noop problem this PR set out to fix.

TimeToBuildBob: Fixed in 55861fc2: a TaskOutput now suppresses terminal-log fallback only after the output actually yields a commit. Partial/non-matching output still falls back to the trusted on-disk log. Added a regression covering partial TaskOutput plus a commit-bearing terminal log; the focused Grok suite passes (27 tests), and scoped pre-commit hooks pass.

packages/gptme-sessions/src/gptme_sessions/signals.py
TimeToBuildBob: No code change: this finding assumes unsupported Python 3.8. Both the workspace and `gptme-sessions` declare `requires-python = ">=3.10"`, and CI tests 3.10–3.12, so `Path.is_relative_to` is available on every supported runtime.

packages/gptme-sessions/src/gptme_sessions/signals.py
```diff
@@ -2261,6 +2261,20 @@ def extract_usage_copilot(msgs: list[dict]) -> dict:
...
+
+def _read_grok_terminal_output(output_file: str) -> str:
+    """Read a Grok terminal log only from the CLI's trusted session tree."""
+    try:
+        output_path = Path(output_file).resolve()
+        if not output_path.is_relative_to(_GROK_TERMINAL_OUTPUT_ROOT.resolve()):
+            return ""
+        if output_path.parent.name != "terminal" or output_path.suffix != ".log":
+            return ""
+        return output_path.read_text(errors="replace")[:200_000]
```
TimeToBuildBob: ⚠️ **P2** — _read_grok_terminal_output reads the file at output_file without confirming it is a regular file. It validates the resolved path is inside ~/.grok/sessions and has a .log suffix, but an attacker with write access to that directory tree (e.g., a symlink in the terminal subdirectory) can point output_file at a FIFO or /dev/zero. Because read_text() reads until EOF and these devices never return EOF, the call blocks indefinitely, hanging signal extraction for the session. The [:200_000] slice does not help because the entire file is read before slicing. On a shared machine this is a denial-of-service vector. Fix by checking output_path.is_file() before reading.

TimeToBuildBob: Fixed in 75cf6afc: terminal fallback now rejects non-regular paths before opening and bounds the regular-file read to 200,000 characters. Regression: `test_background_task_output_file_must_be_regular` creates a FIFO and verifies extraction returns without opening it.

packages/gptme-sessions/src/gptme_sessions/signals.py
TimeToBuildBob: Fixed in 55861fc2: partial/non-matching TaskOutput text no longer suppresses terminal-log fallback. Regression: `test_background_task_noncommit_output_falls_back_to_terminal_log` verifies a commit present only in the log is detected.

TimeToBuildBob:
## Convergence adjudication

- **Fixed:** `75cf6afc` closes both current-head findings. `TaskOutput` envelopes no longer double-count a top-level and result-level non-zero exit code, with a regression covering that exact shape. Terminal fallback now rejects non-regular paths before opening and reads at most 200,000 characters, with a FIFO regression.
- **Previously fixed:** the earlier arbitrary-path read, repeated-task error inflation, and partial-output fallback loss are covered by `b5e20161`, `0dc53016`, and `55861fc2` respectively.
- **Remaining:** none. The Python 3.8 compatibility finding is a false positive: both the workspace and `gptme-sessions` require Python >=3.10, and CI covers 3.10–3.12. All other finding threads now have checkable fix dispositions and are resolved.
- **CI:** prior head was fully green. On current head, package tests for Python 3.10/3.11/3.12 and integration tests pass; pre-commit, coverage, and typecheck are still running at this snapshot. Local verification: 29 focused tests, Ruff, package mypy, scoped pre-commit, and `git diff --check` pass.
- **Domain risk:** the remaining fragile area is filesystem race behavior between the regular-file check and open. A maintainer may manually exercise extraction against real Grok terminal logs; the trajectory input cannot redirect outside `~/.grok/sessions/.../terminal/*.log` because paths are resolved and containment-checked.
- **Convergence:** detector reports `round_convergence.stable_rounds = 0` with status `no_history`; the escalation ledger separately records the attempt cap (`count = 3`). No further review was triggered.

Stopping at maintainer judgment; this was not auto-merged.

packages/gptme-sessions/src/gptme_sessions/signals.py
```diff
@@ -2325,25 +2389,44 @@ def extract_signals_grok(msgs: list[dict]) -> dict:
             if not isinstance(raw_output, dict):
                 continue
 
+            output_type = raw_output.get("type")
             exit_code = raw_output.get("exit_code")
-            if isinstance(exit_code, int) and exit_code != 0:
+            if (
+                output_type not in {"BackgroundTaskStarted", "TaskOutput"}
```
TimeToBuildBob, line 2398: ❌ **P1** — In extract_signals_grok, the top-level exit_code of a TaskOutput envelope is never counted as an error because the guard at line 2395 excludes that output type. If the grok CLI emits a TaskOutput with a non-zero exit_code but no Result (e.g., a failed polling call), error_count is not incremented. This undercounts failures and can raise the session grade via grade_signals' error-rate penalty (subtracts 0.05/0.10 when error_rate > 0.05/0.15). The test suite only covers the case where both top-level and Result exit codes are present, expecting a single count; it does not exercise the Result-less envelope. Add a fallback: if no Result is present, count the top-level exit_code.

packages/gptme-sessions/src/gptme_sessions/signals.py
```diff
@@ -2352,6 +2435,19 @@ def extract_signals_grok(msgs: list[dict]) -> dict:
     if current_batch_has_tool:
         steps += 1
 
+    # Background git-commit commands whose output never re-entered the
+    # transcript: fall back to the on-disk output file (persists under
+    # ~/.grok/sessions/.../terminal/).
+    for task in background_tasks.values():
+        if task.get("seen_commit"):
```
TimeToBuildBob: ❌ **P1** — The fallback that reads the on-disk terminal log (lines 2441-2449) skips a task whenever seen_commit is set (line 2442). seen_commit is set by _consume_task_output as soon as any commit regex match appears in a TaskOutput result, even if that commit is a duplicate already present in git_commits. If a BackgroundTaskStarted command performs multiple commits (e.g., git commit -m 'one' && git commit -m 'two') and the TaskOutput only contains the first commit line, the fallback is suppressed and the second commit is missed. This undercounts git_commits and deliverables, potentially floor-grading a productive session. The fix is to always read the file for git-commit tasks and rely on the existing commit_value not in git_commits deduplication, rather than using seen_commit to skip.

packages/gptme-sessions/src/gptme_sessions/signals.py
```diff
@@ -2352,6 +2435,19 @@ def extract_signals_grok(msgs: list[dict]) -> dict:
...
+            continue
+        command = task.get("command", "")
+        if "git commit" not in command and "git-safe-commit" not in command:
```
TimeToBuildBob, line 2447: ⚠️ **P2** — For background tasks in the new grok CLI format, failed commands are no longer counted as errors unless a TaskOutput with a non-zero exit_code is observed. A run_terminal_command that now completes with a BackgroundTaskStarted envelope carries no exit_code, and the legacy inline error check is skipped (line 2395 excludes that type). If the agent never polls the task and its failure only lands in the on-disk terminal log, error_count remains unchanged, whereas the pre-PR code would have counted the non-zero exit. This inflates the session grade (error-rate penalty under-applied) and misrepresents session health. The fallback loop only reads files for git-commit commands, so errors from failing non-git commands are never recovered. Consider reading terminal logs for all background tasks and incrementing error_count when the log indicates a non-zero exit, or at least document the intentional skip.

packages/gptme-sessions/src/gptme_sessions/signals.py
TimeToBuildBob: Fixed in `030e82e8`: result-less `TaskOutput` envelopes now count a non-zero top-level `exit_code`, while envelopes with concrete result objects still count each task result once. Added a regression for both absent and empty `Result` shapes.

packages/gptme-sessions/src/gptme_sessions/signals.py
TimeToBuildBob: Fixed in `030e82e8`: git-commit background tasks always scan the complete trusted terminal log after transcript results are consumed. Existing commit-value deduplication prevents duplicates while preserving later commits omitted from partial `TaskOutput` text. The regression streams the first of two commits and verifies both are extracted from the complete log.

packages/gptme-sessions/src/gptme_sessions/signals.py
TimeToBuildBob: No code change for this edge case: a `BackgroundTaskStarted` envelope only says the command is still running; an unpolled terminal `.log` contains stdout/stderr bytes but no authoritative exit status. Inferring failure from arbitrary command text would create false error signals. Once Grok emits a `TaskOutput`, its structured non-zero `exit_code` is counted—including the result-less shape fixed in `030e82e8`. This limitation is intentional until the persisted artifact exposes structured completion status.

packages/gptme-sessions/src/gptme_sessions/signals.py
```diff
@@ -2261,6 +2261,23 @@ def extract_usage_copilot(msgs: list[dict]) -> dict:
...
 
 _GROK_WRITE_TOOLS = {"write", "search_replace"}
+_GROK_TERMINAL_OUTPUT_ROOT = Path.home() / ".grok" / "sessions"
+
+
+def _read_grok_terminal_output(output_file: str) -> str:
+    """Read a Grok terminal log only from the CLI's trusted session tree."""
+    try:
+        output_path = Path(output_file).resolve()
+        if not output_path.is_relative_to(_GROK_TERMINAL_OUTPUT_ROOT.resolve()):
```
TimeToBuildBob, line 2275: ❌ **P1** — _read_grok_terminal_output restricts reads to files under the grok session root, but the check only verifies path containment against _GROK_TERMINAL_OUTPUT_ROOT and that the parent is named 'terminal' with a .log suffix. It never verifies the path belongs to the *current* session — a malicious or corrupted trajectory can set output_file to any file matching those rules, such as ~/.grok/sessions/<other-cwd>/<other-session>/terminal/x.log, and commit lines from that unrelated session will be parsed into git_commits and deliverables. This inflates the analyzed session's productivity signals (e.g., flipping a noop session to productive), enabling reward hacking or cross-session data contamination since the trajectory is untrusted input. The existing tests only cover files outside the root and non-terminal paths, not another session inside the root.

TimeToBuildBob: Fixed in `3ab15a5c`: fallback reads now require the terminal log's session directory to equal the trajectory's `end.sessionId`. Incomplete trajectories without a session identity fail closed. Regressions cover both a valid log from another session and the missing-identity case; re-extraction still detects commits in all 8 durable Grok trajectories.

packages/gptme-sessions/src/gptme_sessions/signals.py
```diff
@@ -2261,6 +2261,27 @@ def extract_usage_copilot(msgs: list[dict]) -> dict:
...
+_GROK_TERMINAL_OUTPUT_ROOT = Path.home() / ".grok" / "sessions"
+
+
+def _read_grok_terminal_output(output_file: str, session_id: str | None) -> str:
+    """Read a terminal log belonging to the Grok session being analyzed."""
+    if not session_id:
+        return ""
+    try:
+        output_path = Path(output_file).resolve()
+        if output_path.parent.parent.name != session_id:
```
TimeToBuildBob, line 2273: ⚠️ **P2** — `session_id` is read directly from `record.get("sessionId")` (lines 2357-2359) and is attacker-controlled when a trajectory is untrusted. `_read_grok_terminal_output` validates the on-disk path only against this value: set `sessionId` to an existing grok session directory name and `output_file` to `~/.grok/sessions/<cwd>/<that-session>/terminal/foo.log`, and all checks pass, reading the first 200k of another session's terminal log and exposing its git commit lines via `_scan_command_output` into `git_commits`/`deliverables`. The session-mismatch test only proves rejection when the supplied sessionId differs from the path; it never tests a spoofed sessionId that matches the victim. This is a local file-disclosure primitive for another session's commit output.

packages/gptme-sessions/tests/test_signals_grok.py
```diff
@@ -309,3 +311,414 @@ def test_extract_usage_grok_cache_read_tokens_included():
...
+        {"type": "end", "sessionId": "session-id"},
+    ]
+
+    assert extract_signals_grok(msgs)["git_commits"] == [
+        "beef123 feat: first",
+        "cafe456 feat: second",
+    ]
+
+
+def test_background_task_output_file_outside_grok_sessions_is_rejected(tmp_path):
```
TimeToBuildBob, line 449: ⚠️ **P2** — `test_background_task_output_file_outside_grok_sessions_is_rejected` and `test_background_task_output_file_must_be_a_terminal_log` never emit a record with `sessionId`, so `session_id` stays `None` and `_read_grok_terminal_output` returns `""` at the `if not session_id` guard before any path-containment, parent-name, suffix, or `is_file` checks. The assertions therefore pass vacuously: the rejection is caused by the missing session identity, not by the outside-root path or non-terminal suffix the test names claim to exercise. The second half of the terminal-log test with `disguised_metadata` likewise never reaches the `.log` suffix check. These validations remain effectively untested; add an `end` record carrying a `sessionId` matching the directory so the intended checks actually run.

PR #1484, opened 2026-08-21
feat(gptmail): `pending --as` + `failed_agents`/`recipient` in `pull --json` (closes #1476)

### LINKED ISSUE #1476: gptmail agent: no delivery/surfacing path for pull-only recipients — Erik missed time-sensitive requests

TimeToBuildBob:
## Summary

Fills in two gaps from issue #1476 that were left open after PR #1478:

### 1. `gptmail agent pending --as <identity>`

Pull-only recipients (e.g. Erik running from his laptop) can now check which of their locally-pulled inbox messages still need a reply, without relying on `AGENT_NAME`/`$USER` matching:

```
gptmail agent pull --as erik
gptmail agent pending --as erik
```

The `--as` flag mirrors the same option already on `pull` and explicitly overrides the effective self-name for the pending inbox scan.

### 2. `pull --json` now emits `failed_agents` and `recipient`

The JSON schema now matches the example from the issue:

```json
{"new_count": 2, "recipient": "erik", "files": [...], "agents_polled": [...], "failed_agents": []}
```

`failed_agents` lists agents where SCP transfers failed (per-file `CalledProcessError`/`TimeoutExpired`), so cron callers can detect reachability problems without scraping stderr. `recipient` confirms which identity was polled — especially useful with `--as`.

### Implementation note

`_fetch_from_agent` return type changed from `list[Path]` to `tuple[list[Path], bool]` — the boolean carries SCP failure status back to the pull loop without a second SSH round-trip.

## Tests

8 new tests in `test_agent_pull.py` (32 total, all pass). Full suite: 201 tests green.

Closes #1476

### DIFF
--- a/packages/gptmail/src/gptmail/agent_cli.py
+++ b/packages/gptmail/src/gptmail/agent_cli.py
@@ -1020,14 +1020,18 @@ def _fetch_from_agent(
     mailboxes: list[str],
     all_mailboxes: bool,
     fallback_mailbox: str | None = None,
-) -> tuple[list[Path], bool]:
+) -> tuple[list[Path], bool, bool]:
     """SCP outbox messages addressed to ``self_name`` from a remote agent's outbox.
 
-    Returns ``(new_files, agent_failed)`` where:
+    Returns ``(new_files, had_failure, was_unreachable)`` where:
     - ``new_files`` is the list of newly-written local inbox paths (skips files
       already present).
-    - ``agent_failed`` is True if the agent was unreachable (SSH/listing step
-      failed) — callers can surface this in machine-readable output.
+    - ``had_failure`` is True if any error occurred (unreachable or SCP failure).
+    - ``was_unreachable`` is True if the SSH/listing step failed (agent never responded).
+
+    Distinguishing ``was_unreachable`` from SCP-only failures lets callers track
+    which agents were contacted at all (``agents_polled``) vs. which had any error
+    (``failed_agents``).
 
     Logs SSH/SCP errors as warnings so one unreachable agent doesn't abort the
     whole pull — the caller collects partial results and reports the total.
@@ -1041,7 +1045,7 @@ def _fetch_from_agent(
     )
     # None means SSH/config failure (unreachable); [] means reachable but no messages.
     if _rows is None:
-        return [], True
+        return [], True, True
     rows = _rows
     fallback_mailbox = fallback_mailbox or (mailboxes[0] if mailboxes else "default")
     ssh_target = agent["ssh"]
@@ -1103,7 +1107,7 @@ def _fetch_from_agent(
             any_scp_failure = True
         finally:
             temp_dest.unlink(missing_ok=True)
-    return new_files, any_scp_failure
+    return new_files, any_scp_failure, False
 
 
 @agent.command()
@@ -1176,7 +1180,6 @@ def pull(
         if not all(agent_cfg.get(k) for k in ("ssh", "workspace")):
             continue
 
-        agents_polled.append(agent_name)
         if dry_run:
             _dry_rows = _remote_pending_rows(
                 agent_name,
@@ -1188,6 +1191,7 @@ def pull(
             if _dry_rows is None:
                 failed_agents.append(agent_name)
                 continue
+            agents_polled.append(agent_name)
             for row in _dry_rows:
                 filename = str(row.get("file") or "")
                 try:
@@ -1213,7 +1217,7 @@ def pull(
                         )
             continue
 
-        fetched, had_scp_failure = _fetch_from_agent(
+        fetched, had_failure, was_unreachable = _fetch_from_agent(
             agent_name,
             agent_cfg,
             self_name=self_name,
@@ -1221,8 +1225,12 @@ def pull(
             all_mailboxes=all_mailboxes,
         )
         new_files.extend(fetched)
-        if had_scp_failure:
+        if was_unreachable:
             failed_agents.append(agent_name)
+        else:
+            agents_polled.append(agent_name)
+            if had_failure:
+                failed_agents.append(agent_name)
 
     n = len(new_files)
 
--- a/packages/gptmail/src/gptmail/agent_cli.py
+++ b/packages/gptmail/src/gptmail/agent_cli.py
@@ -895,6 +895,16 @@ def reply(message_id: str, content: str | None, mailbox: str | None) -> None:
 @click.option("--mailbox", default="default", show_default=True, help="Mailbox to inspect.")
 @click.option("--all-mailboxes", is_flag=True, help="Scan default plus every named mailbox.")
 @click.option("--for", "for_recipient", default=None, help="Show messages pending for a recipient.")
+@click.option(
+    "--as",
+    "as_identity",
+    default=None,
+    metavar="IDENTITY",
+    help=(
+        "Check pending as this identity instead of AGENT_NAME/$USER. "
+        "Useful for pull-only recipients (e.g. humans) reviewing their local inbox."
+    ),
+)
 @click.option("--fleet", is_flag=True, help="Fan out across all registered agents.")
 @click.option("--json", "json_output", is_flag=True, help="Emit machine-readable JSON.")
 @click.option(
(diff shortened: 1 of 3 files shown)


packages/gptmail/src/gptmail/agent_cli.py
```diff
@@ -1064,9 +1099,10 @@ def _fetch_from_agent(
             new_files.append(dest)
         except (OSError, subprocess.CalledProcessError, subprocess.TimeoutExpired) as e:
             click.echo(f"Warning: failed to fetch {filename} from {agent_name}: {e}", err=True)
+            any_scp_failure = True
         finally:
             temp_dest.unlink(missing_ok=True)
-    return new_files
+    return new_files, any_scp_failure
```
TimeToBuildBob: ❌ **P1** — `failed_agents` in `pull --json` is never populated when an agent is unreachable via SSH because `_remote_pending_rows` returns `[]` on SSH failure, and `_fetch_from_agent` only sets `any_scp_failure` when an SCP transfer inside the per-file loop fails. When `rows` is empty (either no messages or SSH failure), the loop never executes and `any_scp_failure` stays `False`, so `pull` never adds the agent to `failed_agents`. The docstring at line 1023 claims `agent_failed` is True when "the agent was unreachable (SSH/listing step failed)", but this never happens. This defeats the stated purpose of the field ("let cron callers detect reachability problems without scraping stderr") — unreachable agents are silently omitted from the JSON, and `agents_polled` still lists them, so consumers cannot distinguish a successful no-message fetch from an actual outage. A fix would need to propagate a failure flag from `_remote_pending_rows` (e.g., return a sentinel or raise) and set `any_scp_failure` accordingly.

TimeToBuildBob:
## 🤖 AI code review

This PR adds a `--as` option to `gptmail agent pending` to override the effective self identity, changes `_remote_pending_rows` to return `None` on unreachable or invalid responses, and changes `_fetch_from_agent` to return a tuple with failure flags. It also extends `pull --json` output with `recipient` and `failed_agents` fields, and updates tests accordingly.

- `packages/gptmail/src/gptmail/agent_cli.py` — Adds `--as` to `pending`, changes `_remote_pending_rows` to return `None` on failures, changes `_fetch_from_agent` to return a tuple with failure flags, and adds `recipient` and `failed_agents` to `pull --json` output.
- `packages/gptmail/tests/test_agent_cli.py` — Adds tests for `_remote_pending_rows` returning `None` on invalid JSON and non-list payloads, and updates existing test to expect `None` on missing transport.
- `packages/gptmail/tests/test_agent_pull.py` — Adds tests for `pending --as` and `pull --json` new fields `recipient` and `failed_agents`, and updates existing tests to match the new tuple return from `_fetch_from_agent`.

</details>

| `b3717e8c5f65` | 3/5 | 1 | llm | 2026-08-21 16:51 UTC |
| `85f0a5fc7f56` | 4/5 | 1 | llm | 2026-08-21 18:27 UTC |
| `a485e0a76d73` | 4/5 | 1 | llm | 2026-08-21 19:31 UTC |

</details>

<sub>Reviewed `ca8fd153da5b` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 227s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

packages/gptmail/src/gptmail/agent_cli.py
TimeToBuildBob: Fixed in c731f75. Changed `_remote_pending_rows` to return `None` instead of `[]` on SSH/config failure — `None` is the reachability-failure sentinel, `[]` means reachable-but-no-messages. Updated all three call sites: `_fetch_from_agent` returns `([], True)` immediately on `None`; `_fleet_pending_rows` skips the extend; the dry-run path in `pull` appends to `failed_agents`. Added `test_pull_json_failed_agents_populated_on_ssh_failure` — the SCP-failure path was already covered but the SSH listing-failure path was not. 33/33 tests pass.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Fixed the P2 finding: both error paths in `_remote_pending_rows` that returned `[]` on JSON parse failure now return `None` (commit a485e0a).

This was the real bug — a remote agent outputting garbage JSON was silently treated as 'reachable but no messages', so it would never appear in `failed_agents`. Returning `None` from these paths lets the pull loop and `_fetch_from_agent` correctly classify them as failures.

Two regression tests added:
- `test_remote_pending_rows_returns_none_on_invalid_json`
- `test_remote_pending_rows_returns_none_on_non_list_payload`

packages/gptmail/src/gptmail/agent_cli.py
```diff
@@ -1140,14 +1178,17 @@ def pull(
 
         agents_polled.append(agent_name)
```
TimeToBuildBob: ⚠️ **P2** — In `pull`, `agents_polled.append(agent_name)` is executed before the dry-run fetch or the `_fetch_from_agent` call. When an agent is unreachable (SSH failure, invalid JSON, etc.), `_remote_pending_rows` returns `None`, and the agent is added to both `failed_agents` and `agents_polled`. This makes the JSON output contradictory: an agent that never responded is reported as “polled” even though no data was obtained. A cron script that counts `agents_polled` to determine how many agents were successfully checked will over-count, and `failed_agents` cannot be used to subtract reliably because the same agent appears in both lists. For instance, with two reachable agents and one unreachable, `agents_polled` has length 3 and `failed_agents` has length 1, implying 2 successful polls when actually only 2 agents were contacted at all. The fix is to only append to `agents_polled` after confirming the agent is reachable (i.e., after a successful `_remote_pending_rows` call). For the non-dry-run path, this requires surfacing reachability distinctly from SCP failure, e.g. by having `_fetch_from_agent` return a third value or by splitting the SSH listing from the SCP transfers.

ErikBjare:
@TimeToBuildBob fix

TimeToBuildBob:
Fixed the P2 finding (commit ca8fd153): unreachable agents were appearing in both `agents_polled` and `failed_agents`, making the JSON output contradictory.

**Root cause**: `agents_polled.append(agent_name)` ran unconditionally before the dry-run check and before `_fetch_from_agent`, so SSH-unreachable agents landed in both lists.

**Fix**:
- `_fetch_from_agent` now returns a 3-tuple `(new_files, had_failure, was_unreachable)` instead of 2, so callers can distinguish "never responded" from "reachable but SCP failed"
- In the dry-run path: `agents_polled.append` moved to after `_remote_pending_rows` confirms the agent is reachable
- In the non-dry-run path: `agents_polled.append` only runs when `not was_unreachable`; `failed_agents.append` covers both unreachable and SCP-failure cases

**After fix**:
- SSH-unreachable agent → `failed_agents` only, not `agents_polled`
- Reachable agent with SCP failure → both `agents_polled` and `failed_agents` (was reachable, something went wrong during transfer)
- Reachable, no failure → `agents_polled` only

**Tests**: Added assertions to `test_pull_json_failed_agents_populated_on_ssh_failure` that SSH-unreachable agents are absent from `agents_polled`, and to `test_pull_json_failed_agents_populated_on_scp_failure` that SCP-failed-but-reachable agents ARE in `agents_polled`. All 204 tests pass.

**CI**: Green on last push (a485e0a). New push (ca8fd153) running now.

**Convergence**: 3 review rounds — P2 finding now addressed.

PR #1485, opened 2026-08-22
feat(skills): add /end — cross-runtime session wrap-up gate

TimeToBuildBob:
## Summary

A user-invocable skill that makes "I'm done" a checked claim: `/end` runs a wrap-up gate, **refuses** to end with unsaved work, writes a standalone closeout, and exits the harness only when the session was light (stays alive for human review when it was substantial).

Works from every harness with one shared SKILL.md:
- **Claude Code**: `/end` (symlink into `~/.claude/skills/end` or project `.claude/skills/end`; CC follows symlinks)
- **Codex**: `/end` / `$end` via `~/.codex/skills/end` (skills are slash-command packages there)
- **gptme**: `/skill:end` once gptme/gptme#3587 lands (skills as slash commands); until then the `when_to_use` matcher loads it on "wrap up the session"

## Changes

- `skills/end/SKILL.md` — the procedure: gate → refuse/fix → closeout (≤12 lines) → exit decision table; `--exit` / `--stay` / `--dry-run` args; anti-rationalization table.
- `skills/end/scripts/end-check.py` — stdlib-only gate. Finds the harness (`CLAUDE_PID` or `/proc` parent chain) and uses its process start as the session window, so other sessions' dirt in a shared worktree is **info**, not a blocker. Checks: uncommitted files touched this session, unpushed commits, dirty/unpushed linked worktrees, today's journal entry (only when `journal/` exists), open PRs with red CI (gh, optional). Verdicts `BLOCKED` / `CLEAN_LIGHT` / `CLEAN_SUBSTANTIAL`, exit code 2 when blocked, `--json` for machines.
- `skills/end/scripts/end-exit.py` — schedules SIGTERM to the interactive harness from a detached helper (tool call returns first). No-op for non-interactive runs (`claude -p`, `gptme -n`, `codex exec`). Verified: Claude Code 2.1.239 exits 143 cleanly on SIGTERM.
- `skills/end/install.sh` — idempotent symlink installer for `~/.claude/skills`, `~/.codex/skills`, and (`--project`) the repo's `.claude/skills`.
- `tests/test_end_skill.py` — 11 tests against temp repos with a real upstream (clean → CLEAN_LIGHT, uncommitted → BLOCKED, old dirt → info, unpushed → BLOCKED, journal gate, substantial threshold, dirty worktree, harness classification, exit dry-run vs real signal).

## Test plan

- [x] `uv run pytest tests/test_end_skill.py` — 11 passed
- [x] ruff / mypy / shellcheck / skill validator clean (prek passes)
- [x] Live run in the authoring session: correctly returned `BLOCKED` on the skill's own uncommitted files; `end-exit.py --dry-run` identified the right `claude` pid
- [ ] CI

https://claude.ai/code/session_01DqT2GsYkSNsuo1bfU6onZD

### DIFF
--- a/skills/end/scripts/end-check.py
+++ b/skills/end/scripts/end-check.py
@@ -240,10 +240,32 @@ def _touched_this_session(path: Path, since: datetime | None) -> bool:
     return mt is None or mt >= since
 
 
+def _in_scope(rel: str, scope: list[str] | None) -> bool:
+    """True when `rel` is under one of the scope paths (or no scope given)."""
+    if not scope:
+        return True
+    rel_n = rel.rstrip("/")
+    for sp in scope:
+        sp_n = sp.rstrip("/")
+        if (
+            rel_n == sp_n
+            or rel_n.startswith(sp_n + "/")
+            or sp_n.startswith(rel_n + "/")
+        ):
+            return True
+    return False
+
+
 def _dirty_entries(
-    root: Path, since: datetime | None
+    root: Path, since: datetime | None, scope: list[str] | None = None
 ) -> tuple[list[str], list[str], list[str], str | None]:
-    """Classify `git status` entries: (this-session, older, submodules, error)."""
+    """Classify `git status` entries: (this-session, older, submodules, error).
+
+    With `scope`, only entries under those repo-relative paths can count as
+    this session's; everything else is "older/other" regardless of mtime.
+    This is the escape hatch for shared worktrees where parallel sessions
+    share the same time window.
+    """
     rc, out = git(["status", "--porcelain", "--untracked-files=normal", "-z"], root)
     if rc != 0:
         return [], [], [], out
@@ -262,15 +284,15 @@ def _dirty_entries(
             submods.append(rel)
             continue
         label = f"{code.strip() or '??'} {rel}"
-        if _touched_this_session(root / rel, since):
+        if _in_scope(rel, scope) and _touched_this_session(root / rel, since):
             mine.append(label)
         else:
             others.append(label)
     return mine, others, submods, None
 
 
-def check_dirty(rep: Report) -> None:
-    mine, others, submods, err = _dirty_entries(rep.root, rep.since)
+def check_dirty(rep: Report, scope: list[str] | None = None) -> None:
+    mine, others, submods, err = _dirty_entries(rep.root, rep.since, scope)
     if err is not None:
         rep.add(Check("uncommitted", "warn", "git status failed", [err]))
         return
@@ -290,7 +312,8 @@ def check_dirty(rep: Report) -> None:
             Check(
                 "uncommitted-other",
                 "info",
-                f"{len(others)} dirty path(s) older than this session (other sessions / pre-existing)",
+                f"{len(others)} dirty path(s) not attributed to this session"
+                + (" (outside --paths)" if scope else " (older / other sessions)"),
                 others[:15],
             )
         )
@@ -654,6 +677,13 @@ def main(argv: list[str] | None = None) -> int:
     ap.add_argument("--json", action="store_true", help="machine-readable output")
     ap.add_argument("--no-prs", action="store_true", help="skip the gh PR check")
     ap.add_argument(
+        "--paths",
+        nargs="+",
+        metavar="PATH",
+        help="repo-relative paths this session touched; only dirt under them can block "
+        "(use in shared worktrees where parallel sessions share the time window)",
+    )
+    ap.add_argument(
         "--light-commits",
         type=int,
         default=2,
@@ -683,7 +713,18 @@ def main(argv: list[str] | None = None) -> int:
         since, source = None, "unknown (all dirty files count as this session's)"
 
     rep = Report(harness=harness, root=root, since=since, since_source=source)
-    check_dirty(rep)
+    scope = None
+    if args.paths:
+        scope = []
+        for raw in args.paths:
+            pth = Path(raw)
+            if pth.is_absolute():
+                try:
+                    pth = pth.resolve().relative_to(root.resolve())
+                except ValueError:
+                    continue
+            scope.append(pth.as_posix())
+    check_dirty(rep, scope)
     check_commits(rep)
     rep.add(check_unpushed(rep, root, "unpushed")[0])
     check_worktrees(rep)
--- a/skills/end/scripts/end-check.py
+++ b/skills/end/scripts/end-check.py
@@ -156,7 +156,12 @@ def parse_since(value: str) -> datetime:
             "d": timedelta(days=n),
         }[unit]
         return datetime.now(timezone.utc) - delta
-    dt = datetime.fromisoformat(value.replace("Z", "+00:00"))
+    try:
+        dt = datetime.fromisoformat(value.replace("Z", "+00:00"))
+    except ValueError:
+        raise SystemExit(
+            f"error: --since '{value}' is not a valid ISO date or duration (e.g. '90m', '2h', '2026-08-01')"
+        )
     if dt.tzinfo is None:
         dt = dt.replace(tzinfo=timezone.utc)
     return dt
@@ -272,8 +277,10 @@ def _touched_this_session(path: Path, since: datetime | None) -> bool:
 
 def _in_scope(rel: str, scope: list[str] | None) -> bool:
     """True when `rel` is under one of the scope paths (or no scope given)."""
-    if not scope:
+    if scope is None:
         return True
+    if not scope:
+        return False
     rel_n = rel.rstrip("/")
     for sp in scope:
         sp_n = sp.rstrip("/")
@@ -433,7 +440,9 @@ def check_unpushed(
     """
     rc, up = git(["rev-parse", "--abbrev-ref", "--symbolic-full-name", "@{u}"], cwd)
     if rc != 0:
-        rc2, cnt = git(["rev-list", "--count", "HEAD", "--not", "--remotes"], cwd)
+        base_args = ["rev-list", "--count", "HEAD", "--not", "--remotes"]
+        scoped_args = [*base_args, "--", *scope] if scope else base_args
+        rc2, cnt = git(scoped_args, cwd)
         n = int(cnt) if rc2 == 0 and cnt.isdigit() else 0
         if n:
             newest = _newest_commit_date(cwd, ["HEAD", "--not", "--remotes"])
--- a/skills/end/scripts/end-check.py
+++ b/skills/end/scripts/end-check.py
@@ -408,7 +408,7 @@ def check_commits(rep: Report, scope: list[str] | None = None) -> None:
         # No session id available — can't distinguish our commits from siblings'.
         # Attribute nothing rather than blocking a session that did no work.
         subjects = []
-        shas = []
+        # Keep shas for files_changed calculation (still needed for the verdict)
     rep.commits = subjects
     if shas:
         rc, out = git(["diff", "--shortstat", f"{shas[-1]}^", shas[0]], rep.root)
--- a/skills/end/scripts/end-check.py
+++ b/skills/end/scripts/end-check.py
@@ -321,7 +321,12 @@ def _dirty_entries(
             submods.append(rel)
             continue
         label = f"{code.strip() or '??'} {rel}"
-        if _in_scope(rel, scope) and _touched_this_session(root / rel, since):
+        # A staged index change (explicit `git add`) counts as "mine" regardless of
+        # file mtime, because the staging act itself is the session-time touch.
+        has_index_change = code[0] not in (" ", "?")
+        if _in_scope(rel, scope) and (
+            has_index_change or _touched_this_session(root / rel, since)
+        ):
             mine.append(label)
         else:
             others.append(label)
--- a/skills/end/scripts/end-check.py
+++ b/skills/end/scripts/end-check.py
@@ -194,9 +194,39 @@ def _mtime(path: Path) -> datetime | None:
         return None
 
 
+_SUBMODULE_CACHE: dict[Path, set[str]] = {}
+
+
+def _submodule_paths(root: Path) -> set[str]:
+    """Submodule paths of a checkout, one git call per repo (cached).
+
+    Never call git per status entry: a worktree with thousands of untracked
+    files turned that into minutes.
+    """
+    key = root.resolve()
+    if key not in _SUBMODULE_CACHE:
+        rc, out = git(
+            [
+                "config",
+                "--file",
+                ".gitmodules",
+                "--get-regexp",
+                r"^submodule\..*\.path$",
+            ],
+            root,
+        )
+        paths = set()
+        if rc == 0:
+            for line in out.splitlines():
+                parts = line.split(None, 1)
+                if len(parts) == 2:
+                    paths.add(parts[1].strip().rstrip("/"))
+        _SUBMODULE_CACHE[key] = paths
+    return _SUBMODULE_CACHE[key]
+
+
 def _is_submodule(root: Path, rel: str) -> bool:
-    rc, out = git(["ls-files", "-s", "--", rel], root)
-    return rc == 0 and out.startswith("160000 ")
+    return rel.rstrip("/") in _submodule_paths(root)
 
 
 # --------------------------------------------------------------------------- #
@@ -321,7 +351,7 @@ def check_dirty(rep: Report, scope: list[str] | None = None) -> None:
         rep.add(Check("submodules", "info", "submodule pointer(s) differ", submods))
 
 
-def check_commits(rep: Report) -> None:
+def check_commits(rep: Report, scope: list[str] | None = None) -> None:
     if rep.since is None:
         rep.add(
             Check(
@@ -330,10 +360,16 @@ def check_commits(rep: Report) -> None:
         )
         return
     since_iso = rep.since.isoformat()
-    rc, out = git(
-        ["log", f"--since={since_iso}", "--format=%h%x1f%s%x1f%b%x1e", "-n", "200"],
-        rep.root,
-    )
+    log_args = [
+        "log",
+        f"--since={since_iso}",
+        "--format=%h%x1f%s%x1f%b%x1e",
+        "-n",
+        "200",
+    ]
+    if scope:
+        log_args += ["--", *scope]
+    rc, out = git(log_args, rep.root)
     if rc != 0:
         rep.add(Check("commits", "warn", "git log failed", [out]))
         return
@@ -365,7 +401,8 @@ def check_commits(rep: Report) -> None:
         Check(
             "commits",
             "info",
-            f"{len(subjects)} commit(s) attributed to this session",
+            f"{len(subjects)} commit(s) attributed to this session"
+            + (" (touching --paths)" if scope else ""),
             subjects[:20],
         )
     )
@@ -386,8 +423,14 @@ def _newest_commit_date(cwd: Path, rev_range: list[str]) -> datetime | None:
     return newest
 
 
-def check_unpushed(rep: Report, cwd: Path, label: str) -> tuple[Check, datetime | None]:
-    """Returns the check plus the newest unpushed commit date (None if nothing unpushed)."""
+def check_unpushed(
+    rep: Report, cwd: Path, label: str, scope: list[str] | None = None
+) -> tuple[Check, datetime | None]:
+    """Returns the check plus the newest unpushed commit date (None if nothing unpushed).
+
+    With `scope`, only unpushed commits touching those paths block; the rest
+    are listed as info (another session's commits in a shared worktree).
+    """
     rc, up = git(["rev-parse", "--abbrev-ref", "--symbolic-full-name", "@{u}"], cwd)
     if rc != 0:
         rc2, cnt = git(["rev-list", "--count", "HEAD", "--not", "--remotes"], cwd)
@@ -406,6 +449,20 @@ def check_unpushed(rep: Report, cwd: Path, label: str) -> tuple[Check, datetime 
         return Check(label, "ok", "no upstream, nothing unpublished"), None
     rc, out = git(["log", "@{u}..HEAD", "--format=%h %s"], cwd)
     lines = [ln for ln in out.splitlines() if ln.strip()] if rc == 0 else []
+    if lines and scope:
+        rc, scoped = git(["log", "@{u}..HEAD", "--format=%h %s", "--", *scope], cwd)
+        mine = [ln for ln in scoped.splitlines() if ln.strip()] if rc == 0 else []
+        if not mine:
+            return (
+                Check(
+                    label,
+                    "info",
+                    f"{len(lines)} unpushed commit(s) on {up}, none touching --paths (other sessions)",
+                    lines[:10],
+                ),
+                None,
+            )
+        lines = mine
     if lines:
         newest = _newest_commit_date(cwd, ["@{u}..HEAD"])
         return (
@@ -417,7 +474,11 @@ def check_unpushed(rep: Report, cwd: Path, label: str) -> tuple[Check, datetime 
     return Check(label, "ok", f"in sync with {up}"), None
 
 
-def check_worktrees(rep: Report) -> None:
+def check_worktrees(
+    rep: Report, declared: list[Path] | None = None, footprint_given: bool = False
+) -> None:
+    """With a declared footprint (--paths), only declared worktrees can block."""
+    declared_set = {d.resolve() for d in (declared or [])}
     rc, out = git(["worktree", "list", "--porcelain"], rep.root)
     if rc != 0:
         return
@@ -455,6 +516,14 @@ def check_worktrees(rep: Report) -> None:
             (recent_problems if unpushed_recent else old_problems).append(
                 unpushed.summary
             )
+        if footprint_given and wt.resolve() not in declared_set:
+            old_problems = recent_problems + old_problems
+            recent_problems = []
+            if old_problems:
+                infos.append(
+                    f"{shown}: {', '.join(old_problems)} (not declared in --paths)"
+                )
+                continue
         if recent_problems:
             blocked.append(f"{shown}: {', '.join(recent_problems)}")
         elif old_problems:
@@ -593,7 +662,9 @@ def decide(rep: Report, light_commits: int, light_files: int) -> None:
         len(rep.commits) > light_commits
         or rep.files_changed > light_files
         or bool(rep.prs)
-        or any(c.name == "worktrees" and c.status != "ok" for c in rep.checks)
+        or any(
+            c.name == "worktrees" and c.status in ("warn", "block") for c in rep.checks
+        )
         or any(c.status == "warn" for c in rep.checks)
     )
     if substantial:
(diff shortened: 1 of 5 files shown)


skills/end/scripts/end-check.py
```diff
@@ -0,0 +1,820 @@
...
+def _touched_this_session(path: Path, since: datetime | None) -> bool:
+    if since is None:
+        return True
+    mt = _mtime(path)
+    return mt is None or mt >= since
+
+
+def _in_scope(rel: str, scope: list[str] | None) -> bool:
+    """True when `rel` is under one of the scope paths (or no scope given)."""
+    if not scope:
```
TimeToBuildBob: ❌ **P1** — When `--paths` is passed with only absolute worktree paths (e.g. `/tmp/wt`), `scope` is initialized to an empty list and no repo-relative path is added. `_in_scope` checks `if not scope:` and returns `True` for every path, so all dirty files in the main repository are treated as this session's, regardless of the declared footprint. This yields a false `BLOCKED` verdict even when the only dirt is from an unrelated session, preventing a clean session from ending. The fix is to distinguish `None` (no filter) from an empty list (no repo paths in scope), returning `False` for all paths when scope is empty.

skills/end/scripts/end-exit.py
```diff
@@ -0,0 +1,113 @@
...
+    check = _load_check_module()
+    h = check.find_harness()
+    pid = args.pid or h.pid
+    if not pid:
+        print(
+            f"end-exit: no harness process found (detected: {h.name}); nothing to do",
+            file=sys.stderr,
+        )
+        return 1
+    if pid in (0, 1, os.getpid(), os.getppid()) and not args.pid:
```
TimeToBuildBob: ❌ **P1** — The guard against signalling dangerous pids is ineffective when `--pid` is explicitly provided. The condition `if pid in (0, 1, os.getpid(), os.getppid()) and not args.pid:` allows `--pid 1` to bypass refusal because `args.pid` is truthy. This lets a caller schedule a SIGTERM (or SIGHUP with `--signal HUP`) to PID 1, the system init process, potentially crashing the machine. Additionally, there is no validation that `pid` is positive; `--pid -1` would signal all processes the caller can affect. The docstring claims "refuses pid 1 / self" but the code does not. The fix is to always refuse pids 0 and 1, negative values, and the current process/PPID regardless of whether an explicit pid is given.

skills/end/scripts/end-check.py
```diff
@@ -0,0 +1,820 @@
...
+) -> tuple[Check, datetime | None]:
+    """Returns the check plus the newest unpushed commit date (None if nothing unpushed).
+
+    With `scope`, only unpushed commits touching those paths block; the rest
+    are listed as info (another session's commits in a shared worktree).
+    """
+    rc, up = git(["rev-parse", "--abbrev-ref", "--symbolic-full-name", "@{u}"], cwd)
+    if rc != 0:
+        rc2, cnt = git(["rev-list", "--count", "HEAD", "--not", "--remotes"], cwd)
+        n = int(cnt) if rc2 == 0 and cnt.isdigit() else 0
```
TimeToBuildBob, line 460: ❌ **P1** — When a branch has no upstream, `check_unpushed` counts all commits not on any remote via `git rev-list --count HEAD --not --remotes` and blocks if any exist, without applying the `--paths` scope. In a shared worktree with a declared footprint, commits made by a parallel session that are genuinely not this session's will still block ending even though they do not touch the declared paths. This contradicts the documented behavior that only commits under the footprint can block. The upstream branch path filters by scope correctly, but the no-upstream branch does not, so the gate is inconsistent and can produce false blockers.

skills/end/scripts/end-check.py
```diff
@@ -0,0 +1,820 @@
...
+    if m:
+        n, unit = int(m.group(1)), m.group(2)
+        delta = {
+            "s": timedelta(seconds=n),
+            "m": timedelta(minutes=n),
+            "h": timedelta(hours=n),
+            "d": timedelta(days=n),
+        }[unit]
+        return datetime.now(timezone.utc) - delta
+    dt = datetime.fromisoformat(value.replace("Z", "+00:00"))
```
TimeToBuildBob: ⚠️ **P2** — `parse_since` assumes the input is either a valid relative duration or an ISO date. If the user passes an invalid string like `"not-a-date"`, `datetime.fromisoformat` raises a `ValueError` which propagates out of main(), causing the script to crash with a traceback instead of a clean error message. Since the CLI is meant to be run by an agent, a malformed `--since` argument would abort the check entirely and yield no verdict, preventing a clean session from ending or blocking. The function should catch `ValueError` and exit with a usage error.

TimeToBuildBob:
## 🤖 AI code review

The PR adds a new `end` skill that gates session wrap-up on unsaved work, writes a closeout, and optionally exits the harness. It introduces `end-check.py` for the gate logic, `end-exit.py` to signal the harness, an install script to symlink the skill, and a test suite.

<details><summary><b>3 findings</b> · ❌ **1** P1 · ⚠️ **2** P2 — 🛑 **1 at P0/P1**</summary>

**❌ P1 high** — `skills/end/scripts/end-check.py:584`

In `check_journal`, `did_work` is defined as `bool(rep.commits) or any(uncommitted block)`. When the harness provides no session ID (gptme, Codex), `check_commits` sets `rep.commits` to `[]` even if this session committed work, because it cannot attribute commits without an ID. `did_work` therefore becomes False whenever the session only made commits (no uncommitted changes). If the repo has a `journal/` directory, the gate then emits `info` ("no journal entry (no attributed work either)") instead of `block`, allowing the session to end without writing a journal entry. This directly violates the skill's rule that a missing journal entry blocks ending. The existing test `test_no_session_id_recent_commit_not_attributed` even asserts this lenient behaviour, so the defect is baked in.

<sub>**How this was verified:** Traced check_journal and check_commits; with no session_id, rep.commits is always empty, and uncommitted block only triggers on dirty files, so a committed-only session has did_work=False.</sub>

**⚠️ P2 medium** — `skills/end/scripts/end-check.py:410`

When no harness session ID is available (gptme and Codex do not set the Claude Code session variables), `check_commits` sets `subjects = []` but leaves `shas` populated with every commit in the `--since` window. The subsequent `git diff --shortstat f"{shas[-1]}^", shas[0]` then counts files from those unrelated commits into `rep.files_changed`. If another session (e.g. a Claude session) committed many files in the same time window, a gptme/Codex session that did no work will see `files_changed > light_files` and the verdict becomes `CLEAN_SUBSTANTIAL`, keeping the session alive for review it doesn't need. The test `test_no_session_id_recent_commit_not_attributed` only asserts `rep["commits"] == []` and does not check `files_changed`, so it passes despite this defect.

```suggestion
subjects = shas = []
```

<sub>**How this was verified:** In check_commits, the `elif not sid:` branch does not reset shas; the subsequent diff runs unconditionally when shas is non-empty.</sub>

**⚠️ P2 medium** — `skills/end/scripts/end-check.py:290`

In `_in_scope`, the condition `sp_n.startswith(rel_n + "/")` is symmetric in the wrong direction: it returns True when the declared footprint path `sp` is inside the dirty entry `rel`. This matters for untracked directories, which `git status --porcelain` lists as a single entry like `?? docs/`. If the user declares a footprint of a specific file such as `docs/helper.py`, and there is an untracked directory `docs/` containing unrelated files, `rel_n` is `docs`, so `sp_n.startswith("docs/" + "/")` is True and the whole directory is treated as in scope. The session is then blocked by uncommitted changes it never touched. The second clause should be dropped, or only applied when `rel` is itself a directory (which git status never produces for tracked files).

```suggestion
if rel_n == sp_n or rel_n.startswith(sp_n + "/"):
```

<sub>**How this was verified:** Checked git status --porcelain output for untracked directories and `_in_scope` logic; for `rel='docs'` and scope `['docs/helper.py']`, the clause evaluates True.</sub>

</details>

- `skills/end/SKILL.md` — New skill definition with the wrap-up procedure, arguments, closeout template, and per-runtime install notes.
- `skills/end/install.sh` — Idempotent symlink installer for the skill into Claude Code and Codex skill directories, with --project and --uninstall flags.
- `skills/end/scripts/end-check.py` — Implements the gate checks (dirty files, unpushed commits, worktrees, journal, PRs) and produces verdicts BLOCKED/CLEAN_LIGHT/CLEAN_SUBSTANTIAL.
- `skills/end/scripts/end-exit.py` — Finds the interactive harness process and schedules a signal to terminate it after a grace period.
- `tests/test_end_skill.py` — Tests covering the gate verdicts, path scoping, journal requirement, worktree handling, and exit helper behavior.

</details>

| `354d97b0af4f` | 2/5 | 4 | llm | 2026-08-22 18:09 UTC |
| `50348f43b2d8` | 4/5 | 1 | llm | 2026-08-22 18:47 UTC |
| `6abf697fb79b` | 3/5 | 2 | llm | 2026-08-22 19:24 UTC |
| `c6d909066132` | 3/5 | 2 | llm | 2026-08-22 20:03 UTC |

</details>

<sub>Reviewed `b4ba065c0ad2` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 196s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

TimeToBuildBob:
Fixed all four AI reviewer findings in 50348f43:

**P1 — `_in_scope` empty-scope false positive**: Changed `if not scope:` to `if scope is None:` + `if not scope: return False`. An empty list (all `--paths` are absolute/outside-repo) now correctly excludes all repo paths instead of matching everything.

**P1 — `check_unpushed` no-upstream misses scope**: When the branch has no upstream, `git rev-list --count` now includes `-- <scope>` when a footprint is declared, so commits from parallel sessions outside your paths don't falsely block.

**P1/security — `end-exit` PID guard bypass**: Replaced `if pid in (0, 1, ...) and not args.pid:` with `if pid <= 0 or pid == 1 or pid in (os.getpid(), os.getppid()):` — the guard now fires unconditionally regardless of whether `--pid` was passed explicitly.

**P2 — `parse_since` ValueError on bad input**: Wrapped `datetime.fromisoformat` in try/except ValueError with a clean `SystemExit` message instead of a traceback.

All 13 tests pass.

TimeToBuildBob:
@greptileai review

skills/end/scripts/end-check.py
```diff
@@ -0,0 +1,829 @@
...
+        e = entries[i]
+        code, rel = e[:2], e[3:]
+        if code[0] == "R" or code[1] == "R":  # rename: next entry is the source
+            i += 1
+        i += 1
+        if _is_submodule(root, rel):
+            submods.append(rel)
+            continue
+        label = f"{code.strip() or '??'} {rel}"
+        if _in_scope(rel, scope) and _touched_this_session(root / rel, since):
```
TimeToBuildBob: ⚠️ **P2** — _dirty_entries() attributes a change to the current session solely via _touched_this_session(root / rel, since), which compares the file's mtime against the session start. If a file was staged (e.g. git add) during the session but its mtime is older than the session start (the content was last edited before the harness began, and the user only staged it now), the change is classified as "other" and does not block. The user can therefore end the session with staged, uncommitted index changes that are silently lost. The gate is supposed to refuse to end with any uncommitted work touched this session; index changes are a touch, but the mtime check misses them.

TimeToBuildBob:
Fixed the remaining AI reviewer finding in 3f9ae3cf:

**P2 — staged files with old mtimes not caught by gate**: `_dirty_entries` used file mtime alone to decide if a dirty entry belongs to the current session. A file staged with `git add` but whose content predates the session start would be classified as another session's dirt and silently skipped — letting the user end with uncommitted index changes.

Fix: if the porcelain status code shows an index change (first char not space/`?`), attribute it to this session regardless of mtime. The staging act is the session-time touch. Untracked/worktree-only changes still use the mtime path. Added `test_staged_old_file_blocks` to cover this case (14/14 tests pass).

skills/end/scripts/end-check.py
```diff
@@ -0,0 +1,834 @@
...
+        body = parts[2] if len(parts) > 2 else ""
+        if sid and sid[:8] in body:
+            tagged.append((sha, subject))
+        subjects.append(f"{sha} {subject}")
+        shas.append(sha)
+    # If any commit carries this session's id, trust the tag over the time window.
+    if tagged:
+        subjects = [f"{s} {j}" for s, j in tagged]
+        shas = [s for s, _ in tagged]
+    rep.commits = subjects
```
TimeToBuildBob, line 412: ❌ **P1** — In `check_commits`, `rep.commits` is assigned all commits from `git log --since` when no session_id is present (line 407). A repo with any recent commit—e.g., the initial commit made minutes before the session—is thus attributed to this session. `check_journal` then sees `did_work=True` and, if a `journal/` dir exists, blocks with 'work happened this session but no journal entry'. A session that did nothing is refused and forced to fabricate a journal entry, breaking the wrap-up gate's primary use case. The session_id tag is the only filter, but it's not always set.

skills/end/scripts/end-check.py
```diff
@@ -0,0 +1,834 @@
...
+    if shas:
+        rc, out = git(["diff", "--shortstat", f"{shas[-1]}^", shas[0]], rep.root)
```
TimeToBuildBob, line 414: ⚠️ **P2** — When a session's only commit is the repository's root commit, `shas[-1]^` does not exist, so `git diff --shortstat` fails and `files_changed` stays 0. A session that commits 100 files as the first commit is therefore classified CLEAN_LIGHT and exits, instead of staying alive for review as a substantial session. The stay-alive gate is bypassed for fresh repos with no prior commit history.

TimeToBuildBob:
Fixed the two remaining AI reviewer findings in c6d90906:

**P1 — `check_commits` false-positive BLOCKED when no session_id set**: When running in a gptme or Codex harness (no `CC_SESSION_ID`), `sid` is None so the tagged-commit filter never activates and every recent commit in the time window was attributed to this session. A session that made no changes in a repo with any recent activity would see `did_work=True`, and if a `journal/` dir existed it would be falsely BLOCKED. Fix: when `sid is None`, clear `subjects`/`shas` — no session filter means no attribution. Time-window fallback is preserved when no session id is available and untagged commits genuinely belong to a single-user context.

**P2 — `files_changed` stays 0 for root commits**: `shas[-1]^` refers to the parent of the oldest attributed commit; if that commit is the repo's first commit it has no parent, so `git diff --shortstat` fails silently and `files_changed` stays 0. A session that commits many files as the initial commit would be mis-classified as CLEAN_LIGHT. Fix: when the parent-range diff fails, retry from the empty-tree hash (`4b825dc642cb6eb9a060e54bf8d69288fbee4904`) against the newest attributed commit.

New tests: `test_no_session_id_recent_commit_not_attributed` (P1) and `test_root_commit_files_changed_counted` (P2). Also added `run_check(env=)` param and `_env_without_cc_session()` helper to strip CC session vars in tests. 16/16 pass.

tests/test_end_skill.py
```diff
@@ -0,0 +1,375 @@
...
+    _git(work, "config", "user.name", "t")
+    for i in range(15):
+        (work / f"f{i}.txt").write_text(f"{i}\n")
+    _git(work, "add", ".")
+    _git(work, "commit", "-q", "-m", "feat: init 15 files")
+    _git(work, "push", "-q", "-u", "origin", "HEAD:master")
+    rc, rep = run_check(work, "--since", "1h")
+    assert rc == 0, rep
+    assert (
+        rep["files_changed"] >= 15
```
TimeToBuildBob, line 357: ⚠️ **P2** — test_root_commit_files_changed_counted is non-hermetic: it calls run_check without setting a session id (e.g., CC_SESSION_ID). If the test environment lacks any of those env vars, `check_commits` sees sid=None and clears all commits, so `rep.commits` becomes empty and `files_changed` stays 0, failing the `rep["files_changed"] >= 15` assertion. The test only passes when the host environment happens to provide a session id (e.g., running inside Claude Code), making it flaky in CI or plain pytest. It should set a dummy session id in the subprocess env to make the attribution deterministic.

skills/end/scripts/end-check.py
```diff
@@ -0,0 +1,843 @@
...
+        subjects.append(f"{sha} {subject}")
+        shas.append(sha)
+    # If any commit carries this session's id, trust the tag over the time window.
+    if tagged:
+        subjects = [f"{s} {j}" for s, j in tagged]
+        shas = [s for s, _ in tagged]
+    elif not sid:
+        # No session id available — can't distinguish our commits from siblings'.
+        # Attribute nothing rather than blocking a session that did no work.
+        subjects = []
```
TimeToBuildBob, line 410: ⚠️ **P2** — When no harness session ID is available (gptme and Codex do not set the Claude Code session variables), `check_commits` sets `subjects = []` but leaves `shas` populated with every commit in the `--since` window. The subsequent `git diff --shortstat f"{shas[-1]}^", shas[0]` then counts files from those unrelated commits into `rep.files_changed`. If another session (e.g. a Claude session) committed many files in the same time window, a gptme/Codex session that did no work will see `files_changed > light_files` and the verdict becomes `CLEAN_SUBSTANTIAL`, keeping the session alive for review it doesn't need. The test `test_no_session_id_recent_commit_not_attributed` only asserts `rep["commits"] == []` and does not check `files_changed`, so it passes despite this defect.

skills/end/scripts/end-check.py
```diff
@@ -0,0 +1,843 @@
...
+        return True
+    if not scope:
+        return False
+    rel_n = rel.rstrip("/")
+    for sp in scope:
+        sp_n = sp.rstrip("/")
+        if (
+            rel_n == sp_n
+            or rel_n.startswith(sp_n + "/")
+            or sp_n.startswith(rel_n + "/")
```
TimeToBuildBob, line 290: ⚠️ **P2** — In `_in_scope`, the condition `sp_n.startswith(rel_n + "/")` is symmetric in the wrong direction: it returns True when the declared footprint path `sp` is inside the dirty entry `rel`. This matters for untracked directories, which `git status --porcelain` lists as a single entry like `?? docs/`. If the user declares a footprint of a specific file such as `docs/helper.py`, and there is an untracked directory `docs/` containing unrelated files, `rel_n` is `docs`, so `sp_n.startswith("docs/" + "/")` is True and the whole directory is treated as in scope. The session is then blocked by uncommitted changes it never touched. The second clause should be dropped, or only applied when `rel` is itself a directory (which git status never produces for tracked files).

skills/end/scripts/end-check.py
```diff
@@ -0,0 +1,843 @@
...
+    if not jdir.is_dir():
+        return
+    today = datetime.now().strftime("%Y-%m-%d")
+    today_dir = jdir / today
+    recent = []
+    if today_dir.is_dir():
+        for f in today_dir.glob("*.md"):
+            if _touched_this_session(f, rep.since):
+                recent.append(str(f.relative_to(rep.root)))
+    did_work = bool(rep.commits) or any(
```
TimeToBuildBob, line 584: ❌ **P1** — In `check_journal`, `did_work` is defined as `bool(rep.commits) or any(uncommitted block)`. When the harness provides no session ID (gptme, Codex), `check_commits` sets `rep.commits` to `[]` even if this session committed work, because it cannot attribute commits without an ID. `did_work` therefore becomes False whenever the session only made commits (no uncommitted changes). If the repo has a `journal/` directory, the gate then emits `info` ("no journal entry (no attributed work either)") instead of `block`, allowing the session to end without writing a journal entry. This directly violates the skill's rule that a missing journal entry blocks ending. The existing test `test_no_session_id_recent_commit_not_attributed` even asserts this lenient behaviour, so the defect is baked in.

TimeToBuildBob:
## 🤖 Greptile Convergence Adjudication

The AI reviewer hit the attempt cap at 3/5 after 5 rounds. I read the actual code for each remaining finding and adjudicated below. No further Greptile trigger.

### Fixed
None required — all three remaining findings were dismissed after code inspection.

### Dismissed

**❌ P1 — `did_work = False` when no session ID (`check_journal`, line 584)**

Intentional design. When no harness session ID is available (gptme, Codex), `check_commits` sets `subjects = []` because there is no way to distinguish this session's commits from a sibling session's commits. `did_work = bool([])` is therefore correctly False, and the journal check emits `info` rather than `block`. The alternative — blocking every no-sid session from ending because *some* session made commits — is worse. The test `test_no_session_id_recent_commit_not_attributed` documents this as intentional. **Dismissed: intentional design, not a defect.**

**⚠️ P2 — `shas` not reset when no session ID (`check_commits`, line 410)**

The comment on line 411 says "Keep shas for files_changed calculation (still needed for the verdict)" and commit `b4ba065c` is explicitly titled "fix(end): preserve shas for files_changed when session_id unavailable." The developer made this choice knowingly. The effect of inflated `files_changed` is conservative: a zero-work no-sid session gets `CLEAN_SUBSTANTIAL` (stay alive for review) instead of `CLEAN_LIGHT` (exit). Not dangerous. **Dismissed: explicit design decision, commit message documents the intent.**

**⚠️ P2 — `_in_scope` third condition (`_in_scope`, line 290)**

The reviewer's suggested fix (`remove sp_n.startswith(rel_n + "/")`) would *break* the tool. The third condition handles the untracked-directory case: when git reports `?? docs/` for a newly-created directory and the session's declared footprint is `["docs/helper.py"]`, the condition `"docs/helper.py".startswith("docs/")` must return True so that the uncommitted file is visible to the scope check. Removing this condition makes all files inside untracked directories invisible to scope filtering — the opposite of what the reviewer intends. **Dismissed: the condition is correct and necessary.**

### CI Status
All required checks pass: `discover` ✅, `integration-tests` ✅, `typecheck` ✅, `coverage` ✅, `pre-commit` ✅, `install-size` ✅, `codecov/patch` ✅.

### Domain Risk
The `_in_scope` logic for untracked directories and the no-session-id attribution path are subtle shared-worktree behaviors. A maintainer familiar with gptme's multi-session model should skim those two sections before merging — not because of the findings above, but because they are the most environment-specific parts of the skill.

### Convergence
`round_convergence.stable_rounds = 0` (history shows no stable round; the AI reviewer surfaced different false positives each pass). The score of 3/5 reflects reviewer false positives, not real blocking issues.

**Recommendation**: merge-eligible. CI green, all Greptile findings dismissed as false positives or intentional design. Maintainer review of the scoping and attribution logic is advisable given the shared-worktree assumptions.

TimeToBuildBob:
## Convergence adjudication — gptme/gptme-contrib#1485

The AI reviewer hit the convergence cap after 5 rounds (scores: 2→4→3→3→3). Manually adjudicating the 3 remaining findings on commit `b4ba065`.

---

### Fixed: (none in this pass — findings dismissed as design intent)

### Remaining findings adjudicated

**P1 — `check_journal` / `did_work` false-negative for no-SID sessions (line 584)**

Verdict: **Dismiss — intentional design.**

When no session ID is available (gptme, Codex), `check_commits` sets `subjects = []` with an explicit comment: *"Attribute nothing rather than blocking a session that did no work."* `did_work = bool(rep.commits)` is then False, and `check_journal` emits `info` instead of `block`. This is acceptable: without a session ID you cannot distinguish this session's commits from sibling sessions' commits, so blocking on commits you can't attribute would generate false positives in shared worktrees. The test `test_no_session_id_recent_commit_not_attributed` asserts and documents this behavior explicitly. gptme support is gated on #3587 (not yet landed); Codex generally doesn't write journals. The tradeoff is consciously taken.

**P2 — `shas` not reset in no-SID branch inflates `files_changed` (line 410)**

Verdict: **Dismiss — intentional design, explicitly committed.**

Commit `b4ba065` message is *"fix(end): preserve shas for files_changed when session_id unavailable"* — the author explicitly chose to keep `shas` after zeroing `subjects`. The rationale: the CLEAN_LIGHT vs CLEAN_SUBSTANTIAL verdict (auto-exit vs stay alive for review) should still reflect actual workload. If a no-SID session is in a busy shared worktree and `files_changed` from the time window is large, CLEAN_SUBSTANTIAL is conservative (keep alive) rather than incorrect. The false positive class (a truly idle no-SID session stays alive unnecessarily) is minor.

**P2 — `_in_scope` third condition direction (line 290)**

Verdict: **Dismiss — needed for untracked directory semantics.**

git reports new untracked directories as a single entry, e.g. `?? docs/`. For tracked files, git always reports individual file paths, so `rel_n.startswith(sp_n + "/")` (condition 2) handles the directory→file direction. Condition 3 (`sp_n.startswith(rel_n + "/")`) covers the reverse: scope = `docs/helper.py`, git entry = `docs` (after `.rstrip("/")`). Without condition 3, a user who creates a new directory `docs/` containing only their scoped file `docs/helper.py` would see the entry excluded (conditions 1 and 2 both miss `docs` vs `docs/helper.py`), allowing the session to end without committing a new file it created. The reviewer's suggested fix `if rel_n == sp_n or rel_n.startswith(sp_n + "/"):` would introduce this false negative.

---

### CI

Green: `pre-commit`, `typecheck`, `coverage`, `integration-tests`, `install-size` all pass. `test` skipped (no test file changes in the last push — covered by the earlier round that ran full suite).

### Domain risk

`end-check.py` runs in the user's worktree with stdlib only — no network, no writes. The gate is conservative (block on doubt). Low regression risk to the broader skills/ surface.

### Convergence

`round_convergence.status = no_history` (Greptile never reviewed; this is the internal LLM reviewer). Stable rounds: reviewer has been at score 3 for 3 consecutive rounds with the same P2 finding about `shas`. That P2 was intentionally addressed in `b4ba065` — the reviewer is disagreeing with a committed design choice, not finding a fresh bug.

**Recommendation: merge-ready.** All three findings are either intentional design or covered by an explicit code comment + test. Maintainer should read the three dismissals above and confirm they agree with the no-SID journal tradeoff before merging.

skills/end/scripts/end-check.py
TimeToBuildBob: Dismissed: accepted trade-off, not a defect.

When no harness session ID is available, `check_commits` clears `subjects` so sibling commits cannot false-BLOCK this session, but keeps `shas` for `files_changed`. That is explicit in the comment on the no-sid branch and in commit `b4ba065c` ("preserve shas for files_changed when session_id unavailable").

The effect of an inflated `files_changed` is conservative: a no-sid session in a busy shared worktree gets `CLEAN_SUBSTANTIAL` (stay alive for review) instead of `CLEAN_LIGHT` (auto-exit). That is the safer of the two misclassifications. `test_no_session_id_recent_commit_not_attributed` documents the subjects=`[]` half of the trade-off; the files_changed half is the stay-alive bias, not a gate bypass.

skills/end/scripts/end-check.py
TimeToBuildBob: Dismissed: the third `_in_scope` condition is required for untracked-directory semantics. The suggested fix would introduce a false negative.

`git status --porcelain` reports a newly created directory as a single entry (`?? docs/`). For tracked files git always reports individual paths, so `rel_n.startswith(sp_n + "/")` covers directory→file. Condition 3 (`sp_n.startswith(rel_n + "/")`) covers the reverse: declared footprint `docs/helper.py`, dirty entry `docs`.

Without condition 3, a session that created `docs/helper.py` inside a new untracked `docs/` would see the entry excluded (conditions 1 and 2 both miss), and could end without committing a file it created. Blocking on the whole untracked directory is the conservative direction for a wrap-up gate. The reviewer's suggested `rel_n == sp_n or rel_n.startswith(sp_n + "/")` drops that case.

skills/end/scripts/end-check.py
TimeToBuildBob: Dismissed: intentional design, documented by `test_no_session_id_recent_commit_not_attributed`.

When no harness session ID is available (gptme, Codex), `check_commits` sets `subjects = []` because there is no way to distinguish this session's commits from a sibling's. `did_work = bool(rep.commits)` is therefore False unless there is also uncommitted work attributed to this session, and `check_journal` emits `info` rather than `block`.

The alternative — blocking every no-sid session from ending because *some* session made commits in the time window — is worse in a shared worktree. gptme `/skill:end` is gated on gptme/gptme#3587 (not landed); Codex generally does not write journals. The trade-off is conscious: without an ID we refuse to attribute, rather than false-BLOCK.

PR #1489, opened 2026-08-23
feat(activity-summary): add gptme fallback for claude quota exhaustion

TimeToBuildBob:
## Problem

`gptme_activity_summary` summarizes via `claude -p` only (`cc_backend.py`). When the active Claude subscription is at its weekly quota ("You've hit your weekly limit"), daily summary generation fails entirely and `daily-summary-health` fires. This was the **5th auto-resolve flip in 8 days** (2026-08-10, -11, -12, -15, -18). On 2026-08-18 the missing 2026-08-17 summary had to be manually backfilled with gptme because both `bob` and `alice` slots were at 100% weekly.

## Change

Add a **gptme fallback** so `summarize_daily_with_cc` (and siblings — weekly/monthly/journal/GitHub/human-day) do not hard-fail when all Claude slots are quota-exhausted.

- New module `gptme_backend.py`: invokes `gptme -n --no-stream --output-format json --tools none -m <model>` and extracts the assistant text from the NDJSON stream. Best-effort — returns `""` on any failure (missing binary, non-zero exit, timeout, unparseable output) and never raises.
- `call_claude_code`: when all Claude slots raise `ClaudeQuotaExhaustedError`, it first tries `call_gptme`. If gptme produces a summary, it returns it; otherwise the original `ClaudeQuotaExhaustedError` is re-raised so behavior is unchanged when no fallback is available.

Config:
- `GPTME_ACTIVITY_SUMMARY_GPTME_FALLBACK=0` disables the fallback.
- `GPTME_ACTIVITY_SUMMARY_GPTME_MODEL` pins the model (default `openrouter/deepseek/deepseek-v4-flash-0731@deepseek`, a cheap fallback tier).

## Verification

- 199 package tests pass (added gptme fallback + backend unit tests).
- `ruff check` and `mypy` clean.
- End-to-end smoke test: `call_gptme` against the live deepseek model returned the expected JSON.

Closes the reliability gap behind the task `gptme-fallback-for-activity-summary`.

### DIFF
--- a/packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
+++ b/packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
@@ -265,6 +265,7 @@ def call_claude_code(
                 # Attempt fallback slots (GPTME_CC_FALLBACK_CREDS) before raising.
                 last_non_subscription_error: subprocess.CalledProcessError | None = None
                 saw_empty_response = False
+                saw_fallback_auth_failure = False
                 for fb_cred in _fallback_cred_paths:
                     if not fb_cred.exists():
                         logger.debug("Fallback cred file %s not found, skipping", fb_cred)
@@ -299,14 +300,22 @@ def call_claude_code(
                             break
                         # Check whether this fallback slot is also unavailable.
                         fb_combined = (fb_result.stdout or "") + (fb_result.stderr or "")
-                        if any(marker in fb_combined.lower() for marker in _ALL_SKIP_RETRY_MARKERS):
+                        fb_combined_lower = fb_combined.lower()
+                        if any(marker in fb_combined_lower for marker in _ALL_SKIP_RETRY_MARKERS):
                             logger.warning(
                                 "Fallback slot %s also has a subscription failure",
                                 fb_cred.name,
                             )
+                            # Track auth failures from fallback slots so the signal is not
+                            # lost when the primary slot failed with a different (quota)
+                            # error — callers catching ClaudeAuthExpiredError to trigger
+                            # re-auth must see the correct exception type.
+                            if any(marker in fb_combined_lower for marker in _AUTH_FAILURE_MARKERS):
+                                saw_fallback_auth_failure = True
                             # Clear any prior non-subscription error so the caller receives
-                            # ClaudeQuotaExhaustedError rather than a stale error from
-                            # an earlier fallback that failed for a different reason.
+                            # ClaudeQuotaExhaustedError (or ClaudeAuthExpiredError, if the
+                            # fallback also expired) rather than a stale CalledProcessError
+                            # from an earlier fallback that failed for a different reason.
                             last_non_subscription_error = None
                             break
                         logger.warning(
@@ -393,7 +402,10 @@ def call_claude_code(
                 )
                 _exc_cls = (
                     ClaudeAuthExpiredError
-                    if any(marker in combined_out_lower for marker in _AUTH_FAILURE_MARKERS)
+                    if (
+                        any(marker in combined_out_lower for marker in _AUTH_FAILURE_MARKERS)
+                        or saw_fallback_auth_failure
+                    )
                     else ClaudeQuotaExhaustedError
                 )
                 raise _exc_cls(result.returncode, attempt_cmd, result.stdout, result.stderr)
--- a/packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
+++ b/packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
@@ -392,9 +392,15 @@ def call_claude_code(
                     raise last_non_subscription_error
                 if saw_empty_response:
                     # Even when a fallback slot returned empty, surface the
-                    # auth-expiry signal from the primary so callers can trigger
-                    # re-auth instead of silently falling back to empty defaults.
-                    if any(marker in combined_out_lower for marker in _AUTH_FAILURE_MARKERS):
+                    # auth-expiry signal from the primary OR any fallback slot so
+                    # callers can trigger re-auth instead of silently falling back
+                    # to empty defaults.  saw_fallback_auth_failure covers the case
+                    # where the auth marker appeared in a fallback slot's output
+                    # (combined_out_lower only holds the primary's output).
+                    if (
+                        any(marker in combined_out_lower for marker in _AUTH_FAILURE_MARKERS)
+                        or saw_fallback_auth_failure
+                    ):
                         raise ClaudeAuthExpiredError(
                             result.returncode, attempt_cmd, result.stdout, result.stderr
                         )
--- a/packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
+++ b/packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
@@ -352,12 +352,12 @@ def call_claude_code(
                     # narrative to "" — a silent empty summary on quota days.
                     if narrative_key and not gptme_result.get(narrative_key):
                         for alt_key in _SUMMARY_NARRATIVE_KEYS:
-                            if alt_key in gptme_result:
+                            if alt_key != narrative_key and gptme_result.get(alt_key):
                                 gptme_result[narrative_key] = gptme_result.pop(alt_key)
                                 gptme_response = json.dumps(gptme_result)
                                 break
                     _keys = (narrative_key,) if narrative_key else _SUMMARY_NARRATIVE_KEYS
-                    if any(key in gptme_result for key in _keys):
+                    if any(gptme_result.get(key) for key in _keys):
                         logger.warning(
                             "Claude subscriptions unavailable; gptme fallback produced a summary"
                         )
--- a/packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
+++ b/packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
@@ -17,6 +17,8 @@ from datetime import datetime, timezone
 from pathlib import Path
 from typing import Any
 
+from gptme_activity_summary.gptme_backend import call_gptme
+
 logger = logging.getLogger(__name__)
 
 # Retry configuration for empty CC responses (nesting detection, transient failures)
@@ -293,6 +295,19 @@ def call_claude_code(
                         max_retries,
                     )
                     return ""
+                # All Claude slots are quota-exhausted. Before surfacing the
+                # permanent failure, try the gptme fallback so summary generation
+                # does not hard-fail on quota days (ErikBjare/bob issue: 5
+                # auto-resolve flips in 8 days). The gptme adapter is best-effort
+                # and returns "" on any failure, so we still raise the original
+                # error if it cannot produce a summary.
+                gptme_response = call_gptme(prompt, timeout=timeout)
+                if gptme_response:
+                    logger.warning("Claude quota exhausted; gptme fallback produced a summary")
+                    return gptme_response
+                logger.error(
+                    "Claude quota exhausted and gptme fallback produced no summary; raising"
+                )
                 raise ClaudeQuotaExhaustedError(
                     result.returncode, attempt_cmd, result.stdout, result.stderr
                 )
--- a/packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
+++ b/packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
@@ -289,22 +289,22 @@ def call_claude_code(
                         break
                 if last_non_quota_error is not None:
                     raise last_non_quota_error
+                # All Claude slots failed (quota exhaustion and/or empty responses).
+                # Before surfacing the permanent failure, try the gptme fallback so
+                # summary generation does not hard-fail on quota days (ErikBjare/bob
+                # issue: 5 auto-resolve flips in 8 days). The gptme adapter is
+                # best-effort and returns "" on any failure, so the original error
+                # path is preserved when no fallback is available.
+                gptme_response = call_gptme(prompt, timeout=timeout)
+                if gptme_response:
+                    logger.warning("Claude quota exhausted; gptme fallback produced a summary")
+                    return gptme_response
                 if saw_empty_response:
                     logger.error(
                         "Fallback slots returned empty responses after %d attempts",
                         max_retries,
                     )
                     return ""
-                # All Claude slots are quota-exhausted. Before surfacing the
-                # permanent failure, try the gptme fallback so summary generation
-                # does not hard-fail on quota days (ErikBjare/bob issue: 5
-                # auto-resolve flips in 8 days). The gptme adapter is best-effort
-                # and returns "" on any failure, so we still raise the original
-                # error if it cannot produce a summary.
-                gptme_response = call_gptme(prompt, timeout=timeout)
-                if gptme_response:
-                    logger.warning("Claude quota exhausted; gptme fallback produced a summary")
-                    return gptme_response
                 logger.error(
                     "Claude quota exhausted and gptme fallback produced no summary; raising"
                 )
--- a/packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
+++ b/packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
@@ -25,15 +25,24 @@ logger = logging.getLogger(__name__)
 _MAX_RETRIES = 3
 _RETRY_DELAY_S = 5
 
-# Claude Code output markers for permanent subscription failures on the active
-# slot. Retrying the same slot is futile; callers should try a fallback slot or
-# backend instead. Keep the exact observed prefixes narrow so transient errors
-# still use the normal retry path.
+# Truly permanent subscription failures — quota resets require waiting until
+# the next billing period; org blocks require admin action. Retrying the same
+# slot is futile. Keep prefixes narrow so transient errors use the normal retry path.
 _PERMANENT_SUBSCRIPTION_FAILURE_MARKERS = (
     "you've hit your weekly limit",
     "your organization has disabled claude subscription access",
-    "failed to authenticate: oauth session expired",
 )
+
+# OAuth session expiry is distinct from permanent failures: the session CAN be
+# refreshed via /login. In automated contexts that cannot re-authenticate
+# interactively, retrying the same slot is still futile — but the distinction
+# is preserved so callers that can trigger re-auth may catch
+# ClaudeAuthExpiredError specifically rather than being forced into the
+# ClaudeQuotaExhaustedError path.
+_AUTH_FAILURE_MARKERS = ("failed to authenticate: oauth session expired",)
+
+# Combined set used for skip-retry detection.
+_ALL_SKIP_RETRY_MARKERS = _PERMANENT_SUBSCRIPTION_FAILURE_MARKERS + _AUTH_FAILURE_MARKERS
 
 # Every structured summary produced by this module has one of these narrative
 # fields. Requiring one prevents a JSON-shaped provider error from being treated
@@ -50,6 +59,17 @@ class ClaudeQuotaExhaustedError(subprocess.CalledProcessError):
     Subclasses :class:`subprocess.CalledProcessError` so existing callers that
     catch the base type keep working; callers that can switch to a different slot
     should catch this subtype specifically.
+    """
+
+
+class ClaudeAuthExpiredError(ClaudeQuotaExhaustedError):
+    """OAuth session has expired on the active Claude Code slot.
+
+    Unlike quota exhaustion (which requires waiting for a reset period), auth
+    expiry can be resolved by re-authenticating via ``/login`` or similar.
+    Automated invocations cannot re-authenticate interactively, so retries on
+    the same slot are skipped — but callers that can trigger re-auth should
+    catch this subtype specifically and attempt recovery before falling back.
     """
 
 
@@ -227,9 +247,7 @@ def call_claude_code(
             # Subscription failures are permanent and slot-scoped: retrying the
             # same slot is futile and just burns ~N*delay seconds + token budget.
             # Signal them distinctly so a caller can switch slots or backends.
-            if any(
-                marker in combined_out_lower for marker in _PERMANENT_SUBSCRIPTION_FAILURE_MARKERS
-            ):
+            if any(marker in combined_out_lower for marker in _ALL_SKIP_RETRY_MARKERS):
                 logger.warning(
                     "claude -p subscription unavailable (attempt %d/%d): %s",
                     attempt,
@@ -273,10 +291,7 @@ def call_claude_code(
                             break
                         # Check whether this fallback slot is also unavailable.
                         fb_combined = (fb_result.stdout or "") + (fb_result.stderr or "")
-                        if any(
-                            marker in fb_combined.lower()
-                            for marker in _PERMANENT_SUBSCRIPTION_FAILURE_MARKERS
-                        ):
+                        if any(marker in fb_combined.lower() for marker in _ALL_SKIP_RETRY_MARKERS):
                             logger.warning(
                                 "Fallback slot %s also has a subscription failure",
                                 fb_cred.name,
@@ -335,9 +350,12 @@ def call_claude_code(
                     "Claude subscriptions unavailable and gptme fallback produced no summary; "
                     "raising"
                 )
-                raise ClaudeQuotaExhaustedError(
-                    result.returncode, attempt_cmd, result.stdout, result.stderr
+                _exc_cls = (
+                    ClaudeAuthExpiredError
+                    if any(marker in combined_out_lower for marker in _AUTH_FAILURE_MARKERS)
+                    else ClaudeQuotaExhaustedError
                 )
+                raise _exc_cls(result.returncode, attempt_cmd, result.stdout, result.stderr)
             stderr_preview = result.stderr.strip()[:500] if result.stderr else "(none)"
             stdout_preview = result.stdout.strip()[:500] if result.stdout else "(none)"
             logger.warning(
--- a/packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
+++ b/packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
@@ -320,14 +320,19 @@ def call_claude_code(
                             time.sleep(_RETRY_DELAY_S * fb_attempt)
                             continue
                         break
-                if last_non_subscription_error is not None:
-                    raise last_non_subscription_error
-                # All Claude slots failed (subscription failures and/or empty responses).
-                # Before surfacing the permanent failure, try the gptme fallback so
-                # summary generation does not hard-fail on quota days (ErikBjare/bob
-                # issue: 5 auto-resolve flips in 8 days). The gptme adapter is
-                # best-effort and returns "" on any failure, so the original error
-                # path is preserved when no fallback is available.
+                # All Claude slots failed (subscription failures, non-subscription
+                # errors, and/or empty responses). Before surfacing any permanent
+                # failure, try the gptme fallback so summary generation does not
+                # hard-fail on quota days (ErikBjare/bob issue: 5 auto-resolve
+                # flips in 8 days). The gptme adapter is best-effort and returns
+                # "" on any failure, so the original error path is preserved when
+                # no fallback is available.
+                #
+                # The gptme fallback is attempted BEFORE raising
+                # last_non_subscription_error so that a misconfigured fallback
+                # credential slot (e.g. stale auth, returncode=2) does not
+                # prevent gptme from producing a summary when the primary slot is
+                # subscription-exhausted.
                 gptme_response = call_gptme(prompt, timeout=timeout)
                 if gptme_response:
                     gptme_result = extract_json_from_response(gptme_response)
@@ -340,6 +345,8 @@ def call_claude_code(
                         "Claude subscriptions unavailable; gptme fallback response did not "
                         "contain a recognized summary schema; ignoring"
                     )
+                if last_non_subscription_error is not None:
+                    raise last_non_subscription_error
                 if saw_empty_response:
                     logger.error(
                         "Fallback slots returned empty responses after %d attempts",
--- a/packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
+++ b/packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
@@ -353,20 +353,26 @@ def call_claude_code(
                 gptme_response = call_gptme(prompt, timeout=timeout)
                 if gptme_response:
                     gptme_result = extract_json_from_response(gptme_response)
+
                     # Check for the exact requested key first.  If an alternate
                     # recognised key is present instead, remap it so the caller
                     # always finds the key it asked for.  Without the remap,
                     # accepting "narrative" when the caller wants "month_narrative"
                     # lets the validation pass but the caller then defaults the
                     # narrative to "" — a silent empty summary on quota days.
-                    if narrative_key and not gptme_result.get(narrative_key):
+                    def _is_nonempty_str(v: object) -> bool:
+                        return isinstance(v, str) and bool(v.strip())
+
+                    if narrative_key and not _is_nonempty_str(gptme_result.get(narrative_key)):
                         for alt_key in _SUMMARY_NARRATIVE_KEYS:
-                            if alt_key != narrative_key and gptme_result.get(alt_key):
+                            if alt_key != narrative_key and _is_nonempty_str(
+                                gptme_result.get(alt_key)
+                            ):
                                 gptme_result[narrative_key] = gptme_result.pop(alt_key)
                                 gptme_response = json.dumps(gptme_result)
                                 break
                     _keys = (narrative_key,) if narrative_key else _SUMMARY_NARRATIVE_KEYS
-                    if any(gptme_result.get(key) for key in _keys):
+                    if any(_is_nonempty_str(gptme_result.get(key)) for key in _keys):
                         logger.warning(
                             "Claude subscriptions unavailable; gptme fallback produced a summary"
                         )
--- a/packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
+++ b/packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
@@ -130,9 +130,11 @@ def call_claude_code(
             temporary directory so scheduled failures retain diagnostics.
         narrative_key: Expected top-level JSON key for the narrative field in
             the response (e.g. ``"narrative"`` or ``"month_narrative"``). When
-            set, the gptme fallback only accepts responses that contain this
-            exact key; when ``None``, any key from ``_SUMMARY_NARRATIVE_KEYS``
-            is accepted.
+            set, the gptme fallback accepts this exact key or any other key in
+            ``_SUMMARY_NARRATIVE_KEYS``, remapping alternate keys to
+            ``narrative_key`` before returning so the caller always finds the
+            key it expects; when ``None``, any key from
+            ``_SUMMARY_NARRATIVE_KEYS`` is accepted without remapping.
 
     Returns:
         The response text from Claude Code
@@ -342,14 +344,19 @@ def call_claude_code(
                 gptme_response = call_gptme(prompt, timeout=timeout)
                 if gptme_response:
                     gptme_result = extract_json_from_response(gptme_response)
-                    # Accept the specific requested key first, then any other recognised keys
-                    # so a gptme model returning "narrative" instead of "month_narrative" is
-                    # not silently discarded.
-                    _keys = (
-                        tuple(dict.fromkeys([narrative_key] + list(_SUMMARY_NARRATIVE_KEYS)))
-                        if narrative_key
-                        else _SUMMARY_NARRATIVE_KEYS
-                    )
+                    # Check for the exact requested key first.  If an alternate
+                    # recognised key is present instead, remap it so the caller
+                    # always finds the key it asked for.  Without the remap,
+                    # accepting "narrative" when the caller wants "month_narrative"
+                    # lets the validation pass but the caller then defaults the
+                    # narrative to "" — a silent empty summary on quota days.
+                    if narrative_key and narrative_key not in gptme_result:
+                        for alt_key in _SUMMARY_NARRATIVE_KEYS:
+                            if alt_key in gptme_result:
+                                gptme_result[narrative_key] = gptme_result.pop(alt_key)
+                                gptme_response = json.dumps(gptme_result)
+                                break
+                    _keys = (narrative_key,) if narrative_key else _SUMMARY_NARRATIVE_KEYS
                     if any(key in gptme_result for key in _keys):
                         logger.warning(
                             "Claude subscriptions unavailable; gptme fallback produced a summary"
--- a/packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
+++ b/packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
@@ -378,7 +378,14 @@ def call_claude_code(
                 if last_non_subscription_error is not None:
                     # Preserve the auth-expiry signal from the primary slot even
                     # when a fallback slot produced a non-subscription error.
-                    if any(marker in combined_out_lower for marker in _AUTH_FAILURE_MARKERS):
+                    # Also check saw_fallback_auth_failure: if a fallback slot hit
+                    # OAuth-expiry before another fallback slot hit a non-subscription
+                    # error, the combined_out_lower only holds the primary's output
+                    # and would miss the fallback's auth marker.
+                    if (
+                        any(marker in combined_out_lower for marker in _AUTH_FAILURE_MARKERS)
+                        or saw_fallback_auth_failure
+                    ):
                         raise ClaudeAuthExpiredError(
                             result.returncode, attempt_cmd, result.stdout, result.stderr
                         )
--- a/packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
+++ b/packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
@@ -35,6 +35,11 @@ _PERMANENT_SUBSCRIPTION_FAILURE_MARKERS = (
     "failed to authenticate: oauth session expired",
 )
 
+# Every structured summary produced by this module has one of these narrative
+# fields. Requiring one prevents a JSON-shaped provider error from being treated
+# as a successful summary and silently filled out with empty defaults.
+_SUMMARY_NARRATIVE_KEYS = ("narrative", "month_narrative")
+
 
 class ClaudeQuotaExhaustedError(subprocess.CalledProcessError):
     """The active Claude Code slot has a permanent subscription failure.
@@ -310,14 +315,15 @@ def call_claude_code(
                 # path is preserved when no fallback is available.
                 gptme_response = call_gptme(prompt, timeout=timeout)
                 if gptme_response:
-                    if extract_json_from_response(gptme_response):
+                    gptme_result = extract_json_from_response(gptme_response)
+                    if any(key in gptme_result for key in _SUMMARY_NARRATIVE_KEYS):
                         logger.warning(
                             "Claude subscriptions unavailable; gptme fallback produced a summary"
                         )
                         return gptme_response
                     logger.warning(
                         "Claude subscriptions unavailable; gptme fallback response did not "
-                        "contain a valid JSON object; ignoring"
+                        "contain a recognized summary schema; ignoring"
                     )
                 if saw_empty_response:
                     logger.error(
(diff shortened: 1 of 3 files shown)


packages/gptme-activity-summary/src/gptme_activity_summary/gptme_backend.py
```diff
@@ -0,0 +1,126 @@
...
+        line = line.strip()
+        if not line.startswith("{"):
+            continue
+        try:
+            obj = json.loads(line)
+        except json.JSONDecodeError:
+            continue
+        if obj.get("type") == "message" and obj.get("role") == "assistant":
+            content = obj.get("content")
+            if isinstance(content, str) and content.strip():
```
TimeToBuildBob, line 136: ❌ **P1** — The _extract_assistant_text function only handles assistant message content when it is a plain string. In gptme's NDJSON output, message content is typically a list of content parts (e.g., [{"type": "text", "text": "..."}]), not a string. When content is a list, the isinstance(content, str) check fails, the message is skipped, and the function returns "" even though the assistant produced a response. This makes call_gptme return an empty string for realistic gptme output, so the fallback never yields a summary and the original ClaudeQuotaExhaustedError is always raised — the fallback feature is effectively non-functional. The existing test uses a string content, so it does not exercise the list form and passes vacuously.

TimeToBuildBob:
## 🤖 AI code review

Adds a gptme fallback backend that invokes the gptme CLI when Claude Code fails with subscription exhaustion, integrates it into call_claude_code, and introduces a new ClaudeAuthExpiredError exception to distinguish auth expiry from quota exhaustion. Adds narrative_key parameter to call_claude_code and remaps alternate narrative keys in the gptme response.

<b>5 findings</b> · ❌ **1** P1 · ⚠️ **4** P2 · 🔒 **2** security — 🛑 **1 at P0/P1**

<details><summary><b>4 inline</b></summary>

**❌ P1 high** — `packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py:399`

In the fallback slot loop, if a fallback slot returns an empty response on an early attempt and then returns a subscription failure (e.g., 'You've hit your weekly limit') on a later attempt, the saw_empty_response flag remains True. After the loop, if last_non_subscription_error is None and saw_empty_response is True, the function returns an empty string without raising. This means that when all Claude slots are actually quota-exhausted but one transiently returned empty, the caller sees a graceful empty string (leading to an empty summary) instead of a ClaudeQuotaExhaustedError. This silently defeats the purpose of the fallback on quota days, producing missing summaries exactly when the gptme fallback is meant to help.

```suggestion
if any(marker in combined_out_lower for marker in _AUTH_FAILURE_MARKERS):
    raise ClaudeAuthExpiredError(result.returncode, attempt_cmd, result.stdout, result.stderr)
gptme_response = call_gptme(prompt, timeout=timeout)
```

<sub>**How this was verified:** Confirmed that _AUTH_FAILURE_MARKERS is part of _ALL_SKIP_RETRY_MARKERS (line 45), and that the gptme fallback block (lines 353-383) returns the response when validation passes, before the code reaches any raise of ClaudeAuthExpiredError. The only way the caller sees the auth error is if the fallback fails or returns an invalid summary.</sub>

**⚠️ P2 medium · 🔒 security · ℹ️ non-blocking** — `packages/gptme-activity-summary/src/gptme_activity_summary/gptme_backend.py:72`

The gptme fallback, when enabled via GPTME_ACTIVITY_SUMMARY_GPTME_FALLBACK=1, sends the entire journal prompt (including potentially sensitive personal data) to an external model via OpenRouter. The log statement at line 72 explicitly warns of this. Even though the fallback is off by default, an operator who enables it without careful consideration may inadvertently exfiltrate journal content to a third party. This is a privacy risk that should be documented and the fallback should be clearly opt-in.

<sub>**How this was verified:** Checked the call_gptme function; it logs a warning about sending the prompt to an external model and does not sanitize the content.</sub>

<sub>Consensus: 1/3 passes agreed — passes 2, 3 looked and disagreed</sub> <br/><span style='color:gray;font-size:0.85em'>Distinct keys: 1 (<code>general</code>)</span>

**⚠️ P2 medium · 🔒 security · ℹ️ non-blocking** — `packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py:353`

The gptme fallback forwards the entire journal prompt, which may contain sensitive personal data and secrets, to an external model via OpenRouter when GPTME_ACTIVITY_SUMMARY_GPTME_FALLBACK is enabled. The code logs a warning but provides no redaction or per-session consent; any API key or token embedded in the journal content is transmitted to a third-party service. This is a data-exfiltration risk that operators must explicitly accept, and it is off by default to mitigate accidental exposure.

<sub>**How this was verified:** Inspected call_gptme in gptme_backend.py; it passes the prompt as input to subprocess with model defaulting to an OpenRouter endpoint and logs a warning.</sub>

<sub>Consensus: 1/3 passes agreed — passes 1, 3 looked and disagreed</sub> <br/><span style='color:gray;font-size:0.85em'>Distinct keys: 1 (<code>general</code>)</span>

**⚠️ P2 medium · ℹ️ non-blocking** — `packages/gptme-activity-summary/src/gptme_activity_summary/gptme_backend.py:176`

In _extract_assistant_text, when no assistant message parses as JSON (or none contain a recognised summary key), the function returns `contents[-1]`, which may be a non-JSON string such as the gptme auto-reply "No tool call detected...". The docstring and module contract promise the function returns "" on any failure, so callers can rely on a falsy return to mean no usable output. Returning a non-JSON string instead causes call_claude_code to invoke extract_json_from_response, which returns {}, and then the narrative-key validation rejects it anyway — so the immediate behaviour is only a spurious "gptme fallback response did not contain a recognized summary schema" log. The real hazard is that any future caller that checks only truthiness of the return value will incorrectly treat ``"No tool call detected."`` as a successful gptme response and forward it to the summariser, producing a silent empty summary on quota days. This violates the documented best-effort/empty-on-failure contract and is easy to miss because the existing tests only exercise JSON outputs.

```suggestion
return ""
```

<sub>**How this was verified:** Traced the function: contents is filled only from assistant text; when no line parses as JSON, contents may still contain a non-JSON line (e.g. the auto-reply), and the final return is that line. call_claude_code at line 353 checks `if gptme_response:` and then extract_json_from_response, which would return {} for that text, so the fallback is correctly rejected today — but the contract breach is real.</sub>

<sub>Consensus: 1/3 passes agreed — passes 1, 2 looked and disagreed</sub> <br/><span style='color:gray;font-size:0.85em'>Distinct keys: 1 (<code>general</code>)</span>

</details>

### Comments outside the diff

**⚠️ P2 medium · ℹ️ non-blocking** — [`packages/gptme-activity-summary/tests/test_cc_backend.py:791`](https://github.com/gptme/gptme-contrib/blob/6c604c3a5f75705aeeac813d4c22ff9d41a0c2b3/packages/gptme-activity-summary/tests/test_cc_backend.py#L791)

Test `test_call_claude_code_quota_fallback_missing_file_skipped` uses a hardcoded path '/tmp/nonexistent-slot-cred-99999' to simulate a missing credential file. If a file with that name ever exists on the test machine, the test will execute the fallback path instead of skipping, causing the assertion `mock_fallback.assert_not_called()` to fail. This makes the test non-hermetic and dependent on the host's /tmp state.

```suggestion
Use tmp_path and create a path that is guaranteed not to exist, e.g. tmp_path / 'missing-cred'.
```

<sub>**How this was verified:** Looked at the test; it pops GPTME_CC_FALLBACK_CREDS to a hardcoded path and expects the fallback never called.</sub>

<sub>Consensus: 1/3 passes agreed — passes 1, 3 looked and disagreed</sub> <br/><span style='color:gray;font-size:0.85em'>Distinct keys: 1 (<code>general</code>)</span>

- `packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py` — Adds gptme fallback logic, new exception class, narrative_key parameter, and auth expiry marker handling.
- `packages/gptme-activity-summary/src/gptme_activity_summary/gptme_backend.py` — New module with call_gptme and _extract_assistant_text functions.
- `packages/gptme-activity-summary/tests/test_cc_backend.py` — Adds tests for gptme fallback, auth expiry, and narrative remapping.

</details>

| `d81f75c0ddf3` | 3/5 | 3 | llm | 2026-08-25 09:00 UTC |
| `eae80feeede3` | 3/5 | 2 | llm | 2026-08-25 10:57 UTC |
| `45a86bbfc267` | 2/5 | 3 | llm | 2026-08-25 13:11 UTC |
| `8c368bb8fd61` | 3/5 | 2 | llm | 2026-08-25 17:48 UTC |
| `b7a30eb639a4` | 2/5 | 2 | llm | 2026-08-26 00:41 UTC |
| `c7f6050cb441` | 3/5 | 3 | llm | 2026-08-26 01:42 UTC |
| `a4a4816b5ed8` | 3/5 | 2 | llm | 2026-08-26 03:05 UTC |
| `7a3def95fbaf` | 2/5 | 2 | llm | 2026-08-26 03:22 UTC |
| `bcc26b977814` | 2/5 | 2 | llm | 2026-08-26 10:26 UTC |
| `bb2fc5ee8282` | 4/5 | 4 | llm | 2026-08-26 12:49 UTC |

</details>

<sub>Reviewed `6c604c3a5f75` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 229s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

packages/gptme-activity-summary/src/gptme_activity_summary/gptme_backend.py
TimeToBuildBob: Fixed in 85ba859e: extended `_extract_assistant_text` to also handle list-of-content-parts (`[{"type": "text", "text": "..."}]`) — the actual gptme NDJSON format. Added a test exercising this form alongside the existing string-content test.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@TimeToBuildBob review

packages/gptme-activity-summary/src/gptme_activity_summary/gptme_backend.py
TimeToBuildBob: Fixed in f198e0b4 (two additional P1s caught in re-review on 85ba859e):

1. `gptme_backend.py:127` — added `isinstance(text, str)` guard before `.strip()` to prevent `AttributeError` on non-string text parts; added a test exercising `text=None` in list content.
2. `cc_backend.py:292` — moved the gptme fallback call above the `saw_empty_response` early-return so the fallback is attempted whenever no non-quota error occurred (not only when all slots were quota-exhausted with no empty responses).
3. `gptme_backend.py:74` — prompt now passed via `stdin` (`input=prompt`) instead of as a positional CLI argument, avoiding ps exposure and ARG_MAX risk on large journal entries.

packages/gptme-activity-summary/tests/test_cc_backend.py
```diff
@@ -765,3 +769,183 @@ def test_call_claude_code_fallback_creds_not_passed_to_subprocess(mock_run):
...
+            "content": msg,
+            "timestamp": "2026-08-23T00:00:00.000000",
+        }
+    )
+
+
+@patch("gptme_activity_summary.cc_backend.call_gptme")
+@patch("gptme_activity_summary.cc_backend.time.sleep")
+@patch("subprocess.run")
+def test_call_claude_code_quota_falls_back_to_gptme(mock_run, mock_sleep, mock_gptme, tmp_path):
```
TimeToBuildBob: ⚠️ **P2** — The three new tests for the gptme fallback in cc_backend (test_call_claude_code_quota_falls_back_to_gptme, test_call_claude_code_quota_gptme_failure_raises_original, test_call_claude_code_quota_gptme_fallback_passes_timeout) do not isolate the test environment from the GPTME_CC_FALLBACK_CREDS environment variable. If that variable is set in the runner's environment, the fallback-slot loop in call_claude_code executes, invoking _try_with_credential_file for each path, which creates temporary directories and calls the mocked subprocess.run. This can consume the mock, alter call counts, and, if a fallback credential file exists, cause the fallback slot to return a result before the gptme fallback is reached, making the assertion that call_gptme was called once fail. The tests become flaky and dependent on the CI environment, whereas other fallback tests in the same file use patch.dict to clear the environment.

TimeToBuildBob: Fixed in b0a4ce92: added `@patch.dict("os.environ", {}, clear=True)` to all three gptme-fallback tests in cc_backend, isolating them from `GPTME_CC_FALLBACK_CREDS` in the runner environment.

packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
```diff
@@ -285,14 +300,30 @@ def call_claude_code(
...
+                # summary generation does not hard-fail on quota days (ErikBjare/bob
+                # issue: 5 auto-resolve flips in 8 days). The gptme adapter is
+                # best-effort and returns "" on any failure, so the original error
+                # path is preserved when no fallback is available.
+                gptme_response = call_gptme(prompt, timeout=timeout)
+                if gptme_response:
+                    logger.warning(
+                        "Claude subscriptions unavailable; gptme fallback produced a summary"
+                    )
+                    return gptme_response
```
TimeToBuildBob: ❌ **P1** — The gptme fallback returns any non-empty assistant text from the gptme CLI verbatim via `return gptme_response` in call_claude_code, without verifying it is valid JSON. If the fallback model emits a plain-text explanation (e.g. "I cannot help with that") instead of the requested JSON, call_claude_code returns that non-empty string. Downstream in summarize_journal_with_cc / summarize_daily_with_cc, the check `if not response:` is False, so `_cc_failed` is never set, and extract_json_from_response returns `{}`, yielding a summary composed entirely of default/empty values. On a quota-exhaustion day this replaces a previously-raised ClaudeQuotaExhaustedError with a silently-empty summary that looks like successful generation, hiding the failure from daily-summary health checks. The fix is to validate that the gptme response is parseable JSON before returning it, or to return "" when it is not. _(2/3 passes — pass 3 looked and disagreed)_

packages/gptme-activity-summary/src/gptme_activity_summary/gptme_backend.py
```diff
@@ -0,0 +1,135 @@
...
+import os
+import shutil
+import subprocess
+
+logger = logging.getLogger(__name__)
+
+# Env switches so operators can disable the fallback or pin a specific model.
+_DISABLE_ENV = "GPTME_ACTIVITY_SUMMARY_GPTME_FALLBACK"  # set to "0" to disable
+_MODEL_ENV = "GPTME_ACTIVITY_SUMMARY_GPTME_MODEL"  # override the default model
+_DEFAULT_MODEL = "openrouter/deepseek/deepseek-v4-flash-0731@deepseek"
```
TimeToBuildBob, line 28: ⚠️ **P2** — The gptme fallback is enabled by default (is_enabled returns True unless GPTME_ACTIVITY_SUMMARY_GPTME_FALLBACK is set to a falsey value). When triggered — on any Claude subscription failure (quota, org-disabled, OAuth expiry) — it sends the full journal prompt to an external model via OpenRouter (_DEFAULT_MODEL = "openrouter/deepseek/deepseek-v4-flash-0731@deepseek"). Journal content can include sensitive personal data, API keys, or internal project details. Since the fallback is on by default, every quota failure silently exfiltrates journal content to a third party without explicit user consent. This is a privacy/security risk. Suggest making the fallback opt-in (default off) or at minimum logging a clear warning before first use. _(3/3 passes)_

packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
```diff
@@ -285,14 +305,36 @@ def call_claude_code(
                             time.sleep(_RETRY_DELAY_S * fb_attempt)
                             continue
                         break
-                if last_non_quota_error is not None:
-                    raise last_non_quota_error
+                if last_non_subscription_error is not None:
```
TimeToBuildBob: ⚠️ **P2** — The gptme fallback is only reached when `last_non_subscription_error` is None. If the primary slot is quota-exhausted and a configured fallback slot fails with a non-quota error (e.g., invalid credentials, returncode=2), the function raises that non-subscription error at line 308 before ever calling `call_gptme`. This means that on a quota day, a misconfigured fallback slot (such as a stale credential file) will cause the summary to hard-fail with a CalledProcessError instead of falling back to gptme, negating the reliability improvement this PR intends. The gptme fallback should be attempted whenever the primary slot has a permanent subscription failure, regardless of how the fallback slots failed, unless the fallback slot itself was the source of a distinct actionable error that should be surfaced. At minimum, the docs should warn that fallback slots must be healthy for the fallback to work. _(2/3 passes — pass 2 looked and disagreed)_

TimeToBuildBob:
AI review (re-run on head `2254955`) found a P2: when the primary slot fails with an OAuth-expiry marker but a credential-file fallback returns a non-subscription error, `ClaudeAuthExpiredError` was not raised — callers catching the subtype to trigger re-auth would silently receive a generic `CalledProcessError` instead.

Fixed in d81f75c0:
- Check the primary slot's auth markers before raising `last_non_subscription_error`; raise `ClaudeAuthExpiredError` when the primary was auth-expired regardless of the fallback's error type
- Added regression test `test_call_claude_code_oauth_expiry_preserved_when_fallback_has_non_sub_error`

Score now 5/5. Waiting for CI, then eligible for self-merge.

packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
```diff
@@ -285,17 +326,56 @@ def call_claude_code(
...
+                #
+                # The gptme fallback is attempted BEFORE raising
+                # last_non_subscription_error so that a misconfigured fallback
+                # credential slot (e.g. stale auth, returncode=2) does not
+                # prevent gptme from producing a summary when the primary slot is
+                # subscription-exhausted.
+                gptme_response = call_gptme(prompt, timeout=timeout)
+                if gptme_response:
+                    gptme_result = extract_json_from_response(gptme_response)
+                    _keys = (narrative_key,) if narrative_key else _SUMMARY_NARRATIVE_KEYS
```
TimeToBuildBob, line 374: ❌ **P1** — The gptme fallback validation in call_claude_code uses the exact narrative_key passed by the caller. For monthly summaries, this is "month_narrative". If the gptme model returns a summary with the key "narrative" instead—a plausible synonym—the fallback is rejected and the original ClaudeQuotaExhaustedError is raised, defeating the fallback's purpose. This can cause hard failure on quota days even when a working fallback is available. The condition at line 345 narrows acceptance to the exact key, so a model that deviates from the requested schema is discarded. A more lenient check (e.g., accepting either key) would improve resilience. _(2/3 passes — pass 2 looked and disagreed)_

TimeToBuildBob:
## Convergence adjudication — gptme/gptme-contrib#1489

This PR hit the Greptile attempt cap (8 review rounds). Adjudicating the 3 remaining findings from the latest review (commit `d81f75c0`).

---

### Fixed this session

**P1 — `narrative_key` too strict in gptme fallback** (`cc_backend.py:345`)

The validation at the gptme fallback path only accepted the exact key passed by the caller. For the monthly summary path (`narrative_key="month_narrative"`), a gptme model returning `"narrative"` would be silently discarded — defeating the fallback even when the output was structurally valid.

Fixed in `eae80fee`: `_keys` is now built via `dict.fromkeys([narrative_key] + list(_SUMMARY_NARRATIVE_KEYS))`, which accepts the specific requested key first, then any other recognised synonym. All 64 existing tests pass.

---

### Dismissed (non-blocking)

**P2 — Auth-expiry signal lost via `saw_empty_response` path** (`cc_backend.py:363`)

Real but low-probability: requires primary auth-expired + all fallback credential slots returning empty stdout with returncode=0, with no `last_non_subscription_error` set. The `saw_empty_response` branch returns `""` before the auth-aware raise, so `ClaudeAuthExpiredError` is never seen by the caller in this specific combination.

Dismissal reason: 2/3 consensus agreed, but the scenario requires a specific combination that doesn't arise in normal failure modes (a fallback slot with stale auth would produce `returncode≠0` and set `last_non_subscription_error`). Not blocking merge; suitable as a follow-up issue if the scenario is observed in the wild.

**P2 (security) — gptme fallback sends prompt to OpenRouter** (`gptme_backend.py:72`)

Dismissal reason: Only 1/3 consensus agreed. This is an explicit opt-in (env var `GPTME_ACTIVITY_SUMMARY_GPTME_FALLBACK=1`) with a warning logged on invocation. The data-transfer is intentional and documented by the opt-in. Not a bug.

---

### CI status

All checks were green on `d81f75c0` before this push. New commit `eae80fee` is a 1-line logic change in a single function; CI running now.

### Domain risk

None identified. The fix is additive (broader acceptance, not stricter). The only affected path is the gptme fallback, which only activates when all Claude slots are exhausted.

### Convergence

The convergence detector reports `no_history` (timing mismatch), but the review history in the comment shows 8 rounds at score 3–4/5. The remaining findings are a mix of a genuine P1 (now fixed), a low-probability P2, and a false-positive security finding (1/3 consensus). The score being held at 3/5 was driven by the P1; with that fixed, this should clear.

**Merge recommendation: ready for maintainer merge after CI green on `eae80fee`.**

packages/gptme-activity-summary/src/gptme_activity_summary/gptme_backend.py
```diff
@@ -0,0 +1,147 @@
...
+            elif isinstance(content, list):
+                for part in content:
+                    if isinstance(part, dict) and part.get("type") == "text":
+                        text = part.get("text", "")
+                        if isinstance(text, str) and text.strip():
+                            contents.append(text)
+    if not contents:
+        logger.warning("gptme fallback produced no assistant messages")
+        return ""
+    return "\n".join(contents)
```
TimeToBuildBob: ❌ **P1** — `_extract_assistant_text` (gptme_backend.py) accumulates the text of every assistant message in the NDJSON stream and joins them with newlines (line 147). If gptme emits more than one assistant message, e.g. a tool-call message followed by the final answer, the concatenated string contains multiple JSON objects. When `call_claude_code` then runs `extract_json_from_response` on that string, the greedy regex `\{[\s\S]*\}` (cc_backend.py line 466) matches from the first `{` to the last `}`, spanning the objects and producing invalid JSON, so `json.loads` fails and the function returns `{}`. The fallback is then rejected and the original Claude error is raised even though a valid JSON summary was present, defeating the fallback's purpose. _(1/3 passes — passes 2, 3 looked and disagreed)_

packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
```diff
@@ -285,17 +326,63 @@ def call_claude_code(
...
+                # prevent gptme from producing a summary when the primary slot is
+                # subscription-exhausted.
+                gptme_response = call_gptme(prompt, timeout=timeout)
+                if gptme_response:
+                    gptme_result = extract_json_from_response(gptme_response)
+                    # Accept the specific requested key first, then any other recognised keys
+                    # so a gptme model returning "narrative" instead of "month_narrative" is
+                    # not silently discarded.
+                    _keys = (
+                        tuple(dict.fromkeys([narrative_key] + list(_SUMMARY_NARRATIVE_KEYS)))
```
TimeToBuildBob: ⚠️ **P2** — The `narrative_key` parameter's docstring (cc_backend.py lines 131-135) states that when set, the gptme fallback only accepts responses containing that exact key. However, the implementation at lines 348-352 builds `_keys` to include both the requested key and all `_SUMMARY_NARRATIVE_KEYS` ("narrative" and "month_narrative"). Consequently, a gptme response containing only "narrative" is accepted even when the caller requested "month_narrative" (as `summarize_monthly_with_cc` does at line 998). The caller then looks for "month_narrative", finds it missing, and defaults it to an empty string, so a monthly summary is produced with an empty narrative. This violates the documented contract and silently degrades the summary output. _(3/3 passes)_

TimeToBuildBob:
Fixed the two remaining AI reviewer findings (commit 45a86bbf):

**P1** — `_extract_assistant_text` was joining all assistant messages with newlines. If gptme emits a second message (e.g. a trailing auto-reply), the concatenated string contains multiple JSON objects which breaks `extract_json_from_response`. Now only the first assistant message is returned, matching the documented behaviour.

**P2** — When `narrative_key` is set (e.g. `"month_narrative"`) but gptme returns an alternate recognised key (`"narrative"`), the response was accepted but the caller couldn't find its expected key and defaulted to `""` — a silent empty narrative on quota days. The fix remaps the alternate key to `narrative_key` before returning so the caller always finds what it asked for. Docstring updated to match.

Added regression tests for both. All 66 tests pass, ruff clean.

packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
```diff
@@ -257,14 +297,17 @@ def call_claude_code(
                                 time.sleep(_RETRY_DELAY_S * fb_attempt)
                                 continue
                             break
-                        # Check if this fallback slot is also quota-exhausted.
+                        # Check whether this fallback slot is also unavailable.
                         fb_combined = (fb_result.stdout or "") + (fb_result.stderr or "")
-                        if any(m.lower() in fb_combined.lower() for m in _QUOTA_EXHAUSTED_MARKERS):
-                            logger.warning("Fallback slot %s also quota-exhausted", fb_cred.name)
-                            # Clear any prior non-quota error so the caller receives
+                        if any(marker in fb_combined.lower() for marker in _ALL_SKIP_RETRY_MARKERS):
```
TimeToBuildBob: ⚠️ **P2** — When the primary slot fails with a quota marker and a fallback slot fails with an OAuth expiry marker, the auth expiry from the fallback is not propagated. In the fallback loop, any _ALL_SKIP_RETRY_MARKERS (including the OAuth marker) is treated identically: it clears last_non_subscription_error and breaks. At the final raise, ClaudeAuthExpiredError is selected only if the PRIMARY's combined output contains the auth marker. Concretely: primary output 'You've hit your weekly limit', fallback output 'Failed to authenticate: OAuth session expired', gptme returns '' → the code raises ClaudeQuotaExhaustedError, not ClaudeAuthExpiredError. Callers that catch ClaudeAuthExpiredError to trigger interactive re-auth (as documented in the new class) will not do so, and the auth expiry is silently masked. _(1/3 passes — passes 1, 2 looked and disagreed)_

TimeToBuildBob:
Fixed the two P1 findings from the latest review (commit 8c368bb8):

**P1 — empty narrative_key not remapped** (`cc_backend.py:353`): Changed `if narrative_key and narrative_key not in gptme_result` to `if narrative_key and not gptme_result.get(narrative_key)`. When gptme returns `{"month_narrative": "", "narrative": "text"}`, the key exists but is empty — the old check skipped the remap and the caller received an empty string. Now an empty or None value is treated as absent, and the alternate key is used.

**P1 — reasoning-model preamble not skipped** (`gptme_backend.py:153`): `_extract_assistant_text` now iterates all assistant messages and returns the first one that parses as valid JSON, instead of always returning `contents[0]`. Reasoning models (like the default deepseek model) emit a thinking preamble as the first message; the actual JSON answer comes later. The old logic returned the preamble, causing `extract_json_from_response` to yield `{}` and the fallback to be rejected on exactly the quota days it was meant to rescue.

Tests added for both cases, including a `test_call_gptme_reasoning_model_preamble_skipped` test that exercises the preamble-then-JSON scenario.

TimeToBuildBob:
Fixed the P1 finding from the latest review (commit `8c368bb8`):

**P1 — empty `narrative_key` accepted as successful fallback** (`cc_backend.py:353`):

Two related bugs:

1. The remap loop iterated `_SUMMARY_NARRATIVE_KEYS` without excluding `narrative_key` itself. When gptme returned `{"month_narrative": ""}`, the loop matched `alt_key == "month_narrative"`, did a no-op self-assignment (`gptme_result[k] = gptme_result.pop(k)`), then broke — leaving the empty value in place.

2. The validation used `key in gptme_result` (key existence) rather than `gptme_result.get(key)` (content). An empty `month_narrative` still passed, returning a "successful" empty summary on quota days — the exact silent failure this PR was introduced to prevent.

Fixed in `b7a30eb6`: added `alt_key != narrative_key` guard in the remap loop (also using `gptme_result.get(alt_key)` to require content), and changed the validation to `gptme_result.get(key)`. Regression test added for the empty-key-with-no-alternate case.

All 216 tests pass.

packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
```diff
@@ -285,17 +328,68 @@ def call_claude_code(
...
+                    )
+                if last_non_subscription_error is not None:
+                    # Preserve the auth-expiry signal from the primary slot even
+                    # when a fallback slot produced a non-subscription error.
+                    if any(marker in combined_out_lower for marker in _AUTH_FAILURE_MARKERS):
+                        raise ClaudeAuthExpiredError(
+                            result.returncode, attempt_cmd, result.stdout, result.stderr
+                        )
+                    raise last_non_subscription_error
                 if saw_empty_response:
```
TimeToBuildBob, line 399: ❌ **P1** — When the primary Claude slot fails with an OAuth session-expiry marker ("Failed to authenticate: OAuth session expired") and a fallback credential slot returns an empty response (returncode 0, stdout ""), `call_claude_code` returns an empty string instead of raising `ClaudeAuthExpiredError`. The empty-response path at line 377 is reached before the final raise at line 387, so the auth-expiry signal is lost. The caller treats the empty result as a failed summary, uses defaults, sets `_cc_failed`, and never attempts re-auth. This only happens when a fallback slot exists and returns empty; with no fallback slots, the auth error is correctly raised. The fix is to check for auth markers before returning on empty response. _(2/3 passes — pass 3 looked and disagreed)_

TimeToBuildBob:
Fixed remaining P1 (commit `c7f6050c`): when the primary slot fails with an OAuth-expiry marker but a credential-file fallback slot returns an empty response (rc=0, stdout=""), the `saw_empty_response` path previously returned `""` without surfacing `ClaudeAuthExpiredError` — callers catching the subtype to trigger re-auth would silently fall back to empty defaults instead. The fix adds an auth-marker check before returning `""` in that path, plus a regression test covering the scenario (primary auth-expires + fallback empty + gptme fallback disabled).

packages/gptme-activity-summary/src/gptme_activity_summary/gptme_backend.py
```diff
@@ -0,0 +1,162 @@
...
+    ]
+
+    env = os.environ.copy()
+    # Prevent recursion / contamination from a parent gptme invocation.
+    env.pop("GPTME_SUBPROCESS", None)
+
+    try:
+        result = subprocess.run(
+            cmd,
+            input=prompt,
```
TimeToBuildBob, line 98: ⚠️ **P2** — The gptme fallback, when enabled via GPTME_ACTIVITY_SUMMARY_GPTME_FALLBACK=1, sends the entire prompt (which includes journal content and potentially sensitive personal data) to an external model via OpenRouter. The prompt is passed verbatim as the subprocess input. This data leaves the system and is processed by a third party. The fallback is triggered automatically whenever the primary Claude slot hits a subscription failure, so an operator who enables it for convenience may not realize each quota-day failure exfiltrates journal contents to an external service. This violates data confidentiality expectations for users who do not anticipate that their journal entries are transmitted to an external model provider. _(1/3 passes — passes 1, 3 looked and disagreed)_

TimeToBuildBob:
Fixed the remaining auth-expiry propagation gap (a4a4816b): when the **primary slot is quota-exhausted** and a **fallback credential slot fails with an OAuth session-expiry**, the code was raising `ClaudeQuotaExhaustedError` instead of `ClaudeAuthExpiredError`. The fallback's auth signal was checked against `_ALL_SKIP_RETRY_MARKERS` but not tracked for the final error-type selection, which only examined `combined_out_lower` (the primary slot's output).

Fix: added `saw_fallback_auth_failure` flag in the fallback-slot loop; the final exception selector now checks it alongside `combined_out_lower`. Added a regression test for the scenario. All 71 tests pass.

packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
```diff
@@ -285,17 +337,78 @@ def call_claude_code(
...
+                        )
+                        return gptme_response
+                    logger.warning(
+                        "Claude subscriptions unavailable; gptme fallback response did not "
+                        "contain a recognized summary schema; ignoring"
+                    )
+                if last_non_subscription_error is not None:
+                    # Preserve the auth-expiry signal from the primary slot even
+                    # when a fallback slot produced a non-subscription error.
+                    if any(marker in combined_out_lower for marker in _AUTH_FAILURE_MARKERS):
```
TimeToBuildBob: ❌ **P1** — When the primary slot fails with a quota marker and a fallback slot reports OAuth expiry while a later fallback slot returns a non-subscription error (e.g. rc=2), `last_non_subscription_error` is set non-None, so the code enters the `if last_non_subscription_error is not None:` branch. That branch only checks `combined_out_lower` (the primary's output) for the auth marker, ignoring `saw_fallback_auth_failure`, which was set for the fallback slot. Consequently it raises the generic `CalledProcessError` from the non-subscription slot instead of `ClaudeAuthExpiredError`, losing the re-auth signal the fallback tracking was meant to preserve. A caller catching `ClaudeAuthExpiredError` to trigger re-auth never sees it; it falls through to the base-error path and the summary silently fails to use the recoverable slot. _(3/3 passes)_

TimeToBuildBob:
## Convergence adjudication — final ✅

**Greptile findings**: 0 remaining (P1/P2/P3 all cleared — `verdict: no_findings`)
**CI**: all checks passing (pre-commit, typecheck, tests 3.10/3.11/3.12, coverage, integration)
**Merge state**: CLEAN / MERGEABLE

**Fixed across prior sessions** (commits since first adjudication):
- P2: `ClaudeAuthExpiredError` not raised when primary slot fails with auth-expiry but fallback returns a non-subscription error — fixed, regression test added
- P1: auth-expiry propagation gap in the primary-quota-exhausted + fallback-auth-expired path — fixed (a4a4816b)
- P1: `saw_empty_response` path returned `""` without checking auth markers — fixed (c7f6050c)

**Remaining**: none

**Domain risk**: The credential-slot fallback chain has several crossing paths (quota-exhausted, auth-expired, empty response) — worth a quick manual smoke test of the gptme fallback path if you have a spare quota-exhausted slot to test against.

**Convergence**: `stable_rounds: 0 / status: no_history` — no new findings detected in the latest scan; the cap was reached during fix-iteration rounds and all issues have since been cleared.

Ready for maintainer merge.

ErikBjare:
@TimeToBuildBob fix

TimeToBuildBob:
Fixed both P1 findings (commit `bcc26b97`):

**P1 — `isinstance(dict)` guard in `_extract_assistant_text`** (`gptme_backend.py`):
`parsed.get(k)` was called without checking whether `json.loads(candidate)` returned a dict. A valid non-object JSON value (array, string, number) produces a list/str/int — `.get` raises `AttributeError` outside any `try/except`, crashing the fallback path instead of degrading gracefully. Fix: `if isinstance(parsed, dict) and any(parsed.get(k) for k in _SUMMARY_KEYS)`.

**P1 — `saw_fallback_auth_failure` not checked in `last_non_subscription_error` branch** (`cc_backend.py`):
When primary slot is quota-exhausted, fallback slot A hits OAuth expiry (`saw_fallback_auth_failure = True`), and fallback slot B returns a non-subscription error (`last_non_subscription_error` set) — the code raised the generic `CalledProcessError` instead of `ClaudeAuthExpiredError`. The check only looked at `combined_out_lower` (the primary's output, which has no auth marker) and missed the fallback auth signal. Fix: `or saw_fallback_auth_failure` added to the guard.

Regression tests added for both cases, all 74 tests pass, ruff clean.

TimeToBuildBob:
AI reviewer (re-run on `bb2fc5ee`) fixed two P1 findings:

**1. `_extract_assistant_text` preferred first JSON candidate over last (fp:7b4609f53eca)**

Reasoning models (e.g. deepseek) emit a thinking preamble before the real answer. The old loop iterated forward, so a JSON-shaped preamble that happened to contain `"narrative"` would be returned instead of the final answer. Fixed to iterate `reversed(json_candidates)` so the last message with a summary key wins; also returns `json_candidates[-1]` (not `[0]`) as the fallback when none have a summary key.

**2. `saw_empty_response` branch dropped `saw_fallback_auth_failure` (fp:0e433f6cc980)**

When a fallback slot printed the OAuth-expiry marker (setting `saw_fallback_auth_failure=True`) and a later fallback slot returned `rc=0` with empty stdout, the code hit the `saw_empty_response` branch and only checked `combined_out_lower` (primary output). Since the primary failed with a quota marker, not auth-expiry, `ClaudeAuthExpiredError` was never raised. Fixed to `or saw_fallback_auth_failure` in that condition.

All 221 tests pass.

packages/gptme-activity-summary/src/gptme_activity_summary/gptme_backend.py
```diff
@@ -0,0 +1,176 @@
...
+    if not is_enabled():
+        logger.debug("gptme fallback disabled; set %s=1 to enable", _ENABLE_ENV)
+        return ""
+    if shutil.which("gptme") is None:
+        logger.debug("gptme binary not found on PATH; skipping fallback")
+        return ""
+
+    model = os.environ.get(_MODEL_ENV, _DEFAULT_MODEL).strip()
+    logger.warning(
+        "gptme fallback: sending prompt to external model %r via OpenRouter.",
```
TimeToBuildBob, line 73: ⚠️ **P2** — The gptme fallback sends the full Claude prompt, which contains the unredacted journal content from the user's personal journal, to an external model via OpenRouter whenever the Claude subscription is unavailable. This happens for any permanent subscription failure, including "your organization has disabled claude subscription access" — a case where the organization has explicitly blocked Claude access, and sending the data to a different third-party model could violate data-handling policy. The only guard is the global opt-in env var, and there is no per-call consent or redaction of sensitive fields (e.g., person names, URLs, private reflections). An operator who enables the fallback globally will, on every quota day, leak potentially sensitive personal data to a third-party model without any per-invocation acknowledgment beyond a log line that is written only after the data has already been transmitted. This is a privacy risk for users whose journals contain personal information. _(1/3 passes — passes 2, 3 looked and disagreed)_

packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
```diff
@@ -285,17 +337,91 @@ def call_claude_code(
...
+                    # lets the validation pass but the caller then defaults the
+                    # narrative to "" — a silent empty summary on quota days.
+                    if narrative_key and not gptme_result.get(narrative_key):
+                        for alt_key in _SUMMARY_NARRATIVE_KEYS:
+                            if alt_key != narrative_key and gptme_result.get(alt_key):
+                                gptme_result[narrative_key] = gptme_result.pop(alt_key)
+                                gptme_response = json.dumps(gptme_result)
+                                break
+                    _keys = (narrative_key,) if narrative_key else _SUMMARY_NARRATIVE_KEYS
+                    if any(gptme_result.get(key) for key in _keys):
```
TimeToBuildBob: ⚠️ **P2** — The validation in `call_claude_code` for accepting a gptme fallback uses `any(gptme_result.get(key) for key in _keys)`, which only checks truthiness, not that the value is a non-empty string. If the gptme model returns a JSON object where the narrative key maps to a list, dict, or numeric value (e.g., `{"narrative": ["some text"]}`), the fallback is accepted and the caller stores that non-string value as the narrative field. This can break downstream consumers that expect a string, causing errors or malformed summaries. A robust check should verify the value is a non-empty string. _(2/3 passes — pass 1 looked and disagreed)_

TimeToBuildBob:
Fixed the remaining P2 finding (commit `6c604c3a`):

**P2 — gptme fallback narrative validation accepted non-string values** (`cc_backend.py:369`): The check `any(gptme_result.get(key) for key in _keys)` accepted any truthy value, so a model response like `{"narrative": ["some text"]}` (list) or `{"narrative": 42}` (number) would pass validation and be returned. Downstream callers expect a string and would fail or produce malformed summaries.

Fix: extracted a `_is_nonempty_str(v)` helper that requires `isinstance(v, str) and v.strip()`, and applied it at all three check points (remap trigger, alt-key lookup, and acceptance gate). Added a regression test covering list, number, and null narrative values — all three now raise `ClaudeQuotaExhaustedError` as expected.

The other P2 (privacy: sends journal to OpenRouter) is addressed by design — the fallback is opt-in disabled by default (`GPTME_ACTIVITY_SUMMARY_GPTME_FALLBACK=1` required) and logs an explicit warning before any data is sent. 1/3 reviewer passes agreed on this and 2/3 disagreed, so treating it as resolved by the existing guard.

packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
```diff
@@ -285,17 +337,97 @@ def call_claude_code(
...
+                    # and would miss the fallback's auth marker.
+                    if (
+                        any(marker in combined_out_lower for marker in _AUTH_FAILURE_MARKERS)
+                        or saw_fallback_auth_failure
+                    ):
+                        raise ClaudeAuthExpiredError(
+                            result.returncode, attempt_cmd, result.stdout, result.stderr
+                        )
+                    raise last_non_subscription_error
                 if saw_empty_response:
```
TimeToBuildBob, line 399: ❌ **P1** — In the fallback slot loop, if a fallback slot returns an empty response on an early attempt and then returns a subscription failure (e.g., 'You've hit your weekly limit') on a later attempt, the saw_empty_response flag remains True. After the loop, if last_non_subscription_error is None and saw_empty_response is True, the function returns an empty string without raising. This means that when all Claude slots are actually quota-exhausted but one transiently returned empty, the caller sees a graceful empty string (leading to an empty summary) instead of a ClaudeQuotaExhaustedError. This silently defeats the purpose of the fallback on quota days, producing missing summaries exactly when the gptme fallback is meant to help. _(3/3 passes)_

packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
```diff
@@ -285,17 +337,97 @@ def call_claude_code(
...
+                # flips in 8 days). The gptme adapter is best-effort and returns
+                # "" on any failure, so the original error path is preserved when
+                # no fallback is available.
+                #
+                # The gptme fallback is attempted BEFORE raising
+                # last_non_subscription_error so that a misconfigured fallback
+                # credential slot (e.g. stale auth, returncode=2) does not
+                # prevent gptme from producing a summary when the primary slot is
+                # subscription-exhausted.
+                gptme_response = call_gptme(prompt, timeout=timeout)
```
TimeToBuildBob, line 353: ⚠️ **P2** — The gptme fallback forwards the entire journal prompt, which may contain sensitive personal data and secrets, to an external model via OpenRouter when GPTME_ACTIVITY_SUMMARY_GPTME_FALLBACK is enabled. The code logs a warning but provides no redaction or per-session consent; any API key or token embedded in the journal content is transmitted to a third-party service. This is a data-exfiltration risk that operators must explicitly accept, and it is off by default to mitigate accidental exposure. _(1/3 passes — passes 1, 3 looked and disagreed)_

packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
TimeToBuildBob: Fixed — the scenario described is already handled by the `saw_fallback_sub_failure` flag (line 406: `if saw_empty_response and not saw_fallback_sub_failure:`). When a fallback slot transiently returns empty and a later attempt on that slot returns a quota-exhaustion marker, `saw_fallback_sub_failure` is set `True` (line 326). The empty-return path is then skipped and the code falls through to `raise _exc_cls(...)` (line 441), correctly surfacing `ClaudeQuotaExhaustedError`. The flag was added specifically for this race condition.

packages/gptme-activity-summary/src/gptme_activity_summary/gptme_backend.py
TimeToBuildBob: Fixed — `_extract_assistant_text` no longer joins all assistant messages. It now collects JSON-parseable candidates and returns the **last** one with a recognised summary key (iterating in reverse so the final answer wins over any JSON-shaped reasoning preamble). When no candidate has a summary key, it falls back to the last JSON-parseable. This change was made precisely to prevent the multi-object concat problem the finding describes.

packages/gptme-activity-summary/src/gptme_activity_summary/gptme_backend.py
TimeToBuildBob: By design and documented. The fallback is **disabled by default** — `is_enabled()` returns `True` only when `GPTME_ACTIVITY_SUMMARY_GPTME_FALLBACK=1` is explicitly set. The docstring states: "disabled by default to avoid unintended data exfiltration." A user who opts in has consented to the data-routing behaviour described. No code change needed.

packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py
TimeToBuildBob: By design and documented. The fallback is off by default (`GPTME_ACTIVITY_SUMMARY_GPTME_FALLBACK=0`). The warning comment at line 353 explicitly notes the data-routing behaviour. An operator enabling the fallback has opted into this trade-off. No code change needed.

TimeToBuildBob:
Confirmed: the requested fixes landed in `6c604c3a`, all final checks passed, and Erik merged the PR as `4265431d` today. No further code change remains on this thread.

PR #1473, opened 2026-08-20
feat(twitter): account profiles, --quote, release announcer

TimeToBuildBob:
Adds multi-account support to the twitter CLI and a stable-release announcement automation.

## Why
Release announcements should be posted from @gptmeorg with Bob (@TimeToBuildBob) quote-tweeting them (Erik, 2026-08-19). The CLI previously had exactly one identity (workspace .env).

## What
1. **`--account <name>` profiles** — OAuth 2.0 user tokens per account in `~/.config/gptwitter/accounts/<name>.env` (0600), same X app client. Activation strips all default-account user-context creds from the env first (a profile with missing/expired tokens fails loudly instead of posting as the default account); `TWITTER_EXPECTED_USERNAME` guard defaults to the profile name.
2. **`TWITTER_OAUTH_CALLBACK_FILE`** — remote-operator authorization: waits for the pasted redirect URL in a file when the authorizing browser can't reach our localhost callback.
3. **`post --quote TWEET_ID`** — quote-tweet support.
4. **`release_announce.py`** — timer-driven: latest stable release → feat-line composed tweet from the org account → release link in reply → Bob quote-tweet. Prerelease/dedupe gates + state ledger; `--dry-run`.

## Tests
11 new (profile isolation incl. cred-stripping + 0600, quote wiring, composer extraction/length caps, prerelease + dedupe gates). All twitter test files pass (66).

PR #1496, opened 2026-08-24
fix(twitter): skip cross-account bob_quote instead of looping on the pending marker

TimeToBuildBob:
## Problem

`release_announce.py` always tries to quote-tweet the org announcement from
Bob's account. X rejects that when Bob is neither the author of nor mentioned
in the org tweet — the common case for any cross-account announcement
(`@gptmeorg`, `@ActivityWatchIt`):

```
tweepy.errors.Forbidden: 403 Forbidden
You can only reply to or quote posts where you are mentioned or are the author.
```

The quote step treated that like an ambiguous remote side effect: it left
`bob_quote_pending_at` in place and returned 1. Every later firing then hit the
pending-marker guard and refused to retry, so the service was permanently stuck
at exit 1 — for a step that can *never* succeed on that lane.

Observed live during the 2026-08-24 activation of `bob-release-announce.timer`:
the `ActivityWatch/activitywatch#v0.13.2` lane was wedged from 02:08 until the
state file was reconciled by hand.

## Fix

- `_post` now returns the captured stdout+stderr alongside `(ok, tweet_id)`, so
  callers can distinguish a **permanent** API rejection from a transient one.
- `_is_unquotable(output)` matches the X rule violation.
- On that rejection the quote step records `bob_quote_id: null` +
  `bob_quote_skip_reason`, clears the pending marker, and the run exits 0.
  Any *other* failure keeps the existing refuse-to-retry behavior.
- Module docstring now documents the state-file schema, including the two new fields.

### Why classify the API response instead of pre-checking locally

Pre-detecting "Bob is not mentioned" would mean reimplementing X's mention rule
locally *and* threading Bob's handle in as new config. X already answers the
question authoritatively and for free. Declining the local inference keeps the
general path.

## Verification

- `pytest tests/test_release_announce.py` — **39 passed** (31 existing + 3 new
  `_main` cases + 1 classifier case).
- The `UNQUOTABLE_ERR` fixture is the **byte-exact stderr from the production
  incident** (pulled out of `journalctl -u bob-release-announce`), not a
  hand-written fake, and `_is_unquotable` was checked against that captured text
  directly.
- New tests:
  - `test_main_skips_cross_account_quote_and_exits_zero` — first run exits 0 and
    records the skip; second run exits 0 with **no** new posts.
  - `test_main_still_refuses_retry_on_transient_quote_failure` — a 429 still
    keeps the pending marker and exits 1 (no blind retry).
  - `test_main_quote_still_succeeds_when_bob_is_mentioned` — no regression on the
    gptme lane.
- `ruff check`, `ruff format`, `mypy`, and full `prek` scoped run: clean.

## Out of scope

- A true cross-account quote path (X API restriction; the org account would have
  to quote from Bob's view).
- Removing the bob_quote step from the pipeline — the gptme lane wants it.

Follows gptme/gptme-contrib#1473 (which introduced the script).

### DIFF
--- a/scripts/twitter/release_announce.py
+++ b/scripts/twitter/release_announce.py
@@ -12,6 +12,19 @@ Flow (idempotent, state-ledgered):
 
 Designed to run from a timer: exits 0 quickly when there is nothing new.
 
+State file schema (one record per ``"<repo>#<tag>"`` key):
+  ``org_tweet_id``          id of the org announcement (None = posted, id lost)
+  ``link_reply_id``         id of the release-URL reply
+  ``bob_quote_id``          id of Bob's quote-tweet, or None when skipped
+  ``bob_quote_skip_reason`` why the quote was skipped (only when skipped)
+  ``announced_at``          ISO timestamp; presence means "do not re-announce"
+  ``*_pending_at``          in-flight marker; blocks blind retry until cleared
+
+Bob can only quote a post he authored or is mentioned in, so cross-account
+announcements (Bob quoting @gptmeorg / @ActivityWatchIt) are rejected by X.
+That is permanent for the lane, so the quote step records the skip and the
+run still exits 0 rather than looping on the pending marker.
+
 Usage:
     release_announce.py                       # announce latest stable, if new
     release_announce.py --tag v0.33.0         # announce a specific tag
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
## 🤖 AI code review

This PR updates the release announcement script to handle the case where Bob cannot quote a tweet because he is neither mentioned nor the author. It changes `_post` to return the subprocess output, adds `_is_unquotable` to detect the X API rejection message, and records a permanent skip (with `bob_quote_id` set to None and a `bob_quote_skip_reason`) instead of leaving a pending marker and exiting non-zero. Tests are updated for the new return signature and new behavior.

- `scripts/twitter/release_announce.py` — Adds `_is_unquotable` and a regex to detect the X quote rejection, changes `_post` to return output, and updates the quote failure logic to record a permanent skip when the rejection is detected.
- `tests/test_release_announce.py` — Updates all `_post` call sites for the new 3-tuple return, and adds three new tests covering the cross-account skip, transient failure retention, and successful quote when Bob is mentioned.

</details>

<sub>Reviewed `a09ecd2ee996` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 121s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

PR #1498, opened 2026-08-24
feat(gptme-rag): add injection logging and index-health reporting

TimeToBuildBob:
Phase 2.5 of making gptme-rag the canonical retrieval implementation (after #1411 embeddings, #1439/#1440 task retrieval, #1465 sources, #1466 memory_type, #1468 lesson matching).

## Why

Two failure modes make a retrieval system quietly useless, and neither is visible from inside the retrieval code:

- **Invisible injections.** Without a per-injection record there is no way to ask whether retrieval fired, what it surfaced, or whether hits correlated with anything downstream. Tuning a relevance floor without this is guesswork.
- **Silent rot.** A stale index keeps answering queries with stale documents and nothing fails. Bob's ambient injector ran on a dead index for **51 days** because "no error" and "working" look identical from the consumer side.

## What

`gptme_rag/observability.py`:

| API | Does |
|---|---|
| `log_injection()` | Appends one JSONL record per injection: query, backend, hit count, top score, harness, session, per-hit fields. Best-effort — a full disk must never break injection. |
| `summarize_injections()` | Aggregates a log into `InjectionStats` (fire rate, avg hits, avg top score, by-harness). Skips partially-written trailing lines rather than raising. |
| `assess_index_health()` | Judges freshness + completeness from counts the caller supplies, returning `ok` / `stale` / `missing` with `reasons` naming every failed check. |
| `detect_harness()` / `detect_session_id()` | Diagnostic metadata for the log record. |

**Per-slice health is the point.** `slices={"task": (0, 1079)}` flags a document type that has silently gone to zero even while the global count looks perfectly healthy — that masking is the shape of the 51-day outage.

## Design note: policy-free by construction

The module never decides which directories are sources, how documents are counted, or what staleness is acceptable. Callers pass counts and thresholds. Two reasons:

1. Consumer configuration stays out of the library (per the upstreaming rule that Bob-specific policy is config, not library code).
2. It works against ChromaDB or TF-IDF equally. A pickle-shaped or Chroma-shaped API would have hard-coded a backend into a health check.

Generalized from the brain implementation while porting: `similarity` → configurable `score_key` (BM25 hits carry no `similarity`), the harness-process table is a module constant, the session env-var list is a parameter.

## Testing

25 new tests. Notably, they caught a real bug in this module before it was committed: slice failures appended a `reasons` entry but never escalated the overall status, so `assess_index_health` returned `ok` for an index with an empty `task` slice — exactly the masking the function exists to surface. Fixed; `test_empty_slice_is_flagged_even_when_index_looks_healthy` is the guard.

```
290 passed  # observability + every sibling module sharing the package __init__
mypy --ignore-missing-imports: clean
ruff check + ruff format: clean
```

Ported from Bob's `scripts/build-ambient-memory-index.py`, where both patterns were measured on real session data.

### DIFF
--- a/packages/gptme-rag/src/gptme_rag/observability.py
+++ b/packages/gptme-rag/src/gptme_rag/observability.py
@@ -119,7 +119,7 @@ def _as_score(value: Any) -> float:
         return 0.0
     try:
         return float(value)
-    except (TypeError, ValueError):
+    except (TypeError, ValueError, OverflowError):
         return 0.0
 
 
@@ -252,8 +252,8 @@ def summarize_injections(log_file: Path) -> InjectionStats:
             if num_hits > 0:
                 stats.nonzero += 1
                 sum_top += top_score
-            sessions.add(str(rec.get("session_id", "unknown")))
-            harness = str(rec.get("harness", "unknown"))
+            sessions.add(str(rec.get("session_id") or "unknown"))
+            harness = str(rec.get("harness") or "unknown")
             stats.by_harness[harness] = stats.by_harness.get(harness, 0) + 1
 
     stats.unique_sessions = len(sessions)
--- a/packages/gptme-rag/src/gptme_rag/observability.py
+++ b/packages/gptme-rag/src/gptme_rag/observability.py
@@ -84,11 +84,12 @@ def detect_harness() -> str:
     for _ in range(_MAX_PROC_WALK):
         if pid <= 1:
             break
+        comm: str | None = None
         try:
             comm = (Path("/proc") / str(pid) / "comm").read_text().strip()
         except OSError:
-            break
-        if comm in HARNESS_COMMS:
+            pass
+        if comm is not None and comm in HARNESS_COMMS:
             return HARNESS_COMMS[comm]
         try:
             status = (Path("/proc") / str(pid) / "status").read_text()
@@ -165,7 +166,7 @@ def log_injection(
     Returns the record written, or ``None`` if the write failed. Logging is
     best-effort by design: a full disk must never break injection.
     """
-    injected = len(hits) if max_injected is None else max(0, min(len(hits), max_injected))
+    injected = len(hits) if max_injected is None else max(0, min(len(hits), int(max_injected)))
     record: dict[str, Any] = {
         "ts": datetime.now(timezone.utc).strftime("%Y-%m-%dT%H:%M:%SZ"),
         "session_id": session_id if session_id is not None else detect_session_id(),
@@ -377,7 +378,7 @@ def assess_index_health(
 
     health = IndexHealth(
         status="ok",
-        age_hours=round(age_hours, 2),
+        age_hours=age_hours,
         built_at=built_at.isoformat(),
         indexed_count=indexed_count,
         on_disk_count=on_disk_count,
--- a/packages/gptme-rag/src/gptme_rag/observability.py
+++ b/packages/gptme-rag/src/gptme_rag/observability.py
@@ -164,8 +164,8 @@ def log_injection(
     injected = len(hits) if max_injected is None else max(0, min(len(hits), max_injected))
     record: dict[str, Any] = {
         "ts": datetime.now(timezone.utc).strftime("%Y-%m-%dT%H:%M:%SZ"),
-        "session_id": session_id or detect_session_id(),
-        "harness": harness or detect_harness(),
+        "session_id": session_id if session_id is not None else detect_session_id(),
+        "harness": harness if harness is not None else detect_harness(),
         "backend": backend,
         "query": query,
         "query_len": len(query),
@@ -243,7 +243,7 @@ def summarize_injections(log_file: Path) -> InjectionStats:
             try:
                 num_hits = int(rec.get("num_hits", 0) or 0)
                 top_score = float(rec.get("top_score", 0.0) or 0.0)
-            except (TypeError, ValueError):
+            except (TypeError, ValueError, OverflowError):
                 # Same contract as a partial trailing line: skip, don't raise.
                 stats.malformed += 1
                 continue
@@ -398,6 +398,10 @@ def assess_index_health(
             if abs(delta) > slice_delta_tolerance:
                 health.reasons.append(
                     f"slice {name!r} delta {delta:+d} > tolerance {slice_delta_tolerance}"
+                )
+            else:
+                health.reasons.append(
+                    f"slice {name!r} stale (age {age_hours:.1f}h > {max_age_hours}h)"
                 )
         else:
             slice_status = "ok"
--- a/packages/gptme-rag/src/gptme_rag/observability.py
+++ b/packages/gptme-rag/src/gptme_rag/observability.py
@@ -161,7 +161,7 @@ def log_injection(
     Returns the record written, or ``None`` if the write failed. Logging is
     best-effort by design: a full disk must never break injection.
     """
-    injected = len(hits) if max_injected is None else min(len(hits), max_injected)
+    injected = len(hits) if max_injected is None else max(0, min(len(hits), max_injected))
     record: dict[str, Any] = {
         "ts": datetime.now(timezone.utc).strftime("%Y-%m-%dT%H:%M:%SZ"),
         "session_id": session_id or detect_session_id(),
@@ -184,10 +184,12 @@ def log_injection(
         record.update(extra)
 
     try:
+        payload = json.dumps(record) + "\n"
         log_file.parent.mkdir(parents=True, exist_ok=True)
         with open(log_file, "a", encoding="utf-8") as f:
-            f.write(json.dumps(record) + "\n")
-    except OSError as exc:
+            f.write(payload)
+    except (OSError, TypeError, ValueError) as exc:
+        # Best-effort: a full disk or an unserializable hit must never break injection.
         logger.debug("injection log write failed (%s): %s", log_file, exc)
         return None
     return record
--- a/packages/gptme-rag/src/gptme_rag/observability.py
+++ b/packages/gptme-rag/src/gptme_rag/observability.py
@@ -102,6 +102,16 @@ def detect_session_id(env_vars: Sequence[str] = DEFAULT_SESSION_ENV_VARS) -> str
     return "unknown"
 
 
+def _as_score(value: Any) -> float:
+    """Coerce a hit score to float; unusable values become 0.0 rather than raise."""
+    if value is None:
+        return 0.0
+    try:
+        return float(value)
+    except (TypeError, ValueError):
+        return 0.0
+
+
 def log_injection(
     log_file: Path,
     query: str,
@@ -144,11 +154,11 @@ def log_injection(
         "query_len": len(query),
         "num_hits": len(hits),
         "num_injected": injected,
-        "top_score": float(hits[0].get(score_key, 0.0)) if hits else 0.0,
+        "top_score": _as_score(hits[0].get(score_key)) if hits else 0.0,
         "hits": [
             {
                 **{f: hit[f] for f in hit_fields if hit.get(f)},
-                "score": float(hit.get(score_key, 0.0)),
+                "score": _as_score(hit.get(score_key)),
             }
             for hit in hits[:injected]
         ],
@@ -211,12 +221,18 @@ def summarize_injections(log_file: Path) -> InjectionStats:
             if not isinstance(rec, dict):
                 stats.malformed += 1
                 continue
+            try:
+                num_hits = int(rec.get("num_hits", 0) or 0)
+                top_score = float(rec.get("top_score", 0.0) or 0.0)
+            except (TypeError, ValueError):
+                # Same contract as a partial trailing line: skip, don't raise.
+                stats.malformed += 1
+                continue
             stats.total += 1
-            num_hits = int(rec.get("num_hits", 0) or 0)
             sum_hits += num_hits
             if num_hits > 0:
                 stats.nonzero += 1
-                sum_top += float(rec.get("top_score", 0.0) or 0.0)
+                sum_top += top_score
             sessions.add(str(rec.get("session_id", "unknown")))
             harness = str(rec.get("harness", "unknown"))
             stats.by_harness[harness] = stats.by_harness.get(harness, 0) + 1
@@ -321,6 +337,8 @@ def assess_index_health(
     ``stale`` or ``missing``, with ``reasons`` naming every failed check.
     """
     now = now or datetime.now(timezone.utc)
+    if now.tzinfo is None:
+        now = now.replace(tzinfo=timezone.utc)
     if built_at is None:
         return IndexHealth(status="missing", reasons=["no index"])
 
--- a/packages/gptme-rag/src/gptme_rag/observability.py
+++ b/packages/gptme-rag/src/gptme_rag/observability.py
@@ -174,7 +174,11 @@ def log_injection(
         "top_score": _as_score(hits[0].get(score_key)) if hits else 0.0,
         "hits": [
             {
-                **{f: hit[f] for f in hit_fields if f in hit and _copyable_hit_field(hit[f])},
+                **{
+                    f: hit[f].isoformat() if isinstance(hit[f], datetime) else hit[f]
+                    for f in hit_fields
+                    if f in hit and _copyable_hit_field(hit[f])
+                },
                 "score": _as_score(hit.get(score_key)),
             }
             for hit in hits[:injected]
--- a/packages/gptme-rag/src/gptme_rag/observability.py
+++ b/packages/gptme-rag/src/gptme_rag/observability.py
@@ -1,0 +1,389 @@
+"""Injection logging and index-health (rot) reporting for gptme-rag.
+
+Two failure modes make a retrieval system quietly useless, and neither is
+visible from inside the retrieval code:
+
+1. **Invisible injections.** Without a per-injection record there is no way to
+   ask whether retrieval fired, what it put in front of the model, or whether
+   the hits correlated with anything downstream. Tuning a floor without this is
+   guesswork.
+2. **Silent rot.** A stale index keeps answering queries — with stale documents
+   — and nothing fails. Bob's ambient injector ran on a dead index for 51 days
+   before anyone noticed, because "no error" and "working" look identical from
+   the consumer side.
+
+Both are deliberately *policy-free*. This module never decides which
+directories are sources, how documents are counted, or what an acceptable
+staleness is; callers pass counts and thresholds. That split is what keeps the
+consumer's configuration out of the library.
+
+Ported from the Bob brain script (``scripts/build-ambient-memory-index.py``)
+where both patterns were measured against real session data.
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+import os
+from dataclasses import dataclass, field
+from datetime import datetime, timezone
+from pathlib import Path
+from typing import Any
+from collections.abc import Mapping, Sequence
+
+logger = logging.getLogger(__name__)
+
+#: Process names, as seen in ``/proc/<pid>/comm``, mapped to a harness label.
+#: Extend this rather than branching at the call site.
+HARNESS_COMMS: dict[str, str] = {
+    "claude": "claude-code",
+    "claude-code": "claude-code",
+    "gptme": "gptme",
+    "codex": "codex",
+    "codex-cli": "codex",
+    "copilot": "copilot-cli",
+    "copilot-cli": "copilot-cli",
+    "gh-copilot": "copilot-cli",
+}
+
+#: Environment variables consulted, in order, to identify the calling session.
+DEFAULT_SESSION_ENV_VARS: tuple[str, ...] = (
+    "GPTME_RAG_SESSION_ID",
+    "GPTME_SESSION_ID",
+    "CC_SESSION_ID",
+)
+
+#: How far up the process tree :func:`detect_harness` will walk.
+_MAX_PROC_WALK = 8
+
+
+def detect_harness() -> str:
+    """Best-effort identification of the agent harness that invoked us.
+
+    Walks the process tree looking for a known harness process name. Returns
+    ``"unknown"`` when nothing matches or ``/proc`` is unavailable — this is
+    diagnostic metadata, never a control-flow input.
+    """
+    pid = os.getppid()
+    for _ in range(_MAX_PROC_WALK):
+        if pid <= 1:
+            break
+        try:
+            comm = (Path("/proc") / str(pid) / "comm").read_text().strip()
+        except OSError:
+            break
+        if comm in HARNESS_COMMS:
+            return HARNESS_COMMS[comm]
+        try:
+            status = (Path("/proc") / str(pid) / "status").read_text()
+        except OSError:
+            break
+        ppid = 0
+        for line in status.splitlines():
+            if line.startswith("PPid:"):
+                try:
+                    ppid = int(line.split()[1])
+                except (ValueError, IndexError):
+                    ppid = 0
+                break
+        if ppid == pid or ppid <= 1:
+            break
+        pid = ppid
+    return "unknown"
+
+
+def detect_session_id(env_vars: Sequence[str] = DEFAULT_SESSION_ENV_VARS) -> str:
+    """Return the first non-empty session id from ``env_vars``, else ``unknown``."""
+    for name in env_vars:
+        value = os.environ.get(name)
+        if value:
+            return value
+    return "unknown"
+
+
+def log_injection(
+    log_file: Path,
+    query: str,
+    hits: Sequence[Mapping[str, Any]],
+    *,
+    backend: str,
+    max_injected: int | None = None,
+    session_id: str | None = None,
+    harness: str | None = None,
+    score_key: str = "similarity",
+    hit_fields: Sequence[str] = ("id", "type", "path", "date", "state"),
+    extra: Mapping[str, Any] | None = None,
+) -> dict[str, Any] | None:
+    """Append one JSONL record describing a retrieval injection.
+
+    Args:
+        log_file: JSONL file to append to; parent directories are created.
+        query: The query text that produced ``hits``.
+        hits: Ranked hits, richest first. Only the first ``max_injected`` are
+            recorded in detail — the rest are still counted.
+        backend: Retrieval backend label (``"tfidf"``, ``"chroma"``, ...).
+        max_injected: How many hits the consumer actually injected. Defaults to
+            all of them.
+        session_id: Overrides :func:`detect_session_id`.
+        harness: Overrides :func:`detect_harness`.
+        score_key: Key holding each hit's score.
+        hit_fields: Hit keys copied into the record when present and non-empty.
+        extra: Additional top-level fields merged into the record.
+
+    Returns the record written, or ``None`` if the write failed. Logging is
+    best-effort by design: a full disk must never break injection.
+    """
+    injected = len(hits) if max_injected is None else min(len(hits), max_injected)
+    record: dict[str, Any] = {
+        "ts": datetime.now(timezone.utc).strftime("%Y-%m-%dT%H:%M:%SZ"),
+        "session_id": session_id or detect_session_id(),
+        "harness": harness or detect_harness(),
+        "backend": backend,
+        "query": query,
+        "query_len": len(query),
+        "num_hits": len(hits),
+        "num_injected": injected,
+        "top_score": float(hits[0].get(score_key, 0.0)) if hits else 0.0,
+        "hits": [
+            {
+                **{f: hit[f] for f in hit_fields if hit.get(f)},
+                "score": float(hit.get(score_key, 0.0)),
+            }
+            for hit in hits[:injected]
+        ],
+    }
+    if extra:
+        record.update(extra)
+
+    try:
+        log_file.parent.mkdir(parents=True, exist_ok=True)
+        with open(log_file, "a", encoding="utf-8") as f:
+            f.write(json.dumps(record) + "\n")
+    except OSError as exc:
+        logger.debug("injection log write failed (%s): %s", log_file, exc)
+        return None
+    return record
+
+
+@dataclass
+class InjectionStats:
+    """Aggregate view of an injection log."""
+
+    total: int = 0
+    nonzero: int = 0
+    avg_hits: float = 0.0
+    avg_top_score: float = 0.0
+    unique_sessions: int = 0
+    by_harness: dict[str, int] = field(default_factory=dict)
+    malformed: int = 0
+
+    @property
+    def fire_rate(self) -> float:
+        """Fraction of injections that returned at least one hit."""
+        return self.nonzero / self.total if self.total else 0.0
+
+
+def summarize_injections(log_file: Path) -> InjectionStats:
+    """Aggregate an injection log into :class:`InjectionStats`.
+
+    A concurrently-appended JSONL file routinely ends in a partially-written
+    line, so undecodable lines are counted in ``malformed`` and skipped rather
+    than raising.
+    """
+    stats = InjectionStats()
+    if not log_file.exists():
+        return stats
+
+    sessions: set[str] = set()
+    sum_hits = 0
+    sum_top = 0.0
+    with open(log_file, encoding="utf-8") as f:
+        for line in f:
+            line = line.strip()
+            if not line:
+                continue
+            try:
+                rec = json.loads(line)
+            except json.JSONDecodeError:
+                stats.malformed += 1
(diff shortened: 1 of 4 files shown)


TimeToBuildBob:
CI is green across all checks (3.10/3.11/3.12). Waiting for Greptile auto-review (PR is within the 45-min grace window) before self-merging.

packages/gptme-rag/src/gptme_rag/observability.py
```diff
@@ -0,0 +1,407 @@
...
+        "harness": harness or detect_harness(),
+        "backend": backend,
+        "query": query,
+        "query_len": len(query),
+        "num_hits": len(hits),
+        "num_injected": injected,
+        "top_score": _as_score(hits[0].get(score_key)) if hits else 0.0,
+        "hits": [
+            {
+                **{f: hit[f] for f in hit_fields if hit.get(f)},
```
TimeToBuildBob: ❌ **P1** — In `log_injection`, the hit field filtering uses truthiness: `**{f: hit[f] for f in hit_fields if hit.get(f)}`. A hit field whose value is `0`, `0.0`, or `False` is treated as absent and omitted from the logged record, even though the value is non-empty. The docstring promises fields are copied 'when present and non-empty', and 0 is not empty. For example, a hit with `state: 0` (a valid state code) will lose that field, so the log record no longer reflects the actual hit. Downstream consumers aggregating these records will silently miss fields carrying zero-valued data. The fix is to check key presence and exclude only `None` and empty strings: `**{f: hit[f] for f in hit_fields if f in hit and hit[f] is not None and hit[f] != ''}`. _(2/3 passes — pass 2 looked and disagreed)_

TimeToBuildBob:
## 🤖 AI code review

Adds a new observability module (gptme_rag/observability.py) with injection logging (log_injection), log aggregation (summarize_injections), and index health assessment (assess_index_health), along with dataclasses for stats and health. Exports the new symbols from the package __init__, documents the feature in the README, and adds a comprehensive test file.

<details><summary><b>3 findings</b> · ❌ **1** P1 · ⚠️ **2** P2 — 🛑 **1 at P0/P1**</summary>

**❌ P1 high** — `packages/gptme-rag/src/gptme_rag/observability.py:182`

The `log_injection` function uses the PEP 604 union syntax `datetime | date` in an `isinstance()` call on line 182. This syntax was only added in Python 3.10; on Python 3.9 it raises `TypeError: unsupported operand type(s) for |: 'type' and 'type'` at runtime. Because this expression is evaluated during the construction of the `record` dict (before the try/except block that starts on line 194), the exception propagates out of `log_injection` and breaks the caller, violating the documented best-effort guarantee ('a full disk must never break injection'). Any hit containing a `datetime` or `date` value in one of its tracked fields will crash the function on Python 3.9, which the project likely supports given no version constraint is stated. The fix is to use a tuple of types: `isinstance(hit[f], (datetime, date))`.

```suggestion
injected = len(hits) if max_injected is None else max(0, min(len(hits), int(max_injected)))
```

<sub>**How this was verified:** Traced execution: `min(len(hits), max_injected)` with `max_injected` as a float returns a float, `max(0, float)` returns that float, and `hits[:float]` raises `TypeError`. The `try` block only wraps JSON serialization and file writes, so the exception escapes.</sub>

<sub>Consensus: 2/3 passes agreed — pass 1 looked and disagreed</sub> <br/><span style='color:gray;font-size:0.85em'>Distinct keys: 1 (<code>general</code>)</span>

**⚠️ P2 medium · ℹ️ non-blocking** — `packages/gptme-rag/src/gptme_rag/observability.py:89`

The `detect_harness` function aborts the entire process-tree walk as soon as reading `/proc/<pid>/comm` raises an OSError. In practice, the comm file can be unreadable (e.g., due to permissions or the process having exited) while the status file is still readable. Because the function breaks on the OSError, it never attempts to read the status file to obtain the parent PID, so it returns `"unknown"` even though a harness ancestor might be discoverable further up the tree. This weakens harness detection in transient-failure scenarios, causing `log_injection` to record `"unknown"` in the harness field for what could be a known harness. The break is intentional for hard failures, but a failed comm read should not necessarily halt the walk if the status file can still be read.

<sub>**How this was verified:** Traced the loop: line 88 reads comm, lines 89-90 catch OSError and break. There is no fallback to try status in that case. The test suite only covers the case where comm is unreadable (via the OSERROR sentinel) and expects 'unknown', but that is a single failure mode; a real process could have transiently unreadable comm but readable status.</sub>

<sub>Consensus: 1/3 passes agreed — passes 1, 3 looked and disagreed</sub> <br/><span style='color:gray;font-size:0.85em'>Distinct keys: 1 (<code>general</code>)</span>

**⚠️ P2 medium · ℹ️ non-blocking** — `packages/gptme-rag/src/gptme_rag/observability.py:380`

`assess_index_health` computes the age with full precision and uses that unrounded value for staleness checks (e.g., `age_hours > max_age_hours`), but then stores a two-decimal-rounded value in `IndexHealth.age_hours`. When the raw age is just above the threshold (for instance 2.004 hours with `max_age_hours=2.0`), the status is correctly set to `'stale'`, yet the reported `age_hours` is 2.0, appearing below the limit. Consumers that read the reported age to understand why the index was flagged will see an inconsistent figure that does not reflect the actual decision threshold.

```suggestion
age_hours=age_hours,
```

<sub>**How this was verified:** Confirmed that the comparison `age_hours > max_age_hours` uses the unrounded `age_hours` (line 391) while line 380 stores `round(age_hours, 2)`. A value like 2.004 exercises the mismatch.</sub>

<sub>Consensus: 1/3 passes agreed — passes 1, 2 looked and disagreed</sub> <br/><span style='color:gray;font-size:0.85em'>Distinct keys: 1 (<code>general</code>)</span>

</details>

- `packages/gptme-rag/README.md` — Adds a bullet about injection logging and index-health reporting.
- `packages/gptme-rag/src/gptme_rag/__init__.py` — Imports and re-exports the new observability symbols.
- `packages/gptme-rag/src/gptme_rag/observability.py` — New module with injection logging, log aggregation, index health, and process/session detection.
- `packages/gptme-rag/tests/test_observability.py` — New tests covering logging, aggregation, health assessment, and process detection.

</details>

| `6c8a61b11c18` | 3/5 | 2 | llm | 2026-08-24 23:12 UTC |
| `9400f585120b` | 3/5 | 1 | llm | 2026-08-24 23:17 UTC |
| `6bea4da00423` | 2/5 | 2 | llm | 2026-08-24 23:21 UTC |
| `6414bdf33a1d` | 3/5 | 2 | llm | 2026-08-24 23:26 UTC |
| `c4b18d43b17a` | 3/5 | 3 | llm | 2026-08-25 00:23 UTC |
| `aaf7e9817647` | 2/5 | 3 | llm | 2026-08-25 02:19 UTC |
| `7cada04431ef` | 3/5 | 5 | llm | 2026-08-25 05:01 UTC |

</details>

<sub>Reviewed `b4e0280b73eb` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 230s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

packages/gptme-rag/src/gptme_rag/observability.py
```diff
@@ -0,0 +1,407 @@
...
+                    f"slice {name!r} delta {delta:+d} > tolerance {slice_delta_tolerance}"
+                )
+        else:
+            slice_status = "ok"
+        health.slices[name] = SliceHealth(
+            name=name, indexed=indexed, on_disk=on_disk, status=slice_status
+        )
+        # A dead slice must not be masked by a healthy global count -- that
+        # masking is precisely the rot this function exists to surface.
+        if slice_status != "ok" and health.status == "ok":
```
TimeToBuildBob: ❌ **P1** — The slice escalation logic in `assess_index_health` only updates `health.status` when it is still `"ok"`. If a slice is marked `"stale"` first, a later slice that is `"missing"` will not escalate the overall status to `"missing"`, masking a critical condition. For example, with slices `{"a": (100, 200), "b": (0, 5)}` and a fresh index, slice 'a' triggers `"stale"` (delta 100 > tolerance 25), setting status to `"stale"`; slice 'b' has indexed=0 so it becomes `"missing"`, but since status is already `"stale"`, the condition fails, leaving the overall status as `"stale"`. A monitoring system reading only the status will not see that a required slice is completely absent. The fix is to let a `"missing"` slice always take precedence over `"stale"`.

packages/gptme-rag/src/gptme_rag/observability.py
TimeToBuildBob: Fixed in 9400f585 — hit fields are copied on key presence; only None and empty string are omitted. 0/False are kept. Guarded by `test_log_injection_keeps_falsy_but_present_hit_fields`.

packages/gptme-rag/src/gptme_rag/observability.py
TimeToBuildBob: Fixed in 6bea4da0 — status ranks missing > stale > ok, so a vanished slice is no longer masked by an already-stale global count. Guarded by `test_missing_slice_dominates_stale_global` and `test_missing_slice_dominates_even_after_a_stale_slice`.

packages/gptme-rag/src/gptme_rag/observability.py
```diff
@@ -0,0 +1,423 @@
...
+                **{f: hit[f] for f in hit_fields if f in hit and _copyable_hit_field(hit[f])},
+                "score": _as_score(hit.get(score_key)),
+            }
+            for hit in hits[:injected]
+        ],
+    }
+    if extra:
+        record.update(extra)
+
+    try:
```
TimeToBuildBob, line 195: ❌ **P1** — log_injection wraps the file write in a try/except that only catches OSError. If the `hits` sequence (or `extra`) contains any value that `json.dumps` cannot serialize — for example a datetime, a set, or a custom object — `json.dumps` raises TypeError, which is not caught and propagates out of the function. This breaks the documented contract that logging is best-effort and "a full disk must never break injection." A consumer with a non-serializable hit field will see an exception raised from an observability call, which can abort the injection flow itself. The fix is to also catch TypeError (and ideally broaden to Exception) and return None as the error path already does. _(3/3 passes)_

packages/gptme-rag/src/gptme_rag/observability.py
```diff
@@ -0,0 +1,423 @@
...
+        harness: Overrides :func:`detect_harness`.
+        score_key: Key holding each hit's score.
+        hit_fields: Hit keys copied into the record when present. ``None`` and
+            ``""`` are omitted; ``0`` / ``False`` are kept.
+        extra: Additional top-level fields merged into the record.
+
+    Returns the record written, or ``None`` if the write failed. Logging is
+    best-effort by design: a full disk must never break injection.
+    """
+    injected = len(hits) if max_injected is None else min(len(hits), max_injected)
```
TimeToBuildBob: ❌ **P1** — In `log_injection`, the `injected` count is computed as `min(len(hits), max_injected)` without validating that `max_injected` is non-negative. If a caller passes a negative value (e.g., `-1`), `injected` becomes negative, so `hits[:injected]` slices from the end of the list, recording hits that were not actually injected and omitting the ones that were. The `num_injected` field in the log record becomes a negative number, which is nonsensical and will corrupt downstream aggregation (e.g., `summarize_injections` will sum negative values). For 5 hits and `max_injected=-1`, the log records `num_injected: -1` and a `hits` list of the first 4 items (since `hits[:-1]`), inverting the intended behaviour. Fix by clamping to `max(0, min(len(hits), max_injected))` or rejecting negative values. _(2/3 passes — pass 3 looked and disagreed)_

packages/gptme-rag/src/gptme_rag/observability.py
TimeToBuildBob: Fixed in 6414bdf3 — serialize the record before opening the file, and catch TypeError/ValueError as well as OSError so an unserializable hit cannot abort injection. Guarded by `test_log_injection_never_raises_on_unserializable_hit`.

packages/gptme-rag/src/gptme_rag/observability.py
TimeToBuildBob: Fixed in 6414bdf3 — negative `max_injected` is clamped to 0 so `hits[:injected]` cannot slice from the tail. Guarded by `test_log_injection_clamps_negative_max_injected`.

packages/gptme-rag/src/gptme_rag/observability.py
```diff
@@ -0,0 +1,425 @@
...
+            return value
+    return "unknown"
+
+
+def _as_score(value: Any) -> float:
+    """Coerce a hit score to float; unusable values become 0.0 rather than raise."""
+    if value is None:
+        return 0.0
+    try:
+        return float(value)
```
TimeToBuildBob, line 126: ❌ **P1** — In `_as_score`, the exception handler catches only `TypeError` and `ValueError`. If a hit's score is a very large integer (e.g., `10**1000`), `float(value)` raises `OverflowError`, which is not caught. This propagates out of `_as_score` and into `log_injection` because the record is built before the try block that wraps the file write. As a result, a single hit with a huge score will raise `OverflowError` and crash the caller, breaking injection, which contradicts the function's documented best-effort guarantee. _(2/3 passes — pass 1 looked and disagreed)_

packages/gptme-rag/src/gptme_rag/observability.py
```diff
@@ -0,0 +1,425 @@
...
+            except (TypeError, ValueError):
+                # Same contract as a partial trailing line: skip, don't raise.
+                stats.malformed += 1
+                continue
+            stats.total += 1
+            sum_hits += num_hits
+            if num_hits > 0:
+                stats.nonzero += 1
+                sum_top += top_score
+            sessions.add(str(rec.get("session_id", "unknown")))
```
TimeToBuildBob: ⚠️ **P2** — In `summarize_injections`, the `session_id` and `harness` fields are stringified without handling `None`. If a log record contains `session_id: null` or `harness: null` (e.g., from a different writer or a malformed record), `str(rec.get(...))` yields the literal string `'None'`. This adds `'None'` to the `sessions` set and counts it as a distinct harness bucket, inflating `unique_sessions` and adding a spurious `'None'` entry to `by_harness`. That distorts the aggregate statistics and makes them less reliable for monitoring. The function should treat `None` as missing and fall back to `'unknown'`, using the same `or` pattern applied to `num_hits` and `top_score`. _(2/3 passes — pass 3 looked and disagreed)_

packages/gptme-rag/src/gptme_rag/observability.py
TimeToBuildBob: Not fixing this round. OverflowError on float(10**1000) is not a production path — retrieval scores are ordinary floats, and 2/3 consensus plus a mismatched suggested patch (str(query)) is the 5/5 treadmill, not a real P1. Stopping further reviewer loops on this head.

packages/gptme-rag/src/gptme_rag/observability.py
TimeToBuildBob: P2 accepted as a later polish. str(None) becoming the harness bucket 'None' is stats distortion, not a merge blocker. Will not spin another review cycle for it.

packages/gptme-rag/src/gptme_rag/observability.py
TimeToBuildBob: Fixed in c4b18d43 — added `OverflowError` to the caught exceptions in `_as_score`. The docstring promises best-effort (no raise); OverflowError on extreme integer scores should follow the same path as TypeError/ValueError. Guarded by `test_as_score_handles_overflow`.

packages/gptme-rag/src/gptme_rag/observability.py
TimeToBuildBob: Fixed in c4b18d43 — applied the `or "unknown"` pattern to both `session_id` and `harness` so a null value in the log record falls back to `"unknown"` rather than producing the literal `"None"` bucket. Guarded by `test_summarize_treats_null_session_and_harness_as_unknown`.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@TimeToBuildBob review

packages/gptme-rag/src/gptme_rag/observability.py
```diff
@@ -0,0 +1,425 @@
...
+        if abs(delta) > count_delta_tolerance:
+            health.status = "stale"
+            health.reasons.append(f"count delta {delta:+d} > tolerance {count_delta_tolerance}")
+
+    for name, (indexed, on_disk) in (slices or {}).items():
+        delta = on_disk - indexed
+        if indexed == 0:
+            slice_status = "missing"
+            health.reasons.append(f"slice {name!r} is empty")
+        elif abs(delta) > slice_delta_tolerance or age_hours > max_age_hours:
```
TimeToBuildBob, line 407: ⚠️ **P2** — In assess_index_health, when a slice is not missing but becomes stale solely because the global age_hours exceeds max_age_hours (i.e., abs(delta) <= slice_delta_tolerance), no reason is appended to health.reasons for that slice. The function's documented contract states 'reasons naming every failed check', so a consumer parsing reasons cannot tell which slices are stale. For example, an index with two slices each having abs(delta) within tolerance, but the index is 10 hours old, produces reasons = ['age 10.0h > 2.0h'] and both slices marked stale; the slice names are absent. This degrades the actionable output the module is meant to provide, though it does not affect the overall status computation. _(1/3 passes — passes 1, 3 looked and disagreed)_

packages/gptme-rag/src/gptme_rag/observability.py
```diff
@@ -0,0 +1,425 @@
...
+            ``""`` are omitted; ``0`` / ``False`` are kept.
+        extra: Additional top-level fields merged into the record.
+
+    Returns the record written, or ``None`` if the write failed. Logging is
+    best-effort by design: a full disk must never break injection.
+    """
+    injected = len(hits) if max_injected is None else max(0, min(len(hits), max_injected))
+    record: dict[str, Any] = {
+        "ts": datetime.now(timezone.utc).strftime("%Y-%m-%dT%H:%M:%SZ"),
+        "session_id": session_id or detect_session_id(),
```
TimeToBuildBob: ⚠️ **P2** — In log_injection, lines 167-168 use session_id or detect_session_id() and harness or detect_harness(). This treats an empty string as falsy and replaces it with the detected value, so a caller that explicitly passes session_id="" or harness="" (intending to suppress detection) gets an unexpected auto-detected value instead. The docstrings for these parameters state they "Override" the detection functions, meaning any provided value, including an empty string, should be used as-is. This can silently mislabel records with a harness or session from the environment when the caller tried to leave the field blank. _(1/3 passes — passes 1, 2 looked and disagreed)_

TimeToBuildBob:
Fixed the open AI review findings (commit `aaf7e981`):

1. **P1 OverflowError in `summarize_injections`**: Added `OverflowError` to the `except` clause at line 246 so a giant-integer `top_score` in a log record gets counted as malformed and skipped rather than crashing the aggregator.

2. **P1 datetime hit fields lost silently**: Coerce `datetime` values to ISO string in the hit dict comprehension before `json.dumps`. Previously a `date: datetime(...)` field would raise `TypeError`, the entire record would be discarded, and the return value silently became `None`.

3. **P2 Missing per-slice reason for age-only staleness**: When a slice is stale because the global index age exceeds `max_age_hours` but the per-slice delta is within tolerance, a per-slice reason like `"slice 'task' stale (age 5.0h > 2.0h)"` is now appended so consumers can identify which slices are affected.

4. **P2 Empty `session_id`/`harness` replaced by auto-detection**: Changed `session_id or detect_session_id()` to `session_id if session_id is not None else detect_session_id()` so an explicitly passed `""` is recorded as-is instead of being overridden.

All four fixes have regression tests (45 total, all green on 3.10/3.11/3.12).

packages/gptme-rag/src/gptme_rag/observability.py
```diff
@@ -0,0 +1,433 @@
...
+
+
+def detect_harness() -> str:
+    """Best-effort identification of the agent harness that invoked us.
+
+    Walks the process tree looking for a known harness process name. Returns
+    ``"unknown"`` when nothing matches or ``/proc`` is unavailable — this is
+    diagnostic metadata, never a control-flow input.
+    """
+    pid = os.getppid()
```
TimeToBuildBob: ❌ **P1** — detect_harness() calls os.getppid() unconditionally, which is not defined on Windows (it raises AttributeError). Since log_injection() invokes detect_harness() when harness=None is not passed, any consumer on Windows will crash with an uncaught AttributeError on every injection, preventing logging entirely. The docstring claims best-effort 'unknown' when /proc is unavailable, but the crash happens before /proc is touched. Wrap the getppid() call in try/except AttributeError and return 'unknown'. _(1/3 passes — passes 1, 3 looked and disagreed)_

packages/gptme-rag/src/gptme_rag/observability.py
```diff
@@ -0,0 +1,433 @@
...
+    than raising.
+    """
+    stats = InjectionStats()
+    if not log_file.exists():
+        return stats
+
+    sessions: set[str] = set()
+    sum_hits = 0
+    sum_top = 0.0
+    with open(log_file, encoding="utf-8") as f:
```
TimeToBuildBob: ❌ **P1** — summarize_injections() opens the file with encoding='utf-8' and iterates lines. If any line contains invalid UTF-8 bytes, Python raises UnicodeDecodeError, which is not caught. The docstring explicitly states 'undecodable lines are counted in malformed and skipped', but only json.JSONDecodeError is handled. A concurrent append or a corrupted byte sequence will abort the whole aggregation with an unhandled exception. Open with errors='replace' or catch UnicodeDecodeError per line and count it into malformed. _(1/3 passes — passes 1, 3 looked and disagreed)_

TimeToBuildBob:
Two more P1 fixes pushed in commit \`7cada044\`:

- **`detect_harness()`**: wrap `os.getppid()` in `try/except AttributeError` so it returns `"unknown"` on Windows instead of crashing (getppid is POSIX-only)
- **`summarize_injections()`**: open the log with `errors="replace"` so invalid UTF-8 bytes from concurrent appends count as malformed and are skipped, matching the docstring's stated contract

All 45 tests still pass.

packages/gptme-rag/src/gptme_rag/observability.py
```diff
@@ -0,0 +1,437 @@
...
+        "backend": backend,
+        "query": query,
+        "query_len": len(query),
+        "num_hits": len(hits),
+        "num_injected": injected,
+        "top_score": _as_score(hits[0].get(score_key)) if hits else 0.0,
+        "hits": [
+            {
+                **{
+                    f: hit[f].isoformat() if isinstance(hit[f], datetime) else hit[f]
```
TimeToBuildBob: ❌ **P1** — In log_injection, the per-hit field copy only converts datetime.datetime instances to ISO strings, but the common datetime.date type is not handled. Since the default hit_fields includes "date", a hit carrying a date object (not a datetime) is passed through to json.dumps, which raises TypeError. The exception is caught and log_injection returns None, so the entire injection record is silently dropped. This defeats the purpose of the logging feature: an injection with a date field yields no record, and summarize_injections will not see it. The fix is to also handle datetime.date (which also has .isoformat()) by checking isinstance(hit[f], (datetime, date)) or importing date. _(2/3 passes — pass 2 looked and disagreed)_

packages/gptme-rag/src/gptme_rag/observability.py
```diff
@@ -0,0 +1,437 @@
...
+    record: dict[str, Any] = {
+        "ts": datetime.now(timezone.utc).strftime("%Y-%m-%dT%H:%M:%SZ"),
+        "session_id": session_id if session_id is not None else detect_session_id(),
+        "harness": harness if harness is not None else detect_harness(),
+        "backend": backend,
+        "query": query,
+        "query_len": len(query),
+        "num_hits": len(hits),
+        "num_injected": injected,
+        "top_score": _as_score(hits[0].get(score_key)) if hits else 0.0,
```
TimeToBuildBob: ⚠️ **P2** — When max_injected is 0 (including a negative value clamped to 0), the record's top_score is still taken from the first hit in the full hits list, even though no hit is actually injected. This makes the log claim a high-scoring document was injected when the consumer injected nothing, which can mislead downstream analysis of retrieval quality. The docstring says only the first max_injected hits are recorded in detail, but top_score is not conditioned on that; it should be 0.0 when nothing is injected. A concrete example: max_injected=0 with hits=[{'similarity': 0.9}] yields record with num_injected=0, hits=[], but top_score=0.9. The test test_log_injection_clamps_negative_max_injected does not check top_score, so the bug is uncaught. _(1/3 passes — passes 2, 3 looked and disagreed)_

packages/gptme-rag/src/gptme_rag/observability.py
```diff
@@ -0,0 +1,437 @@
...
+                # Same contract as a partial trailing line: skip, don't raise.
+                stats.malformed += 1
+                continue
+            stats.total += 1
+            sum_hits += num_hits
+            if num_hits > 0:
+                stats.nonzero += 1
+                sum_top += top_score
+            sessions.add(str(rec.get("session_id") or "unknown"))
+            harness = str(rec.get("harness") or "unknown")
```
TimeToBuildBob: ⚠️ **P2** — Empty-string session_id and harness values written by log_injection are treated as 'unknown' by summarize_injections. log_injection writes session_id='' and harness='' verbatim when explicitly passed (test_log_injection_empty_string_session_and_harness_kept), but summarize_injections uses `rec.get(...) or 'unknown'`, so an empty string is falsy and becomes 'unknown'. This loses the distinction between 'no value set' and 'explicitly empty', which may matter for consumers who set an empty harness to mean 'no harness assigned'. The aggregator should preserve empty strings if log_injection writes them, or log_injection should normalize to 'unknown' before writing. _(1/3 passes — passes 1, 3 looked and disagreed)_

packages/gptme-rag/src/gptme_rag/observability.py
```diff
@@ -0,0 +1,437 @@
...
+
+#: Environment variables consulted, in order, to identify the calling session.
+DEFAULT_SESSION_ENV_VARS: tuple[str, ...] = (
+    "GPTME_RAG_SESSION_ID",
+    "GPTME_SESSION_ID",
+    "CC_SESSION_ID",
+)
+
+#: How far up the process tree :func:`detect_harness` will walk.
+_MAX_PROC_WALK = 8
```
TimeToBuildBob: ⚠️ **P2** — The harness detection walk is bounded by a hardcoded _MAX_PROC_WALK of 8. If the real process tree is deeper than 8 levels and the harness process is beyond that depth, detect_harness returns 'unknown' even though a harness is present. This can lead to misleading harness attribution in injection logs, which may frustrate the very debugging this feature is meant to support. While 8 is usually sufficient, in deeply nested shell environments (e.g., a script launched from another script through several levels of su/sudo/nohup) the parent chain can exceed 8. The function should either allow a configurable depth or walk until an OSError/non-numeric PPid is found rather than stopping at an arbitrary count. _(1/3 passes — passes 1, 3 looked and disagreed)_

packages/gptme-rag/src/gptme_rag/observability.py
```diff
@@ -0,0 +1,437 @@
...
+    than raising.
+    """
+    stats = InjectionStats()
+    if not log_file.exists():
+        return stats
+
+    sessions: set[str] = set()
+    sum_hits = 0
+    sum_top = 0.0
+    with open(log_file, encoding="utf-8", errors="replace") as f:
```
TimeToBuildBob, line 239: ⚠️ **P2** — summarize_injections does not guard against log_file being a directory. It checks if not log_file.exists() and then unconditionally calls open(log_file, ...). If the path exists and is a directory, open raises IsADirectoryError (a subclass of OSError), which propagates out of the function. The function is documented as best-effort for malformed lines but crashes on a directory path. The fix is to catch OSError around the open (and possibly around the whole read) and treat it as an empty log, or at least return an empty InjectionStats. This is a robustness issue for callers that pass an invalid path. _(1/3 passes — passes 1, 2 looked and disagreed)_

ErikBjare:
@TimeToBuildBob fix

TimeToBuildBob:
Pushed commit \`b4e0280b\` addressing all 5 findings from the \`7cada044\` AI review:

- **P1 e76650f6** — Added \`date\` to the \`datetime\` import and \`isinstance(hit[f], datetime | date)\` so \`datetime.date\` hit fields are coerced to ISO string, not dropped.
- **P2 8f01218d** — Conditioned \`top_score\` on \`injected > 0\` so it stays \`0.0\` when \`max_injected=0\`.
- **P2 61384d1e** — Changed \`str(rec.get("session_id") or "unknown")\` to \`str(_sid) if _sid is not None else "unknown"\` so empty-string values are preserved.
- **P2 b8f0bc8b** — Increased \`_MAX_PROC_WALK\` from 8 to 32 for deeper process trees.
- **P2 c48c16fa** — Added \`log_file.is_dir()\` guard in \`summarize_injections\` before \`open()\`.

All 49 tests pass.

@TimeToBuildBob review

packages/gptme-rag/src/gptme_rag/observability.py
```diff
@@ -0,0 +1,439 @@
...
+        pid = os.getppid()
+    except AttributeError:
+        # os.getppid is not available on Windows
+        return "unknown"
+    for _ in range(_MAX_PROC_WALK):
+        if pid <= 1:
+            break
+        try:
+            comm = (Path("/proc") / str(pid) / "comm").read_text().strip()
+        except OSError:
```
TimeToBuildBob, line 90: ⚠️ **P2** — The `detect_harness` function aborts the entire process-tree walk as soon as reading `/proc/<pid>/comm` raises an OSError. In practice, the comm file can be unreadable (e.g., due to permissions or the process having exited) while the status file is still readable. Because the function breaks on the OSError, it never attempts to read the status file to obtain the parent PID, so it returns `"unknown"` even though a harness ancestor might be discoverable further up the tree. This weakens harness detection in transient-failure scenarios, causing `log_injection` to record `"unknown"` in the harness field for what could be a known harness. The break is intentional for hard failures, but a failed comm read should not necessarily halt the walk if the status file can still be read. _(1/3 passes — passes 1, 3 looked and disagreed)_

packages/gptme-rag/src/gptme_rag/observability.py
```diff
@@ -0,0 +1,439 @@
...
+    if built_at is None:
+        return IndexHealth(status="missing", reasons=["no index"])
+
+    if built_at.tzinfo is None:
+        built_at = built_at.replace(tzinfo=timezone.utc)
+    age_hours = max(0.0, (now - built_at).total_seconds() / 3600.0)
+
+    health = IndexHealth(
+        status="ok",
+        age_hours=round(age_hours, 2),
```
TimeToBuildBob: ⚠️ **P2** — `assess_index_health` computes the age with full precision and uses that unrounded value for staleness checks (e.g., `age_hours > max_age_hours`), but then stores a two-decimal-rounded value in `IndexHealth.age_hours`. When the raw age is just above the threshold (for instance 2.004 hours with `max_age_hours=2.0`), the status is correctly set to `'stale'`, yet the reported `age_hours` is 2.0, appearing below the limit. Consumers that read the reported age to understand why the index was flagged will see an inconsistent figure that does not reflect the actual decision threshold. _(1/3 passes — passes 1, 2 looked and disagreed)_

TimeToBuildBob:
Pushed commit `300dc1e2` addressing the latest review findings:

- **P2 detect_harness resilience**: Changed the `OSError` on `comm` read from a loop-breaking `break` to a `pass`, so the walk continues up the process tree even when a `comm` file is transiently unreadable. The `status` file (which provides PPID) is still read normally; only if *that* fails does the walk stop.
- **P2 age_hours rounding**: Removed `round(..., 2)` when storing in `IndexHealth.age_hours`. The comparison at line 391 uses the full-precision value, so reporting a rounded value created a mismatch where the reported field didn't reflect the actual decision (e.g. `age_hours=2.0` reported as non-stale despite the check flagging `2.004h > 2.0h`).
- **Defensive `int()` cast in `log_injection`**: `max_injected` is typed `int | None`, but added `int(max_injected)` before the slice to guard against a caller passing a float — `hits[:float]` raises `TypeError` at runtime.

Re: the P1 finding about `isinstance(hit[f], datetime | date)`: the package requires Python >=3.10 (confirmed in `pyproject.toml`), so PEP 604 union syntax in `isinstance()` is valid here. CI runs on 3.10, 3.11, 3.12 and all pass. This finding doesn't apply.

All 49 tests pass.

@TimeToBuildBob review

PR #1501, opened 2026-08-25
feat(sessions): record eval attempt classification on the session record

TimeToBuildBob:
## Problem

Session runners already classify every attempt as either a measurable sample or an infra failure (rate limit, auth failure, dead backend, zero-token exit) — `autonomous-run.sh` computes `IS_INFRA_FAILURE` across seven guards and uses it to skip bandit and grade updates. That verdict is then **thrown away**.

Because it is never persisted, every read-side consumer re-derives infra-ness from duration/token heuristics. In Bob's workspace that predicate is defined once and called from **7** separate sites — 7 chances to forget, and forgetting reports contaminated quality numbers *silently*.

## What this adds (write side only)

- `SessionRecord.attempt_kind`: `repetition` | `infra_retry` | `unknown`, defaulting to `None` so every existing record parses unchanged. `None` means **unrecorded**, not `repetition` — consumers must keep their own fallback.
- `SessionStore.stamp_attempt_kind(session_id, kind)`: a second-pass stamp held under the store lock. It has to be a second pass: the runner's guards need the final exit code, token count and `failure_reason`, all of which `post_session()` itself produces, so the record already exists by the time the verdict is known. Returns `False` (rather than raising) when no record matches, since `post_session()` can fail. Rejects an unknown kind loudly — a typo would otherwise persist as an unread value forever.
- `gptme-sessions stamp-attempt-kind <session-id> <kind>` so shell runners can call it without an inline `python -c`.

## Explicitly out of scope

No read-side consumer is migrated here. Migration is gated on first **measuring agreement** between the recorded field and the existing heuristic on real records — the disagreement count is the finding, not a test failure. Doing both in one PR would ship an unverified swap.

## Tests

`packages/gptme-sessions/tests/test_attempt_kind.py` — 12 cases: infra session reads back `infra_retry`, normal session reads back `repetition`, value persists to JSONL, neighbours untouched, missing session returns `False`, unknown/empty kinds raise and leave the record unmodified.

```
12 passed          # tests/test_attempt_kind.py
529 passed         # test_record + test_store_locking + test_sessions + test_cli
```

Origin: idea #1144, from apache/maka peer research — Maka's eval boundary separates *repetition* from *infra retry* at the attempt, not at read time.

### DIFF
--- a/packages/gptme-sessions/src/gptme_sessions/cli.py
+++ b/packages/gptme-sessions/src/gptme_sessions/cli.py
@@ -3181,6 +3181,8 @@ def stamp_attempt_kind(ctx: click.Context, session_id: str, attempt_kind: str) -
     \b
         gptme-sessions stamp-attempt-kind a1b2 infra_retry
     """
+    if not session_id:
+        raise click.UsageError("SESSION_ID must not be empty.")
     store = SessionStore(sessions_dir=ctx.obj["sessions_dir"])
     if store.stamp_attempt_kind(session_id, attempt_kind):
         click.echo(f"stamped attempt_kind={attempt_kind} on session {session_id}")
--- a/packages/gptme-sessions/src/gptme_sessions/cli.py
+++ b/packages/gptme-sessions/src/gptme_sessions/cli.py
@@ -34,7 +34,7 @@ from .replay import (
     resolve_replay_target,
     resolve_session_record_prefix,
 )
-from .record import SessionRecord, normalize_run_type
+from .record import ATTEMPT_KINDS, SessionRecord, normalize_run_type
 from .signals import extract_from_path
 from .store import (
     SessionStore,
@@ -3165,6 +3165,27 @@ def blame(
         raise click.ClickException(str(exc)) from exc
 
     click.echo(render_json(result) if as_json else render_text(result))
+
+
+@cli.command("stamp-attempt-kind")
+@click.argument("session_id")
+@click.argument("attempt_kind", type=click.Choice(sorted(ATTEMPT_KINDS)))
+@click.pass_context
+def stamp_attempt_kind(ctx: click.Context, session_id: str, attempt_kind: str) -> None:
+    """Record the eval attempt classification on an existing session record.
+
+    Called by the session runner after its infra-failure guards have run —
+    those guards need the final exit code, tokens and failure_reason, so the
+    verdict is only known after post_session() appended the record.
+
+    \b
+        gptme-sessions stamp-attempt-kind a1b2 infra_retry
+    """
+    store = SessionStore(sessions_dir=ctx.obj["sessions_dir"])
+    if store.stamp_attempt_kind(session_id, attempt_kind):
+        click.echo(f"stamped attempt_kind={attempt_kind} on session {session_id}")
+    else:
+        raise click.ClickException(f"no session record found for session_id={session_id!r}")
 
 
 def main() -> int:
--- a/packages/gptme-sessions/src/gptme_sessions/cli.py
+++ b/packages/gptme-sessions/src/gptme_sessions/cli.py
@@ -3184,14 +3184,29 @@ def stamp_attempt_kind(ctx: click.Context, session_id: str, attempt_kind: str) -
     if not session_id:
         raise click.UsageError("SESSION_ID must not be empty.")
     store = SessionStore(sessions_dir=ctx.obj["sessions_dir"])
-    try:
-        record = resolve_session_record_prefix(store.load_all(), session_id)
-    except ValueError as exc:
-        raise click.ClickException(str(exc))
-    if store.stamp_attempt_kind(record.session_id, attempt_kind):
-        click.echo(f"stamped attempt_kind={attempt_kind} on session {record.session_id}")
+    records = store.load_all()
+    # Deduplicate by unique session_id before checking for ambiguity — the
+    # store may hold duplicate rows for the same session_id (e.g. if
+    # post_session() ran twice), and stamp_attempt_kind is designed to stamp
+    # all of them; resolve_session_record_prefix would falsely raise on those.
+    matches = [r for r in records if r.session_id.startswith(session_id)]
+    unique_ids = list(dict.fromkeys(r.session_id for r in matches))
+    if not unique_ids:
+        raise click.ClickException(
+            f"No session found matching {session_id!r}. "
+            "Run 'gptme-sessions query' to list available session IDs."
+        )
+    if len(unique_ids) > 1:
+        raise click.ClickException(
+            f"Ambiguous prefix {session_id!r} matches {len(unique_ids)} sessions: "
+            + ", ".join(unique_ids)
+            + ". Run 'gptme-sessions query' to list available session IDs."
+        )
+    full_id = unique_ids[0]
+    if store.stamp_attempt_kind(full_id, attempt_kind):
+        click.echo(f"stamped attempt_kind={attempt_kind} on session {full_id}")
     else:
-        raise click.ClickException(f"no session record found for session_id={session_id!r}")
+        raise click.ClickException(f"no session record found for session_id={full_id!r}")
 
 
 def main() -> int:
--- a/packages/gptme-sessions/src/gptme_sessions/cli.py
+++ b/packages/gptme-sessions/src/gptme_sessions/cli.py
@@ -3184,8 +3184,12 @@ def stamp_attempt_kind(ctx: click.Context, session_id: str, attempt_kind: str) -
     if not session_id:
         raise click.UsageError("SESSION_ID must not be empty.")
     store = SessionStore(sessions_dir=ctx.obj["sessions_dir"])
-    if store.stamp_attempt_kind(session_id, attempt_kind):
-        click.echo(f"stamped attempt_kind={attempt_kind} on session {session_id}")
+    try:
+        record = resolve_session_record_prefix(store.load_all(), session_id)
+    except ValueError as exc:
+        raise click.ClickException(str(exc))
+    if store.stamp_attempt_kind(record.session_id, attempt_kind):
+        click.echo(f"stamped attempt_kind={attempt_kind} on session {record.session_id}")
     else:
         raise click.ClickException(f"no session record found for session_id={session_id!r}")
 
--- a/packages/gptme-sessions/src/gptme_sessions/record.py
+++ b/packages/gptme-sessions/src/gptme_sessions/record.py
@@ -350,6 +350,10 @@ class SessionRecord:
         # Discard unrecognized dropout_depth values so typos don't silently reach consumers.
         if self.dropout_depth is not None and self.dropout_depth not in DROPOUT_DEPTH_VALUES:
             self.dropout_depth = None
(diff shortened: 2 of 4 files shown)


packages/gptme-sessions/src/gptme_sessions/cli.py
```diff
@@ -3167,6 +3167,27 @@ def blame(
...
+
+    Called by the session runner after its infra-failure guards have run —
+    those guards need the final exit code, tokens and failure_reason, so the
+    verdict is only known after post_session() appended the record.
+
+    \b
+        gptme-sessions stamp-attempt-kind a1b2 infra_retry
+    """
+    store = SessionStore(sessions_dir=ctx.obj["sessions_dir"])
+    if store.stamp_attempt_kind(session_id, attempt_kind):
```
TimeToBuildBob: ⚠️ **P2** — The `stamp-attempt-kind` CLI command passes the `session_id` argument directly to `store.stamp_attempt_kind` without checking for emptiness. The store method raises `ValueError` if `session_id` is empty, but the CLI does not catch it. If a user runs `gptme-sessions stamp-attempt-kind "" repetition`, the exception propagates uncaught, producing a traceback instead of a friendly error. This breaks the CLI's error-handling contract (which uses `ClickException` for expected errors) and could expose internal details. The fix is to validate `session_id` in the CLI (e.g., `if not session_id: raise click.UsageError("Session ID must not be empty.")`) or catch `ValueError` and convert to `ClickException`.

packages/gptme-sessions/src/gptme_sessions/store.py
```diff
@@ -254,6 +254,44 @@ def rewrite(self, records: list[SessionRecord]) -> Path:
...
+        if attempt_kind not in ATTEMPT_KINDS:
+            raise ValueError(
+                f"unknown attempt_kind {attempt_kind!r}; expected one of {sorted(ATTEMPT_KINDS)}"
+            )
+        if not session_id:
+            raise ValueError("session_id is required to stamp attempt_kind")
+
+        with self.lock():
+            records = self.load_all()
+            target = next((r for r in reversed(records) if r.session_id == session_id), None)
```
TimeToBuildBob: ⚠️ **P2** — In `SessionStore.stamp_attempt_kind`, the method finds the last record with the given `session_id` using `next((r for r in reversed(records) if r.session_id == session_id), None)`. Only that single record's `attempt_kind` is set. If the store contains duplicate rows for the same `session_id` (which is possible because `append` does not enforce uniqueness), the other rows remain with `attempt_kind=None`. A read-side consumer that groups by `session_id` would then see one row marked and another unmarked, leading to a mix of `infra_retry` and `None` for the same session, which could cause misinterpretation of the data. The method should either stamp all matching records or document that only the latest is stamped. The comment about rewriting the full list indicates awareness of duplicates but the implementation only touches one.

TimeToBuildBob:
## 🤖 AI code review

Adds an `attempt_kind` field (repetition/infra_retry/unknown) to SessionRecord, a `stamp_attempt_kind` method to SessionStore that rewrites the store under lock, and a new CLI command `stamp-attempt-kind` that resolves a session ID prefix and stamps the classification. Also adds validation to discard invalid attempt_kind values on load, and a test suite covering the write side.

**⚠️ P2 medium** — `packages/gptme-sessions/src/gptme_sessions/cli.py:3187`

The CLI command `stamp-attempt-kind` calls `store.load_all()` without acquiring the store lock, then later calls `store.stamp_attempt_kind(full_id, attempt_kind)`, which acquires the lock and re-loads the store. If a concurrent process appends a record whose session_id matches the requested prefix between the initial load and the stamp, the command's prefix resolution is based on a stale snapshot: it may see a single match and resolve to a full_id, while the stamp method's fresh load could see two matches (or the new record not yet present) — the resolved full_id might not be the session the user intended, or the stamp might not be applied to the new record. This is a TOCTOU race. Other read-modify-write commands like `annotate` hold the lock across the entire cycle; this one does not. The consequence is that under concurrent writes, the command can stamp a session with a classification based on out-of-date information, or fail to stamp a newly appended session that also matches the prefix. Suggest holding `store.lock()` around the load→resolve→stamp sequence.

```suggestion
with store.lock():
    records = store.load_all()
    matches = [r for r in records if r.session_id.startswith(session_id)]
    unique_ids = list(dict.fromkeys(r.session_id for r in matches))
    ...
    full_id = unique_ids[0]
    if store.stamp_attempt_kind(full_id, attempt_kind):
        ...
```

<sub>**How this was verified:** Examined the CLI command body and the store's stamp method. load_all() on line 3187 is outside any lock; store.stamp_attempt_kind acquires the lock and reloads inside the lock. A concurrent writer can land between these two reads.</sub>

</details>

- `packages/gptme-sessions/src/gptme_sessions/cli.py` — Adds the `stamp-attempt-kind` click command that resolves a session prefix, deduplicates by session_id, and calls store.stamp_attempt_kind.
- `packages/gptme-sessions/src/gptme_sessions/record.py` — Adds ATTEMPT_KINDS frozenset, the attempt_kind field (default None), and validation in __post_init__ to discard unrecognized values.
- `packages/gptme-sessions/src/gptme_sessions/store.py` — Adds stamp_attempt_kind which loads all records, sets attempt_kind on matching session_ids, and rewrites the store under the lock.
- `packages/gptme-sessions/tests/test_attempt_kind.py` — Adds 12 tests covering the attempt_kind field, validation, store stamping, duplicate rows, and CLI prefix resolution.

</details>

| `7b1630e8faef` | 4/5 | 2 | llm | 2026-08-25 03:14 UTC |
| `d5040554aedd` | 5/5 | 0 | llm | 2026-08-25 05:21 UTC |
| `d5040554aedd` | 2/5 | 2 | llm | 2026-08-25 05:54 UTC |
| `f74cba3cdbc8` | 4/5 | 1 | llm | 2026-08-25 06:56 UTC |

</details>

<sub>Reviewed `21679a542fd1` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 230s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

TimeToBuildBob:
@greptileai review

packages/gptme-sessions/src/gptme_sessions/cli.py
TimeToBuildBob: Fixed in d5040554. Added `if not session_id: raise click.UsageError("SESSION_ID must not be empty.")` before the store call, so an empty argument produces a UsageError instead of a ValueError traceback.

packages/gptme-sessions/src/gptme_sessions/store.py
TimeToBuildBob: Fixed in d5040554. Changed from `next((r for r in reversed(records) if ...))` to `[r for r in records if ...]` and stamping all matching records, so duplicate rows for the same session_id all get the attempt_kind instead of just the most recent one.

packages/gptme-sessions/src/gptme_sessions/cli.py
```diff
@@ -3167,6 +3167,29 @@ def blame(
...
+@click.pass_context
+def stamp_attempt_kind(ctx: click.Context, session_id: str, attempt_kind: str) -> None:
+    """Record the eval attempt classification on an existing session record.
+
+    Called by the session runner after its infra-failure guards have run —
+    those guards need the final exit code, tokens and failure_reason, so the
+    verdict is only known after post_session() appended the record.
+
+    \b
+        gptme-sessions stamp-attempt-kind a1b2 infra_retry
```
TimeToBuildBob, line 3182: ❌ **P1** — The new stamp-attempt-kind command passes the raw session_id argument directly to SessionStore.stamp_attempt_kind, which matches on exact equality (r.session_id == session_id). The command's own docstring example 'gptme-sessions stamp-attempt-kind a1b2 infra_retry' uses a short prefix, and the existing show and annotate commands resolve ID prefixes via resolve_session_record_prefix. With a prefix, stamp_attempt_kind finds no matching record and the CLI raises 'no session record found', so the documented invocation fails. Fix by resolving the prefix before stamping. _(2/3 passes — pass 1 looked and disagreed)_

packages/gptme-sessions/src/gptme_sessions/record.py
```diff
@@ -206,6 +219,12 @@ class SessionRecord:
     exit_code: int | None = None  # process exit code (124 = timeout)
     failure_reason: str | None = None  # coarse harness failure class (see failure_capture.py)
     error: str | None = None  # stderr tail or trajectory error snippet when harness failed
+    # Eval attempt classification, written once by the session runner which
+    # already computes it (see ``ATTEMPT_KINDS``).  ``None`` on every record
+    # written before this field existed, so consumers must treat ``None`` as
+    # "unrecorded" and fall back to their own heuristic rather than assuming
+    # ``repetition``.
+    attempt_kind: str | None = None
```
TimeToBuildBob, line 227: ❌ **P1** — The new attempt_kind field is added to SessionRecord without the same validation that harm_category and dropout_depth receive in __post_init__ (lines 348-352). Those fields discard unrecognized values so consumers never see unexpected keys. attempt_kind is loaded silently from from_dict, so a hand-edited JSONL or a previous buggy writer could persist an invalid value like 'infra-retry'; consumers checking attempt_kind == 'infra_retry' would miss it and treat the session as a repetition, contaminating quality analytics. The stamp method rejects unknown kinds, but the field can already be invalid at load time. Add validation similar to the others. _(2/3 passes — pass 1 looked and disagreed)_

packages/gptme-sessions/src/gptme_sessions/cli.py
TimeToBuildBob: Fixed in f74cba3c. The CLI now calls `resolve_session_record_prefix(store.load_all(), session_id)` before stamping, identical to the pattern used by `show` and `annotate`. Prefix resolution raises `ValueError` on no-match or ambiguous prefix, which the CLI converts to `ClickException`. The resolved full ID is then passed to `stamp_attempt_kind`.

packages/gptme-sessions/src/gptme_sessions/record.py
TimeToBuildBob: Fixed in f74cba3c. Added validation in `__post_init__` matching the `harm_category` and `dropout_depth` pattern: `if self.attempt_kind is not None and self.attempt_kind not in ATTEMPT_KINDS: self.attempt_kind = None`. A hand-edited JSONL with an invalid value like `'infra-retry'` is now discarded at load time rather than silently passing to consumers. Added two new tests: one confirming invalid kinds are discarded, one confirming all valid kinds survive `from_dict`.

packages/gptme-sessions/src/gptme_sessions/cli.py
TimeToBuildBob: Fixed in 21679a5: stamp-attempt-kind CLI now deduplicates matching session rows directly via load_all() without calling resolve_session_record_prefix, so duplicate rows are stamped correctly instead of raising a ValueError on ambiguity. The TOCTOU concern from the latest pass (P2) is the remaining open design decision about whether to hold the lock across the load→resolve→stamp cycle.

PR #1504, opened 2026-08-25
feat(pm-bandit): wire RunItem into bandit shadow mode

TimeToBuildBob:
Add BOB_PM_BANDIT_SHADOW=1 support to run_item.py for Stage 1 observation-only mode.

### DIFF
--- a/packages/gptme-runloops/src/gptme_runloops/run_item.py
+++ b/packages/gptme-runloops/src/gptme_runloops/run_item.py
@@ -2905,6 +2905,9 @@ def run_work_file(
                 if item_outcome.exit_code != 0 and overall_exit == 0:
                     overall_exit = item_outcome.exit_code
 
+                item_effect = run_post_session(plan, item, item_outcome, config, hooks)
+                item_effects.append(item_effect)
+
                 # Stage 1 shadow: record outcome for later analysis
                 if bandit_shadow and bandit is not None:
                     try:
@@ -2913,25 +2916,17 @@ def run_work_file(
                         work_type = classify_item_work_type(
                             list(item.types), repo=item.repo
                         )
-                        # Map outcome to reward: 1.0 for productive, 0.0 otherwise
-                        reward = (
-                            1.0
-                            if (
-                                item_outcome.exit_code == 0
-                                and not item_outcome.counted_failure
-                            )
-                            else 0.0
-                        )
+                        # Use the observable-effect signal as the reward: a
+                        # clean exit is not evidence work landed (gptme/gptme#3468,
+                        # 2026-08-10). effect=observed means commits/replies were
+                        # actually delivered; anything else maps to 0.0.
+                        reward = 1.0 if item_effect == EFFECT_OBSERVED else 0.0
                         bandit.record_outcome(work_type, model, reward)
                         _log(
                             f"BOB_PM_BANDIT_SHADOW: recorded outcome {work_type} -> {model} = {reward}"
                         )
                     except Exception as exc:
                         _log(f"WARN: bandit outcome recording failed: {exc}")
-
-                item_effects.append(
-                    run_post_session(plan, item, item_outcome, config, hooks)
-                )
             finally:
                 # bash EXIT-trap parity: the claim is abandoned on every exit
                 # path, including SIGTERM (the CLI converts it to SystemExit
--- a/packages/gptme-runloops/src/gptme_runloops/run_item.py
+++ b/packages/gptme-runloops/src/gptme_runloops/run_item.py
@@ -2849,7 +2849,9 @@ def run_work_file(
                 try:
                     from gptme_runloops.pm_dispatch import classify_item_work_type
 
-                    work_type = classify_item_work_type(item.types, repo=item.repo)
+                    work_type = classify_item_work_type(
+                        list(item.types), repo=item.repo
+                    )
                     available = [m for m in [model] if m]
                     if available:
                         bandit_model = bandit.resolve_model(work_type, available)
@@ -2908,7 +2910,9 @@ def run_work_file(
                     try:
                         from gptme_runloops.pm_dispatch import classify_item_work_type
 
-                        work_type = classify_item_work_type(item.types, repo=item.repo)
+                        work_type = classify_item_work_type(
+                            list(item.types), repo=item.repo
+                        )
                         # Map outcome to reward: 1.0 for productive, 0.0 otherwise
                         reward = (
                             1.0
--- a/packages/gptme-runloops/src/gptme_runloops/run_item.py
+++ b/packages/gptme-runloops/src/gptme_runloops/run_item.py
@@ -94,6 +94,7 @@ from gptme_runloops.merge_lifecycle import (
     WorkItem,
     run_merge_lifecycle,
 )
+from gptme_runloops.pm_bandit import PmModelBandit
 from gptme_runloops.pm_dispatch import (
     DISPATCH_COOLDOWN_DIR_ENV,
     EFFECT_NONE,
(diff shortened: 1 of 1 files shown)


packages/gptme-runloops/src/gptme_runloops/run_item.py
```diff
@@ -2867,6 +2902,29 @@ def run_work_file(
...
+                        # Map outcome to reward: 1.0 for productive, 0.0 otherwise
+                        reward = (
+                            1.0
+                            if (
+                                item_outcome.exit_code == 0
+                                and not item_outcome.counted_failure
+                            )
+                            else 0.0
+                        )
+                        bandit.record_outcome(work_type, model, reward)
```
TimeToBuildBob, line 2924: ⚠️ **P2** — The shadow-mode outcome recording calls bandit.record_outcome(work_type, model, reward) unconditionally, even when model is empty, whereas the shadow-mode model-choice logging at line 2853-2854 guards on `if available:` and skips when model is empty (available becomes []). When run_work_file is invoked with model="" (the default), the bandit records outcomes keyed by an empty-string model while never logging any bandit prediction for any model. If the bandit's internal state treats model names as keys, this creates a spurious empty-model entry that later analysis or routing would consume, and since resolve_model was never called for that empty model, there is no baseline expected reward for it. The consequence is that shadow observations are polluted with a meaningless model key, making any comparison between the bandit's predicted model and the actual model vacuous for runs that do not specify a model, defeating the purpose of the shadow-mode data collection.

TimeToBuildBob:
## 🤖 AI code review

The PR adds optional shadow-mode support (BOB_PM_BANDIT_SHADOW=1) to run_work_file in run_item.py. It loads a PmModelBandit instance, logs the bandit's model choice per item, records the observable-effect signal as a reward after each session, and does not yet use the bandit for routing.

<details><summary><b>2 findings</b> · ⚠️ **2** P2</summary>

**⚠️ P2 medium** — `packages/gptme-runloops/src/gptme_runloops/run_item.py:2923`

The reward mapping in the shadow recording treats any effect other than EFFECT_OBSERVED as a failure (reward=0.0), including EFFECT_UNKNOWN. EFFECT_UNKNOWN is returned when the system cannot determine whether the work landed (e.g., no delivery check ran and no PR-state diff is available, as happens for agent_msg_reply or notification items). These items may have succeeded but left no verifiable evidence; mapping them to 0.0 biases the bandit to avoid those work types even when they succeed. This corrupts the shadow training data, so that when the bandit is later used for real routing it will under-rank successful actions. The fix is to only record outcomes when the effect is known (observed or none), or to give unknown a separate neutral reward, and to exclude them from the reward computation.

```suggestion
reward = 1.0 if item_effect == EFFECT_OBSERVED else (0.0 if item_effect == EFFECT_NONE else None)
```

<sub>**How this was verified:** Checked the run_post_session return values and the effect signal constants; EFFECT_UNKNOWN is a distinct value returned for items with no verification evidence. Many item types (agent_msg_reply, notification) do not run a delivery check and have no PR-state diff, so they produce EFFECT_UNKNOWN.</sub>

<sub>Consensus: 1/2 passes agreed — pass 2 looked and disagreed</sub>

**⚠️ P2 medium** — `packages/gptme-runloops/src/gptme_runloops/run_item.py:2924`

The bandit shadow mode writes observations to a shared directory (config.state_dir / 'pm-dispatch') without any file locking or per-process isolation. When multiple run_work_file instances run concurrently (which is the normal case with per-slot locks, since each slot has its own lock but they do not contend), their PmModelBandit.record_outcome calls can interleave writes to the same JSONL ledger, corrupting the observation data. This can cause incorrect model selection when the bandit is later used in routing, undermining the purpose of the shadow experiment. The PR should either use a per-run subdirectory or apply an exclusive lock around the ledger write.

<sub>**How this was verified:** Checked the bandit creation (line 2758) and record_outcome call (line 2924); state_dir is derived from config.state_dir which is shared across runs, and no locking or per-run isolation exists.</sub>

<sub>Consensus: 1/2 passes agreed — pass 1 looked and disagreed</sub>

</details>

- `packages/gptme-runloops/src/gptme_runloops/run_item.py` — Adds PmModelBandit import and shadow-mode blocks that load the bandit, log its per-item model choice, and record post-session outcomes as rewards.

</details>

| `6b7680039977` | 4/5 | 1 | llm | 2026-08-25 04:48 UTC |
| `e734896000df` | 3/5 | 1 | llm | 2026-08-25 05:11 UTC |

</details>

<sub>Reviewed `f7519c3a07b3` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 101s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

packages/gptme-runloops/src/gptme_runloops/run_item.py
```diff
@@ -2867,6 +2904,31 @@ def run_work_file(
...
+                        from gptme_runloops.pm_dispatch import classify_item_work_type
+
+                        work_type = classify_item_work_type(
+                            list(item.types), repo=item.repo
+                        )
+                        # Map outcome to reward: 1.0 for productive, 0.0 otherwise
+                        reward = (
+                            1.0
+                            if (
+                                item_outcome.exit_code == 0
```
TimeToBuildBob: ❌ **P1** — The reward for the bandit is computed as 1.0 only when `item_outcome.exit_code == 0` and not `counted_failure`, which does not reflect whether the session actually moved the target (pushed a commit, posted a reply, etc.). As the file itself notes at lines 2470-2473, a worker can exit 0 while every push fails and the log claims success. This mislabels no-op sessions as productive, and because the bandit learns from these rewards, it will over-select models that exit cleanly without delivering. In shadow mode this corrupts the training data; when the bandit later drives routing, it will favor models that are good at exiting 0 rather than at accomplishing work. A concrete example: a worker that failed to push but exited 0 records reward 1.0, biasing the bandit toward that model for that work type.

TimeToBuildBob:
@greptileai review

packages/gptme-runloops/src/gptme_runloops/run_item.py
TimeToBuildBob: Fixed in f7519c3: moved `record_outcome()` after `run_post_session()` and replaced the `exit_code == 0` heuristic with the `EFFECT_OBSERVED` signal. A clean exit is not evidence work landed — this now uses the same observable-effect signal the rest of the pipeline uses.

packages/gptme-runloops/src/gptme_runloops/run_item.py
```diff
@@ -2867,9 +2904,29 @@ def run_work_file(
...
+                        from gptme_runloops.pm_dispatch import classify_item_work_type
+
+                        work_type = classify_item_work_type(
+                            list(item.types), repo=item.repo
+                        )
+                        # Use the observable-effect signal as the reward: a
+                        # clean exit is not evidence work landed (gptme/gptme#3468,
+                        # 2026-08-10). effect=observed means commits/replies were
+                        # actually delivered; anything else maps to 0.0.
+                        reward = 1.0 if item_effect == EFFECT_OBSERVED else 0.0
```
TimeToBuildBob, line 2923: ⚠️ **P2** — The reward mapping in the shadow recording treats any effect other than EFFECT_OBSERVED as a failure (reward=0.0), including EFFECT_UNKNOWN. EFFECT_UNKNOWN is returned when the system cannot determine whether the work landed (e.g., no delivery check ran and no PR-state diff is available, as happens for agent_msg_reply or notification items). These items may have succeeded but left no verifiable evidence; mapping them to 0.0 biases the bandit to avoid those work types even when they succeed. This corrupts the shadow training data, so that when the bandit is later used for real routing it will under-rank successful actions. The fix is to only record outcomes when the effect is known (observed or none), or to give unknown a separate neutral reward, and to exclude them from the reward computation. _(1/2 passes — pass 2 looked and disagreed)_

packages/gptme-runloops/src/gptme_runloops/run_item.py
```diff
@@ -2867,9 +2904,29 @@ def run_work_file(
...
+                        bandit.record_outcome(work_type, model, reward)
```
TimeToBuildBob, line 2924: ⚠️ **P2** — The bandit shadow mode writes observations to a shared directory (config.state_dir / 'pm-dispatch') without any file locking or per-process isolation. When multiple run_work_file instances run concurrently (which is the normal case with per-slot locks, since each slot has its own lock but they do not contend), their PmModelBandit.record_outcome calls can interleave writes to the same JSONL ledger, corrupting the observation data. This can cause incorrect model selection when the bandit is later used in routing, undermining the purpose of the shadow experiment. The PR should either use a per-run subdirectory or apply an exclusive lock around the ledger write. _(1/2 passes — pass 1 looked and disagreed)_

PR #85, opened 2025-12-26
feat(plugins): add cc-analyze plugin for Claude Code subagent analysis

TimeToBuildBob:
## Summary

Add new plugin for spawning focused Claude Code analysis tasks from gptme.

## Features

- `analyze()` function to run Claude Code with a prompt synchronously or in background
- `check_session()` to monitor background tasks
- `kill_session()` to terminate background tasks
- Supports timeout configuration
- Background mode via tmux for long-running tasks

## Use Cases

- Security audits
- Code reviews
- Test coverage analysis
- Architecture documentation

## Tests

8 tests passing - covering all core functionality.

## Related

Addresses ErikBjare/bob#209 - Erik's directive to create cc_analyze as a gptme-contrib plugin

----

> [!IMPORTANT]
> Adds `cc-analyze` plugin for Claude Code analysis with task management functions and tests.
> 
>   - **New Plugin**:
>     - Adds `cc-analyze` plugin for Claude Code analysis in `plugins/cc-analyze`.
>     - Implements `analyze()`, `check_session()`, and `kill_session()` in `cc_analyze.py` for task management.
>     - Supports synchronous and background task execution with timeout configuration.
>   - **Tests**:
>     - Adds 8 tests in `test_cc_analyze.py` covering core functionalities like CLI availability, synchronous and background execution, and session management.
>   - **Configuration**:
>     - Updates `.pre-commit-config.yaml` to use Python 3.11 and adds `tomli` dependency.
>     - Modifies `Makefile` to disable `unused-ignore` error in mypy.
>     - Adds `gptme-cc-analyze` to `uv.lock` with dependencies and test extras.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=gptme%2Fgptme-contrib&utm_source=github&utm_medium=referral)<sup> for ff7d4dfa5487837d76bb588102b301a99ddc668c. You can [customize](https://app.ellipsis.dev/gptme/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

PR #1505, opened 2026-08-25
feat(precommit): add config-driven root-structure validator

TimeToBuildBob:
Upstreams the shared root-structure validator that #1445 Phase 2 depends on.

## Why

`check_root_structure.py` takes its allowlist as repeated `--allow` args in each
consuming repo. That is fine for one repo and bad for four: gptme (#3580),
gptme-cloud (#867), and gptme-agent-template (#85) have each merged the guard in
hardcoded-args form, so the allowlists now drift independently — the same
accretion #1445 was opened about.

This adds `validate_root_structure_yaml.py`: identical logic, allowlist read from
a per-repo `root-structure-allowlist.yaml`. Consuming repos then own a **data
file**, not a copy of the checker.

The immediate trigger: gptme/gptme-agent-template#88 vendored a local copy of
this validator because it was not available here. Once this merges, that PR can
drop the copy and reference the shared hook instead.

## What

- `scripts/precommit/validators/validate_root_structure_yaml.py` — the validator.
  Reads root entries from the **git index** (`git ls-files`) rather than walking
  the filesystem, so gitignored scratch in a dirty worktree cannot affect the
  result and no `git check-ignore` subprocess-per-entry is needed.
- Hook id `validate-root-structure` in `.pre-commit-hooks.yaml`, so consumers wire
  one line. Requires PyYAML in the hook environment.
- `mypy_path` gains `scripts/precommit/validators` so validator tests can import
  their subject the same way `tests/test_check_root_structure.py` already does
  for `scripts/precommit`.
- 11 tests.

## Two bugs fixed relative to the draft this was extracted from

Both were found by writing the tests, and both are verified to fail against the
pre-fix version:

1. **A valid but empty `allowed_entries: []` exited 1 with no explanation.**
   `load_allowlist` returned an empty frozenset for both "load failed" and
   "loaded, empty", and the caller `if not allowed:` treated them alike — so it
   returned 1 *before* printing which entries were unexpected, and skipped the
   "create the config" hint too because the file did exist. Silent failure.
2. **A scalar `allowed_entries: README.md` was iterated as a string**, making the
   allowlist a set of individual characters (`{R, E, A, D, M, ., m, d}`) instead
   of raising a config error.

`load_allowlist` now returns `None` on load failure and a frozenset on success
(possibly empty), and rejects a non-list `allowed_entries`.

## Verification

```
$ uv run --active python3 -m pytest tests/test_validate_root_structure_yaml.py tests/test_check_root_structure.py -q
26 passed
```

`prek run` clean on the changed files (ruff, ruff-format, mypy, jscpd, and the
rest of the hook set).

Regression evidence for the two bug fixes — the new tests run against the
unfixed validator:

```
FAILED test_empty_allowlist_is_not_reported_as_missing
FAILED test_non_list_allowed_entries_fails
2 failed, 9 passed
```

## Follow-up (not in this PR)

gptme-agent-template, gptme, and gptme-cloud each need a
`root-structure-allowlist.yaml` and a hook swap. Tracked in
`tasks/root-structure-validator-propagation.md`; gptme-agent-template#88 is the
first consumer.

ref: #1445

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -10,3 +10,16 @@
   pass_filenames: false
   always_run: true
   additional_dependencies: []
+
+- id: validate-root-structure
+  name: Check root directory structure (config-driven)
+  description: >
+    Config-driven variant of ``check-root-structure``. Reads the allowlist from
+    a per-repo YAML file (default ``root-structure-allowlist.yaml``) instead of
+    repeated ``--allow`` args, so consuming repos own a data file rather than a
+    copy of the checker. Requires PyYAML in the hook environment.
+  entry: scripts/precommit/validators/validate_root_structure_yaml.py
+  language: script
+  pass_filenames: false
+  always_run: true
+  additional_dependencies: []
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -17,9 +17,9 @@
     Config-driven variant of ``check-root-structure``. Reads the allowlist from
     a per-repo YAML file (default ``root-structure-allowlist.yaml``) instead of
     repeated ``--allow`` args, so consuming repos own a data file rather than a
-    copy of the checker. Requires PyYAML in the hook environment.
+    copy of the checker.
   entry: scripts/precommit/validators/validate_root_structure_yaml.py
-  language: script
+  language: python
   pass_filenames: false
   always_run: true
-  additional_dependencies: []
+  additional_dependencies: [PyYAML]
--- a/scripts/precommit/validators/validate_root_structure_yaml.py
+++ b/scripts/precommit/validators/validate_root_structure_yaml.py
@@ -97,7 +97,7 @@ def load_allowlist(config_path: Path) -> "frozenset[str] | None":
         return None
 
     try:
-        with open(config_path) as f:
+        with open(config_path, encoding="utf-8", errors="surrogateescape") as f:
             data = yaml.safe_load(f)
 
         if not isinstance(data, dict) or "allowed_entries" not in data:
--- a/scripts/precommit/validators/validate_root_structure_yaml.py
+++ b/scripts/precommit/validators/validate_root_structure_yaml.py
@@ -1,0 +1,197 @@
+#!/usr/bin/env python3
+"""Enforce allowed set of top-level files and directories via YAML config file.
+
+This validator reads an allowlist from a YAML config file (default:
+`root-structure-allowlist.yaml` in the repo root) and rejects any tracked
+top-level entries not on that list. The allowlist approach scales across
+multiple repos without code duplication.
+
+Each repo using this validator owns a `root-structure-allowlist.yaml` file
+listing its permitted top-level entries, one per line.
+
+Usage (from repo root):
+    python3 gptme-contrib/scripts/precommit/validators/validate_root_structure_yaml.py
+
+Or with a custom config path:
+    python3 gptme-contrib/scripts/precommit/validators/validate_root_structure_yaml.py \\
+      --config my-root-allowlist.yaml
+
+As a pre-commit hook in .pre-commit-config.yaml:
+    - repo: https://github.com/gptme/gptme-contrib
+      rev: <SHA or tag>
+      hooks:
+      - id: validate-root-structure
+        name: Check root directory structure
+        entry: python3 scripts/precommit/validators/validate_root_structure_yaml.py
+        language: system
+        pass_filenames: false
+        always_run: true
+        args:
+          - --config=root-structure-allowlist.yaml
+
+Config file format (YAML, one entry per line):
+    ---
+    allowed_entries:
+      - .github
+      - .gitignore
+      - README.md
+      - src
+      - tests
+"""
+
+import argparse
+import subprocess
+import sys
+from pathlib import Path
+
+import yaml
+
+
+def get_repo_root() -> Path:
+    """Return the repository root via git rev-parse."""
+    result = subprocess.run(
+        ["git", "rev-parse", "--show-toplevel"],
+        capture_output=True,
+        check=True,
+    )
+    return Path(result.stdout.rstrip(b"\r\n").decode("utf-8", errors="surrogateescape"))
+
+
+def get_tracked_root_entries(repo_root: Path) -> set[str]:
+    """Return first path components of all files in the git index."""
+    result = subprocess.run(
+        ["git", "ls-files", "--cached", "-z"],
+        cwd=repo_root,
+        capture_output=True,
+        check=True,
+    )
+    entries: set[str] = set()
+    for path in result.stdout.split(b"\0"):
+        if path:
+            entries.add(
+                path.split(b"/", maxsplit=1)[0].decode(
+                    "utf-8", errors="surrogateescape"
+                )
+            )
+    return entries
+
+
+def load_allowlist(config_path: Path) -> "frozenset[str] | None":
+    """Load allowed entries from YAML config file.
+
+    Expects a file with structure:
+        ---
+        allowed_entries:
+          - entry1
+          - entry2
+
+    Returns the allowed entries, or None if the config could not be loaded.
+    An empty-but-valid ``allowed_entries`` list returns an empty frozenset,
+    which is distinct from a load failure.
+    """
+    if not config_path.exists():
+        print(
+            f"validate-root-structure: config file not found: {config_path}",
+            file=sys.stderr,
+        )
+        return None
+
+    try:
+        with open(config_path) as f:
+            data = yaml.safe_load(f)
+
+        if not isinstance(data, dict) or "allowed_entries" not in data:
+            print(
+                f"validate-root-structure: config file missing 'allowed_entries' key: {config_path}",
+                file=sys.stderr,
+            )
+            return None
+
+        entries = data["allowed_entries"]
+        if entries is None:
+            entries = []
+        if not isinstance(entries, list):
+            print(
+                f"validate-root-structure: 'allowed_entries' must be a list: {config_path}",
+                file=sys.stderr,
+            )
+            return None
+
+        return frozenset(str(entry) for entry in entries)
+    except yaml.YAMLError as e:
+        print(
+            f"validate-root-structure: failed to parse YAML config: {e}",
+            file=sys.stderr,
+        )
+        return None
+    except OSError as e:
+        print(
+            f"validate-root-structure: error reading config: {e}",
+            file=sys.stderr,
+        )
+        return None
+
+
+def main(argv: "list[str] | None" = None) -> int:
+    parser = argparse.ArgumentParser(
+        description=__doc__,
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+    )
+    parser.add_argument(
+        "--config",
+        type=str,
+        default="root-structure-allowlist.yaml",
+        help="Path to YAML allowlist file (default: root-structure-allowlist.yaml in repo root)",
+    )
+    args = parser.parse_args(argv)
+
+    try:
+        repo_root = get_repo_root()
+    except subprocess.CalledProcessError as error:
+        print(
+            f"validate-root-structure: unable to find git repo root: {error}",
+            file=sys.stderr,
+        )
+        return 1
+
+    # Resolve config path relative to repo root if not absolute
+    config_path = Path(args.config)
+    if not config_path.is_absolute():
+        config_path = repo_root / config_path
+
+    allowed = load_allowlist(config_path)
+    if allowed is None:
+        # load_allowlist already explained what went wrong.
+        if not config_path.exists():
+            print(
+                f"If this is intentional, create {config_path} with allowed entries.",
+                file=sys.stderr,
+            )
+        return 1
+
+    try:
+        entries = get_tracked_root_entries(repo_root)
+    except subprocess.CalledProcessError as error:
+        print(
+            f"validate-root-structure: git ls-files failed: {error}",
+            file=sys.stderr,
+        )
+        return 1
+
+    unexpected = entries - allowed
+    if not unexpected:
+        return 0
+
+    print("validate-root-structure: unexpected top-level entries found:")
+    for name in sorted(unexpected):
+        print(f"  {name}")
+    print()
+    print(
+        f"If this is intentional, add the entry to {config_path} and explain\n"
+        "the choice in your commit message."
+    )
+    return 1
+
+
+if __name__ == "__main__":
+    sys.exit(main())
--- a/scripts/precommit/validators/validate_root_structure_yaml.py
+++ b/scripts/precommit/validators/validate_root_structure_yaml.py
@@ -142,7 +142,7 @@ def main(argv: "list[str] | None" = None) -> int:
 
     try:
         repo_root = get_repo_root()
-    except subprocess.CalledProcessError as error:
+    except (subprocess.CalledProcessError, OSError) as error:
         print(
             f"validate-root-structure: unable to find git repo root: {error}",
             file=sys.stderr,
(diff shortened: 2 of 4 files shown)


TimeToBuildBob:
@greptileai review

.pre-commit-hooks.yaml
```diff
@@ -10,3 +10,16 @@
...
+  description: >
+    Config-driven variant of ``check-root-structure``. Reads the allowlist from
+    a per-repo YAML file (default ``root-structure-allowlist.yaml``) instead of
+    repeated ``--allow`` args, so consuming repos own a data file rather than a
+    copy of the checker. Requires PyYAML in the hook environment.
+  entry: scripts/precommit/validators/validate_root_structure_yaml.py
+  language: script
+  pass_filenames: false
+  always_run: true
+  additional_dependencies: []
```
TimeToBuildBob: ❌ **P1** — The new hook `validate-root-structure` is declared with `language: script` and `additional_dependencies: []`, but the validator script (`validate_root_structure_yaml.py`) unconditionally `import yaml` at module level. With `language: script`, pre-commit does not create a virtualenv or install any packages; it executes the script with the system interpreter. Unless PyYAML is already present in the hook's runtime environment, the hook fails with `ModuleNotFoundError: No module named 'yaml'` on every run, making the hook unusable in any repo lacking PyYAML. The description even states 'Requires PyYAML in the hook environment', but nothing enforces it. The fix is to add `PyYAML` to `additional_dependencies` for this hook (pre-commit will then install it into a virtualenv), or switch to `language: python` and specify the entry as a module with an appropriate `additional_dependencies`. This affects all consumers who adopt the hook.

scripts/precommit/validators/validate_root_structure_yaml.py
```diff
@@ -0,0 +1,197 @@
...
+    """
+    if not config_path.exists():
+        print(
+            f"validate-root-structure: config file not found: {config_path}",
+            file=sys.stderr,
+        )
+        return None
+
+    try:
+        with open(config_path) as f:
```
TimeToBuildBob: ⚠️ **P2** — In `load_allowlist`, the config file is opened with the default `open(config_path)`, which uses the locale encoding (typically UTF-8). If the config file contains a byte sequence that is not valid UTF-8, `open` raises `UnicodeDecodeError`. The function only catches `yaml.YAMLError` and `OSError`; `UnicodeDecodeError` is a subclass of `ValueError` and is not caught, so the hook crashes with an unhandled traceback instead of reporting a clean error and exiting 1. This is especially relevant because the script elsewhere goes to lengths to handle non-UTF-8 paths via surrogateescape, but the config content is not protected. The fix is to open with `errors='surrogateescape'` or `errors='replace'`, or to catch `UnicodeDecodeError` and print a proper message. This will cause the hook to fail on malformed config files that are not strictly UTF-8.

TimeToBuildBob:
## 🤖 AI code review

Adds a new pre-commit hook 'validate-root-structure' that reads an allowlist of permitted top-level entries from a YAML file in the repo root, checks them against the tracked files in the git index, and exits non-zero if any unlisted entry is found. The hook, its validator script, tests, and a mypy path update are included.

- `.pre-commit-hooks.yaml` — Adds the 'validate-root-structure' hook definition with PyYAML dependency.
- `mypy.ini` — Adds 'scripts/precommit/validators' to mypy_path for test imports.
- `scripts/precommit/validators/validate_root_structure_yaml.py` — New validator script that loads an allowlist from YAML and compares with git-index root entries.
- `tests/test_validate_root_structure_yaml.py` — Adds 11 tests for the validator covering config handling, normalization, and git integration.

</details>

| `20de67826b99` | 3/5 | 2 | llm | 2026-08-25 08:28 UTC |
| `40e371835f64` | 3/5 | 1 | llm | 2026-08-25 09:11 UTC |
| `7252ff897a80` | 4/5 | 2 | llm | 2026-08-25 20:02 UTC |
| `f37643c9d8b0` | 3/5 | 2 | llm | 2026-08-26 04:55 UTC |

</details>

<sub>Reviewed `b7a0661a55b4` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 187s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

TimeToBuildBob:
Fixed both findings in 40e3718:

**P1** (`language: script` → `language: python` + `additional_dependencies: [PyYAML]`): `language: script` ignores `additional_dependencies` entirely — pre-commit only installs deps for environment-managed languages (python, node, etc.). Switched to `language: python` so pre-commit creates a virtualenv and installs PyYAML there. Also removed the now-redundant "Requires PyYAML in the hook environment" from the description.

**P2** (`open(config_path)` → `open(config_path, encoding="utf-8", errors="surrogateescape")`): Added explicit UTF-8 + surrogateescape so non-UTF-8 bytes surface via YAML error rather than an unhandled `UnicodeDecodeError` traceback. This matches the script's existing surrogateescape handling for file paths.

tests/test_validate_root_structure_yaml.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+def test_relative_config_resolves_against_repo_root(tmp_path):
+    """A relative --config path is resolved from the repo root, not the cwd."""
+    write_config(tmp_path, "allowed_entries:\n  - README.md\n")
+    with (
+        patch.object(validator, "get_repo_root", return_value=tmp_path),
+        patch.object(validator, "get_tracked_root_entries", return_value={"README.md"}),
+    ):
+        assert validator.main(["--config", "root-structure-allowlist.yaml"]) == 0
```
TimeToBuildBob, line 130: ❌ **P1** — test_relative_config_resolves_against_repo_root patches get_repo_root to return tmp_path but does not patch get_tracked_root_entries. When main() reaches the get_tracked_root_entries call, it executes git ls-files --cached -z in tmp_path, which is not a git repository. This raises subprocess.CalledProcessError, which main() catches and returns 1. The test asserts the return value is 0, so it fails in any environment where tmp_path is not a git repo. The test should also mock get_tracked_root_entries to return a set that matches the allowlist, e.g. {'README.md'}, so it only tests the relative-path resolution logic without depending on git.

TimeToBuildBob:
## ✅ All AI review findings addressed + tests passing

**Greptile findings status:**
- **P1** (language:script ignores deps): Fixed in commit 40e37183 — switched to language:python with PyYAML dependency
- **P2** (UnicodeDecodeError not caught): Fixed in commit 40e37183 — explicit UTF-8 encoding with surrogateescape
- **P1** (test doesn't mock get_tracked_root_entries): Fixed in branch HEAD — test now mocks both functions

**Test verification:** All 11 tests passing including the previously failing test_relative_config_resolves_against_repo_root ✓

The validator is now production-ready:
- Hook will reliably install PyYAML via pre-commit virtualenv
- Config file encoding errors are handled gracefully
- Test suite correctly isolates from git fs operations

TimeToBuildBob:
Fixed two real issues found by the local AI reviewer in this session:

**Commit f60dc438** — Simplified the consumer usage example in the docstring.
The example previously showed `language`, `entry`, `pass_filenames`, and `args` inline — settings that come from `.pre-commit-hooks.yaml` and must not be overridden in the consumer config. Reduced the example to just `id: validate-root-structure`.

**Commit 7252ff89** — Documented and tested the bootstrap requirement that the config file itself must appear in the allowlist. Since `root-structure-allowlist.yaml` lives at the repo root and is tracked by git, it shows up in `entries`; unless listed in `allowed_entries`, the hook immediately flags its own config as unexpected. Updated the example config in the docstring to include the filename, and added `test_config_file_itself_must_be_in_allowlist` to guard this invariant.

All 12 tests pass. Local reviewer scored 4/5 (P2 only) before the second fix; next sweep should clear to 5/5.

TimeToBuildBob:
CI is fully green on the latest commits (7252ff89). Self-merge not eligible: `.pre-commit-hooks.yaml` and `mypy.ini` are outside the allowed self-merge categories and need a human review to land.

All review findings addressed, 12 tests pass. Ready for Erik to merge when he gets a chance.

scripts/precommit/validators/validate_root_structure_yaml.py
```diff
@@ -0,0 +1,192 @@
...
+        "--config",
+        type=str,
+        default="root-structure-allowlist.yaml",
+        help="Path to YAML allowlist file (default: root-structure-allowlist.yaml in repo root)",
+    )
+    args = parser.parse_args(argv)
+
+    try:
+        repo_root = get_repo_root()
+    except subprocess.CalledProcessError as error:
```
TimeToBuildBob: ⚠️ **P2** — `main()` only wraps `get_repo_root()` and `get_tracked_root_entries()` in `try/except subprocess.CalledProcessError`. Both functions call `subprocess.run(..., check=True)` on the `git` binary. If `git` is not installed or not on `PATH`, `subprocess.run` raises `FileNotFoundError` (a subclass of `OSError`), which is not caught by the `except` clause. The exception propagates out of `main()`, and since `__main__` calls `sys.exit(main())`, the process crashes with a Python traceback instead of printing the intended clean error message and returning exit code 1. The PR description claims the validator 'fails cleanly when git cannot resolve a repository root', but that only holds when git exists and returns a non-zero status (e.g., not a repo). A missing `git` binary is a realistic edge case for a standalone script, and the hook should catch `OSError` (or specifically `FileNotFoundError`) and print a user-facing error, then exit 1. This can be fixed by catching `(subprocess.CalledProcessError, OSError)` for both calls. _(1/2 passes — pass 2 looked and disagreed)_

scripts/precommit/validators/validate_root_structure_yaml.py
```diff
@@ -0,0 +1,192 @@
...
+import sys
+from pathlib import Path
+
+import yaml
+
+
+def get_repo_root() -> Path:
+    """Return the repository root via git rev-parse."""
+    result = subprocess.run(
+        ["git", "rev-parse", "--show-toplevel"],
```
TimeToBuildBob, line 48: ⚠️ **P2** — The validator calls subprocess.run with check=True in get_repo_root and get_tracked_root_entries, but main() only catches subprocess.CalledProcessError. If git is not installed or not on PATH, subprocess.run raises FileNotFoundError (a subclass of OSError), which is not caught, causing the script to terminate with an unhandled traceback instead of a clean error message. This is a robustness defect for environments where the hook is run outside of a normal git setup, but it does not affect the intended pre-commit usage since git is always present there. _(1/2 passes — pass 1 looked and disagreed)_

scripts/precommit/validators/validate_root_structure_yaml.py
TimeToBuildBob: Fixed in f37643c9: both `get_repo_root()` and `get_tracked_root_entries()` callers now catch `(subprocess.CalledProcessError, OSError)` so a missing git binary produces a clean error message instead of an unhandled traceback.

scripts/precommit/validators/validate_root_structure_yaml.py
TimeToBuildBob: Fixed in f37643c9: both except clauses now catch `(subprocess.CalledProcessError, OSError)`. The second pass noted this doesn't affect normal pre-commit usage (git is always present), but the fix is trivial and makes the validator clean in standalone contexts too.

TimeToBuildBob:
Automated merge handoff for \`f37643c9d8\`:

- CI gate: checks in progress (discover passed)
- Review threads: all resolved
  - \`fe134266dc32\` and \`1f7806dfdeb8\`: both identify the same issue (OSError not caught when git missing) — fixed in this commit by changing both \`except subprocess.CalledProcessError\` to \`except (subprocess.CalledProcessError, OSError)\`
- Self-merge policy: check-only (precommit validator, not a core plugin path)

This head is ready for maintainer review and merge when CI completes.

scripts/precommit/validators/validate_root_structure_yaml.py
```diff
@@ -0,0 +1,192 @@
...
+        if entries is None:
+            entries = []
+        if not isinstance(entries, list):
+            print(
+                f"validate-root-structure: 'allowed_entries' must be a list: {config_path}",
+                file=sys.stderr,
+            )
+            return None
+
+        return frozenset(str(entry) for entry in entries)
```
TimeToBuildBob: ❌ **P1** — The allowlist entries are used as-is, but `get_tracked_root_entries` normalizes each path to its first component without any leading or trailing slashes. If a user lists an entry with a trailing slash (e.g., `src/`), the allowlist contains `'src/'`, while the actual tracked entry is `'src'`. The set difference `entries - allowed` then flags `'src'` as unexpected, causing the hook to fail even though the directory is correctly allowlisted. This is likely because users naturally write directory names with a slash. The same issue occurs with leading `./` on allowlist entries. The fix is to strip leading `./` and trailing `/` from each allowlist entry before building the frozenset. _(1/2 passes — pass 1 looked and disagreed)_

TimeToBuildBob: ⚠️ **P2** — In `load_allowlist`, each element of `allowed_entries` is stringified with `str(entry)` without checking that it is a string. If a config contains a non-scalar entry (e.g. `allowed_entries: [README.md, [foo]]`), the list is converted to the string `"['foo']"`, which can never match a real root entry. The validator then reports the root entry `foo` as unexpected, despite the config author intending to allow it, and gives no indication that the config is malformed. This produces false failures for an invalid config that should be rejected with a clear error. The fix is to validate that every element of `allowed_entries` is a string and return a configuration error otherwise. _(1/2 passes — pass 2 looked and disagreed)_

TimeToBuildBob:
Fixed the two active AI reviewer findings in commit b7a0661a:

**P1** (`be4fc21dc40c`) — Allowlist entries with trailing `/` or leading `./` now normalize to match the first-component output of `get_tracked_root_entries`. `src/` and `./src` both match `src`. Two new tests cover these cases.

**P2** (`50a20ef7901b`) — Non-string entries in `allowed_entries` (e.g. a nested list `[foo]`) now exit 1 with a clear error message instead of being silently `str()`-ified into `"['foo']"`. New test covers this case.

All 15 tests pass.

scripts/precommit/validators/validate_root_structure_yaml.py
TimeToBuildBob: Fixed in b7a0661a: allowlist entries are now normalized (strip leading `./` and trailing `/`) in `load_allowlist` before building the frozenset. Two new tests cover `src/` and `./src` → `src` cases.

TimeToBuildBob: Fixed in b7a0661a: `load_allowlist` now validates that every element of `allowed_entries` is a string; a non-string entry (e.g. a nested list) exits 1 with a clear error message. New test covers this case.

TimeToBuildBob:
## Merge recommendation: all findings addressed, ready to land

**AI review findings (local multi-pass reviewer, 2 rounds):**

| Finding | Severity | Status |
|---|---|---|
| `be4fc21dc40c` — allowlist trailing `/` / `./` normalization | P1 blocking | Fixed in b7a0661a |
| `50a20ef7901b` — non-string `allowed_entries` not rejected | P2 warning | Fixed in b7a0661a |
| `fe134266dc32` — missing git `OSError` in `main()` | P2 warning | Fixed in f37643c9 |
| `1f7806dfdeb8` — same (pass 2 agreed 1/2) | P2 warning | Fixed in f37643c9 |
| `1760a79dafb0` — `test_relative_config_resolves_against_repo_root` missing mock | P1 blocking | Fixed (prior session) |
| `9f6e89e002e6` — `UnicodeDecodeError` not caught | P2 warning | Addressed |
| `298afd6be564` — `language: script` + PyYAML | P1 blocking | Fixed in 40e3718 |

**Blocking bugs remaining: 0. Security issues: 0.**

**CI:** all green on b7a0661a (coverage, typecheck, pre-commit, integration, install-size).

**Greptile:** triggered once; awaiting response (anti-spam guard prevents further pings — escalate to a human if greptile review is required before merge).

**Tests:** 15 tests pass.

Self-merge not eligible (`.pre-commit-hooks.yaml` and `mypy.ini` changes require human review). Ready for Erik to merge when convenient.

PR #1509, opened 2026-08-25
feat(gptodo): add --skip-claim-held to ready command

TimeToBuildBob:
## Summary

Adds `--skip-claim-held` flag to `gptodo ready`. When set, it queries the coordination DB and filters out tasks whose `cascade:task:<name>` key is currently claimed by another concurrent session.

This is the follow-up to #1487, which Erik pointed out should live in `gptodo ready` directly rather than a standalone wrapper script. The coordination filtering is the only additive piece over what `gptodo ready` already does (dependency + state filtering).

## Behaviour

- Requires `gptme-coordination` (graceful degradation: prints a warning and continues if unavailable)
- Only skips tasks held via `cascade:task:<name>` claims (the standard autonomous-session claim namespace)
- Human-readable output shows a count of skipped tasks; JSON/JSONL output is unaffected

## Usage

```bash
gptodo ready --skip-claim-held
gptodo ready --skip-claim-held --jsonl | head -5
```

## Test

All 522 existing gptodo tests pass. The flag is opt-in and has no effect when coordination is unavailable.

## Related

- Follow-up to #1487

### DIFF
--- a/packages/gptodo/src/gptodo/cli.py
+++ b/packages/gptodo/src/gptodo/cli.py
@@ -3394,7 +3394,17 @@ def tags(state: str | None, show_tasks: bool, filter_tags: tuple[str, ...]):
     default=None,
     help="Exclude tasks in this pool (e.g. '--exclude-pool frontier')",
 )
-def ready(state, output_json, output_jsonl, use_cache, pool_filter, exclude_pool):
+@click.option(
+    "--skip-claim-held",
+    "skip_claim_held",
+    is_flag=True,
+    default=False,
+    help=(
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
Fixed the typecheck failure (commit `4d1654d3`): `WorkClaimManager` expects a `CoordinationDB` instance, not a `Path`. Added `CoordinationDB` to the import and wrapped `db_path` before passing it in. Local typecheck now passes — waiting for CI to confirm.

TimeToBuildBob:
Closing in favor of #1510 which uses `--skip-claimed` (Erik's preferred naming).

PR #1510, opened 2026-08-25
feat(gptodo): add --skip-claimed flag to gptodo ready

TimeToBuildBob:
## Summary

Upstreams the cascade-claim filtering from the now-closed `ready-tasks.py` wrapper script (#1487).

- Adds `--skip-claimed` to `gptodo ready` that filters out tasks already held by another coordination session
- Reads `state/coordination/coord.db` directly via stdlib `sqlite3` — no dependency on the `coordination` package
- Gracefully degrades when the DB is absent (silent no-op, returns all tasks)
- Excludes own-agent claims via `CASCADE_COORDINATION_AGENT` / `AGENT_ID` env vars

## Why

When #1487 was closed, Erik asked: "Shouldn't this be `gptodo ready`?" The filtering logic was the only value-add beyond what `gptodo ready` already had. This PR adds it where it belongs.

## Test plan

- `TestGetClaimedTaskIds`: unit tests for the helper (absent DB, prefix stripping, expiry, own-claim exclusion)
- `TestReadySkipClaimed`: integration tests via CliRunner (hides claimed, shows without flag, degrades gracefully)

```
7 passed in 0.75s
```

### DIFF
--- a/packages/gptodo/src/gptodo/cli.py
+++ b/packages/gptodo/src/gptodo/cli.py
@@ -3383,7 +3383,7 @@ def _get_claimed_task_ids(repo_root: Path) -> set[str]:
             rows = conn.execute(
                 """SELECT task_id FROM work
                 WHERE task_id LIKE 'cascade:task:%'
-                  AND status IN ('claimed', 'completed')
+                  AND status = 'claimed'
                   AND expires_at >= datetime('now')
                   AND (? = '' OR claimer != ?)""",
                 (my_agent, my_agent),
--- a/packages/gptodo/src/gptodo/cli.py
+++ b/packages/gptodo/src/gptodo/cli.py
@@ -3384,8 +3384,8 @@ def _get_claimed_task_ids(repo_root: Path) -> set[str]:
                 """SELECT task_id FROM work
                 WHERE task_id LIKE 'cascade:task:%'
                   AND status = 'claimed'
-                  AND expires_at >= datetime('now')
-                  AND (? = '' OR claimer != ?)""",
+                  AND (expires_at IS NULL OR expires_at >= datetime('now'))
+                  AND (? = '' OR claimer IS NULL OR claimer != ?)""",
                 (my_agent, my_agent),
             ).fetchall()
     except Exception:
--- a/packages/gptodo/src/gptodo/cli.py
+++ b/packages/gptodo/src/gptodo/cli.py
@@ -3350,6 +3350,39 @@ def tags(state: str | None, show_tasks: bool, filter_tags: tuple[str, ...]):
     )
 
 
+def _get_claimed_task_ids(repo_root: Path) -> set[str]:
+    """Return task IDs actively claimed by another agent in the coordination DB.
+
+    Queries ``state/coordination/coord.db`` using stdlib sqlite3 — no coordination
+    package dependency.  Returns an empty set if the DB is absent or unreadable.
+    """
+    import sqlite3
+
+    db_path = repo_root / "state" / "coordination" / "coord.db"
+    if not db_path.exists():
+        return set()
+
+    prefix = "cascade:task:"
+    my_agent = (
+        os.environ.get("CASCADE_COORDINATION_AGENT") or os.environ.get("AGENT_ID") or ""
+    ).strip()
+
+    try:
+        with sqlite3.connect(str(db_path)) as conn:
+            rows = conn.execute(
+                """SELECT task_id FROM work
+                WHERE task_id LIKE 'cascade:task:%'
+                  AND status IN ('claimed', 'completed')
+                  AND expires_at >= datetime('now')
+                  AND (? = '' OR claimer != ?)""",
+                (my_agent, my_agent),
+            ).fetchall()
+    except Exception:
+        return set()
+
+    return {r[0][len(prefix) :] if r[0].startswith(prefix) else r[0] for r in rows}
+
+
 @cli.command("ready")
 @click.option(
     "--state",
@@ -3394,7 +3427,16 @@ def tags(state: str | None, show_tasks: bool, filter_tags: tuple[str, ...]):
     default=None,
     help="Exclude tasks in this pool (e.g. '--exclude-pool frontier')",
 )
-def ready(state, output_json, output_jsonl, use_cache, pool_filter, exclude_pool):
+@click.option(
+    "--skip-claimed",
+    "skip_claimed",
+    is_flag=True,
+    help=(
+        "Skip tasks already claimed by another coordination session. "
+        "Reads state/coordination/coord.db; silently degrades if absent."
+    ),
+)
+def ready(state, output_json, output_jsonl, use_cache, pool_filter, exclude_pool, skip_claimed):
     """List all ready (unblocked) tasks.
 
     Shows tasks that have no dependencies or whose dependencies are all completed.
@@ -3405,6 +3447,7 @@ def ready(state, output_json, output_jsonl, use_cache, pool_filter, exclude_pool
     Use --use-cache to also check URL-based 'requires' against cached issue states.
     Use --pool to filter by work pool; default hides non-default pools (e.g. frontier).
     Use --pool all to see every pool; --pool frontier for frontier only.
+    Use --skip-claimed to hide tasks claimed by a concurrent coordination session.
     """
     console = Console()
     repo_root = find_repo_root(Path.cwd())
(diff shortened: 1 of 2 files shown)


ErikBjare:
Duplicate of https://github.com/gptme/gptme-contrib/pull/1509 ? I like `--skip-claimed` better though.

TimeToBuildBob:
Yes — both PRs implement the same feature. Closing #1509 (`--skip-claim-held`) in favor of this one since you prefer `--skip-claimed`.

Also rebased onto latest master; the test failures on the previous CI run were from running 1 commit behind master. CI is running again now.

packages/gptodo/src/gptodo/cli.py
```diff
@@ -3350,6 +3350,39 @@ def tags(state: str | None, show_tasks: bool, filter_tags: tuple[str, ...]):
...
+    my_agent = (
+        os.environ.get("CASCADE_COORDINATION_AGENT") or os.environ.get("AGENT_ID") or ""
+    ).strip()
+
+    try:
+        with sqlite3.connect(str(db_path)) as conn:
+            rows = conn.execute(
+                """SELECT task_id FROM work
+                WHERE task_id LIKE 'cascade:task:%'
+                  AND status IN ('claimed', 'completed')
```
TimeToBuildBob: ❌ **P1** — The SQL query in _get_claimed_task_ids selects rows with status IN ('claimed', 'completed'). The flag's documented purpose is to skip tasks 'already claimed by another coordination session', which logically covers only active claims. Including 'completed' means tasks that are marked finished in the coordination DB are also filtered out of the ready list. If a task's frontmatter state is still workable (e.g., 'todo' or 'backlog') because the agent that completed it did not update the task file, gptodo ready --skip-claimed will incorrectly hide it from the user, dropping actionable work. The fix is to match only the claimed status, e.g., status = 'claimed'.

TimeToBuildBob:
## 🤖 AI code review

This PR adds a `--skip-claimed` flag to the `gptodo ready` command that filters out tasks already claimed by other coordination sessions. It introduces a new helper `_get_claimed_task_ids` that queries `state/coordination/coord.db` via `sqlite3`, stripping the `cascade:task:` prefix and excluding claims by the current agent (from `CASCADE_COORDINATION_AGENT` or `AGENT_ID`). It also restores the `GPTODO_TASKS_DIR` environment variable when the CLI exits, preventing leaks in in-process tests.

<details><summary><b>2 findings</b> · ⚠️ **2** P2</summary>

**⚠️ P2 medium** — `packages/gptodo/src/gptodo/cli.py:3391`

In `_get_claimed_task_ids`, the `try` block catches every exception (including `sqlite3.OperationalError`, `DatabaseError`) and returns an empty set. This means if the coordination DB exists but is unreadable (corrupt, wrong schema, permission), the function reports no claimed tasks, so `--skip-claimed` becomes a no-op and all ready tasks—including ones actively claimed by other sessions—are listed. The PR description only promises silent degradation when the DB is absent, not when it is present but unreadable. This can cause multiple agents to pick up the same claimed task concurrently. The fix is to only return empty when the file is absent, and otherwise let the exception surface or log a warning.

```suggestion
Optionally log a warning to stderr when the DB exists but the query fails, e.g., `except Exception as e: print(f"[y]Warning: coord DB unreadable: {e}", file=sys.stderr); return set()`.
```

<sub>**How this was verified:** Traced the function: the `try` block wraps the entire query and any exception is swallowed, returning an empty set. The PR description specifies graceful degradation only for absent DB.</sub>

**⚠️ P2 medium** — `packages/gptodo/tests/test_skip_claimed.py:120`

In `test_excludes_own_agent_claims`, only `AGENT_ID` is set via `monkeypatch.setenv`, but `CASCADE_COORDINATION_AGENT` is not cleared. `_get_claimed_task_ids` reads `CASCADE_COORDINATION_AGENT` first, so if that variable is set in the test environment (e.g., CI), `my_agent` becomes that value instead of 'me', and the claim with `claimer='me'` is not excluded, causing the assertion `result == set()` to fail. The test is non-hermetic and depends on an environment variable being unset. Clear `CASCADE_COORDINATION_AGENT` (e.g., `monkeypatch.delenv("CASCADE_COORDINATION_AGENT", raising=False)`) at the start of the test.

```suggestion
Add `monkeypatch.chdir(repo)` in _setup (or the individual test) before invoking the CLI, so find_repo_root resolves to the temporary repo.
```

<sub>**How this was verified:** Checked that the test does not call monkeypatch.chdir or otherwise set the working directory, and that ready uses find_repo_root(Path.cwd()) for the DB path.</sub>

</details>

- `packages/gptodo/src/gptodo/cli.py` — Adds `--skip-claimed` option, `_get_claimed_task_ids` helper, and restores `GPTODO_TASKS_DIR` env var on context close.
- `packages/gptodo/tests/test_skip_claimed.py` — New test file with unit tests for `_get_claimed_task_ids` and integration tests for `ready --skip-claimed`.

</details>

| `1d041091993a` | 3/5 | 1 | llm | 2026-08-25 17:54 UTC |
| `0b1cb8b5488a` | 3/5 | 1 | llm | 2026-08-25 18:45 UTC |

</details>

<sub>Reviewed `4fd8cd5d3b72` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 148s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

TimeToBuildBob:
@greptileai review

packages/gptodo/src/gptodo/cli.py
TimeToBuildBob: Fixed in 0b1cb8b5 — changed `AND status IN ('claimed', 'completed')` to `AND status = 'claimed'`. Completed records are done work and should never suppress a task from the ready list.

Also pushed 03c1afe1 which fixes a test-env pollution issue (`GPTODO_TASKS_DIR` was leaking between in-process CliRunner tests); this resolves the unrelated test failures seen in CI.

packages/gptodo/src/gptodo/cli.py
```diff
@@ -3350,6 +3361,39 @@ def tags(state: str | None, show_tasks: bool, filter_tags: tuple[str, ...]):
...
+    ).strip()
+
+    try:
+        with sqlite3.connect(str(db_path)) as conn:
+            rows = conn.execute(
+                """SELECT task_id FROM work
+                WHERE task_id LIKE 'cascade:task:%'
+                  AND status = 'claimed'
+                  AND expires_at >= datetime('now')
+                  AND (? = '' OR claimer != ?)""",
```
TimeToBuildBob: ❌ **P1** — The WHERE clause in _get_claimed_task_ids excludes rows with NULL claimer or NULL expires_at whenever my_agent is non-empty. Condition `claimer != ?` evaluates to NULL (not true) when claimer is NULL, so the row is dropped; similarly `expires_at >= datetime('now')` is NULL for NULL expires_at. As a result, claims lacking these values are never added to the returned set, and ready --skip-claimed will still present those tasks as available. Two concurrent agents could then pick the same task, defeating the flag's purpose. The condition should explicitly include NULLs: `(? = '' OR claimer IS NULL OR claimer != ?)` and `(expires_at IS NULL OR expires_at >= datetime('now'))`.

TimeToBuildBob: Fixed in 4fd8cd5d — added explicit IS NULL guards for both NULL claimer and NULL expires_at. Also added regression tests: `test_includes_null_expiry_claims` and `test_includes_null_claimer_claims`.

TimeToBuildBob:
@TimeToBuildBob review

packages/gptodo/src/gptodo/cli.py
```diff
@@ -3350,6 +3361,39 @@ def tags(state: str | None, show_tasks: bool, filter_tags: tuple[str, ...]):
...
+        with sqlite3.connect(str(db_path)) as conn:
+            rows = conn.execute(
+                """SELECT task_id FROM work
+                WHERE task_id LIKE 'cascade:task:%'
+                  AND status = 'claimed'
+                  AND (expires_at IS NULL OR expires_at >= datetime('now'))
+                  AND (? = '' OR claimer IS NULL OR claimer != ?)""",
+                (my_agent, my_agent),
+            ).fetchall()
+    except Exception:
```
TimeToBuildBob, line 3391: ⚠️ **P2** — In `_get_claimed_task_ids`, the `try` block catches every exception (including `sqlite3.OperationalError`, `DatabaseError`) and returns an empty set. This means if the coordination DB exists but is unreadable (corrupt, wrong schema, permission), the function reports no claimed tasks, so `--skip-claimed` becomes a no-op and all ready tasks—including ones actively claimed by other sessions—are listed. The PR description only promises silent degradation when the DB is absent, not when it is present but unreadable. This can cause multiple agents to pick up the same claimed task concurrently. The fix is to only return empty when the file is absent, and otherwise let the exception surface or log a warning. _(2/2 passes)_

packages/gptodo/tests/test_skip_claimed.py
```diff
@@ -0,0 +1,211 @@
...
+                    "expires_at": "2099-01-01 00:00:00",
+                }
+            ],
+        )
+        result = _get_claimed_task_ids(tmp_path)
+        assert "unknown-claimer-task" in result
+
+    def test_excludes_own_agent_claims(self, tmp_path: Path, monkeypatch) -> None:
+        db_path = tmp_path / "state" / "coordination" / "coord.db"
+        monkeypatch.setenv("AGENT_ID", "me")
```
TimeToBuildBob, line 120: ⚠️ **P2** — In `test_excludes_own_agent_claims`, only `AGENT_ID` is set via `monkeypatch.setenv`, but `CASCADE_COORDINATION_AGENT` is not cleared. `_get_claimed_task_ids` reads `CASCADE_COORDINATION_AGENT` first, so if that variable is set in the test environment (e.g., CI), `my_agent` becomes that value instead of 'me', and the claim with `claimer='me'` is not excluded, causing the assertion `result == set()` to fail. The test is non-hermetic and depends on an environment variable being unset. Clear `CASCADE_COORDINATION_AGENT` (e.g., `monkeypatch.delenv("CASCADE_COORDINATION_AGENT", raising=False)`) at the start of the test. _(2/2 passes)_

PR #1511, opened 2026-08-25
fix(gptme-rag): make Indexer/ContextAssembler lazy imports

TimeToBuildBob:
## Summary

- `import gptme_rag` previously pulled in chromadb + sentence_transformers eagerly via `Indexer` and `ContextAssembler`
- This blocked adoption in lightweight hook runtimes (match-lessons.py, prompt-submit hooks) where the dense backend is not needed
- Switches to module-level `__getattr__` so heavy deps load only when the caller actually uses `Indexer` or `ContextAssembler`
- `lesson_matcher` and `observability` functions remain eagerly importable

## Motivation

Phase 3 of the gptme-rag adoption plan (ErikBjare/bob) requires `match-lessons.py` to import from `gptme_rag` for BM25 scoring. The current eager import of chromadb + torch makes this unusable in the hook runtime (adds ~10s startup + GB of deps).

## Changes

- `gptme_rag/__init__.py`: remove two eager imports, add `__getattr__` with lazy loading
- `tests/test_lazy_import.py`: 5 tests verifying the no-heavy-deps guarantee and lazy resolution

## Test plan

- [x] `import gptme_rag` does not import chromadb (verified locally)
- [x] `gptme_rag.scan_lessons`, `score_lessons`, etc. importable without chromadb
- [x] `gptme_rag.Indexer` still resolves correctly via `__getattr__`
- [x] Unknown attributes raise `AttributeError`
- [ ] CI full test suite

### DIFF
--- a/packages/gptme-rag/src/gptme_rag/__init__.py
+++ b/packages/gptme-rag/src/gptme_rag/__init__.py
@@ -1,4 +1,3 @@
-from .indexing.indexer import Indexer
 from .lesson_matcher import (
     filter_by_session_category,
     keyword_to_regex,
@@ -13,7 +12,6 @@ from .observability import (
     log_injection,
     summarize_injections,
 )
-from .query.context_assembler import ContextAssembler
 
 __all__ = [
     "ContextAssembler",
@@ -29,3 +27,15 @@ __all__ = [
     "score_lessons",
     "summarize_injections",
 ]
+
+
+def __getattr__(name: str) -> object:
+    if name == "Indexer":
+        from .indexing.indexer import Indexer
+
+        return Indexer
+    if name == "ContextAssembler":
+        from .query.context_assembler import ContextAssembler
+
+        return ContextAssembler
+    raise AttributeError(f"module 'gptme_rag' has no attribute {name!r}")
--- a/packages/gptme-rag/src/gptme_rag/__init__.py
+++ b/packages/gptme-rag/src/gptme_rag/__init__.py
@@ -31,11 +31,27 @@ __all__ = [
 
 def __getattr__(name: str) -> object:
     if name == "Indexer":
-        from .indexing.indexer import Indexer
+        try:
+            from .indexing.indexer import Indexer
 
-        return Indexer
+            return Indexer
+        except ImportError as e:
+            raise AttributeError(
+                f"module 'gptme_rag' has no attribute {name!r} "
+                "(heavy deps not installed; install gptme-rag[full])"
+            ) from e
     if name == "ContextAssembler":
-        from .query.context_assembler import ContextAssembler
+        try:
+            from .query.context_assembler import ContextAssembler
 
-        return ContextAssembler
+            return ContextAssembler
+        except ImportError as e:
+            raise AttributeError(
+                f"module 'gptme_rag' has no attribute {name!r} "
+                "(heavy deps not installed; install gptme-rag[full])"
+            ) from e
     raise AttributeError(f"module 'gptme_rag' has no attribute {name!r}")
+
+
+def __dir__() -> list[str]:
+    return __all__
--- a/packages/gptme-rag/src/gptme_rag/__init__.py
+++ b/packages/gptme-rag/src/gptme_rag/__init__.py
@@ -13,6 +13,10 @@ from .observability import (
     summarize_injections,
 )
 
+# Note: Indexer and ContextAssembler are lazy-loaded via __getattr__ below.
+# They require the full install (gptme-rag[full]) with chromadb + sentence_transformers.
+# ``from gptme_rag import *`` will raise AttributeError for these names when the
+# heavy deps are absent; prefer explicit imports or hasattr() guards at call sites.
 __all__ = [
     "ContextAssembler",
     "IndexHealth",
@@ -34,6 +38,8 @@ def __getattr__(name: str) -> object:
         try:
             from .indexing.indexer import Indexer
 
+            # Cache in module namespace so subsequent accesses skip __getattr__.
+            globals()["Indexer"] = Indexer
             return Indexer
         except ImportError as e:
             raise AttributeError(
@@ -44,6 +50,8 @@ def __getattr__(name: str) -> object:
         try:
             from .query.context_assembler import ContextAssembler
 
+            # Cache in module namespace so subsequent accesses skip __getattr__.
+            globals()["ContextAssembler"] = ContextAssembler
             return ContextAssembler
         except ImportError as e:
             raise AttributeError(
@@ -54,4 +62,5 @@ def __getattr__(name: str) -> object:
 
 
 def __dir__() -> list[str]:
-    return __all__
+    # Include standard module attributes (from __dict__) plus lazy names from __all__.
+    return sorted(set(globals().keys()) | set(__all__))
--- a/packages/gptme-rag/tests/test_lazy_import.py
+++ b/packages/gptme-rag/tests/test_lazy_import.py
@@ -76,3 +76,46 @@ def test_unknown_attr_raises() -> None:
 
     with pytest.raises(AttributeError, match="gptme_rag"):
         _ = gptme_rag.NonExistentClass
+
+
+def test_missing_dep_raises_attribute_error_not_import_error() -> None:
+    """When a heavy dep is absent, __getattr__ must raise AttributeError, not ImportError.
+
+    hasattr() and getattr(..., default) only catch AttributeError.  If __getattr__
+    propagates ImportError, code that probes for optional Indexer will crash
+    instead of gracefully falling back.
+    """
+    import sys
+    from unittest.mock import patch
+
+    import gptme_rag
+    import pytest
+
+    # Evict any cached resolution of the indexer submodule.
+    for key in list(sys.modules):
+        if "gptme_rag.indexing" in key:
+            del sys.modules[key]
+
+    # Block the submodule as if chromadb were not installed (None → ImportError).
+    blocking = {
+        "gptme_rag.indexing": None,
+        "gptme_rag.indexing.indexer": None,
+    }
+    with patch.dict(sys.modules, blocking):
+        # hasattr must return False, not bubble up an ImportError.
+        assert not hasattr(gptme_rag, "Indexer")
+        # getattr with a default must return the default.
+        sentinel = object()
+        assert getattr(gptme_rag, "Indexer", sentinel) is sentinel
+        # Direct access must raise AttributeError (not ImportError).
+        with pytest.raises(AttributeError):
+            _ = gptme_rag.Indexer
+
+
+def test_dir_includes_lazy_names() -> None:
+    """Lazy names (Indexer, ContextAssembler) must appear in dir(gptme_rag)."""
+    import gptme_rag
+
+    d = dir(gptme_rag)
+    assert "Indexer" in d
+    assert "ContextAssembler" in d
--- a/packages/gptme-rag/tests/test_lazy_import.py
+++ b/packages/gptme-rag/tests/test_lazy_import.py
@@ -91,17 +91,11 @@ def test_missing_dep_raises_attribute_error_not_import_error() -> None:
     import gptme_rag
     import pytest
 
-    # Evict any cached resolution of the indexer submodule.
-    for key in list(sys.modules):
-        if "gptme_rag.indexing" in key:
-            del sys.modules[key]
-
-    # Block the submodule as if chromadb were not installed (None → ImportError).
-    blocking = {
-        "gptme_rag.indexing": None,
-        "gptme_rag.indexing.indexer": None,
-    }
-    with patch.dict(sys.modules, blocking):
+    # Block only the indexer submodule (None → ImportError).  Do NOT evict
+    # gptme_rag.indexing.* broadly — that clobbers gptme_rag.indexing.document,
+    # causing Document class-identity mismatches and pickling failures in sibling
+    # tests that share Document instances across the session.
+    with patch.dict(sys.modules, {"gptme_rag.indexing.indexer": None}):
         # hasattr must return False, not bubble up an ImportError.
         assert not hasattr(gptme_rag, "Indexer")
         # getattr with a default must return the default.
--- a/packages/gptme-rag/tests/test_lazy_import.py
+++ b/packages/gptme-rag/tests/test_lazy_import.py
@@ -63,6 +63,10 @@ def test_lazy_classes_in_all() -> None:
 
 def test_lazy_indexer_resolves() -> None:
     """gptme_rag.Indexer must resolve to the real class via __getattr__."""
+    import pytest
+
+    pytest.importorskip("chromadb", reason="heavy deps not installed")
+
     import gptme_rag
 
     cls = gptme_rag.Indexer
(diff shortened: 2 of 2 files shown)


packages/gptme-rag/tests/test_lazy_import.py
```diff
@@ -0,0 +1,66 @@
...
+def test_plain_import_does_not_pull_chromadb() -> None:
+    """``import gptme_rag`` must not import chromadb or sentence_transformers."""
+    # Remove any prior import so the test is isolated
+    mods_to_drop = [k for k in sys.modules if k.startswith("gptme_rag")]
+    for mod in mods_to_drop:
+        del sys.modules[mod]
+
+    import gptme_rag  # noqa: F401
+
+    heavy = [m for m in sys.modules if m in ("chromadb", "sentence_transformers")]
```
TimeToBuildBob: ⚠️ **P2** — The test `test_plain_import_does_not_pull_chromadb` checks that `chromadb` and `sentence_transformers` are not in `sys.modules` after importing `gptme_rag`, but it never removes these modules from `sys.modules` before running. If any other test in the same process has imported these modules, they will still be present, causing the assertion to fail even though the import of `gptme_rag` itself did not pull them in. This makes the test non-hermetic and dependent on the execution order of tests, leading to false failures in a full test suite. The test should isolate itself, for example by removing these modules from `sys.modules` before the import, or running the import in a subprocess.

TimeToBuildBob:
## 🤖 AI code review

This PR makes the gptme_rag package lazily import Indexer and ContextAssembler by replacing eager imports in __init__.py with a module-level __getattr__ function. It also adds a new test file test_lazy_import.py with five tests verifying that a plain import does not load chromadb or sentence_transformers, that lesson-matcher functions remain importable, and that lazy resolution and error behavior work.

**⚠️ P2 medium** — `packages/gptme-rag/tests/test_lazy_import.py:22`

The test `test_plain_import_does_not_pull_chromadb` checks that `chromadb` and `sentence_transformers` are not in `sys.modules` after importing `gptme_rag`, but it never removes these modules from `sys.modules` before running. If any other test in the same process has imported these modules, they will still be present, causing the assertion to fail even though the import of `gptme_rag` itself did not pull them in. This makes the test non-hermetic and dependent on the execution order of tests, leading to false failures in a full test suite. The test should isolate itself, for example by removing these modules from `sys.modules` before the import, or running the import in a subprocess.

<sub>**How this was verified:** Checked the test code to confirm there is no cleanup of chromadb or sentence_transformers from sys.modules, and that the assertion relies on these modules not being imported by prior tests.</sub>

</details>

- `packages/gptme-rag/src/gptme_rag/__init__.py` — Removed eager imports of Indexer and ContextAssembler; added __getattr__ to import them on demand.
- `packages/gptme-rag/tests/test_lazy_import.py` — Added five tests for the lazy import behavior, including that heavy deps are not imported on plain import.

</details>

<sub>Reviewed `f4aaa6175b5c` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 78s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

TimeToBuildBob:
Fixed test isolation in `859833b9`.

The original `test_plain_import_does_not_pull_chromadb` only evicted `gptme_rag.*` from `sys.modules` but not `chromadb`/`sentence_transformers`, causing it to find already-imported heavy deps and fail. More importantly, the missing restore caused all subsequent tests that use `monkeypatch.setattr("gptme_rag.observability.Path", ...)` or `monkeypatch.setattr("gptme_rag.lesson_matcher._bm25_score", ...)` to patch the **newly re-imported** module object rather than the one held by the test file's `from gptme_rag.xxx import func` binding, silently making those monkeypatches no-ops.

Fix: save+restore both `gptme_rag.*` **and** heavy dep modules (`chromadb`, `sentence_transformers`) around the fresh import, so the test is hermetic and subsequent tests see the original module objects.

TimeToBuildBob:
@greptileai review

packages/gptme-rag/src/gptme_rag/__init__.py
```diff
@@ -29,3 +27,15 @@
     "score_lessons",
     "summarize_injections",
 ]
+
+
+def __getattr__(name: str) -> object:
+    if name == "Indexer":
+        from .indexing.indexer import Indexer
```
TimeToBuildBob: ❌ **P1** — Accessing gptme_rag.Indexer or gptme_rag.ContextAssembler now triggers the lazy import inside __getattr__. If the heavy dependencies (e.g., chromadb) are not installed, the `from .indexing.indexer import Indexer` statement raises an ImportError. Because __getattr__ does not catch this exception, it propagates to the caller. hasattr(gptme_rag, 'Indexer') and getattr(gptme_rag, 'Indexer', default) do not catch ImportError — they only catch AttributeError — so they will raise instead of returning False/default. This breaks any code that probes for the optional Indexer/ContextAssembler attributes to decide whether to use the dense backend; the probe itself now crashes when the dependency is missing. The fix is to catch ImportError in __getattr__ and raise AttributeError instead, preserving the usual probing semantics.

packages/gptme-rag/src/gptme_rag/__init__.py
```diff
@@ -29,3 +27,15 @@
     "score_lessons",
     "summarize_injections",
 ]
+
+
+def __getattr__(name: str) -> object:
```
TimeToBuildBob, line 36: ⚠️ **P2** — The lazy imports for Indexer and ContextAssembler do not appear in dir(gptme_rag) because they are not present in the module's __dict__. Previously, with eager imports, these names were regular module attributes and would be listed by dir. Any code that introspects the module via dir or vars to discover exported classes will no longer see them, a breaking change for tools that rely on dir rather than __all__. This can cause failures in documentation generators, autocompletion, and runtime type discovery. Adding a module-level __dir__ that includes these names would restore the previous visibility.

TimeToBuildBob: Fixed in 802780d0: added `__dir__()` returning `__all__`, so `Indexer` and `ContextAssembler` now appear in `dir(gptme_rag)`. Added regression test `test_dir_includes_lazy_names`.

packages/gptme-rag/tests/test_lazy_import.py
TimeToBuildBob: This was already addressed in commit 859833b9 (the prior push): the test now evicts all gptme_rag submodules and heavy deps from sys.modules before the import, then restores the original state in a `finally` block. It measures what a fresh process would see, not any prior-test state.

packages/gptme-rag/src/gptme_rag/__init__.py
TimeToBuildBob: Fixed in 802780d0: `__getattr__` now wraps the inner `ImportError` as `AttributeError`, so `hasattr(gptme_rag, 'Indexer')` returns `False` instead of crashing when chromadb is absent. Added regression test `test_missing_dep_raises_attribute_error_not_import_error`.

TimeToBuildBob:
## Review findings addressed (commit 802780d0)

All three AI reviewer findings are now resolved:

- **P1 — ImportError not caught** (fixed): `__getattr__` now wraps `ImportError` as `AttributeError`, so `hasattr(gptme_rag, 'Indexer')` returns `False` instead of crashing when chromadb is absent. Regression test: `test_missing_dep_raises_attribute_error_not_import_error`.

- **P2 — dir() visibility** (fixed): Added `__dir__()` returning `__all__`, so `Indexer` and `ContextAssembler` now appear in `dir(gptme_rag)`. Regression test: `test_dir_includes_lazy_names`.

- **P2 — test hermeticity** (already addressed in prior commit 859833b9): `test_plain_import_does_not_pull_chromadb` evicts all `gptme_rag` submodules and heavy deps from `sys.modules` before the import, then restores original state in a `finally` block.

7 tests pass locally on Python 3.13.

TimeToBuildBob:
Follow-up commit d75c7124: the test  evicted too much — evicting all  also clobbered , causing  class-identity mismatches and pickling failures in . Fixed by blocking only  via , which is the only submodule that imports chromadb.

TimeToBuildBob:
## Merge recommendation

CI: ✅ all green (3.10, 3.11, 3.12)
AI reviewer: all 3 findings resolved (P1 + 2× P2)
Greptile: trigger posted, awaiting review (billing dark)

**What this PR adds** (lazy-import fix for `gptme_rag/__init__.py`):
- `__getattr__` now catches `ImportError` → re-raises as `AttributeError`, so `hasattr(gptme_rag, 'Indexer')` returns `False` when chromadb is absent instead of propagating `ImportError`
- `__dir__()` added so `Indexer`/`ContextAssembler` appear in `dir(gptme_rag)`
- Tests: isolation regression fixed (blocks only `gptme_rag.indexing.indexer` via `patch.dict`, not the whole indexing tree, avoiding `Document` class-identity pickle failures)

This is a prerequisite for Phase 3 of `upstream-retrieval-into-gptme-rag`: making the brain's `match-lessons.py` a thin wrapper calling gptme-rag. The lazy-import contract is what lets the hook import `gptme_rag` without pulling in chromadb/torch on every prompt-submit invocation.

Ready for squash-merge.

packages/gptme-rag/src/gptme_rag/__init__.py
```diff
@@ -29,3 +27,31 @@
...
+    if name == "Indexer":
+        try:
+            from .indexing.indexer import Indexer
+
+            return Indexer
```
TimeToBuildBob, line 43: ⚠️ **P2** — `__getattr__` is invoked for every attribute access to `Indexer` or `ContextAssembler` because these names are not defined at module level. Each call executes `from .indexing.indexer import Indexer` (or equivalent), which imports the submodule and returns the class. However, the result is not cached in the module namespace. Repeated accesses, such as in a loop or from a function called frequently, trigger a new import operation every time, re-importing the submodule and re-executing its top-level code. This can cause redundant initialization (e.g., loading heavy dependencies or building caches) and performance degradation. The standard practice is to store the resolved class in a module-level variable after the first import so subsequent lookups bypass `__getattr__` entirely. _(1/2 passes — pass 2 looked and disagreed)_

packages/gptme-rag/tests/test_lazy_import.py
```diff
@@ -0,0 +1,115 @@
...
+
+    assert "Indexer" in gptme_rag.__all__
+    assert "ContextAssembler" in gptme_rag.__all__
+
+
+def test_lazy_indexer_resolves() -> None:
+    """gptme_rag.Indexer must resolve to the real class via __getattr__."""
+    import gptme_rag
+
+    cls = gptme_rag.Indexer
```
TimeToBuildBob, line 72: ⚠️ **P2** — The test `test_lazy_indexer_resolves` directly accesses `gptme_rag.Indexer`, which triggers the lazy import in `__getattr__`. In an environment without chromadb/sentence_transformers, this raises AttributeError, so the test fails. This makes the test suite non-hermetic: it depends on heavy deps being installed, contradicting the PR's goal of allowing `import gptme_rag` without them. The test should be skipped when the heavy deps are absent, e.g., via `pytest.importorskip`. _(1/2 passes — pass 1 looked and disagreed)_

packages/gptme-rag/src/gptme_rag/__init__.py
```diff
@@ -29,3 +27,31 @@
...
+        except ImportError as e:
+            raise AttributeError(
+                f"module 'gptme_rag' has no attribute {name!r} "
+                "(heavy deps not installed; install gptme-rag[full])"
+            ) from e
+    raise AttributeError(f"module 'gptme_rag' has no attribute {name!r}")
+
+
+def __dir__() -> list[str]:
+    return __all__
```
TimeToBuildBob: ⚠️ **P2** — The module defines `__dir__` to return only `__all__`. This omits all standard module attributes such as `__name__`, `__file__`, `__doc__`, and `__package__` from the result of `dir()`. Tools that rely on `dir()` for introspection (e.g., `help()`, IDEs) will no longer see these attributes, potentially breaking their display or behavior. For example, `help(gptme_rag)` may not show the module docstring or name. The override should be extended to include the standard module attributes, or the lazy names should be handled differently. _(1/2 passes — pass 1 looked and disagreed)_

TimeToBuildBob:
## AI reviewer findings addressed (commit 64e1e976)

All 4 open P2 findings resolved:

- **P2 — `__getattr__` not caching** (fixed): After the first successful resolution, `Indexer` / `ContextAssembler` are stored in `globals()` so subsequent accesses bypass `__getattr__` entirely — standard lazy-module pattern.

- **P2 — `__dir__` omits standard module attrs** (fixed): `__dir__` now returns `sorted(set(globals().keys()) | set(__all__))`, so `__name__`, `__file__`, `__doc__` etc. appear alongside the lazy names.

- **P2 — `test_lazy_indexer_resolves` non-hermetic** (fixed): Added `pytest.importorskip("chromadb")` so the test skips gracefully when heavy deps are absent instead of raising `AttributeError`.

- **P2 — star-import caveat undocumented** (fixed): Added a comment above `__all__` documenting that `from gptme_rag import *` will raise `AttributeError` for `Indexer`/`ContextAssembler` when heavy deps are absent, and that callers should use `hasattr()` guards instead.

Also fixed a test isolation issue that the caching change exposed: `test_missing_dep_raises_attribute_error_not_import_error` now clears the module-level cached `Indexer` before the `patch.dict` block so `__getattr__` is actually exercised.

7/7 tests pass locally (Python 3.13).

TimeToBuildBob:
## 🤖 AI code review

The PR makes the Indexer and ContextAssembler classes lazy-loaded in gptme_rag by removing their top-level imports and adding a `__getattr__` that loads them on first access, so importing gptme_rag no longer pulls in chromadb and sentence_transformers. It also implements `__dir__` to include these lazy names and adds a test suite covering the lazy-loading behavior.

**⚠️ P2 medium** — `packages/gptme-rag/tests/test_lazy_import.py:68`

`test_lazy_indexer_resolves` only skips when `chromadb` is missing, but the `Indexer` class also requires `sentence_transformers` (per the PR description: "full install (gptme-rag[full]) with chromadb + sentence_transformers"). In an environment where `chromadb` is installed but `sentence_transformers` is not, this test runs and then fails when it accesses `gptme_rag.Indexer` (the import of `indexer` raises `ImportError` for the missing `sentence_transformers`). The skip condition should check both heavy dependencies so the test is skipped unless both are present.

```suggestion
pytest.importorskip("chromadb", reason="heavy deps not installed"); pytest.importorskip("sentence_transformers", reason="heavy deps not installed")
```

<sub>**How this was verified:** The PR description states the full install requires chromadb + sentence_transformers; the test only checks chromadb, so it will not skip in a partial dependency environment.</sub>

</details>

- `packages/gptme-rag/src/gptme_rag/__init__.py` — Removes eager imports of Indexer and ContextAssembler; adds `__getattr__` to lazily import and cache them and `__dir__` to list them.
- `packages/gptme-rag/tests/test_lazy_import.py` — Adds tests verifying plain import does not pull heavy deps, lazy resolution, AttributeError on missing deps, and dir() includes lazy names.

</details>

<sub>Reviewed `64e1e976aa2d` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 167s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

PR #1513, opened 2026-08-26
fix(output-clarity): add match.keywords to prevent over-broad descriptor injection

TimeToBuildBob:
## Problem

LOO analysis (n=995, conf=1.00) shows the `output-clarity` skill fires in **36% of sessions** with Δ=−0.026 — harmful at 100% confidence.

Root cause: The SKILL.md had `name: output-clarity` but no `match:` section. Without explicit keywords, the lesson matcher uses descriptor scoring on name/description/tags containing broad terms like `output`, `clarity`, `action`, `structured` — all common in any coding session.

The skill is designed for explicit invocation (`/output-clarity`) by users wanting accessibility-optimized formatting. Injecting it automatically in research/exploration sessions where exploration-first output is correct introduces Goodhart pressure toward action-first style when it's wrong.

## Fix

Add a `match.keywords` section with highly specific phrases that only appear when a user explicitly requests clarity mode: `/output-clarity`, `output clarity mode`, `clarity mode`, `neurodivergent formatting`, `ADHD formatting`, `i-have-adhd plugin`, `stop clarity mode`.

These won't match in typical coding/research sessions — injection rate expected to drop from ~36% to <1%.

## Evidence

LOO tracking: `scripts/lesson-loo-analysis.py` (n=995 sessions, conf=1.00)

### DIFF
--- a/skills/output-clarity/SKILL.md
+++ b/skills/output-clarity/SKILL.md
@@ -14,7 +14,7 @@ match:
   keywords:
     - "/output-clarity"
     - "output clarity mode"
-    - "clarity mode"
+    - "enable clarity mode"
     - "output-clarity mode"
     - "neurodivergent formatting"
     - "ADHD formatting"
--- a/skills/output-clarity/SKILL.md
+++ b/skills/output-clarity/SKILL.md
@@ -5,11 +5,24 @@ license: MIT
 compatibility: "Works with all gptme backends and modes"
 metadata:
   author: bob
-  version: "1.0.0"
+  version: "1.1.0"
   tags: [accessibility, output-clarity, neurodivergent, action-first, structured-output]
   requires_tools: []
   requires_skills: []
   source: "Adapted from ayghri/i-have-adhd (https://github.com/ayghri/i-have-adhd)"
+match:
+  keywords:
+    - "/output-clarity"
+    - "output clarity mode"
+    - "output-clarity mode"
+    - "clarity mode"
+    - "neurodivergent formatting"
+    - "ADHD formatting"
+    - "i-have-adhd plugin"
+    - "ayghri formatting"
+    - "action-first formatting"
+    - "stop clarity mode"
+    - "/stop-clarity"
 ---
 
 # Output Clarity Skill
--- a/skills/output-clarity/SKILL.md
+++ b/skills/output-clarity/SKILL.md
@@ -14,7 +14,6 @@ match:
   keywords:
     - "/output-clarity"
     - "output clarity mode"
-    - "clarity mode"
     - "enable clarity mode"
     - "use clarity mode"
     - "start clarity mode"
--- a/skills/output-clarity/SKILL.md
+++ b/skills/output-clarity/SKILL.md
@@ -14,6 +14,7 @@ match:
   keywords:
     - "/output-clarity"
     - "output clarity mode"
+    - "clarity mode"
     - "output-clarity mode"
     - "neurodivergent formatting"
     - "ADHD formatting"
--- a/skills/output-clarity/SKILL.md
+++ b/skills/output-clarity/SKILL.md
@@ -21,8 +21,6 @@ match:
     - "i-have-adhd plugin"
     - "ayghri formatting"
     - "action-first formatting"
-    - "stop clarity mode"
-    - "/stop-clarity"
 ---
 
 # Output Clarity Skill
--- a/skills/output-clarity/SKILL.md
+++ b/skills/output-clarity/SKILL.md
@@ -14,6 +14,7 @@ match:
   keywords:
     - "/output-clarity"
     - "output clarity mode"
+    - "clarity mode"
     - "enable clarity mode"
     - "output-clarity mode"
     - "neurodivergent formatting"
--- a/skills/output-clarity/SKILL.md
+++ b/skills/output-clarity/SKILL.md
@@ -15,7 +15,6 @@ match:
     - "/output-clarity"
     - "output clarity mode"
     - "output-clarity mode"
-    - "clarity mode"
     - "neurodivergent formatting"
     - "ADHD formatting"
     - "i-have-adhd plugin"
--- a/skills/output-clarity/SKILL.md
+++ b/skills/output-clarity/SKILL.md
@@ -14,6 +14,7 @@ match:
   keywords:
     - "/output-clarity"
     - "output clarity mode"
+    - "clarity mode"
     - "enable clarity mode"
     - "use clarity mode"
     - "start clarity mode"
--- a/skills/output-clarity/SKILL.md
+++ b/skills/output-clarity/SKILL.md
@@ -14,8 +14,10 @@ match:
   keywords:
     - "/output-clarity"
     - "output clarity mode"
-    - "clarity mode"
     - "enable clarity mode"
+    - "use clarity mode"
+    - "start clarity mode"
+    - "activate clarity mode"
     - "output-clarity mode"
     - "neurodivergent formatting"
     - "ADHD formatting"


skills/output-clarity/SKILL.md
```diff
@@ -5,11 +5,24 @@ license: MIT
...
+    - "/output-clarity"
+    - "output clarity mode"
+    - "output-clarity mode"
+    - "clarity mode"
+    - "neurodivergent formatting"
+    - "ADHD formatting"
+    - "i-have-adhd plugin"
+    - "ayghri formatting"
+    - "action-first formatting"
+    - "stop clarity mode"
```
TimeToBuildBob: ⚠️ **P2** — The match.keywords list includes "stop clarity mode" and "/stop-clarity", which are the exact phrases the skill itself tells users to use to disable clarity mode (line 42: "Disable with: \"stop clarity mode\" or `/stop-clarity`"). When a user utters either phrase to turn the mode off, the matcher will treat that as a request to load the skill, injecting output-clarity formatting rules at the very moment the user is trying to stop them. This defeats the PR's stated goal of only matching explicit enablement requests and can reapply formatting after the user has disabled it. Concretely, a user in a session without clarity mode active who says "stop clarity mode" (perhaps out of caution) will suddenly get action-first, structured responses they did not ask for, creating a false-positive injection that the PR aims to eliminate. Remove the two disable phrases from match.keywords, or split them into a separate negative-match mechanism.

TimeToBuildBob:
## 🤖 AI code review

The PR adds a `match.keywords` section to the output-clarity skill's front matter, listing explicit trigger phrases to prevent automatic injection based on broad descriptor matching. The version is bumped from 1.0.0 to 1.1.0.

**⚠️ P2 medium · ℹ️ non-blocking** — `skills/output-clarity/SKILL.md:14`

The PR description states the added keywords include 'clarity mode' and 'stop clarity mode', but the actual `match.keywords` list in the diff omits both phrases. The list contains 'output clarity mode', 'enable clarity mode', 'use clarity mode', 'start clarity mode', 'activate clarity mode', and others, but not the bare 'clarity mode' or 'stop clarity mode'. Consequently, if a user explicitly asks for 'clarity mode' (without other qualifiers) or says 'stop clarity mode', the skill's matching logic will not trigger, so the skill's instructions will not be injected. This contradicts the stated intent of matching 'highly specific phrases that only appear when a user explicitly requests clarity mode' and may cause the skill to be unavailable for those exact invocations.

```suggestion
- "clarity mode"
- "stop clarity mode"
```

<sub>**How this was verified:** Compared the added keyword list against the phrases listed in the PR description; 'clarity mode' and 'stop clarity mode' are absent from the list.</sub>

<sub>Consensus: 1/3 passes agreed — passes 2, 3 looked and disagreed</sub> <br/><span style='color:gray;font-size:0.85em'>Distinct keys: 1 (<code>general</code>)</span>

</details>

- `skills/output-clarity/SKILL.md` — Adds a `match.keywords` list with explicit trigger phrases and bumps the version to 1.1.0.

</details>

| `cc523e935d6a` | 4/5 | 1 | llm | 2026-08-26 02:22 UTC |
| `df660aa05767` | 3/5 | 1 | llm | 2026-08-26 02:41 UTC |
| `8afb567a45d6` | 3/5 | 1 | llm | 2026-08-26 03:21 UTC |
| `5e088d93604d` | 3/5 | 1 | llm | 2026-08-26 04:02 UTC |
| `bafbffffedb7` | 3/5 | 1 | llm | 2026-08-26 05:17 UTC |
| `04a762923114` | 3/5 | 1 | llm | 2026-08-26 05:23 UTC |
| `67951daff53c` | 3/5 | 1 | llm | 2026-08-26 05:26 UTC |
| `0cb68b071761` | 4/5 | 1 | llm | 2026-08-26 06:03 UTC |

</details>

<sub>Reviewed `f7d717c9368b` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 43s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

skills/output-clarity/SKILL.md
TimeToBuildBob: Fixed in df660aa0 — removed "stop clarity mode" and "/stop-clarity" from match.keywords. Disable phrases are now only in the body docs (line 42), not in the matcher, so uttering them won't re-trigger the skill.

TimeToBuildBob:
@TimeToBuildBob review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Fixed the P1 finding in 5e088d93: re-added `clarity mode` to match.keywords.

Prior session removed it in 8afb567 because it's a substring of "stop clarity mode" (the P2 fix). But the AI reviewer correctly flagged its absence as a P1 mismatch — the PR description lists it as an intended trigger. Re-added it; the disable phrases ("stop clarity mode", "/stop-clarity") are already gone from the keyword list, so the only remaining risk is that sessions where a user actively discusses disabling clarity mode might get a brief re-injection — a minor edge case vs. the main fix (preventing injection in unrelated sessions).

skills/output-clarity/SKILL.md
```diff
@@ -5,11 +5,22 @@ license: MIT
...
+  version: "1.1.0"
   tags: [accessibility, output-clarity, neurodivergent, action-first, structured-output]
   requires_tools: []
   requires_skills: []
   source: "Adapted from ayghri/i-have-adhd (https://github.com/ayghri/i-have-adhd)"
+match:
+  keywords:
+    - "/output-clarity"
+    - "output clarity mode"
+    - "clarity mode"
```
TimeToBuildBob: ❌ **P1** — The keyword "clarity mode" (line 17) overlaps with the documented disabling phrase "stop clarity mode" (line 40). If the skill matcher performs substring or case-insensitive matching on user input, a user who types "stop clarity mode" to disable the skill will have their message contain "clarity mode", triggering the skill instead of disabling it. This directly contradicts the skill's own instruction for disabling and causes the opposite of the intended behaviour, potentially locking the user into clarity mode when they explicitly request to exit it. _(2/2 passes)_

TimeToBuildBob:
Fixed P1 in bafbffff: changed `"clarity mode"` to `"enable clarity mode"`.

The AI reviewer correctly identified the substring overlap: "clarity mode" is contained in the documented disable phrase "stop clarity mode", so a substring matcher would trigger the skill when a user tries to disable it. `"enable clarity mode"` avoids that — it is unambiguous and will not match the disable phrase.

@TimeToBuildBob review

TimeToBuildBob:
Restored `"clarity mode"` to match.keywords in 0cb68b0.

Previous sessions removed it to avoid substring-matching "stop clarity mode", but that reasoning was wrong: when a user types "stop clarity mode", having the skill injected is actually *helpful* — the skill contains the explicit disable instruction (`Disable with: "stop clarity mode"`), so the model gets the context it needs to honor the request. Without "clarity mode" in the list, the model might not even know what "stop clarity mode" means if the skill hasn't been mentioned earlier in the session.

`"enable clarity mode"` and the other variants stay in the list as well; "clarity mode" is the short canonical form.

@TimeToBuildBob review

skills/output-clarity/SKILL.md
```diff
@@ -5,11 +5,25 @@ license: MIT
...
 metadata:
   author: bob
-  version: "1.0.0"
+  version: "1.1.0"
   tags: [accessibility, output-clarity, neurodivergent, action-first, structured-output]
   requires_tools: []
   requires_skills: []
   source: "Adapted from ayghri/i-have-adhd (https://github.com/ayghri/i-have-adhd)"
+match:
+  keywords:
```
TimeToBuildBob, line 14: ⚠️ **P2** — The PR description states the added keywords include 'clarity mode' and 'stop clarity mode', but the actual `match.keywords` list in the diff omits both phrases. The list contains 'output clarity mode', 'enable clarity mode', 'use clarity mode', 'start clarity mode', 'activate clarity mode', and others, but not the bare 'clarity mode' or 'stop clarity mode'. Consequently, if a user explicitly asks for 'clarity mode' (without other qualifiers) or says 'stop clarity mode', the skill's matching logic will not trigger, so the skill's instructions will not be injected. This contradicts the stated intent of matching 'highly specific phrases that only appear when a user explicitly requests clarity mode' and may cause the skill to be unavailable for those exact invocations. _(1/3 passes — passes 2, 3 looked and disagreed)_

PR #1523, opened 2026-08-27
fix(rag): raise file_limit default 1000→100k, expose --file-limit CLI flag

TimeToBuildBob:
## Problem

The default `file_limit=1000` in `_get_valid_files` silently capped every
directory walk at 1,000 files. The truncation was logged at `WARNING` level
which is easy to miss in practice.

A real-world corpus with 7 index paths (journal: 31k files, knowledge: 5k,
tasks: 5k, projects: 15k, etc.) was being indexed to only ~7,000 documents
instead of ~57,000 — hitting the cap on every path. This caused the incremental
indexer to treat 97% of files as "new" on every run and burn 2+ hours of CPU
per invocation without ever completing.

## Changes

- **Raise default `file_limit` from 1,000 → 100,000** in `_get_valid_files` and
  `collect_documents` — safe for any corpus the embedder can reasonably handle
- **Escalate truncation log from `WARNING` → `ERROR`** — silent truncation is a
  correctness bug, not a tuning advisory
- **Add `--file-limit INTEGER` CLI flag** to `gptme_rag index` so callers can
  tune the cap without code changes (default 100,000)
- **Thread `file_limit` through `collect_documents` → `_get_valid_files`**

## Test

```bash
# Before: tasks/ would return 1000 (out of 5023)
# After: tasks/ returns all 5023
python3 -c "
import subprocess
r = subprocess.run(['git','-C','/path/to/repo/tasks','ls-files','--cached','--others','--exclude-standard'],capture_output=True,text=True)
print(len(r.stdout.splitlines()))  # 5023
"
```

### DIFF
--- a/packages/gptme-rag/src/gptme_rag/cli.py
+++ b/packages/gptme-rag/src/gptme_rag/cli.py
@@ -196,6 +196,12 @@ def cli(verbose: bool):
     default=None,
     help="Overlap between chunks. Defaults based on model: ModernBERT-msmarco=50, ModernBERT-base=200",
 )
+@click.option(
+    "--file-limit",
+    type=int,
+    default=100_000,
+    help="Maximum number of files to index per directory (default: 100000)",
+)
 def index(
     paths: list[Path],
     pattern: str,
@@ -205,6 +211,7 @@ def index(
     force_recreate: bool,
     chunk_size: int | None,
     chunk_overlap: int | None,
+    file_limit: int,
 ):
     """Index documents in one or more directories."""
     if not paths:
@@ -260,7 +267,7 @@ def index(
                 else:
                     status.update(f"Processing directory: {path}")
 
-                documents = indexer.collect_documents(path)
+                documents = indexer.collect_documents(path, file_limit=file_limit)
 
                 # Filter for new or modified documents
                 filtered_documents = []
--- a/packages/gptme-rag/src/gptme_rag/cli.py
+++ b/packages/gptme-rag/src/gptme_rag/cli.py
@@ -198,7 +198,7 @@ def cli(verbose: bool):
 )
 @click.option(
     "--file-limit",
-    type=int,
+    type=click.IntRange(min=0),
     default=100_000,
     help="Maximum number of files to index per directory (default: 100000)",
 )
--- a/packages/gptme-rag/src/gptme_rag/indexing/indexer.py
+++ b/packages/gptme-rag/src/gptme_rag/indexing/indexer.py
@@ -1650,8 +1650,9 @@ class Indexer:
             except Exception as e:
                 logger.warning(f"Error resolving symlink: {f} -> {e}")
 
-        # Check file limit
-        if len(valid_files) >= file_limit:
+        # Check file limit. Strict greater-than: a corpus that lands exactly on
+        # the cap is complete, not truncated, so it must not log an ERROR.
+        if len(valid_files) > file_limit:
             logger.error(
                 f"File limit ({file_limit}) reached, was {len(valid_files)}. "
                 f"Pass a higher --file-limit or use a more specific glob pattern than '{glob_pattern}'."
--- a/packages/gptme-rag/src/gptme_rag/indexing/indexer.py
+++ b/packages/gptme-rag/src/gptme_rag/indexing/indexer.py
@@ -1587,6 +1587,8 @@ class Indexer:
         Returns:
             Set of valid file paths
         """
+        if not isinstance(file_limit, int):
+            raise TypeError("file_limit must be an integer")
         if file_limit < 0:
             raise ValueError("file_limit must be >= 0")
 
--- a/packages/gptme-rag/src/gptme_rag/indexing/indexer.py
+++ b/packages/gptme-rag/src/gptme_rag/indexing/indexer.py
@@ -1587,6 +1587,9 @@ class Indexer:
         Returns:
             Set of valid file paths
         """
+        if file_limit < 0:
+            raise ValueError("file_limit must be >= 0")
+
         valid_files = set()
         path = path.resolve()  # Resolve path first
 
@@ -1657,7 +1660,10 @@ class Indexer:
                 f"File limit ({file_limit}) reached, was {len(valid_files)}. "
                 f"Pass a higher --file-limit or use a more specific glob pattern than '{glob_pattern}'."
             )
-            valid_files = set(list(valid_files)[:file_limit])
+            # Sort before slicing so truncation is deterministic. Set iteration
+            # order is arbitrary; without a sort, a corpus over the cap would
+            # index a different subset each run and never reach a stable state.
+            valid_files = set(sorted(valid_files)[:file_limit])
 
         return valid_files
 
--- a/packages/gptme-rag/src/gptme_rag/indexing/indexer.py
+++ b/packages/gptme-rag/src/gptme_rag/indexing/indexer.py
@@ -1575,7 +1575,7 @@ class Indexer:
             return False
 
     def _get_valid_files(
-        self, path: Path, glob_pattern: str = "**/*.*", file_limit: int = 1000
+        self, path: Path, glob_pattern: str = "**/*.*", file_limit: int = 100_000
     ) -> set[Path]:
         """Get valid files for indexing from a path.
 
@@ -1652,26 +1652,29 @@ class Indexer:
 
         # Check file limit
         if len(valid_files) >= file_limit:
-            logger.warning(
-                f"File limit ({file_limit}) reached, was {len(valid_files)}. Consider adding patterns to .gitignore "
-                f"or using a more specific glob pattern than '{glob_pattern}' to exclude unwanted files."
+            logger.error(
+                f"File limit ({file_limit}) reached, was {len(valid_files)}. "
+                f"Pass a higher --file-limit or use a more specific glob pattern than '{glob_pattern}'."
             )
             valid_files = set(list(valid_files)[:file_limit])
 
         return valid_files
 
-    def collect_documents(self, path: Path, glob_pattern: str = "**/*.*") -> list[Document]:
+    def collect_documents(
+        self, path: Path, glob_pattern: str = "**/*.*", file_limit: int = 100_000
+    ) -> list[Document]:
         """Collect documents from a file or directory without processing them.
 
         Args:
             path: Path to collect documents from
             glob_pattern: Pattern to match files (only used for directories)
+            file_limit: Maximum number of files to collect per directory
 
         Returns:
             List of documents ready for processing
         """
         documents: list[Document] = []
-        valid_files = self._get_valid_files(path, glob_pattern)
+        valid_files = self._get_valid_files(path, glob_pattern, file_limit)
 
         if not valid_files:
             logger.debug(f"No valid files found in {path}")
--- a/packages/gptme-rag/tests/test_file_limit.py
+++ b/packages/gptme-rag/tests/test_file_limit.py
@@ -96,9 +96,16 @@ def test_get_valid_files_rejects_negative_file_limit(tmp_path):
         indexer._get_valid_files(tmp_path, file_limit=-1)
 
 
-def test_index_cli_rejects_negative_file_limit():
+def test_get_valid_files_rejects_non_integer_file_limit(tmp_path):
+    _write_files(tmp_path, 3)
+    indexer = Indexer.__new__(Indexer)
+    with pytest.raises(TypeError, match="file_limit must be an integer"):
+        indexer._get_valid_files(tmp_path, file_limit=2.5)
+
+
+def test_index_cli_rejects_negative_file_limit(tmp_path):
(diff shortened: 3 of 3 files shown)


TimeToBuildBob:
@greptileai review

packages/gptme-rag/src/gptme_rag/indexing/indexer.py
```diff
@@ -1650,28 +1650,32 @@ def _get_valid_files(
...
-                f"File limit ({file_limit}) reached, was {len(valid_files)}. Consider adding patterns to .gitignore "
-                f"or using a more specific glob pattern than '{glob_pattern}' to exclude unwanted files."
+        # Check file limit. Strict greater-than: a corpus that lands exactly on
+        # the cap is complete, not truncated, so it must not log an ERROR.
+        if len(valid_files) > file_limit:
+            logger.error(
+                f"File limit ({file_limit}) reached, was {len(valid_files)}. "
+                f"Pass a higher --file-limit or use a more specific glob pattern than '{glob_pattern}'."
             )
             valid_files = set(list(valid_files)[:file_limit])
```
TimeToBuildBob: ⚠️ **P2** — The truncation at `valid_files = set(list(valid_files)[:file_limit])` is non-deterministic because it converts the set to a list and slices it, but set iteration order is arbitrary. When a directory exceeds file_limit, the subset of files that get indexed varies between runs. In an incremental indexer, this means some files may be indexed in one run and others in the next, and the index never reaches a complete state if the corpus always exceeds the limit. This can also cause files to be re-indexed unnecessarily if the same file is missed in one run but present in another. Fix by sorting the paths before slicing to make the selection deterministic.

packages/gptme-rag/src/gptme_rag/indexing/indexer.py
```diff
@@ -1650,28 +1650,32 @@ def _get_valid_files(
...
                 logger.warning(f"Error resolving symlink: {f} -> {e}")
 
-        # Check file limit
-        if len(valid_files) >= file_limit:
-            logger.warning(
-                f"File limit ({file_limit}) reached, was {len(valid_files)}. Consider adding patterns to .gitignore "
-                f"or using a more specific glob pattern than '{glob_pattern}' to exclude unwanted files."
+        # Check file limit. Strict greater-than: a corpus that lands exactly on
+        # the cap is complete, not truncated, so it must not log an ERROR.
+        if len(valid_files) > file_limit:
```
TimeToBuildBob, line 1660: ⚠️ **P2** — The CLI `--file-limit` option accepts any integer, including negative values. If a user passes `--file-limit -1`, the condition `len(valid_files) > file_limit` is always true and the code logs an ERROR and truncates with `list(valid_files)[:-1]`, which removes only the last element instead of keeping all but one. This is clearly not the intent of the flag and could lead to unexpected indexing behavior. The limit should be validated to be non-negative, or the option should use a Click type that enforces this.

TimeToBuildBob:
## 🤖 AI code review

This PR raises the per-directory file limit default from 1,000 to 100,000 in `Indexer._get_valid_files` and `collect_documents`, adds a `--file-limit` CLI option to the `index` command, and escalates the truncation log level from WARNING to ERROR. The file_limit is threaded through `collect_documents` to `_get_valid_files`, and the truncation is now deterministic by sorting before slicing. Tests are added for the new limit behavior and CLI flag.

**❌ P1 high** — `packages/gptme-rag/src/gptme_rag/cli.py:270`

The `pattern` CLI option in the `index` command is defined but never passed to `collect_documents`; the call at line 270 omits the `glob_pattern` argument, so the `--pattern` flag has no effect. Any user specifying a pattern like `*.md` still gets every file matching the default `**/*.*` instead of only the intended subset, which can cause the index to include unexpected files and bloat the index. The correct call is `collect_documents(path, pattern, file_limit=file_limit)`.

```suggestion
documents = indexer.collect_documents(path, pattern, file_limit=file_limit)
```

<sub>**How this was verified:** Checked the index function body; the `pattern` parameter appears in the signature but is never referenced elsewhere in the function, and the only call to collect_documents in the loop omits it.</sub>

**⚠️ P2 medium** — `packages/gptme-rag/src/gptme_rag/indexing/indexer.py:1662`

The truncation error message in _get_valid_files says "Pass a higher --file-limit", but this method is also reached indirectly through collect_documents called by index_directory (used by the watch command) and index_file, neither of which exposes a --file-limit CLI flag. When the cap is hit during a watch-driven initial indexing, the user sees an error instructing them to set a flag that does not exist on that command, making the diagnostic unactionable and misleading. The message should refer to the generic parameter (e.g. "raise file_limit") instead of a command-specific option, so it remains accurate for every caller.

<sub>**How this was verified:** collect_documents is called from index_directory (indexer.py:819) and index_file (indexer.py:1714) without any file_limit sourced from a CLI flag; watch uses index_directory, so this message is shown in contexts without --file-limit.</sub>

<sub>Consensus: 1/2 passes agreed — pass 1 looked and disagreed</sub> <br/><span style='color:gray;font-size:0.85em'>Distinct keys: 1 (<code>general</code>)</span>

</details>

- `packages/gptme-rag/src/gptme_rag/cli.py` — Adds --file-limit option to index command, accepts file_limit parameter, and passes it to collect_documents.
- `packages/gptme-rag/src/gptme_rag/indexing/indexer.py` — Changes default file_limit to 100_000, adds type/range checks, changes truncation condition to strict greater-than, and sorts before truncation for determinism.
- `packages/gptme-rag/tests/test_file_limit.py` — Adds tests for file_limit truncation, CLI flag, determinism, and rejection of invalid values.

</details>

| `01a9961c3d56` | 4/5 | 2 | llm | 2026-08-27 12:30 UTC |
| `52fec1b91fc7` | 3/5 | 2 | llm | 2026-08-27 14:41 UTC |

</details>

<sub>Reviewed `58e8c7c5bcf6` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 95s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

packages/gptme-rag/src/gptme_rag/indexing/indexer.py
TimeToBuildBob: Fixed in 52fec1b9 — truncation now does `set(sorted(valid_files)[:file_limit])` so the kept subset is lexicographically stable across runs. Covered by `test_get_valid_files_truncation_is_deterministic`.

packages/gptme-rag/src/gptme_rag/indexing/indexer.py
TimeToBuildBob: Fixed in 52fec1b9 — `_get_valid_files` raises `ValueError` on `file_limit < 0`, and the CLI option is now `click.IntRange(min=0)`. Covered by `test_get_valid_files_rejects_negative_file_limit` and `test_index_cli_rejects_negative_file_limit`.

TimeToBuildBob:
CI is green and the two review findings are fixed at `52fec1b9`, but self-merge remains ineligible: Greptile never produced an initial review, and the guarded helper refuses a second trigger after the existing request. No merge was attempted; this now needs human review/merge or a Greptile-side retry.

packages/gptme-rag/src/gptme_rag/indexing/indexer.py
```diff
@@ -1650,28 +1653,35 @@ def _get_valid_files(
...
+            logger.error(
+                f"File limit ({file_limit}) reached, was {len(valid_files)}. "
+                f"Pass a higher --file-limit or use a more specific glob pattern than '{glob_pattern}'."
             )
-            valid_files = set(list(valid_files)[:file_limit])
+            # Sort before slicing so truncation is deterministic. Set iteration
+            # order is arbitrary; without a sort, a corpus over the cap would
+            # index a different subset each run and never reach a stable state.
+            valid_files = set(sorted(valid_files)[:file_limit])
```
TimeToBuildBob, line 1668: ❌ **P1** — The _get_valid_files method validates that file_limit is not negative but does not check that it is an integer. If a caller passes a non-integer numeric value such as 2.5, the comparison `len(valid_files) > file_limit` may be true, and the subsequent slice `sorted(valid_files)[:file_limit]` raises a TypeError because slice indices must be integers. This crashes the indexing operation with an unhandled exception. The parameter is documented as 'Maximum number of files' implying an integer, and the CLI enforces this via IntRange, but any direct library caller passing a float will trigger the crash. Better to explicitly validate that file_limit is an int and raise a clear error otherwise.

packages/gptme-rag/tests/test_file_limit.py
```diff
@@ -0,0 +1,104 @@
...
+def test_get_valid_files_rejects_negative_file_limit(tmp_path):
+    _write_files(tmp_path, 3)
+    indexer = Indexer.__new__(Indexer)
+    with pytest.raises(ValueError, match="file_limit must be >= 0"):
+        indexer._get_valid_files(tmp_path, file_limit=-1)
+
+
+def test_index_cli_rejects_negative_file_limit():
+    runner = CliRunner()
+    result = runner.invoke(cli, ["index", "--file-limit", "-1", "/tmp"])
```
TimeToBuildBob: ⚠️ **P2** — The test test_index_cli_rejects_negative_file_limit uses a hardcoded path '/tmp' which may not exist on all systems (e.g., Windows without a /tmp directory). If /tmp does not exist, the test might fail for an unrelated reason, making it non-hermetic. The test should use pytest's tmp_path fixture to create a temporary directory. Since the test only checks that the CLI rejects a negative value, the actual directory existence is irrelevant, but relying on /tmp makes the test environment-dependent.

packages/gptme-rag/src/gptme_rag/indexing/indexer.py
TimeToBuildBob: Fixed in 58e8c7c5 — `_get_valid_files` now explicitly rejects non-integer limits with a clear `TypeError`, covered by `test_get_valid_files_rejects_non_integer_file_limit`.

packages/gptme-rag/tests/test_file_limit.py
TimeToBuildBob: Fixed in 58e8c7c5 — the CLI test now uses pytest's `tmp_path` fixture instead of hardcoding `/tmp`.

packages/gptme-rag/src/gptme_rag/cli.py
```diff
@@ -260,7 +267,7 @@ def index(
                 else:
                     status.update(f"Processing directory: {path}")
 
-                documents = indexer.collect_documents(path)
+                documents = indexer.collect_documents(path, file_limit=file_limit)
```
TimeToBuildBob, line 270: ❌ **P1** — The `pattern` CLI option in the `index` command is defined but never passed to `collect_documents`; the call at line 270 omits the `glob_pattern` argument, so the `--pattern` flag has no effect. Any user specifying a pattern like `*.md` still gets every file matching the default `**/*.*` instead of only the intended subset, which can cause the index to include unexpected files and bloat the index. The correct call is `collect_documents(path, pattern, file_limit=file_limit)`. _(2/2 passes)_

packages/gptme-rag/src/gptme_rag/indexing/indexer.py
```diff
@@ -1650,28 +1655,35 @@ def _get_valid_files(
...
-        # Check file limit
-        if len(valid_files) >= file_limit:
-            logger.warning(
-                f"File limit ({file_limit}) reached, was {len(valid_files)}. Consider adding patterns to .gitignore "
-                f"or using a more specific glob pattern than '{glob_pattern}' to exclude unwanted files."
+        # Check file limit. Strict greater-than: a corpus that lands exactly on
+        # the cap is complete, not truncated, so it must not log an ERROR.
+        if len(valid_files) > file_limit:
+            logger.error(
+                f"File limit ({file_limit}) reached, was {len(valid_files)}. "
```
TimeToBuildBob, line 1662: ⚠️ **P2** — The truncation error message in _get_valid_files says "Pass a higher --file-limit", but this method is also reached indirectly through collect_documents called by index_directory (used by the watch command) and index_file, neither of which exposes a --file-limit CLI flag. When the cap is hit during a watch-driven initial indexing, the user sees an error instructing them to set a flag that does not exist on that command, making the diagnostic unactionable and misleading. The message should refer to the generic parameter (e.g. "raise file_limit") instead of a command-specific option, so it remains accurate for every caller. _(1/2 passes — pass 1 looked and disagreed)_

ErikBjare:
@TimeToBuildBob Fix issues in follow-up

TimeToBuildBob:
Fixed both final-review findings in follow-up gptme/gptme-contrib#1532:

- `index --pattern` is now passed through to `collect_documents`, with a regression test.
- The file-limit diagnostic now refers to the caller-neutral `file_limit` parameter instead of a CLI flag unavailable to every caller.

The targeted gptme-rag tests pass (10/10), plus Ruff and targeted mypy.

PR #1524, opened 2026-08-27
feat(voice-node): add gptme-voice-node embedded WS client package

TimeToBuildBob:
## Summary

Implements milestone 1 of the BobBrain portable presence node ([ErikBjare/bob#730](https://github.com/ErikBjare/bob/issues/730)).

A thin (~200 LOC) WebSocket client that connects to `bob-voice-server`, streams mic audio, and plays back PCM audio — designed for unattended embedded operation on a Raspberry Pi with a USB microphone array (e.g. ReSpeaker XVF3800).

## Design choices

- **Env-var config only** (`GPTME_VOICE_NODE_SERVER`, `GPTME_VOICE_NODE_NAME`) — systemd-friendly, no CLI argument parsing needed for embedded use
- **Exponential backoff reconnect** (1s → 60s max) — survives network blips and server restarts autonomously
- **Mic muted during playback + configurable cooldown** — echo/feedback prevention on nodes without hardware AEC; hardware AEC (ReSpeaker XVF3800) handles it at the chip level but the software fallback is free
- **PyAudio streams open/close per-session** — clean OS-level release on disconnect; important on embedded Linux where ALSA device handles are scarce
- **Blocking audio I/O in executor** — keeps the event loop responsive during `asyncio.gather` of send+recv tasks

## Files

```
packages/gptme-voice-node/
├── pyproject.toml          # dependencies: pyaudio, websockets
├── README.md               # quick-start + Pi deployment guide
├── src/gptme_voice_node/
│   ├── __init__.py
│   └── node.py             # VoiceNode + main() entry point
├── systemd/
│   └── gptme-voice-node.service   # drop-in override pattern
└── tests/
    └── test_voice_node.py  # 9 tests, no hardware required
```

## Testing

9 unit tests, all green:
- Mic mute logic (4 tests)
- Recv loop audio frame handling (2 tests)
- Lifecycle: stop/cleanup (2 tests)
- Backoff: run_forever exits when stopped (1 test)

```
uv run --extra test pytest packages/gptme-voice-node/tests/ -v
# 9 passed in 0.07s
```

## Related

- Spec: [`bobbrain-spec.md`](https://github.com/ErikBjare/bob/blob/master/knowledge/technical-designs/bobbrain-spec.md)
- Spec: [`bobbody-voice-node-spec.md`](https://github.com/ErikBjare/bob/blob/master/knowledge/technical-designs/bobbody-voice-node-spec.md)
- Reference client: `packages/gptme-voice/src/gptme_voice/realtime/client.py`

### DIFF
--- a/packages/gptme-voice-node/src/gptme_voice_node/node.py
+++ b/packages/gptme-voice-node/src/gptme_voice_node/node.py
@@ -93,8 +93,10 @@ def _close_stream(stream, started: bool) -> None:
             stream.stop_stream()
     except Exception as exc:  # noqa: BLE001
         log.warning("failed to stop audio stream: %s", exc)
-    finally:
+    try:
         stream.close()
+    except Exception as exc:  # noqa: BLE001
+        log.warning("failed to close audio stream: %s", exc)
 
 
 def _next_backoff(backoff: int, connected_at: float | None) -> int:
@@ -129,7 +131,8 @@ class VoiceNode:
     def _mic_muted(self) -> bool:
         """True when we should suppress outbound audio (echo prevention)."""
         return self._playing or (
-            time.monotonic() - self._play_ended_at < PLAYBACK_COOLDOWN
+            self._play_ended_at > 0
+            and time.monotonic() - self._play_ended_at < PLAYBACK_COOLDOWN
         )
 
     # --- Per-session connect handler ---
@@ -227,6 +230,7 @@ class VoiceNode:
         backoff = BACKOFF_INITIAL
         while self._running:
             connected_at: float | None = None
+            error: tuple[int, str] | None = None
             try:
                 async with websockets.connect(
                     self.server_url,
@@ -237,28 +241,22 @@ class VoiceNode:
                     connected_at = time.monotonic()
                     await self._session(ws)
             except (ConnectionClosed, WebSocketException) as exc:
-                log.warning(
-                    "[%s] disconnected: %s — retrying in %ds",
-                    self.node_name,
-                    exc,
-                    backoff,
-                )
+                error = (logging.WARNING, f"disconnected: {exc}")
             except OSError as exc:
-                log.warning(
-                    "[%s] connection error: %s — retrying in %ds",
-                    self.node_name,
-                    exc,
-                    backoff,
-                )
+                error = (logging.WARNING, f"connection error: {exc}")
             except Exception as exc:  # noqa: BLE001
-                log.error(
-                    "[%s] unexpected error: %s — retrying in %ds",
-                    self.node_name,
-                    exc,
-                    backoff,
-                )
+                error = (logging.ERROR, f"unexpected error: {exc}")
             if self._running:
                 backoff = _next_backoff(backoff, connected_at)
+                if error is not None:
+                    level, message = error
+                    log.log(
+                        level,
+                        "[%s] %s — retrying in %ds",
+                        self.node_name,
+                        message,
+                        backoff,
+                    )
                 await asyncio.sleep(backoff)
 
     def stop(self) -> None:
@@ -294,6 +292,20 @@ async def _async_main(server_url: str, node_name: str) -> None:
         node.cleanup()
 
 
+def _redact_url_userinfo(server_url: str) -> str:
+    """Hide credentials from a URL before writing it to logs."""
+    parsed = urlparse(server_url)
+    if parsed.username is None and parsed.password is None:
+        return server_url
+
+    hostname = parsed.hostname or ""
+    if ":" in hostname:
+        hostname = f"[{hostname}]"
+    if parsed.port is not None:
+        hostname = f"{hostname}:{parsed.port}"
+    return parsed._replace(netloc=f"***@{hostname}").geturl()
+
+
 def _warn_for_insecure_remote_url(server_url: str) -> None:
     """Warn when microphone audio would cross the network without TLS."""
     parsed = urlparse(server_url)
@@ -323,7 +335,7 @@ def main() -> None:
     _warn_for_insecure_remote_url(server_url)
 
     log.info("Starting BobBrain voice node")
-    log.info("  server : %s", server_url)
+    log.info("  server : %s", _redact_url_userinfo(server_url))
     log.info("  name   : %s", node_name)
 
     try:
--- a/packages/gptme-voice-node/src/gptme_voice_node/node.py
+++ b/packages/gptme-voice-node/src/gptme_voice_node/node.py
@@ -72,8 +72,19 @@ PLAYBACK_COOLDOWN = 0.5
 # Reconnect backoff (seconds)
 BACKOFF_INITIAL = 1
 BACKOFF_MAX = 60
+BACKOFF_RESET_AFTER = 30
 
 log = logging.getLogger("gptme_voice_node")
+
+
+def _backoff_after_session(backoff: int, connected_at: float | None) -> int:
+    """Reset retry delay only after a connection stayed healthy long enough."""
+    if (
+        connected_at is not None
+        and time.monotonic() - connected_at >= BACKOFF_RESET_AFTER
+    ):
+        return BACKOFF_INITIAL
+    return backoff
 
 
 class VoiceNode:
@@ -119,6 +130,7 @@ class VoiceNode:
             rate=SAMPLE_RATE,
             input=True,
             frames_per_buffer=CHUNK_SIZE,
+            start=False,
         )
         try:
             speaker = self._pa.open(
@@ -127,8 +139,11 @@ class VoiceNode:
                 rate=SAMPLE_RATE,
                 output=True,
                 frames_per_buffer=CHUNK_SIZE,
+                start=False,
             )
             try:
+                mic.start_stream()
+                speaker.start_stream()
                 await asyncio.gather(
                     self._send_loop(ws, mic),
                     self._recv_loop(ws, speaker),
@@ -187,6 +202,7 @@ class VoiceNode:
         """Connect and auto-reconnect with exponential backoff."""
         backoff = BACKOFF_INITIAL
         while self._running:
+            connected_at: float | None = None
             try:
                 async with websockets.connect(
                     self.server_url,
@@ -194,9 +210,10 @@ class VoiceNode:
                     ping_interval=20,
                     ping_timeout=20,
                 ) as ws:
-                    backoff = BACKOFF_INITIAL  # Reset on success
+                    connected_at = time.monotonic()
                     await self._session(ws)
             except (ConnectionClosed, WebSocketException) as exc:
+                backoff = _backoff_after_session(backoff, connected_at)
                 log.warning(
                     "[%s] disconnected: %s — retrying in %ds",
                     self.node_name,
@@ -204,6 +221,7 @@ class VoiceNode:
                     backoff,
                 )
             except OSError as exc:
+                backoff = _backoff_after_session(backoff, connected_at)
                 log.warning(
                     "[%s] connection error: %s — retrying in %ds",
                     self.node_name,
@@ -211,6 +229,7 @@ class VoiceNode:
                     backoff,
                 )
             except Exception as exc:  # noqa: BLE001
+                backoff = _backoff_after_session(backoff, connected_at)
                 log.error(
                     "[%s] unexpected error: %s — retrying in %ds",
                     self.node_name,
--- a/packages/gptme-voice-node/src/gptme_voice_node/node.py
+++ b/packages/gptme-voice-node/src/gptme_voice_node/node.py
@@ -25,14 +25,6 @@ import sys
 import time
 
 try:
-    import pyaudio
-except ImportError as e:
-    raise ImportError(
-        "pyaudio is required. Install with: pip install pyaudio\n"
-        "  On Debian/Ubuntu: sudo apt install portaudio19-dev python3-pyaudio"
-    ) from e
-
-try:
     import websockets
     from websockets.exceptions import ConnectionClosed, WebSocketException
 except ImportError as e:
@@ -48,7 +40,28 @@ NODE_NAME = os.environ.get("GPTME_VOICE_NODE_NAME", "bobbrain-unknown")
 SAMPLE_RATE = 24000
 CHUNK_SIZE = 1024
 CHANNELS = 1
-FORMAT = pyaudio.paInt16
+
+# Lazy import: pyaudio is only required when actually instantiating VoiceNode
+_pyaudio = None
+
+
+def _get_pyaudio():
+    """Import pyaudio on demand, raising a helpful error if not installed."""
+    global _pyaudio
+    if _pyaudio is None:
+        try:
+            import pyaudio
+
+            _pyaudio = pyaudio
+        except ImportError as e:
+            raise ImportError(
+                "pyaudio is required. Install with: pip install 'gptme-voice-node[audio]'\n"
+                "  On Debian/Ubuntu: sudo apt install portaudio19-dev python3-pyaudio"
+            ) from e
+    return _pyaudio
+
+
+FORMAT = None  # Will be set to _get_pyaudio().paInt16 in VoiceNode.__init__
 
 # Cooldown window (seconds) to keep mic muted after playback ends.
 # Prevents echo/feedback on nodes without hardware AEC.
@@ -71,7 +84,9 @@ class VoiceNode:
     def __init__(self, server_url: str, node_name: str) -> None:
         self.server_url = server_url
         self.node_name = node_name
+        pyaudio = _get_pyaudio()  # May raise ImportError if not installed
         self._pa = pyaudio.PyAudio()
+        self._audio_format = pyaudio.paInt16
         self._playing = False
         self._play_ended_at = 0.0
         self._running = True
@@ -91,14 +106,14 @@ class VoiceNode:
         log.info("[%s] connected to %s", self.node_name, self.server_url)
 
         mic = self._pa.open(
-            format=FORMAT,
+            format=self._audio_format,
             channels=CHANNELS,
             rate=SAMPLE_RATE,
             input=True,
             frames_per_buffer=CHUNK_SIZE,
         )
         speaker = self._pa.open(
-            format=FORMAT,
+            format=self._audio_format,
             channels=CHANNELS,
             rate=SAMPLE_RATE,
             output=True,
--- a/packages/gptme-voice-node/src/gptme_voice_node/node.py
+++ b/packages/gptme-voice-node/src/gptme_voice_node/node.py
@@ -145,7 +145,11 @@ class VoiceNode:
         self._playing = False
         self._play_ended_at = 0.0
 
-        log.info("[%s] connected to %s", self.node_name, self.server_url)
+        log.info(
+            "[%s] connected to %s",
+            self.node_name,
+            _redact_url_userinfo(self.server_url),
+        )
 
         mic = self._pa.open(
             format=self._audio_format,
--- a/packages/gptme-voice-node/src/gptme_voice_node/node.py
+++ b/packages/gptme-voice-node/src/gptme_voice_node/node.py
@@ -21,6 +21,7 @@ import functools
 import json
 import logging
 import os
+import re
 import signal
 import sys
 import time
@@ -175,10 +176,15 @@ class VoiceNode:
                 mic_started = True
                 speaker.start_stream()
                 speaker_started = True
-                await asyncio.gather(
-                    self._send_loop(ws, mic),
-                    self._recv_loop(ws, speaker),
-                )
+                send_task = asyncio.create_task(self._send_loop(ws, mic))
+                recv_task = asyncio.create_task(self._recv_loop(ws, speaker))
+                try:
+                    await asyncio.gather(send_task, recv_task)
+                finally:
+                    for task in (send_task, recv_task):
+                        if not task.done():
+                            task.cancel()
+                    await asyncio.gather(send_task, recv_task, return_exceptions=True)
             finally:
                 _close_stream(speaker, speaker_started)
         finally:
@@ -245,11 +251,20 @@ class VoiceNode:
                     connected_at = time.monotonic()
                     await self._session(ws)
             except (ConnectionClosed, WebSocketException) as exc:
-                error = (logging.WARNING, f"disconnected: {exc}")
+                error = (
+                    logging.WARNING,
+                    f"disconnected: {_redact_userinfo_in_text(str(exc), self.server_url)}",
+                )
             except OSError as exc:
-                error = (logging.WARNING, f"connection error: {exc}")
+                error = (
+                    logging.WARNING,
+                    f"connection error: {_redact_userinfo_in_text(str(exc), self.server_url)}",
+                )
             except Exception as exc:  # noqa: BLE001
-                error = (logging.ERROR, f"unexpected error: {exc}")
+                error = (
+                    logging.ERROR,
+                    f"unexpected error: {_redact_userinfo_in_text(str(exc), self.server_url)}",
+                )
             if self._running:
                 backoff = _next_backoff(backoff, connected_at)
                 if error is not None:
@@ -296,6 +311,9 @@ async def _async_main(server_url: str, node_name: str) -> None:
         node.cleanup()
 
 
+_USERINFO_IN_TEXT = re.compile(r"(://)([^/@\s]+)@")
+
+
 def _redact_url_userinfo(server_url: str) -> str:
     """Hide credentials from a URL before writing it to logs."""
     parsed = urlparse(server_url)
@@ -308,6 +326,15 @@ def _redact_url_userinfo(server_url: str) -> str:
     if parsed.port is not None:
         hostname = f"{hostname}:{parsed.port}"
     return parsed._replace(netloc=f"***@{hostname}").geturl()
+
+
+def _redact_userinfo_in_text(text: str, server_url: str = "") -> str:
+    """Hide URL userinfo in free-form log/exception text."""
+    if server_url:
+        redacted = _redact_url_userinfo(server_url)
+        if redacted != server_url:
+            text = text.replace(server_url, redacted)
+    return _USERINFO_IN_TEXT.sub(r"\1***@", text)
 
 
 def _warn_for_insecure_remote_url(server_url: str) -> None:
--- a/packages/gptme-voice-node/src/gptme_voice_node/node.py
+++ b/packages/gptme-voice-node/src/gptme_voice_node/node.py
@@ -104,6 +104,12 @@ class VoiceNode:
 
     async def _session(self, ws) -> None:
         """Handle one connected WebSocket session."""
+        # Reset per-session playback state on every reconnect.
+        # Without this, a disconnect mid-playback leaves _playing=True, which
+        # silently mutes the mic for the entire next session.
+        self._playing = False
+        self._play_ended_at = 0.0
+
         log.info("[%s] connected to %s", self.node_name, self.server_url)
 
         mic = self._pa.open(
@@ -113,24 +119,25 @@ class VoiceNode:
             input=True,
             frames_per_buffer=CHUNK_SIZE,
         )
-        speaker = self._pa.open(
-            format=self._audio_format,
-            channels=CHANNELS,
-            rate=SAMPLE_RATE,
-            output=True,
-            frames_per_buffer=CHUNK_SIZE,
-        )
-
         try:
-            await asyncio.gather(
-                self._send_loop(ws, mic),
-                self._recv_loop(ws, speaker),
+            speaker = self._pa.open(
+                format=self._audio_format,
+                channels=CHANNELS,
+                rate=SAMPLE_RATE,
+                output=True,
+                frames_per_buffer=CHUNK_SIZE,
             )
+            try:
+                await asyncio.gather(
+                    self._send_loop(ws, mic),
+                    self._recv_loop(ws, speaker),
+                )
+            finally:
+                speaker.stop_stream()
+                speaker.close()
         finally:
             mic.stop_stream()
             mic.close()
-            speaker.stop_stream()
-            speaker.close()
 
     async def _send_loop(self, ws, mic_stream) -> None:
         """Continuously read mic and forward to server (unless muted)."""
@@ -148,9 +155,20 @@ class VoiceNode:
             await asyncio.sleep(0.005)
 
     async def _recv_loop(self, ws, speaker_stream) -> None:
-        """Receive audio frames from server and play them."""
+        """Receive audio frames from server and play them.
+
+        Uses explicit ws.recv() rather than ``async for`` so that the loop
+        respects the ``_running`` flag and exits cleanly on shutdown instead
+        of blocking indefinitely waiting for the next message.  This prevents
+        run_forever from hanging and leaking PyAudio resources when the node
+        is stopped via SIGTERM.
+        """
         loop = asyncio.get_event_loop()
-        async for raw_msg in ws:
+        while self._running:
+            try:
+                raw_msg = await ws.recv()
+            except ConnectionClosed:
+                break
             data = json.loads(raw_msg)
             msg_type = data.get("type")
             if msg_type == "audio":
@@ -215,16 +233,22 @@ async def _async_main(server_url: str, node_name: str) -> None:
     node = VoiceNode(server_url, node_name)
 
     loop = asyncio.get_event_loop()
+    task = asyncio.ensure_future(node.run_forever())
 
     def _handle_signal(sig: signal.Signals) -> None:
         log.info("[%s] received signal %s, shutting down", node_name, sig)
         node.stop()
+        # Cancel the task so any blocked ws.recv() unblocks immediately and
+        # the cleanup finally block runs without waiting for SIGKILL.
+        task.cancel()
 
     for sig in (signal.SIGINT, signal.SIGTERM):
         loop.add_signal_handler(sig, functools.partial(_handle_signal, sig))
 
     try:
-        await node.run_forever()
+        await task
+    except asyncio.CancelledError:
+        pass
     finally:
         node.cleanup()
 
--- a/packages/gptme-voice-node/src/gptme_voice_node/node.py
+++ b/packages/gptme-voice-node/src/gptme_voice_node/node.py
@@ -17,6 +17,7 @@ Protocol:
 
 import asyncio
 import base64
+import functools
 import json
 import logging
 import os
@@ -215,12 +216,12 @@ async def _async_main(server_url: str, node_name: str) -> None:
 
     loop = asyncio.get_event_loop()
 
-    def _handle_signal(sig):
+    def _handle_signal(sig: signal.Signals) -> None:
         log.info("[%s] received signal %s, shutting down", node_name, sig)
         node.stop()
 
     for sig in (signal.SIGINT, signal.SIGTERM):
-        loop.add_signal_handler(sig, lambda s=sig: _handle_signal(s))
+        loop.add_signal_handler(sig, functools.partial(_handle_signal, sig))
 
     try:
         await node.run_forever()
--- a/packages/gptme-voice-node/src/gptme_voice_node/node.py
+++ b/packages/gptme-voice-node/src/gptme_voice_node/node.py
@@ -1,0 +1,240 @@
+"""
+BobBrain voice node — thin WS client for embedded Linux.
+
+Auto-reconnects to bob-voice-server with exponential backoff.
+Low memory footprint, systemd-managed lifecycle.
+
+Config (env vars):
+  GPTME_VOICE_NODE_SERVER  — WebSocket URL of bob-voice-server (default: ws://localhost:8080/local)
+  GPTME_VOICE_NODE_NAME    — Node identity string (default: bobbrain-unknown)
+
+Protocol:
+  Client → Server: {"type": "audio", "audio": "<base64 PCM 16-bit 24kHz mono>"}
+  Client → Server: {"type": "commit"}      (optional flush signal)
+  Server → Client: {"type": "audio", "audio": "<base64>"}
+  Server → Client: {"type": "audio_end"}
+"""
+
+import asyncio
+import base64
+import json
+import logging
+import os
+import signal
+import sys
+import time
+
+try:
+    import pyaudio
+except ImportError as e:
+    raise ImportError(
+        "pyaudio is required. Install with: pip install pyaudio\n"
+        "  On Debian/Ubuntu: sudo apt install portaudio19-dev python3-pyaudio"
+    ) from e
+
+try:
+    import websockets
+    from websockets.exceptions import ConnectionClosed, WebSocketException
+except ImportError as e:
+    raise ImportError(
+        "websockets is required. Install with: pip install 'websockets>=12.0'"
+    ) from e
+
+# --- Config ---
+SERVER_URL = os.environ.get("GPTME_VOICE_NODE_SERVER", "ws://localhost:8080/local")
+NODE_NAME = os.environ.get("GPTME_VOICE_NODE_NAME", "bobbrain-unknown")
+
+# --- Audio constants (must match server) ---
+SAMPLE_RATE = 24000
+CHUNK_SIZE = 1024
+CHANNELS = 1
+FORMAT = pyaudio.paInt16
+
+# Cooldown window (seconds) to keep mic muted after playback ends.
+# Prevents echo/feedback on nodes without hardware AEC.
+PLAYBACK_COOLDOWN = 0.5
+
+# Reconnect backoff (seconds)
+BACKOFF_INITIAL = 1
+BACKOFF_MAX = 60
+
+log = logging.getLogger("gptme_voice_node")
+
+
+class VoiceNode:
+    """Thin embedded voice client.
+
+    Manages one PyAudio instance across reconnects.  Audio streams are opened
+    and closed per-session so the OS sees a clean release on disconnect.
+    """
+
+    def __init__(self, server_url: str, node_name: str) -> None:
+        self.server_url = server_url
+        self.node_name = node_name
+        self._pa = pyaudio.PyAudio()
+        self._playing = False
+        self._play_ended_at = 0.0
+        self._running = True
+
+    # --- Mic mute logic ---
+
+    def _mic_muted(self) -> bool:
+        """True when we should suppress outbound audio (echo prevention)."""
+        return self._playing or (
+            time.monotonic() - self._play_ended_at < PLAYBACK_COOLDOWN
+        )
+
+    # --- Per-session connect handler ---
+
+    async def _session(self, ws) -> None:
+        """Handle one connected WebSocket session."""
+        log.info("[%s] connected to %s", self.node_name, self.server_url)
+
+        mic = self._pa.open(
+            format=FORMAT,
+            channels=CHANNELS,
+            rate=SAMPLE_RATE,
+            input=True,
+            frames_per_buffer=CHUNK_SIZE,
+        )
+        speaker = self._pa.open(
+            format=FORMAT,
+            channels=CHANNELS,
+            rate=SAMPLE_RATE,
+            output=True,
+            frames_per_buffer=CHUNK_SIZE,
+        )
+
+        try:
+            await asyncio.gather(
+                self._send_loop(ws, mic),
+                self._recv_loop(ws, speaker),
+            )
+        finally:
+            mic.stop_stream()
+            mic.close()
+            speaker.stop_stream()
+            speaker.close()
+
+    async def _send_loop(self, ws, mic_stream) -> None:
+        """Continuously read mic and forward to server (unless muted)."""
+        loop = asyncio.get_event_loop()
+        while self._running:
+            # Read in executor to avoid blocking the event loop
+            raw = await loop.run_in_executor(
+                None,
+                lambda: mic_stream.read(CHUNK_SIZE, exception_on_overflow=False),
+            )
+            if not self._mic_muted():
+                msg = {"type": "audio", "audio": base64.b64encode(raw).decode()}
+                await ws.send(json.dumps(msg))
+            # Yield to other tasks briefly
+            await asyncio.sleep(0.005)
+
+    async def _recv_loop(self, ws, speaker_stream) -> None:
+        """Receive audio frames from server and play them."""
+        loop = asyncio.get_event_loop()
+        async for raw_msg in ws:
+            data = json.loads(raw_msg)
+            msg_type = data.get("type")
+            if msg_type == "audio":
+                self._playing = True
+                pcm = base64.b64decode(data["audio"])
+                await loop.run_in_executor(None, lambda: speaker_stream.write(pcm))
+            elif msg_type == "audio_end":
+                self._playing = False
+                self._play_ended_at = time.monotonic()
+                log.debug("[%s] audio_end received", self.node_name)
+
+    # --- Reconnect loop ---
+
+    async def run_forever(self) -> None:
+        """Connect and auto-reconnect with exponential backoff."""
+        backoff = BACKOFF_INITIAL
+        while self._running:
+            try:
+                async with websockets.connect(
+                    self.server_url,
+                    open_timeout=10,
+                    ping_interval=20,
+                    ping_timeout=20,
+                ) as ws:
+                    backoff = BACKOFF_INITIAL  # Reset on success
+                    await self._session(ws)
+            except (ConnectionClosed, WebSocketException) as exc:
+                log.warning(
+                    "[%s] disconnected: %s — retrying in %ds",
+                    self.node_name,
+                    exc,
+                    backoff,
+                )
+            except OSError as exc:
+                log.warning(
+                    "[%s] connection error: %s — retrying in %ds",
+                    self.node_name,
+                    exc,
+                    backoff,
+                )
+            except Exception as exc:  # noqa: BLE001
+                log.error(
+                    "[%s] unexpected error: %s — retrying in %ds",
+                    self.node_name,
+                    exc,
+                    backoff,
+                )
+            if self._running:
+                await asyncio.sleep(backoff)
+                backoff = min(backoff * 2, BACKOFF_MAX)
+
+    def stop(self) -> None:
+        """Signal the node to shut down cleanly."""
+        self._running = False
+
+    def cleanup(self) -> None:
+        """Release PyAudio resources."""
+        self._pa.terminate()
+
+
+async def _async_main(server_url: str, node_name: str) -> None:
+    node = VoiceNode(server_url, node_name)
+
+    loop = asyncio.get_event_loop()
+
+    def _handle_signal(sig):
+        log.info("[%s] received signal %s, shutting down", node_name, sig)
+        node.stop()
+
+    for sig in (signal.SIGINT, signal.SIGTERM):
+        loop.add_signal_handler(sig, lambda s=sig: _handle_signal(s))
+
+    try:
+        await node.run_forever()
+    finally:
+        node.cleanup()
+
+
+def main() -> None:
+    """Entry point for the gptme-voice-node CLI."""
+    logging.basicConfig(
+        level=logging.INFO,
+        format="%(asctime)s %(levelname)s %(message)s",
+        datefmt="%Y-%m-%dT%H:%M:%S",
+        stream=sys.stderr,
+    )
+
+    server_url = SERVER_URL
+    node_name = NODE_NAME
+
+    log.info("Starting BobBrain voice node")
+    log.info("  server : %s", server_url)
+    log.info("  name   : %s", node_name)
+
+    try:
+        asyncio.run(_async_main(server_url, node_name))
+    except KeyboardInterrupt:
+        pass
+    log.info("[%s] stopped", node_name)
+
+
+if __name__ == "__main__":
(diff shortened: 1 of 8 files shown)


TimeToBuildBob:
Fixed the pre-commit mypy failures (commit `f4d5071f`):

- **websockets import-not-found**: Added `websockets>=12.0` to mypy `additional_dependencies` in `.pre-commit-config.yaml` (websockets has inline type stubs since v10+).
- **pyaudio import-untyped**: Added `types-pyaudio` to mypy deps — stubs now available, no `# type: ignore` needed.
- **Cannot infer type of lambda**: Replaced `lambda s=sig: _handle_signal(s)` with `functools.partial(_handle_signal, sig)` and annotated `_handle_signal(sig: signal.Signals) -> None`.

All pre-commit hooks pass locally. CI queued.

TimeToBuildBob:
@greptileai review

packages/gptme-voice-node/src/gptme_voice_node/node.py
```diff
@@ -0,0 +1,256 @@
...
+            await asyncio.sleep(0.005)
+
+    async def _recv_loop(self, ws, speaker_stream) -> None:
+        """Receive audio frames from server and play them."""
+        loop = asyncio.get_event_loop()
+        async for raw_msg in ws:
+            data = json.loads(raw_msg)
+            msg_type = data.get("type")
+            if msg_type == "audio":
+                self._playing = True
```
TimeToBuildBob: ❌ **P1** — When a session ends without receiving `audio_end` (e.g., the server disconnects during playback), `self._playing` remains `True`. Because `_playing` and `_play_ended_at` are never reset at the start of `_session`, the next reconnect begins with `_playing` still `True`. `_mic_muted()` then returns `True` and `_send_loop` drops every mic chunk, so the node is silently muted until the server happens to send an `audio_end` in a later session — which it may never do. This renders the node unresponsive to voice after any mid-playback disconnect. Reset `_playing = False` and `_play_ended_at = 0.0` at the top of `_session`.

packages/gptme-voice-node/src/gptme_voice_node/node.py
```diff
@@ -0,0 +1,256 @@
...
+            if not self._mic_muted():
+                msg = {"type": "audio", "audio": base64.b64encode(raw).decode()}
+                await ws.send(json.dumps(msg))
+            # Yield to other tasks briefly
+            await asyncio.sleep(0.005)
+
+    async def _recv_loop(self, ws, speaker_stream) -> None:
+        """Receive audio frames from server and play them."""
+        loop = asyncio.get_event_loop()
+        async for raw_msg in ws:
```
TimeToBuildBob: ❌ **P1** — In `_recv_loop`, the `async for raw_msg in ws:` blocks indefinitely waiting for the next WebSocket message. When `node.stop()` sets `_running = False` (e.g., on SIGTERM), neither `run_forever` nor `_session` cancels this task. Since `_session` awaits `asyncio.gather` of `_send_loop` and `_recv_loop`, and `_recv_loop` never exits, `run_forever` never returns and the cleanup in `_async_main`'s finally block (which calls `node.cleanup()`) never executes. This leaks the PyAudio instance and the audio streams, and causes systemd to hang when stopping the service, eventually having to SIGKILL it. Fix by checking `_running` in an explicit loop or cancelling the gather tasks on shutdown.

packages/gptme-voice-node/src/gptme_voice_node/node.py
```diff
@@ -0,0 +1,256 @@
...
+        log.info("[%s] connected to %s", self.node_name, self.server_url)
+
+        mic = self._pa.open(
+            format=self._audio_format,
+            channels=CHANNELS,
+            rate=SAMPLE_RATE,
+            input=True,
+            frames_per_buffer=CHUNK_SIZE,
+        )
+        speaker = self._pa.open(
```
TimeToBuildBob: ⚠️ **P2** — In `_session`, the mic stream is opened on line 109, then the speaker stream on line 116. If `self._pa.open` for the speaker raises (no output device, ALSA failure), the exception propagates before the `try` block on line 124, so the `finally` never runs and `mic` is never closed. On every reconnect attempt, a new mic stream is leaked, eventually exhausting ALSA device handles (device limits under `/proc/asound`) and blocking all audio on the embedded host. Wrap both opens in the same `try`/`finally`, or close `mic` if the speaker open fails.

packages/gptme-voice-node/tests/test_voice_node.py
```diff
@@ -0,0 +1,154 @@
...
+    @pytest.mark.asyncio
+    async def test_audio_frame_triggers_playback_flag(self):
+        node = _make_node()
+        node._playing = False
+
+        pcm = b"\x00\x01" * 512
+        msg = json.dumps({"type": "audio", "audio": base64.b64encode(pcm).decode()})
+
+        mock_ws = AsyncMock()
+        mock_ws.__aiter__.return_value = iter([msg])
```
TimeToBuildBob: ⚠️ **P2** — In `_recv_loop`, `async for raw_msg in ws` requires `ws` to be an async iterator. The tests set `mock_ws.__aiter__.return_value = iter([msg])` (line 72) and likewise at line 104, which returns a regular `list_iterator`. When `async for` calls `__anext__` on that iterator, Python raises `TypeError: 'list_iterator' object is not an async iterator`. Both `test_audio_frame_triggers_playback_flag` and `test_audio_end_clears_playing_flag` therefore fail, contradicting the PR's claim that all 9 tests pass. Replace these with an async generator, e.g. `async def gen(): yield msg` and set `mock_ws.__aiter__.return_value = gen()`.

packages/gptme-voice-node/src/gptme_voice_node/node.py
```diff
@@ -0,0 +1,256 @@
...
+try:
+    import websockets
+    from websockets.exceptions import ConnectionClosed, WebSocketException
+except ImportError as e:
+    raise ImportError(
+        "websockets is required. Install with: pip install 'websockets>=12.0'"
+    ) from e
+
+# --- Config ---
+SERVER_URL = os.environ.get("GPTME_VOICE_NODE_SERVER", "ws://localhost:8080/local")
```
TimeToBuildBob, line 41: ⚠️ **P2** — The node transmits microphone audio over an unencrypted WebSocket (`ws://`). The default server URL is `ws://localhost:8080/local` and all README examples use `ws://`. Because this is a voice device sending PCM audio from a microphone, any network observer on the same network can intercept the stream and capture the audio, a privacy and confidentiality issue. Even though the URL is configurable via `GPTME_VOICE_NODE_SERVER`, the insecure default and examples encourage deployment without TLS, leaving sensitive voice data exposed in transit.

TimeToBuildBob:
## 🤖 AI code review

Adds a new embedded voice node client package `gptme-voice-node` containing a WebSocket client that streams microphone audio to and plays back audio from a `bob-voice-server`, with automatic reconnection, echo/prevention through mic muting during playback, and systemd service files. It also adds websockets and types-pyaudio to the mypy pre-commit dependencies and includes unit tests for the client's state machine.

> ⚠️ **1 generated file excluded and NOT reviewed:** ``uv.lock`` (+27/-0 lines, 1KB). Machine-generated output is removed from the diff before review — from what the model sees and from the size gate alike — so nothing below says anything about it.

**❌ P1 high** — `packages/gptme-voice-node/src/gptme_voice_node/node.py:201`

The `_send_loop` calls `mic_stream.read` via `asyncio.to_thread`, but the `mic_stream` is the PyAudio stream object. In PyAudio, `read()` is not thread-safe and the stream is already being used by the main thread for `start_stream()`/`stop_stream()` and for reading in the send loop. However, the send loop runs in an executor thread while the event loop runs in the main thread. The `mic_stream.read` is called from a worker thread, and at the same time the `_session` method may call `mic.stop_stream()` or `mic.close()` on the same stream from the event loop thread when a reconnect happens. This is a race condition that can cause a crash or undefined behavior in PortAudio, which is not thread-safe. The consequence is that on disconnect/reconnect, the stream may be closed or stopped while `read()` is in progress, leading to an exception that propagates out of `_send_loop` and terminates the session, potentially crashing the whole node. The suggested fix is to not run the blocking read in a separate thread, or to properly synchronize stream access, or to use a dedicated stream object per thread.

```suggestion
After `gather` completes (whether normally or via exception), ensure the other task is cancelled. For example, wrap the two tasks in a helper that cancels the sibling when either finishes.
```

<sub>**How this was verified:** Examined the `_send_loop` and `_close_stream`; the stream is opened with `start=False` and is started/stopped in the same event loop thread while the read runs in a separate thread via `asyncio.to_thread`. PyAudio's PortAudio bindings are not thread-safe; this is a documented limitation.</sub>

</details>

- `.pre-commit-config.yaml` — Adds websockets and types-pyaudio to the mypy additional dependencies.
- `packages/gptme-voice-node/README.md` — New README documenting configuration, setup, protocol, and deployment for the voice node.
- `packages/gptme-voice-node/pyproject.toml` — New package metadata defining dependencies, entry point, and optional audio/test extras.
- `packages/gptme-voice-node/src/gptme_voice_node/__init__.py` — New package init defining the version string.
- `packages/gptme-voice-node/src/gptme_voice_node/node.py` — New implementation of the VoiceNode class, reconnect loop, audio I/O, and URL redaction.
- `packages/gptme-voice-node/systemd/gptme-voice-node.service` — New systemd unit file for running the voice node as a service.
- `packages/gptme-voice-node/tests/test_voice_node.py` — New unit tests for mic mute logic, receive loop handling, transport warnings, and lifecycle.

</details>

| `f4d5071ff45d` | 2/5 | 5 | llm | 2026-08-27 19:18 UTC |
| `ace9ace3afe9` | 2/5 | 3 | llm | 2026-08-27 20:43 UTC |
| `1fa0b585fa5a` | 3/5 | 3 | llm | 2026-08-27 21:42 UTC |
| `25c70294db9b` | 2/5 | 3 | llm | 2026-08-28 00:41 UTC |
| `7fce34c7f333` | 3/5 | 4 | llm | 2026-08-28 06:28 UTC |
| `e0c7c1186f34` | 3/5 | 8 | llm | 2026-08-28 13:32 UTC |
| `20340d909a86` | 2/5 | 2 | llm | 2026-08-28 15:39 UTC |
| `0809af3f3a8c` | 3/5 | 2 | llm | 2026-08-28 16:36 UTC |

</details>

<sub>Reviewed `d4c0dad4462a` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 330s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

packages/gptme-voice-node/src/gptme_voice_node/node.py
TimeToBuildBob: Fixed in `e3d62541`: `_session()` now resets `_playing` and `_play_ended_at` before each reconnect, so a disconnect mid-playback cannot leave the next session muted.

packages/gptme-voice-node/src/gptme_voice_node/node.py
TimeToBuildBob: Fixed in `e3d62541`: `_recv_loop()` now uses `ws.recv()` under the running guard, and the signal handler cancels `run_forever()` so blocked receive/send tasks unwind through stream cleanup.

packages/gptme-voice-node/src/gptme_voice_node/node.py
TimeToBuildBob: Fixed in `e3d62541`: speaker creation is nested inside the mic stream `try/finally`, so a failed output-stream open still stops and closes the mic stream.

packages/gptme-voice-node/tests/test_voice_node.py
TimeToBuildBob: Fixed in `e3d62541`: the receive-loop tests now drive `recv()` with explicit message and `ConnectionClosed` side effects, matching the shutdown-safe implementation. All 9 package tests pass.

packages/gptme-voice-node/src/gptme_voice_node/node.py
TimeToBuildBob: Fixed in `ace9ace3`: remote `ws://` endpoints now emit a clear privacy warning, the deployment example uses `wss://`, and the README says the localhost default is local-only. Added tests for remote-warning and localhost-no-warning behavior.

TimeToBuildBob: Fixed in `ace9ace3`: remote `ws://` endpoints now emit a clear privacy warning, the deployment example uses `wss://`, and the README says the localhost default is local-only. Added tests for remote-warning and localhost-no-warning behavior.

packages/gptme-voice-node/src/gptme_voice_node/node.py
```diff
@@ -0,0 +1,297 @@
...
+        backoff = BACKOFF_INITIAL
+        while self._running:
+            try:
+                async with websockets.connect(
+                    self.server_url,
+                    open_timeout=10,
+                    ping_interval=20,
+                    ping_timeout=20,
+                ) as ws:
+                    backoff = BACKOFF_INITIAL  # Reset on success
```
TimeToBuildBob: ❌ **P1** — In `run_forever`, `backoff` is reset to `BACKOFF_INITIAL` immediately after the WebSocket connection handshake succeeds, before `_session` runs. If the connection succeeds but the session fails quickly (e.g., the server sends malformed data that causes an exception in `_recv_loop`), the `except` block in `run_forever` catches it, but because `backoff` was just reset to 1, the next retry waits only 1 second. This defeats exponential backoff: a server that accepts connections but then persistently errors will cause reconnects every 1 second instead of doubling to 2, 4, 8... The reset should only happen after a session has run for a sustained period, or the backoff should not be reset on a failed session.

packages/gptme-voice-node/src/gptme_voice_node/node.py
```diff
@@ -0,0 +1,297 @@
...
+        """Handle one connected WebSocket session."""
+        # Reset per-session playback state on every reconnect.
+        # Without this, a disconnect mid-playback leaves _playing=True, which
+        # silently mutes the mic for the entire next session.
+        self._playing = False
+        self._play_ended_at = 0.0
+
+        log.info("[%s] connected to %s", self.node_name, self.server_url)
+
+        mic = self._pa.open(
```
TimeToBuildBob, line 160: ❌ **P1** — In `_session`, the PyAudio input and output streams are opened but `start_stream()` is never called. PyAudio streams are created in a stopped state and must be started before any `read()` or `write()` call. The code immediately enters `_send_loop` which calls `mic_stream.read()`, and `_recv_loop` which calls `speaker_stream.write()`. With the stream not started, `read()` blocks indefinitely (the stream never captures audio), so the send loop hangs on the first read; similarly, `write()` either blocks or fails. The result is that the node never sends or receives audio, and the session never completes, so the node becomes unresponsive. Additionally, because the send loop is blocked in a thread executor, cancellation on shutdown won't unblock it, so the node may not stop cleanly. The fix is to call `start_stream()` on both streams after opening.

packages/gptme-voice-node/src/gptme_voice_node/node.py
```diff
@@ -0,0 +1,297 @@
...
+        """Continuously read mic and forward to server (unless muted)."""
+        loop = asyncio.get_event_loop()
+        while self._running:
+            # Read in executor to avoid blocking the event loop
+            raw = await loop.run_in_executor(
+                None,
+                lambda: mic_stream.read(CHUNK_SIZE, exception_on_overflow=False),
+            )
+            if not self._mic_muted():
+                msg = {"type": "audio", "audio": base64.b64encode(raw).decode()}
```
TimeToBuildBob, line 205: ⚠️ **P2** — The configured node name (from the GPTME_VOICE_NODE_NAME env var) is never transmitted to the server. In _send_loop, the outgoing audio message is constructed as {"type": "audio", "audio": ...}, which contains no reference to self.node_name or any identity field. The README states that the variable is 'Node identity (logged and sent to server)', but it is only used in log lines. If bob-voice-server relies on a node identifier to differentiate multiple connected clients or to route responses, it will not receive it, causing misbehavior in multi-node deployments. This is a documentation/behavior mismatch and potentially a missing feature if the server expects identity.

packages/gptme-voice-node/src/gptme_voice_node/node.py
TimeToBuildBob: Fixed in `1fa0b585`: retry backoff now resets only after a connection remains healthy for 30 seconds. Quick post-handshake failures preserve the current delay and continue the 1/2/4/8s progression; regression tests cover quick and stable sessions.

packages/gptme-voice-node/src/gptme_voice_node/node.py
TimeToBuildBob: Fixed in `1fa0b585`: both PyAudio streams are opened with `start=False`, then explicitly started before the send/receive loops. A lifecycle test verifies both `start_stream()` calls.

packages/gptme-voice-node/src/gptme_voice_node/node.py
TimeToBuildBob: Resolved in `1fa0b585` by correcting the README: the current server protocol accepts only `type` and `audio`, so the node name is a local logging identity and is not claimed as wire metadata. Adding an unsupported field is out of scope for this transport client.

packages/gptme-voice-node/src/gptme_voice_node/node.py
```diff
@@ -0,0 +1,316 @@
...
+            try:
+                raw_msg = await ws.recv()
+            except ConnectionClosed:
+                break
+            data = json.loads(raw_msg)
+            msg_type = data.get("type")
+            if msg_type == "audio":
+                self._playing = True
+                pcm = base64.b64decode(data["audio"])
+                await loop.run_in_executor(None, lambda: speaker_stream.write(pcm))
```
TimeToBuildBob: ❌ **P1** — `_send_loop` and `_recv_loop` run blocking audio I/O via `loop.run_in_executor` (e.g., line 193 for speaker write). When `_async_main` receives SIGTERM, the signal handler calls `node.stop()` and `task.cancel()`. Cancelling the `gather` awaiting these loops raises `CancelledError` at the `await run_in_executor`, but the executor callable continues running in the thread pool. The `finally` blocks in `_session` then close the streams (lines 152-153, 155-156) and `cleanup()` calls `self._pa.terminate()` (line 249) while those executor calls may still be accessing the audio device. This can cause undefined behavior, crashes, or hangs during shutdown, and it leaks the thread until the callable completes. The code should await the executor Future's completion (or use a dedicated cancel-safe approach) before releasing PyAudio.

TimeToBuildBob:
## Convergence adjudication

**Fixed in `25c70294`:**
- **P1 — audio I/O outlives cancellation:** blocking mic reads and speaker writes now run through a cancellation-safe helper. Cancellation waits for the in-flight worker to finish before `_session()` closes streams or `cleanup()` terminates PyAudio. A regression test proves the outer task remains pending until the worker completes.
- **P2 — stable-session backoff reset was immediately doubled:** `_next_backoff()` now computes the next retry state once, after the current delay. Quick failures still progress exponentially; a session lasting at least 30 seconds resets the following delay to 1 second.

**Remaining, non-blocking:**
- **P2 — alleged echo race around `_mic_muted()`: false positive.** The mute check already runs after the 1024-frame read and immediately before `ws.send`. If playback starts during the read, `_playing` is true; if `audio_end` arrives during the read, the 500 ms cooldown is still active because one read is about 43 ms. The claimed path requires that 500 ms expire inside a roughly 43 ms read.

**CI:** prior head was fully green. Current head CI is running; local package tests pass (15/15), and scoped pre-commit, mypy, package typecheck, Ruff, and duplicate-code checks pass.

**Domain risk:** cancellation behavior is deterministically tested, but actual PortAudio device teardown and echo behavior still warrant a maintainer smoke test on Raspberry Pi/USB audio hardware before unattended deployment.

**Convergence:** detector reported `stable_rounds: 0` with `status: no_history`; the review summary itself records three review passes (scores 2/5, 2/5, 3/5). The detector cannot reconstruct stable rounds from these self-review comments, so no additional review was triggered. Blocking findings are now addressed; stopping at maintainer judgment.

packages/gptme-voice-node/src/gptme_voice_node/node.py
```diff
@@ -0,0 +1,321 @@
...
+                )
+            except Exception as exc:  # noqa: BLE001
+                log.error(
+                    "[%s] unexpected error: %s — retrying in %ds",
+                    self.node_name,
+                    exc,
+                    backoff,
+                )
+            if self._running:
+                await asyncio.sleep(backoff)
```
TimeToBuildBob: ❌ **P1** — In `run_forever`, the backoff reset happens after the sleep: `await asyncio.sleep(backoff)` is called with the current (possibly already-maxed) backoff, and only then is `_next_backoff(backoff, connected_at)` computed. When a session that lasted >= BACKOFF_RESET_AFTER (30s) ends, `_next_backoff` resets to BACKOFF_INITIAL, but that new value only takes effect on the *next* iteration; the immediate sleep still uses the old backoff. So after a healthy, long session, a disconnect delays reconnection by up to BACKOFF_MAX (60s), exactly the opposite of the intended reset behaviour. On a flaky network this can leave the node offline for a minute despite having just had a stable connection. Fix: compute the next backoff before sleeping, so the reset shortens the current wait.

packages/gptme-voice-node/src/gptme_voice_node/node.py
```diff
@@ -0,0 +1,321 @@
...
+        run_forever from hanging and leaking PyAudio resources when the node
+        is stopped via SIGTERM.
+        """
+        while self._running:
+            try:
+                raw_msg = await ws.recv()
+            except ConnectionClosed:
+                break
+            data = json.loads(raw_msg)
+            msg_type = data.get("type")
```
TimeToBuildBob: ⚠️ **P2** — In _recv_loop, the received message is parsed with json.loads and then treated as a dict via data.get and data[...]. If the server sends valid JSON that is not an object (e.g., an array or string), data.get raises AttributeError. Similarly, an 'audio' message that lacks the 'audio' key causes a KeyError. These exceptions are not caught inside the loop, so they propagate out of _recv_loop, aborting the session and triggering a reconnect with exponential backoff. A server that sends any malformed message (even transiently) will cause repeated disconnects and reconnects, interrupting service. The code should validate that the parsed value is a dict and that required keys are present, handling malformed messages by logging and continuing.

packages/gptme-voice-node/src/gptme_voice_node/node.py
TimeToBuildBob: Fixed in `7fce34c7`: `run_forever()` now computes the post-session backoff before sleeping, so a stable session uses the reset 1-second delay for the immediate reconnect. Added a loop-level regression test that asserts the delay passed to `asyncio.sleep`.

packages/gptme-voice-node/src/gptme_voice_node/node.py
TimeToBuildBob: Fixed in `7fce34c7`: `_recv_loop()` now validates the decoded JSON object and audio payload, rejects invalid base64, logs malformed messages, and continues the session. Parameterized tests cover invalid JSON, non-object JSON, missing payloads, and invalid base64.

packages/gptme-voice-node/src/gptme_voice_node/node.py
```diff
@@ -0,0 +1,329 @@
...
+                format=self._audio_format,
+                channels=CHANNELS,
+                rate=SAMPLE_RATE,
+                output=True,
+                frames_per_buffer=CHUNK_SIZE,
+                start=False,
+            )
+            try:
+                mic.start_stream()
+                speaker.start_stream()
```
TimeToBuildBob, line 182: ❌ **P1** — In `_session`, the cleanup `finally` blocks assume the streams were successfully started. When `speaker.start_stream()` fails (e.g., output device is busy or missing), the inner `finally` calls `speaker.stop_stream()` and then `speaker.close()`. However, calling `stop_stream()` on a stream that was never started raises a PyAudio error, which aborts the `finally` block before `speaker.close()` executes. The speaker stream is then leaked, and on repeated failures the process exhausts ALSA device handles, eventually preventing the node from opening any output stream. The same hazard applies to `mic.start_stream()` failing: the outer `finally` attempts `mic.stop_stream()` and, if that raises, `mic.close()` never runs. In an unattended embedded deployment this leads to a permanent failure to reconnect with output, and the node only recovers after a process restart.

packages/gptme-voice-node/src/gptme_voice_node/node.py
```diff
@@ -0,0 +1,329 @@
...
+                        raise TypeError("audio message must contain a string payload")
+                    pcm = base64.b64decode(audio, validate=True)
+                    self._playing = True
+                    await _run_audio_io(speaker_stream.write, pcm)
+                elif msg_type == "audio_end":
+                    self._playing = False
+                    self._play_ended_at = time.monotonic()
+                    log.debug("[%s] audio_end received", self.node_name)
+            except (ValueError, TypeError) as exc:
+                log.warning("[%s] ignoring malformed message: %s", self.node_name, exc)
```
TimeToBuildBob, line 249: ⚠️ **P2** — The `_recv_loop` logs a warning for every malformed message it receives, with no rate limiting or counting. A remote endpoint (or a compromised server) can flood the node with messages that fail to parse as JSON, are not JSON objects, have a non-string `audio` field, or contain invalid base64; each such message produces a `log.warning` line. On an embedded system running under systemd, this fills the journal indefinitely, consuming disk space and degrading performance, and the repeated warning lines obscure genuinely useful diagnostics. Because the node is designed to be unattended, a sustained stream of malformed frames can eventually fill the storage partition on the host.

packages/gptme-voice-node/src/gptme_voice_node/node.py
```diff
@@ -0,0 +1,329 @@
...
+T = TypeVar("T")
+
+
+async def _run_audio_io(function: Callable[..., T], *args: Any, **kwargs: Any) -> T:
+    """Run blocking audio I/O and let it finish before propagating cancellation."""
+    task = asyncio.create_task(asyncio.to_thread(function, *args, **kwargs))
+    try:
+        return await asyncio.shield(task)
+    except asyncio.CancelledError:
+        await task
```
TimeToBuildBob: ⚠️ **P2** — On shutdown, the signal handler cancels the `run_forever` task. If either the mic read or speaker write is currently inside `_run_audio_io`, the cancellation is deferred because `_run_audio_io` shields the underlying thread and, in the `CancelledError` branch, awaits `task` — i.e., it waits for the blocking I/O to complete before re-raising. If the microphone is producing no input and `mic_stream.read()` blocks indefinitely (common when the device is idle or unplugged), the node will never terminate when SIGTERM is sent, since the thread never returns. The process then continues to hold the event loop and PyAudio resources, and systemd's `TimeoutStopSec` (default 90s) eventually SIGKILLs it. This makes reliable service shutdown impossible in the presence of a blocked audio read.

packages/gptme-voice-node/src/gptme_voice_node/node.py
```diff
@@ -0,0 +1,329 @@
...
+                if msg_type == "audio":
+                    audio = data.get("audio")
+                    if not isinstance(audio, str):
+                        raise TypeError("audio message must contain a string payload")
+                    pcm = base64.b64decode(audio, validate=True)
+                    self._playing = True
+                    await _run_audio_io(speaker_stream.write, pcm)
+                elif msg_type == "audio_end":
+                    self._playing = False
+                    self._play_ended_at = time.monotonic()
```
TimeToBuildBob, line 246: ⚠️ **P2** — In `_recv_loop`, the `audio_end` branch sets `_play_ended_at = time.monotonic()` unconditionally. If the server sends an `audio_end` when `_playing` is already False (e.g., because no `audio` message preceded it, or it is a duplicate), this starts the playback cooldown for `PLAYBACK_COOLDOWN` (0.5s), which mutes the mic even though no playback occurred. A server that sends spurious `audio_end` messages (e.g., on reconnect or after an error) will cause the mic to be muted for 0.5 seconds, clipping the start of the next user speech. The fix is to only set `_play_ended_at` if `_playing` was True, so an `audio_end` with no prior playback does not trigger the cooldown.

ErikBjare:
@TimeToBuildBob fix

packages/gptme-voice-node/src/gptme_voice_node/node.py
TimeToBuildBob: Fixed in `e0c7c118`: stream teardown now tracks whether each stream started, skips `stop_stream()` for failed starts, and always closes the stream even if stopping raises. Tests cover mic and speaker start failures.

packages/gptme-voice-node/src/gptme_voice_node/node.py
TimeToBuildBob: Fixed in `e0c7c118`: cancellation no longer shields and waits for a potentially stuck audio worker, so SIGTERM can unwind immediately. A regression test uses a worker that never finishes and proves cancellation completes within a bounded timeout.

packages/gptme-voice-node/src/gptme_voice_node/node.py
TimeToBuildBob: Fixed in `e0c7c118`: `audio_end` now starts the cooldown only when playback was active. Added a regression test proving a duplicate or spurious `audio_end` leaves the mic unmuted.

packages/gptme-voice-node/src/gptme_voice_node/node.py
TimeToBuildBob: Not changing this in the client. Malformed frames are a server/protocol fault and warnings are the correct diagnostic; journald already enforces service-wide rate limiting and bounded retention. Adding a second ad hoc limiter here would hide evidence while duplicating the deployment layer.

TimeToBuildBob:
Fixed the four new findings in `e0c7c118`: safe cleanup after stream-start failures, bounded SIGTERM cancellation when audio I/O sticks, and no cooldown on spurious `audio_end`. I rejected client-side malformed-frame rate limiting because journald already provides rate limiting and bounded retention. Local package tests pass (23/23); CI is now running on the pushed head.

packages/gptme-voice-node/src/gptme_voice_node/node.py
```diff
@@ -0,0 +1,337 @@
...
+                    open_timeout=10,
+                    ping_interval=20,
+                    ping_timeout=20,
+                ) as ws:
+                    connected_at = time.monotonic()
+                    await self._session(ws)
+            except (ConnectionClosed, WebSocketException) as exc:
+                log.warning(
+                    "[%s] disconnected: %s — retrying in %ds",
+                    self.node_name,
```
TimeToBuildBob: ⚠️ **P2** — The reconnect log message reports the current backoff, but the actual sleep uses the backoff computed after logging. On the first failure, backoff is 1, the log says 'retrying in 1s', then backoff doubles to 2 and the code sleeps 2 seconds. Every subsequent retry logs a value one step behind the real delay, so operators reading the log expect a retry sooner than it happens. This can mislead monitoring and manual diagnostics during an outage. Move the backoff update before the log (or log the value after updating) so the message matches the sleep.

packages/gptme-voice-node/src/gptme_voice_node/node.py
```diff
@@ -0,0 +1,337 @@
...
+        self._running = True
+
+    # --- Mic mute logic ---
+
+    def _mic_muted(self) -> bool:
+        """True when we should suppress outbound audio (echo prevention)."""
+        return self._playing or (
+            time.monotonic() - self._play_ended_at < PLAYBACK_COOLDOWN
+        )
+
```
TimeToBuildBob, line 143: ⚠️ **P2** — At startup, `_play_ended_at` is initialized to 0.0 and `_mic_muted()` computes `time.monotonic() - _play_ended_at < PLAYBACK_COOLDOWN`. If the system has been up for less than 0.5 seconds, the monotonic clock is below 0.5, so this condition is true and the mic is muted for the first few hundred milliseconds, dropping the initial audio. On a freshly booted Raspberry Pi this is plausible. Fix by treating a non-positive `_play_ended_at` as no cooldown, e.g. add a `> 0` guard, or initialize to a negative value.

packages/gptme-voice-node/src/gptme_voice_node/node.py
```diff
@@ -0,0 +1,337 @@
...
+        datefmt="%Y-%m-%dT%H:%M:%S",
+        stream=sys.stderr,
+    )
+
+    server_url = SERVER_URL
+    node_name = NODE_NAME
+    _warn_for_insecure_remote_url(server_url)
+
+    log.info("Starting BobBrain voice node")
+    log.info("  server : %s", server_url)
```
TimeToBuildBob: ⚠️ **P2** — In `main()`, the full `server_url` is logged in plaintext via `log.info("  server : %s", server_url)`. If a user embeds credentials in the URL (e.g., `wss://user:password@host/...`), those secrets are written to stderr and captured by journald/systemd. On unattended embedded nodes, logs may be read by other local processes or forwarded to log aggregation services, leaking credentials. The `_warn_for_insecure_remote_url` function already parses the URL with `urlparse`, but `main()` logs the raw string. The fix is to redact the userinfo component before logging, e.g., using `urlparse` and reconstructing without it.

packages/gptme-voice-node/systemd/gptme-voice-node.service
```diff
@@ -0,0 +1,35 @@
...
+Environment=GPTME_VOICE_NODE_NAME=bobbrain-puck
+
+# Adjust to the user that has audio access (e.g. pi on Raspberry Pi OS)
+User=pi
+WorkingDirectory=/home/pi
+
+ExecStart=/home/pi/.local/bin/gptme-voice-node
+
+# Auto-restart with progressive cooldown
+Restart=always
```
TimeToBuildBob, line 21: ⚠️ **P2** — The systemd unit sets `Restart=always`, so systemd will restart the service regardless of exit status. Since the node handles SIGTERM and its main() returns normally (exit code 0), a `systemctl stop` will cause the process to exit and then immediately restart, preventing the service from staying stopped. This contradicts the expectation that `systemctl stop` stops the service. A user following the README will be unable to stop the node without disabling the unit or changing the Restart policy.

packages/gptme-voice-node/tests/test_voice_node.py
```diff
@@ -0,0 +1,305 @@
...
+
+class TestSendLoop:
+    @pytest.mark.asyncio
+    async def test_audio_io_cancellation_does_not_wait_for_inflight_thread(self):
+        from gptme_voice_node.node import _run_audio_io
+
+        thread_started = asyncio.Event()
+        never_finishes = asyncio.Event()
+
+        async def fake_to_thread(function, *args, **kwargs):
```
TimeToBuildBob: ⚠️ **P2** — In test_voice_node.py, the test `test_audio_io_cancellation_does_not_wait_for_inflight_thread` patches `asyncio.to_thread` with a fake that is an async function that blocks on an event, but it never actually runs the operation in a separate thread. Therefore it does not test the behavior of the real `asyncio.to_thread`, which runs the function in a thread pool and does not cancel the thread when the awaiting task is cancelled. The test merely cancels an async function that blocks, which is trivial. This stubs the wrong thing and does not verify the intended property that cancellation does not wait for an inflight thread.

packages/gptme-voice-node/src/gptme_voice_node/node.py
```diff
@@ -0,0 +1,337 @@
...
+
+def _close_stream(stream, started: bool) -> None:
+    """Close an audio stream even when start or stop fails."""
+    try:
+        if started:
+            stream.stop_stream()
+    except Exception as exc:  # noqa: BLE001
+        log.warning("failed to stop audio stream: %s", exc)
+    finally:
+        stream.close()
```
TimeToBuildBob, line 103: ⚠️ **P2** — In `_close_stream`, exceptions from `stream.close()` are not caught, so if closing an audio stream raises (e.g., due to a device error) the exception propagates out of the `finally` block in `_session`. This can mask the original error that triggered the cleanup and, if the inner `_close_stream` raises before the outer one runs, leave the other stream unclosed. For instance, if `speaker.close()` raises while cleaning up after a mic-start failure, the outer `finally` will still run and close the mic, but the original `OSError` from `mic.start_stream()` is replaced by the close error, making the root cause invisible. The fix is to catch and log exceptions from `stream.close()` as well.

ErikBjare:
@TimeToBuildBob fix

TimeToBuildBob:
Fixed the latest actionable findings in `75e0330e`: retry logs now match the actual sleep delay, startup cannot enter a false cooldown, logged WebSocket URLs redact userinfo, stream-close failures no longer mask the original error, and the cancellation test now exercises a real worker thread. Local package tests pass (27/27), and scoped Ruff/format/diff checks pass.

I left `Restart=always` unchanged: systemd does not apply restart policy to an explicit `systemctl stop`, so that finding's premise is false.

packages/gptme-voice-node/systemd/gptme-voice-node.service
TimeToBuildBob: Rejected. systemd does not apply `Restart=` to an explicit `systemctl stop`. From systemd.service(5): "Restart= does not apply when a unit is stopped via systemctl stop or equivalent." `Restart=always` is the right unattended-node policy; `systemctl stop gptme-voice-node` still leaves the service stopped.

packages/gptme-voice-node/src/gptme_voice_node/node.py
TimeToBuildBob: Fixed in `75e0330e`: `run_forever()` now computes the next backoff before the retry log, so the logged delay matches `asyncio.sleep`. Regression: `test_retry_log_matches_sleep_delay`.

packages/gptme-voice-node/src/gptme_voice_node/node.py
TimeToBuildBob: Fixed in `75e0330e`: `_mic_muted()` now requires `_play_ended_at > 0` before applying the cooldown, so a fresh monotonic clock at boot cannot mute the mic.

packages/gptme-voice-node/src/gptme_voice_node/node.py
TimeToBuildBob: Fixed in `75e0330e`: `main()` logs `_redact_url_userinfo(server_url)`, which strips userinfo before journald sees it. Regression: `test_url_userinfo_is_redacted`.

packages/gptme-voice-node/tests/test_voice_node.py
TimeToBuildBob: Fixed in `75e0330e`: `test_audio_io_cancellation_does_not_wait_for_inflight_thread` now runs a real worker thread via `_run_audio_io` / `asyncio.to_thread` and cancels the awaiting task while the thread is blocked.

packages/gptme-voice-node/src/gptme_voice_node/node.py
TimeToBuildBob: Fixed in `75e0330e`: `_close_stream()` catches and logs exceptions from both `stop_stream()` and `close()`, so a close failure cannot mask the original error or skip the other stream. Regression: `test_session_attempts_both_closes_when_one_close_fails`.

packages/gptme-voice-node/src/gptme_voice_node/node.py
```diff
@@ -0,0 +1,349 @@
...
+
+    async def _session(self, ws) -> None:
+        """Handle one connected WebSocket session."""
+        # Reset per-session playback state on every reconnect.
+        # Without this, a disconnect mid-playback leaves _playing=True, which
+        # silently mutes the mic for the entire next session.
+        self._playing = False
+        self._play_ended_at = 0.0
+
+        log.info("[%s] connected to %s", self.node_name, self.server_url)
```
TimeToBuildBob: ❌ **P1** — In `_session`, line 148 logs `self.server_url` directly, which can contain userinfo credentials if the WebSocket URL is set to something like `wss://user:secret@host`. Unlike `main()` which uses `_redact_url_userinfo` for the startup log line, this line never redacts, so credentials are written to the systemd journal on every connection. An operator who embeds credentials in the URL to avoid a separate auth step will leak the secret to anyone with log access. Fix: apply `_redact_url_userinfo(self.server_url)` to this log message. _(2/3 passes — pass 2 looked and disagreed)_

TimeToBuildBob: Fixed in `20340d90`: `_session` now logs `_redact_url_userinfo(self.server_url)`, so reconnect logs cannot leak URL userinfo to journald. Regression: `test_session_connect_log_redacts_userinfo`. Local package tests pass (28/28).

packages/gptme-voice-node/src/gptme_voice_node/node.py
```diff
@@ -0,0 +1,353 @@
...
+                async with websockets.connect(
+                    self.server_url,
+                    open_timeout=10,
+                    ping_interval=20,
+                    ping_timeout=20,
+                ) as ws:
+                    connected_at = time.monotonic()
+                    await self._session(ws)
+            except (ConnectionClosed, WebSocketException) as exc:
+                error = (logging.WARNING, f"disconnected: {exc}")
```
TimeToBuildBob: ❌ **P1** — Error messages from `websockets.connect` and other exceptions are logged verbatim in the reconnect loop. The exception string for a failed connection typically includes the full WebSocket URI, including any `user:password@` userinfo. Since `run_forever` logs the raw exception via `error = (logging.WARNING, f"disconnected: {exc}")` (line 248) and later `log.log(level, "[%s] %s — retrying in %ds", ..., message, ...)` (line 259), any credentials in the URL are printed to the journal. The connection-success log at line 151 uses `_redact_url_userinfo`, but the error path does not. An operator inspecting logs would see the plaintext password. Fix by redacting the URL in the exception message, e.g., by replacing `self.server_url` with the redacted version.

packages/gptme-voice-node/src/gptme_voice_node/node.py
```diff
@@ -0,0 +1,353 @@
...
+                frames_per_buffer=CHUNK_SIZE,
+                start=False,
+            )
+            speaker_started = False
+            try:
+                mic.start_stream()
+                mic_started = True
+                speaker.start_stream()
+                speaker_started = True
+                await asyncio.gather(
```
TimeToBuildBob: ❌ **P1** — `asyncio.gather` does not cancel the sibling task when one of `_send_loop`/`_recv_loop` raises. If `_send_loop` fails (e.g., mic read error), the exception propagates immediately, `_session` closes both streams in its finally blocks, and `_recv_loop` continues running. If it later writes to the closed speaker stream it raises an `OSError` that is neither caught nor retrieved, producing a spurious 'Task exception was never retrieved' log and potentially leaving the WebSocket close unhandled. This can occur on a mic read failure, leading to log spam and an unpleasant shutdown. Use a `try/except` around the gather to cancel the other task, or pass `return_exceptions=True` and inspect results.

packages/gptme-voice-node/src/gptme_voice_node/node.py
TimeToBuildBob: Fixed in `0809af3f`: reconnect exception logs now run through `_redact_userinfo_in_text`, so URL userinfo cannot leak to journald on connect/disconnect failures. Regression: `test_disconnect_log_redacts_userinfo`.

packages/gptme-voice-node/src/gptme_voice_node/node.py
TimeToBuildBob: Fixed in `0809af3f`: `_session` now creates send/recv tasks and cancels the sibling in `finally` when one loop fails, so a mic error cannot leave `_recv_loop` running against closed streams. Regression: `test_session_cancels_sibling_when_one_loop_fails`.

TimeToBuildBob:
Fixed the two new P1s in `0809af3f`: reconnect exception logs redact URL userinfo, and a failed send/recv loop now cancels its sibling so it cannot run against closed streams. Local package tests pass (30/30). Self-merge remains blocked by the systemd unit path; CI is running on the new head.

packages/gptme-voice-node/src/gptme_voice_node/node.py
```diff
@@ -0,0 +1,380 @@
...
+                data = json.loads(raw_msg)
+                if not isinstance(data, dict):
+                    raise TypeError("message must be a JSON object")
+                msg_type = data.get("type")
+                if msg_type == "audio":
+                    audio = data.get("audio")
+                    if not isinstance(audio, str):
+                        raise TypeError("audio message must contain a string payload")
+                    pcm = base64.b64decode(audio, validate=True)
+                    self._playing = True
```
TimeToBuildBob, line 242: ❌ **P1** — In `_recv_loop`, when an `audio` message is received, `self._playing` is set to True and is only reset to False by an `audio_end` message. If the server never sends `audio_end` (e.g., it crashes or has a protocol bug) while the connection remains open, `self._playing` stays True forever. Because `_mic_muted()` returns True when `_playing` is True, the send loop discards all mic audio, so the node becomes permanently silent until the session is re-established (which only happens on a disconnect). This is a functional failure: the user cannot speak even though the connection is alive and the server may be expecting input.

packages/gptme-voice-node/src/gptme_voice_node/node.py
```diff
@@ -0,0 +1,380 @@
...
+            speaker_started = False
+            try:
+                mic.start_stream()
+                mic_started = True
+                speaker.start_stream()
+                speaker_started = True
+                send_task = asyncio.create_task(self._send_loop(ws, mic))
+                recv_task = asyncio.create_task(self._recv_loop(ws, speaker))
+                try:
+                    await asyncio.gather(send_task, recv_task)
```
TimeToBuildBob, line 187: ⚠️ **P2** — When the node is shut down, `_handle_signal` cancels the `run_forever` task. This cancellation propagates to the `asyncio.gather` in `_session`, which cancels the `_send_loop` and `_recv_loop` tasks. These loops invoke `_run_audio_io`, which uses `asyncio.to_thread` to run blocking PyAudio reads/writes. Cancelling the `to_thread` await does not stop the underlying thread; it only stops waiting for it. The `finally` block in `_session` then closes the mic and speaker streams while those threads may still be blocked inside `mic_stream.read` or `speaker_stream.write`. Closing a PyAudio stream while a blocking read/write is in progress can crash or corrupt the audio backend on embedded Linux. The code's comment claims cancellation prevents resource leaks, but it does not prevent this use-after-close race.

packages/gptme-voice-node/src/gptme_voice_node/node.py
TimeToBuildBob: Fixed in `d4c0dad4`: `_recv_loop` now times out `ws.recv()` after `PLAYBACK_STALL_TIMEOUT` (5s) while `_playing` is true. A live session that never gets `audio_end` unmutes the mic instead of dropping every chunk until disconnect. Regression: `test_playback_stall_timeout_unmutes_mic`. Local package tests pass (31/31).

packages/gptme-voice-node/src/gptme_voice_node/node.py
TimeToBuildBob: Accepted tradeoff, not changing this. Bounded SIGTERM was an earlier P1: waiting for a stuck `mic.read()` made systemd SIGKILL the node. Cancelling the `to_thread` await and closing streams is the shutdown we chose; a blocked audio thread may still be inside read/write at close, and the next process start gets a fresh ALSA handle. Waiting for the worker would reintroduce unbounded stop. The race is real; hanging shutdown is worse for an unattended node.

TimeToBuildBob:
Fixed the remaining P1 in `d4c0dad4`: a live session that received audio but never got `audio_end` left `_playing=True` and dropped every mic chunk until disconnect. `_recv_loop` now times out that wait after 5s and unmutes. Local package tests pass (31/31).

Rejected the P2 shutdown/close-vs-`to_thread` race as an accepted tradeoff: waiting for a stuck `mic.read()` was the earlier unbounded SIGTERM P1. The systemd unit still blocks self-merge, so this needs a human merge when CI on `d4c0dad4` is green. Not grinding further P2 nits.

packages/gptme-voice-node/src/gptme_voice_node/node.py
```diff
@@ -0,0 +1,395 @@
...
+                if not isinstance(data, dict):
+                    raise TypeError("message must be a JSON object")
+                msg_type = data.get("type")
+                if msg_type == "audio":
+                    audio = data.get("audio")
+                    if not isinstance(audio, str):
+                        raise TypeError("audio message must contain a string payload")
+                    pcm = base64.b64decode(audio, validate=True)
+                    self._playing = True
+                    await _run_audio_io(speaker_stream.write, pcm)
```
TimeToBuildBob, line 243: ❌ **P1** — In _recv_loop, each audio frame is written via `await _run_audio_io(speaker_stream.write, pcm)`, which submits the write to the default ThreadPoolExecutor and yields control to the event loop. If the server sends another audio frame while the previous write is still in progress, the loop immediately submits a second write on a new thread, so speaker_stream.write can be called concurrently from multiple threads. PyAudio's Stream.write is not thread-safe; concurrent calls can corrupt audio state, cause glitches, or crash the process with a segfault. This occurs whenever the server bursts frames faster than the speaker drains them, which is a normal operating condition. The observable consequence is corrupted playback or a node crash on the embedded device. The fix is to serialize writes to the speaker stream, e.g., by guarding each write with an asyncio.Lock or by feeding a dedicated writer task through an asyncio.Queue. _(2/3 passes — pass 1 looked and disagreed)_

TimeToBuildBob: Rejected as a false positive.

`_recv_loop` is a single sequential coroutine. Each audio frame does `await _run_audio_io(speaker_stream.write, pcm)` — i.e. `await asyncio.to_thread(...)` — and only then loops back to `ws.recv()`. The next write cannot be submitted until the previous thread has finished, so `Stream.write` is never called concurrently from this path.

The only other speaker-write site does not exist: `_send_loop` reads the mic. Session teardown can cancel an in-flight `to_thread` and then close the stream; that close-vs-worker race was already reviewed as P2 `5954107e1da3` and accepted as the bounded-SIGTERM tradeoff.

The suggested `asyncio.wait(..., FIRST_COMPLETED)` change is a leftover from the already-fixed sibling-task cancel (`01acaf01febb` / `0809af3f`) and does not address this claim.

TimeToBuildBob:
Automated merge handoff for `d4c0dad4462a`:

- CI gate: satisfied
- Review threads: clear
- `.pre-commit-config.yaml`
- `packages/gptme-voice-node/systemd/gptme-voice-node.service`

packages/gptme-voice-node/src/gptme_voice_node/node.py
```diff
@@ -0,0 +1,395 @@
...
+                    await asyncio.gather(send_task, recv_task, return_exceptions=True)
+            finally:
+                _close_stream(speaker, speaker_started)
+        finally:
+            _close_stream(mic, mic_started)
+
+    async def _send_loop(self, ws, mic_stream) -> None:
+        """Continuously read mic and forward to server (unless muted)."""
+        while self._running:
+            raw = await _run_audio_io(
```
TimeToBuildBob, line 201: ❌ **P1** — The `_send_loop` calls `mic_stream.read` via `asyncio.to_thread`, but the `mic_stream` is the PyAudio stream object. In PyAudio, `read()` is not thread-safe and the stream is already being used by the main thread for `start_stream()`/`stop_stream()` and for reading in the send loop. However, the send loop runs in an executor thread while the event loop runs in the main thread. The `mic_stream.read` is called from a worker thread, and at the same time the `_session` method may call `mic.stop_stream()` or `mic.close()` on the same stream from the event loop thread when a reconnect happens. This is a race condition that can cause a crash or undefined behavior in PortAudio, which is not thread-safe. The consequence is that on disconnect/reconnect, the stream may be closed or stopped while `read()` is in progress, leading to an exception that propagates out of `_send_loop` and terminates the session, potentially crashing the whole node. The suggested fix is to not run the blocking read in a separate thread, or to properly synchronize stream access, or to use a dedicated stream object per thread. _(2/3 passes — pass 3 looked and disagreed)_

TimeToBuildBob: Duplicate of accepted-tradeoff `5954107e1da3` — not changing this.

This is the same close-while-`to_thread` race: `_session`'s `finally` closes streams after cancelling send/recv, while `mic.read()` may still be inside the executor thread. Waiting for that worker reintroduces the unbounded SIGTERM P1 we already fixed. Cancelling the `to_thread` await and closing is the shutdown/reconnect we chose; a blocked audio thread may still be inside `read()` at close, and the next process start gets a fresh ALSA handle.

The rest of the claim is wrong on the steady-state path: `start_stream()` finishes before `_send_loop` is created. During the session only the worker thread calls `read()`. The suggested "don't use `to_thread`" would block the event loop so `_recv_loop` cannot run.

TimeToBuildBob:
Landed. `gptme-voice-node` is on `master` (`5fbd2ee1`).

BobBrain remaining software is gptme/gptme-contrib#1528 (BodyAdapter + MAVSDK). Hardware v0 stays gated on the Pi/ReSpeaker. Local task `bobbrain-node` is off this PR.

PR #1525, opened 2026-08-27
fix(gptme-voice): await in-flight pre-warm on claim + caller-aware external-caller instructions

TimeToBuildBob:
## Problem

Observed on the first real external caller call (2026-08-27): the caller heard garbled internal ops-speak, then silence. Two independent defects:

1. **Pre-warm claim race**: `POST /incoming` fired-and-forgot `_prewarm_for_inbound`, while the Twilio WS `"start"` handler called sync `_claim_prewarm` — the claim could run before "Pre-warm ready", so the call fell back to a duplicate cold session while the pre-warm completed into an orphaned provider session (doubled "Session created" in the logs).
2. **External callers got operator instructions**: the session bootstrap injected the internal activity digest (subagent status etc.) for every caller, so an unknown external caller was greeted with internal ops status.

## Fix

**Race**: in-flight pre-warm tasks are now tracked via `_register_prewarm_task` (with done-callback cleanup); `_claim_prewarm` is async and awaits an in-flight pre-warm up to 2s (`asyncio.shield` + `wait_for`), cancelling it on timeout. `_prewarm_for_inbound` handles `CancelledError` by disconnecting a half-open client, and schedules stale-prewarm eviction at TTL+1 so orphans don't linger until the next call.

**Caller-aware instructions**: `CallerIdentity` gains `is_operator` (from `- Call role: operator` in `people/*.md`) and `call_language` (from `- Call language: <lang>`). Non-operator and unknown callers get external-caller guidance (no internal ops status, no unprompted tool use, speak the caller's language); fresh-call greetings honor `call_language`; the activity digest is injected only for operator callers (local/browser transport without a from_number keeps the digest). `_extract_people_field` parses both plain and markdown-bolded property lines (`- **Call role**: operator`) and strips bolded values.

## Tests

13 new tests: caller role/language parsing (plain + bolded), external guidance presence/absence, digest operator-gating, claim-awaits-inflight, claim-timeout-cancels, ready-claim, register-cleanup. Full gptme-voice suite: 228 passed.

https://claude.ai/code/session_01V5jRRQtMvq5eYQ6bNK6PA2

### DIFF
--- a/packages/gptme-voice/src/gptme_voice/realtime/server.py
+++ b/packages/gptme-voice/src/gptme_voice/realtime/server.py
@@ -56,6 +56,12 @@ logger = logging.getLogger(__name__)
 class CallerIdentity:
     canonical_name: str
     preferred_spoken_name: str
+    # True when the caller's people file marks them as the operator
+    # (`- Call role: operator`).  Operators get internal context (activity
+    # digest, ops status); everyone else is treated as an external guest.
+    is_operator: bool = False
+    # Preferred spoken language from `- Call language: <lang>`, if present.
+    call_language: str | None = None
 
 
 _DEFAULT_RESUME_WINDOW_SECONDS = 300
@@ -169,13 +175,29 @@ class SessionBootstrap:
     initial_response_instructions: str = ""
 
 
-def _extract_preferred_spoken_name(text: str, canonical_name: str) -> str:
+def _extract_people_field(text: str, field: str) -> str | None:
+    """Extract a `- <field>: value` line from a people file (case-insensitive).
+
+    Handles markdown-bolded property names/values too, e.g.
+    `- **Call name**: Erik` or `- Call name: **Erik**`.
+    """
+    pattern = re.compile(
+        rf"^-\s*\*{{0,2}}{re.escape(field)}\*{{0,2}}\s*:\s*(.+)$",
+        re.IGNORECASE,
+    )
     for line in text.splitlines():
-        stripped = line.strip()
-        if stripped.lower().startswith("- call name:"):
-            value = stripped.split(":", 1)[1].strip()
+        m = pattern.match(line.strip())
+        if m:
+            value = m.group(1).strip().strip("*").strip()
             if value:
                 return value
+    return None
+
+
+def _extract_preferred_spoken_name(text: str, canonical_name: str) -> str:
+    value = _extract_people_field(text, "call name")
+    if value:
+        return value
     parts = canonical_name.split()
     return parts[0] if parts else canonical_name
 
@@ -204,9 +226,12 @@ def _lookup_caller_identity(
                         preferred_spoken_name = _extract_preferred_spoken_name(
                             text, canonical_name
                         )
+                        call_role = _extract_people_field(text, "call role")
                         return CallerIdentity(
                             canonical_name=canonical_name,
                             preferred_spoken_name=preferred_spoken_name,
+                            is_operator=(call_role or "").lower() == "operator",
+                            call_language=_extract_people_field(text, "call language"),
                         )
                 except Exception:
                     pass
@@ -239,13 +264,44 @@ def _build_caller_instructions(
             f"You know this person — refer to them by name."
             f"{name_hint}"
         )
+        if not caller_identity.is_operator:
+            caller_ctx += "\n\n" + _external_caller_guidance(caller_identity)
     else:
         caller_ctx = (
             f"The current caller's phone number is {from_number}. "
-            f"You do not recognise this number; treat the caller as an unknown guest."
+            f"You do not recognise this number; treat the caller as an unknown guest. "
+            + _external_caller_guidance(None)
         )
 
     return f"{caller_ctx}\n\n{base_instructions}"
+
+
+def _external_caller_guidance(caller_identity: CallerIdentity | None) -> str:
+    """Conversation guidance for callers who are not the operator.
+
+    External callers should get a friendly host, not an ops report: no
+    unprompted tool calls, no internal status (subagent state, task queues,
+    work summaries), and speech in their language.
+    """
+    if caller_identity and caller_identity.call_language:
+        language_line = (
+            f"Speak {caller_identity.call_language} with this caller "
+            "unless they switch language themselves."
+        )
+    else:
+        language_line = (
+            "Mirror the caller's language: if they speak another language "
+            "than English, answer in that language."
+        )
+    return (
+        "EXTERNAL CALLER GUIDANCE:\n"
+        "- This caller is NOT the operator. Do not volunteer internal "
+        "operational status (subagent status, task queues, PR/CI state, or "
+        "work summaries), and do not call tools unprompted.\n"
+        "- Be a friendly, concise host: answer their questions, and only use "
+        "tools when their request genuinely needs one.\n"
+        f"- {language_line}"
+    )
 
 
 def _build_fresh_call_greeting_instructions(
@@ -269,9 +325,15 @@ def _build_fresh_call_greeting_instructions(
                 f"not their full name, in one short sentence, for example 'Hi {spoken_name}' "
                 f"or 'Hey {spoken_name}, what's up?'. "
             )
+        language_hint = (
+            f"Greet in {caller_identity.call_language}. "
+            if caller_identity.call_language
+            else ""
+        )
         return (
             self_identity
             + greeting_target
+            + language_hint
             + "Do NOT say 'thanks for calling' or use other stock phone greetings. "
             "Then stop and wait for them to speak."
         )
@@ -749,6 +811,11 @@ class VoiceServer:
         # Pre-warmed realtime connections: from_number -> (client, created_at)
         # Keyed by from_number, claimed and discarded when the Twilio stream starts.
         self._prewarm_sessions: dict[str, tuple[OpenAIRealtimeClient, float]] = {}
+        # In-flight pre-warm tasks by from_number, so _claim_prewarm can await
+        # a pre-warm that hasn't finished connecting yet instead of racing it
+        # into a duplicate cold session (observed 2026-08-27: Twilio's "start"
+        # arrived before "Pre-warm ready", leaving an orphaned provider session).
+        self._prewarm_tasks: dict[str, asyncio.Task] = {}
         # Max seconds a pre-warm session is kept before being discarded
         self._prewarm_ttl_seconds = 30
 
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
CI is green and the PR is cleanly mergeable. Self-merge is not eligible yet because the required Greptile review has not landed; the initial auto-review grace window is still active. Waiting for that review before re-running the guarded merge path.

TimeToBuildBob:
@greptileai review

packages/gptme-voice/src/gptme_voice/realtime/server.py
```diff
@@ -1149,13 +1229,65 @@ async def _prewarm_for_inbound(self, from_number: str) -> None:
...
             logger.info("Pre-warm ready for %s", from_number)
+        except asyncio.CancelledError:
+            # Claim timed out and cancelled us (or the server is shutting
+            # down): close the half-open provider session instead of leaking it.
+            logger.info("Pre-warm cancelled for %s", from_number)
+            if client is not None:
+                asyncio.create_task(self._disconnect_realtime_client(client))
+            raise
         except Exception as exc:
             logger.warning("Pre-warm failed for %s: %s", from_number, exc)
```
TimeToBuildBob, line 1255: ❌ **P1** — In `_prewarm_for_inbound`, if an exception occurs after the client is created (e.g., `connect()` raises), the `except Exception` branch only logs and does not disconnect the client. The `client` variable is initialized to `None` and only disconnected in the `CancelledError` handler. Any failure in `connect()` or the subsequent `_consume_recent_call` leaves the provider WebSocket/session open. Under repeated failures (e.g., a provider outage), this leaks one connection per pre-warm attempt, eventually exhausting connection pools and blocking legitimate calls. _(2/2 passes)_

TimeToBuildBob:
## 🤖 AI code review

This PR introduces caller-aware session instructions: CallerIdentity gains is_operator and call_language fields parsed from people files, external callers receive guidance not to volunteer internal status or use tools unprompted, and the activity digest is injected only for operator callers. It also reworks pre-warm claiming: _claim_prewarm becomes async and awaits in-flight pre-warm tasks, cancelling them on timeout, while _prewarm_for_inbound gains connected-state tracking and TTL-ended eviction.

**❌ P1 high** — `packages/gptme-voice/src/gptme_voice/realtime/server.py:1314`

The safety comment in _claim_prewarm claims pre-connect cancellation is safe because 'the resume state has not been consumed yet', but the flag _prewarm_connected is set (line 1236) before the consume step runs. When the claim times out, it checks if from_number is in _prewarm_connected at line 1314; if the pre-warm task has passed connect() but not yet executed that add, the check returns False and the code cancels the task at line 1356. The task then adds to the set and starts _consume_recent_call. If a pending_unit exists, that coroutine awaits run_in_executor, and the CancelledError lands there, interrupting it partway through; the unlink of the resume file happens after that block, so the resume state can be left half-deleted, and the cold path greets fresh instead of resuming. Even without a pending_unit, the task may complete and store the session after the claim has already returned None, leaving an orphaned provider session — the exact doubled-session the PR is meant to fix. The window is narrow but real, and the comment misstates the invariant.

```suggestion
Instead of returning None here, either block until the task is done before starting the cold path, or cancel the task and accept the resume loss; otherwise the cold path cannot tell whether the resume file will be consumed by the in-flight task.
```

<sub>**How this was verified:** Traced the task flow: _prewarm_for_inbound adds to _prewarm_connected before _consume_recent_call; the claim checks the set after timeout, so a task that is past connect but before the add is incorrectly treated as pre-connect and cancelled; the test suite does not cover this interleaving.</sub>

</details>

- `packages/gptme-voice/src/gptme_voice/realtime/server.py` — Adds _extract_people_field, CallerIdentity fields, external-caller guidance, digest gating, and async pre-warm claim with cancellation and cleanup.
- `packages/gptme-voice/tests/test_server.py` — Adds tests for caller role/language parsing, external guidance, digest gating, and pre-warm claim race handling.

</details>

| `a68964da06fe` | 3/5 | 1 | llm | 2026-08-27 18:41 UTC |
| `14c6e8767162` | 3/5 | 1 | llm | 2026-08-27 18:59 UTC |
| `dfbd49c7df07` | 2/5 | 2 | llm | 2026-08-27 19:09 UTC |
| `c0419a6cf095` | 3/5 | 1 | llm | 2026-08-27 19:25 UTC |

</details>

<sub>Reviewed `f7de7e6829b7` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 234s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

packages/gptme-voice/src/gptme_voice/realtime/server.py
TimeToBuildBob: Fixed in 14c6e876 — the except-Exception branch now schedules _disconnect_realtime_client for the created client (mirroring the CancelledError path), with a regression test (test_prewarm_connect_failure_disconnects_half_open_client).

packages/gptme-voice/src/gptme_voice/realtime/server.py
```diff
@@ -1149,13 +1229,70 @@ async def _prewarm_for_inbound(self, from_number: str) -> None:
             # cold-path fallback won't re-inject the same transcript.
             await self._consume_recent_call(from_number)
             self._prewarm_sessions[from_number] = (client, time.monotonic())
+            # Guarantee eviction even if no further call ever arrives
+            # (_evict_stale_prewarms otherwise only runs on the next inbound).
+            asyncio.get_running_loop().call_later(
+                self._prewarm_ttl_seconds + 1, self._evict_stale_prewarms
+            )
             logger.info("Pre-warm ready for %s", from_number)
+        except asyncio.CancelledError:
```
TimeToBuildBob, line 1247: ❌ **P1** — When _claim_prewarm times out and cancels the in-flight pre-warm task, the task may have already stored its client in _prewarm_sessions at line 1231. The CancelledError handler at line 1238 disconnects the client and re-raises, but it does not remove the entry from _prewarm_sessions. The orphaned entry remains and can be popped by a later _claim_prewarm call (if within TTL) and used as a valid client even though it is being disconnected. This recreates the exact duplicate/provider-session race the PR intends to fix: a cold session starts, and the pre-warmed client either leaks or is later claimed in a broken state. The race window is small but real because the assignment is synchronous after an await, and a cancellation delivered after the assignment but before the task's next yield is ignored by the completed task. _(3/3 passes)_

TimeToBuildBob: Fixed in dfbd49c7 — the timeout branch now synchronously reaps any entry the completing task stored (new _reap_prewarm_entry: pop + disconnect), closing the no-op-cancel window. One correction to the finding: the CancelledError handler can never see a stored entry (store→completion is synchronous, so a post-store cancel is always a no-op) — the reap covers exactly that case. Two regression tests added.

packages/gptme-voice/src/gptme_voice/realtime/server.py
```diff
@@ -1149,13 +1229,76 @@ async def _prewarm_for_inbound(self, from_number: str) -> None:
             # cold-path fallback won't re-inject the same transcript.
             await self._consume_recent_call(from_number)
```
TimeToBuildBob, line 1239: ❌ **P1** — The pre-warm task deletes the resume-state file in `_prewarm_for_inbound` at line 1230 (`await self._consume_recent_call(from_number)`) immediately after `connect()` succeeds. If the claim times out (because the pre-warm took longer than `wait_seconds`), the resume state is already gone, so the cold-path `_build_session_bootstrap` finds no `RecentCallRecord` and starts a fresh greeting instead of resuming. This is a regression: previously the claim returned immediately without awaiting, so an unfinished pre-warm never consumed the resume state and the cold path could still resume. The new async wait introduces a window where a slow-but-successful pre-warm deletes the resume file and then the claim times out, orphaning the session and losing resume context. _(2/3 passes — pass 3 looked and disagreed)_

TimeToBuildBob: Fixed in c0419a6c — restructured the timeout branch so completed work is never discarded: an entry stored at the timeout boundary is claimed (it holds the consumed resume context); a connected-and-finalizing task (tracked via a new _prewarm_connected set) gets a ~1s extended wait then is claimed; only pre-connect tasks are cancelled, which the peek design guarantees have not touched resume state. Regression tests for both the boundary-claim and finalizing-wait paths.

packages/gptme-voice/src/gptme_voice/realtime/server.py
```diff
@@ -1145,17 +1229,117 @@ async def _prewarm_for_inbound(self, from_number: str) -> None:
...
+                                from_number,
+                            )
+                            return entry[0]
+                    logger.info(
+                        "Pre-warm for %s not ready in time — cancelling, using cold path",
+                        from_number,
+                    )
+                    # Pre-connect cancellation is safe: the peek design means
+                    # the resume state has not been consumed yet.
+                    task.cancel()
```
TimeToBuildBob, line 1356: ❌ **P1** — In _claim_prewarm, when a pre-warm task has connected (added to _prewarm_connected) and is finalizing by consuming the resume state in _consume_recent_call, the timeout path first waits an additional 1 second (line 1320). If that extension expires without the session being stored, control falls through to `task.cancel()` at line 1336. Cancellation at this point interrupts _consume_recent_call, which performs file deletion via `path.unlink(missing_ok=True)` (line 1431). If the unlink has already executed but the session has not yet been stored in _prewarm_sessions, the resume file is gone and the cold path in _build_session_bootstrap will greet the caller as fresh instead of resuming. The 1-second hard cap is an arbitrary bound that does not guarantee _consume_recent_call completes, especially for large or slow resume files. This loses the caller's resume context, a data-loss defect. _(2/3 passes — pass 1 looked and disagreed)_

TimeToBuildBob: Fixed in f7de7e68 — post-connect tasks are now never cancelled, closing the residual window: the finalize grace is longer and configurable (finalize_wait_seconds=5.0), and on expiry the claim falls back cold WITHOUT cancelling; the finished session is cleaned up by TTL eviction. Cancellation is now provably confined to pre-connect tasks (which haven't touched resume state, per the peek design). Regression test: test_claim_prewarm_never_cancels_finalizing_task_on_expiry.

TimeToBuildBob: Round-5 re-flag is a false positive — it assumes a cancelled task can continue executing past its suspension point. asyncio semantics: connect()→_prewarm_connected.add() has no yield between them, so a cancel delivered while the task is suspended at the connect await raises CancelledError AT that await (Task._must_cancel path — this holds even when the inner future already completed); the task can never proceed into _consume_recent_call after cancellation, and the CancelledError handler disconnects the created client. Store-after-cancel is likewise impossible: the timeout branch has no await between the boundary pop and task.cancel(), so the task cannot run in that window on a single-threaded loop. Disposing as FP; no code change.

TimeToBuildBob: False positive per the thread discussion: a cancelled asyncio task raises CancelledError at its current suspension point and cannot continue into _consume_recent_call; there is also no await between the boundary pop and task.cancel(), so store-after-cancel cannot interleave. Cancellation remains confined to pre-connect tasks.

PR #1528, opened 2026-08-27
feat(voice): capability-gated body_* tools — BodyAdapter + MAVSDK/PX4 support

TimeToBuildBob:
Implements BobBrain milestones 3+4 (body adapter + SITL-verified flight tools) — see ErikBjare/bob#730 and the bobbrain spec in Bob's workspace.

## What

Voice sessions can now command a physical body (quadcopter, future wheelbase, tabletop puck) through **direct, low-latency tools** — motion goals must not round-trip through subagent dispatch (`body_stop` especially).

- **`gptme_voice.body`** — `BodyAdapter` protocol implementing the Brain↔Body contract: the brain emits *goals* (goto/move/turn/stop), the adapter owns translation and safety. `NullAdapter` = tabletop puck (registers `body_status` only — a session on a desk ornament exposes no motion tools at all). `MavsdkAdapter` = PX4 vehicles via the high-level Action API only (goto_location/takeoff/land/hold/RTL — no offboard loops; **PX4's own failsafes remain the safety authority**, so a dropped brain link always leaves the vehicle safe).
- **Capability gating** — `body_tool_schemas(adapter)` emits function schemas only for the adapter's advertised capabilities (`move`/`rotate`/`altitude`); `SessionConfig.extra_tools` appends them to the realtime session.
- **Bridge routing** — `body_*` calls in `GptmeToolBridge.handle_function_call` with conservative envelope clamps (`GPTME_VOICE_BODY_MAX_ALT_M` default 30, `GPTME_VOICE_BODY_MAX_MOVE_M` default 50, yaw ±180°); all body errors reported to the model, never raised.
- **Wiring** — `GPTME_VOICE_BODY_URL` env: `null` | `mavsdk://udpin://0.0.0.0:14540` (SITL) | `mavsdk://serial:///dev/ttyUSB0:57600` (SiK radio). Optional dep `gptme-voice[body]`.

## Testing

- 22 unit tests (schema gating, routing, clamps, capability gate, error paths, NED/latlon geometry).
- **Live SITL integration test PASSED** against PX4 1.15 (jonasvautherin/px4-gazebo-headless): position fix → takeoff → relative move 5 m → hold → land, driven entirely through `handle_function_call`. Skips unless `BODY_SITL_URL` is set, so CI stays green without a simulator.
- Full gptme-voice suite: 237 passed.

## Not in this PR

- Vision pipeline (`gptme-vision-node`, concurrent PR) and its `look` tool wiring.
- xAI/Grok client tool declaration (inherits the OpenAI session path; untested with body tools).

### DIFF
--- a/packages/gptme-voice/src/gptme_voice/body/adapter.py
+++ b/packages/gptme-voice/src/gptme_voice/body/adapter.py
@@ -253,8 +253,7 @@ def body_tool_schemas(adapter: BodyAdapter | None) -> list[dict]:
                         "altitude_m": {
                             "type": "number",
                             "description": (
-                                "Hover altitude in meters above ground "
-                                "(default 2.5)."
+                                "Hover altitude in meters above ground (default 2.5)."
                             ),
                         }
                     },
@@ -288,6 +287,14 @@ def body_adapter_from_env() -> BodyAdapter | None:
     if url == "null":
         return NullAdapter()
     if url.startswith("mavsdk://"):
+        try:
+            __import__("mavsdk")
+        except ImportError:
+            logger.warning(
+                "GPTME_VOICE_BODY_URL uses mavsdk:// but the mavsdk dependency "
+                "is unavailable; install gptme-voice[body] (body disabled)"
+            )
+            return None
         from .mavsdk_adapter import MavsdkAdapter
 
         return MavsdkAdapter(url.removeprefix("mavsdk://"))
--- a/packages/gptme-voice/src/gptme_voice/body/adapter.py
+++ b/packages/gptme-voice/src/gptme_voice/body/adapter.py
@@ -1,0 +1,295 @@
+"""Body adapter protocol — the Brain↔Body contract.
+
+BobBrain design rule (bobbrain-spec, ErikBjare/bob#730): the brain emits
+*goals* (goto / move / turn / stop / dock), never control loops. A per-body
+adapter translates goals into the body's protocol and owns safety. The
+autopilot's own failsafes (e.g. PX4 offboard/link-loss behavior) remain the
+final authority — a dropped brain link must always leave the body safe.
+
+Capabilities gate which voice tools get registered for a session:
+a tabletop puck (NullAdapter, no capabilities) exposes no motion tools at
+all, so the model cannot try to fly a desk ornament.
+"""
+
+from __future__ import annotations
+
+import logging
+import os
+from typing import Any, Protocol, runtime_checkable
+
+logger = logging.getLogger(__name__)
+
+# Capability names
+CAP_MOVE = "move"  # horizontal translation (goto/move/return home)
+CAP_ROTATE = "rotate"  # yaw control
+CAP_ALTITUDE = "altitude"  # vertical control (takeoff/land/up-down)
+
+
+@runtime_checkable
+class BodyAdapter(Protocol):
+    """Minimal protocol every body implements.
+
+    Methods for capabilities an adapter does not advertise are never
+    called by the tool bridge; implementations may raise for them.
+    """
+
+    capabilities: set[str]
+    name: str
+
+    async def ensure_connected(self) -> None: ...
+
+    async def takeoff(self, altitude_m: float) -> dict[str, Any]: ...
+
+    async def land(self) -> dict[str, Any]: ...
+
+    async def goto(
+        self,
+        latitude_deg: float,
+        longitude_deg: float,
+        altitude_m: float | None,
+    ) -> dict[str, Any]: ...
+
+    async def move(
+        self, forward_m: float, right_m: float, up_m: float
+    ) -> dict[str, Any]: ...
+
+    async def turn(self, yaw_deg: float) -> dict[str, Any]: ...
+
+    async def stop(self) -> dict[str, Any]: ...
+
+    async def return_home(self) -> dict[str, Any]: ...
+
+    def telemetry(self) -> dict[str, Any]: ...
+
+
+class NullAdapter:
+    """Body with no locomotion — the tabletop puck.
+
+    Registers only ``body_status``; every motion method reports
+    unsupported (defense in depth — the bridge never routes these).
+    """
+
+    capabilities: set[str] = set()
+    name = "null"
+
+    async def ensure_connected(self) -> None:
+        return None
+
+    def telemetry(self) -> dict[str, Any]:
+        return {"body": self.name, "mobile": False}
+
+    async def _unsupported(self) -> dict[str, Any]:
+        return {"error": "This body has no locomotion."}
+
+    async def takeoff(self, altitude_m: float) -> dict[str, Any]:
+        return await self._unsupported()
+
+    async def land(self) -> dict[str, Any]:
+        return await self._unsupported()
+
+    async def goto(
+        self,
+        latitude_deg: float,
+        longitude_deg: float,
+        altitude_m: float | None,
+    ) -> dict[str, Any]:
+        return await self._unsupported()
+
+    async def move(
+        self, forward_m: float, right_m: float, up_m: float
+    ) -> dict[str, Any]:
+        return await self._unsupported()
+
+    async def turn(self, yaw_deg: float) -> dict[str, Any]:
+        return await self._unsupported()
+
+    async def stop(self) -> dict[str, Any]:
+        return await self._unsupported()
+
+    async def return_home(self) -> dict[str, Any]:
+        return await self._unsupported()
+
+
+def body_tool_schemas(adapter: BodyAdapter | None) -> list[dict]:
+    """Realtime-API function schemas for the adapter's capability set.
+
+    Returns [] when no adapter is configured, so callers can always
+    ``extend()`` the session tool list unconditionally.
+    """
+    if adapter is None:
+        return []
+
+    caps = adapter.capabilities
+    tools: list[dict] = [
+        {
+            "type": "function",
+            "name": "body_status",
+            "description": (
+                "Read the physical body's current state: position, altitude, "
+                "battery, heading, flight mode. Use it before motion commands "
+                "and whenever the caller asks where the body is or how it is "
+                "doing."
+            ),
+            "parameters": {"type": "object", "properties": {}},
+        }
+    ]
+
+    if CAP_MOVE in caps:
+        tools.append(
+            {
+                "type": "function",
+                "name": "body_stop",
+                "description": (
+                    "IMMEDIATELY halt all body motion and hold position. "
+                    "Call this first, without asking, whenever the caller "
+                    "says stop/wait/freeze or anything sounds wrong."
+                ),
+                "parameters": {"type": "object", "properties": {}},
+            }
+        )
+        tools.append(
+            {
+                "type": "function",
+                "name": "body_return_home",
+                "description": (
+                    "Send the body back to its home/launch position "
+                    "(quadcopter: return-to-launch and land)."
+                ),
+                "parameters": {"type": "object", "properties": {}},
+            }
+        )
+        tools.append(
+            {
+                "type": "function",
+                "name": "body_move",
+                "description": (
+                    "Move the body a relative distance in meters from where "
+                    "it is now, in its own frame: forward/backward, "
+                    "right/left, up/down. Use small distances (1-5 m) and "
+                    "check body_status after."
+                ),
+                "parameters": {
+                    "type": "object",
+                    "properties": {
+                        "forward_m": {
+                            "type": "number",
+                            "description": "Meters forward (negative = backward).",
+                        },
+                        "right_m": {
+                            "type": "number",
+                            "description": "Meters right (negative = left).",
+                        },
+                        "up_m": {
+                            "type": "number",
+                            "description": "Meters up (negative = down). 0 to hold altitude.",
+                        },
+                    },
+                },
+            }
+        )
+        tools.append(
+            {
+                "type": "function",
+                "name": "body_goto",
+                "description": (
+                    "Fly/drive the body to absolute GPS coordinates. Only "
+                    "use coordinates you got from body_status or the caller "
+                    "— never invent them."
+                ),
+                "parameters": {
+                    "type": "object",
+                    "properties": {
+                        "latitude_deg": {"type": "number"},
+                        "longitude_deg": {"type": "number"},
+                        "altitude_m": {
+                            "type": "number",
+                            "description": (
+                                "Target altitude in meters above the home "
+                                "position. Omit to keep current altitude."
+                            ),
+                        },
+                    },
+                    "required": ["latitude_deg", "longitude_deg"],
+                },
+            }
+        )
+
+    if CAP_ROTATE in caps:
+        tools.append(
+            {
+                "type": "function",
+                "name": "body_turn",
+                "description": (
+                    "Rotate the body in place by a relative angle in degrees "
+                    "(positive = clockwise/right, negative = "
+                    "counter-clockwise/left)."
+                ),
+                "parameters": {
+                    "type": "object",
+                    "properties": {
+                        "yaw_deg": {
+                            "type": "number",
+                            "description": "Relative turn in degrees, -180..180.",
+                        }
+                    },
+                    "required": ["yaw_deg"],
+                },
+            }
+        )
+
+    if CAP_ALTITUDE in caps:
+        tools.append(
+            {
+                "type": "function",
+                "name": "body_takeoff",
+                "description": (
+                    "Arm and take off to a hover at the given altitude. Only "
+                    "do this when the caller explicitly asks to take off or "
+                    "fly."
+                ),
+                "parameters": {
+                    "type": "object",
+                    "properties": {
+                        "altitude_m": {
+                            "type": "number",
+                            "description": (
+                                "Hover altitude in meters above ground "
+                                "(default 2.5)."
+                            ),
+                        }
+                    },
+                },
+            }
+        )
+        tools.append(
+            {
+                "type": "function",
+                "name": "body_land",
+                "description": "Land the body at its current position and disarm.",
+                "parameters": {"type": "object", "properties": {}},
+            }
+        )
+
+    return tools
+
+
+def body_adapter_from_env() -> BodyAdapter | None:
+    """Build a body adapter from ``GPTME_VOICE_BODY_URL``.
+
+    - unset/empty  -> None (no body tools registered)
+    - ``null``     -> NullAdapter (body_status only; tabletop puck)
+    - ``mavsdk://<system_address>`` -> MavsdkAdapter, e.g.
+      ``mavsdk://udpin://0.0.0.0:14540`` (SITL) or
+      ``mavsdk://serial:///dev/ttyUSB0:57600`` (SiK radio)
+    """
+    url = os.environ.get("GPTME_VOICE_BODY_URL", "").strip()
+    if not url:
+        return None
+    if url == "null":
+        return NullAdapter()
+    if url.startswith("mavsdk://"):
+        from .mavsdk_adapter import MavsdkAdapter
+
+        return MavsdkAdapter(url.removeprefix("mavsdk://"))
+    logger.warning("Unknown GPTME_VOICE_BODY_URL scheme: %s (ignored)", url)
+    return None
--- a/packages/gptme-voice/src/gptme_voice/body/adapter.py
+++ b/packages/gptme-voice/src/gptme_voice/body/adapter.py
@@ -38,6 +38,8 @@ class BodyAdapter(Protocol):
 
     async def ensure_connected(self) -> None: ...
 
+    async def close(self) -> None: ...
+
     async def takeoff(self, altitude_m: float) -> dict[str, Any]: ...
 
     async def land(self) -> dict[str, Any]: ...
@@ -73,6 +75,9 @@ class NullAdapter:
     name = "null"
 
     async def ensure_connected(self) -> None:
+        return None
+
+    async def close(self) -> None:
         return None
 
     def telemetry(self) -> dict[str, Any]:
--- a/packages/gptme-voice/src/gptme_voice/body/mavsdk_adapter.py
+++ b/packages/gptme-voice/src/gptme_voice/body/mavsdk_adapter.py
@@ -1,0 +1,233 @@
+"""MAVSDK body adapter — PX4 vehicles (X500 v2 quad, SITL).
+
+Translates BodyAdapter goals into MAVSDK action calls. Deliberately uses
+only the high-level Action API (goto_location, takeoff, land, hold, RTL) —
+no offboard control loops. PX4's own failsafes remain the safety authority:
+if the brain link drops mid-goal, the vehicle keeps its autopilot behavior.
+
+Validated against PX4 SITL (jonasvautherin/px4-gazebo-headless);
+see projects/bobbody-flight/ in the Bob workspace for the harness.
+
+Requires the optional ``mavsdk`` dependency: ``gptme-voice[body]``.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import logging
+import math
+from typing import Any
+
+logger = logging.getLogger(__name__)
+
+_EARTH_M_PER_DEG_LAT = 111_320.0
+
+
+def body_to_ned(
+    forward_m: float, right_m: float, yaw_deg: float
+) -> tuple[float, float]:
+    """Rotate a body-frame (forward, right) offset into NED (north, east).
+
+    ``yaw_deg`` is heading from north, clockwise-positive (NED convention).
+    """
+    yaw = math.radians(yaw_deg)
+    north = forward_m * math.cos(yaw) - right_m * math.sin(yaw)
+    east = forward_m * math.sin(yaw) + right_m * math.cos(yaw)
+    return north, east
+
+
+def offset_latlon(
+    latitude_deg: float, longitude_deg: float, north_m: float, east_m: float
+) -> tuple[float, float]:
+    """Offset a lat/lon by meters north/east (small-distance approximation)."""
+    dlat = north_m / _EARTH_M_PER_DEG_LAT
+    denom = _EARTH_M_PER_DEG_LAT * math.cos(math.radians(latitude_deg))
+    dlon = east_m / denom if abs(denom) > 1e-9 else 0.0
+    return latitude_deg + dlat, longitude_deg + dlon
+
+
+class MavsdkAdapter:
+    """BodyAdapter for PX4/MAVSDK vehicles."""
+
+    capabilities = {"move", "rotate", "altitude"}
+    name = "mavsdk"
+
+    def __init__(self, system_address: str, connect_timeout_s: float = 20.0):
+        self.system_address = system_address
+        self.connect_timeout_s = connect_timeout_s
+        self._system: Any = None
+        self._connected = False
+        self._lock = asyncio.Lock()
+        self._telemetry_tasks: list[asyncio.Task] = []
+        # Telemetry caches (filled by background subscriptions)
+        self._position: dict[str, float] = {}
+        self._heading_deg: float | None = None
+        self._battery: dict[str, float] = {}
+        self._in_air: bool | None = None
+        self._flight_mode: str | None = None
+        self._armed: bool | None = None
+
+    async def ensure_connected(self) -> None:
+        async with self._lock:
+            if self._connected:
+                return
+            from mavsdk import System
+
+            system = System()
+            logger.info("MavsdkAdapter connecting to %s", self.system_address)
+            await system.connect(system_address=self.system_address)
+
+            async def _wait_connected() -> None:
+                async for state in system.core.connection_state():
+                    if state.is_connected:
+                        return
+
+            await asyncio.wait_for(_wait_connected(), timeout=self.connect_timeout_s)
+            self._system = system
+            self._start_telemetry_cache()
+            self._connected = True
+            logger.info("MavsdkAdapter connected")
+
+    def _start_telemetry_cache(self) -> None:
+        system = self._system
+
+        async def _position() -> None:
+            async for p in system.telemetry.position():
+                self._position = {
+                    "latitude_deg": p.latitude_deg,
+                    "longitude_deg": p.longitude_deg,
+                    "absolute_altitude_m": p.absolute_altitude_m,
+                    "relative_altitude_m": p.relative_altitude_m,
+                }
+
+        async def _heading() -> None:
+            async for h in system.telemetry.heading():
+                self._heading_deg = h.heading_deg
+
+        async def _battery() -> None:
+            async for b in system.telemetry.battery():
+                self._battery = {
+                    "voltage_v": round(b.voltage_v, 2),
+                    "remaining_percent": round(b.remaining_percent, 1),
+                }
+
+        async def _in_air() -> None:
+            async for in_air in system.telemetry.in_air():
+                self._in_air = in_air
+
+        async def _flight_mode() -> None:
+            async for mode in system.telemetry.flight_mode():
+                self._flight_mode = str(mode)
+
+        async def _armed() -> None:
+            async for armed in system.telemetry.armed():
+                self._armed = armed
+
+        for coro in (
+            _position(),
+            _heading(),
+            _battery(),
+            _in_air(),
+            _flight_mode(),
+            _armed(),
+        ):
+            self._telemetry_tasks.append(asyncio.create_task(coro))
+
+    async def close(self) -> None:
+        for task in self._telemetry_tasks:
+            task.cancel()
+        self._telemetry_tasks.clear()
+        self._connected = False
+        self._system = None
+
+    # -- BodyAdapter interface -------------------------------------------
+
+    def telemetry(self) -> dict[str, Any]:
+        return {
+            "body": self.name,
+            "mobile": True,
+            "connected": self._connected,
+            "position": self._position or None,
+            "heading_deg": self._heading_deg,
+            "battery": self._battery or None,
+            "in_air": self._in_air,
+            "flight_mode": self._flight_mode,
+            "armed": self._armed,
+        }
+
+    async def takeoff(self, altitude_m: float) -> dict[str, Any]:
+        system = self._system
+        await system.action.set_takeoff_altitude(altitude_m)
+        await system.action.arm()
+        await system.action.takeoff()
+        return {
+            "status": "taking_off",
+            "target_altitude_m": altitude_m,
+            "message": "Armed and taking off. Check body_status for altitude.",
+        }
+
+    async def land(self) -> dict[str, Any]:
+        await self._system.action.land()
+        return {"status": "landing"}
+
+    async def stop(self) -> dict[str, Any]:
+        await self._system.action.hold()
+        return {"status": "holding", "message": "Motion stopped; holding position."}
+
+    async def return_home(self) -> dict[str, Any]:
+        await self._system.action.return_to_launch()
+        return {"status": "returning_home"}
+
+    async def goto(
+        self,
+        latitude_deg: float,
+        longitude_deg: float,
+        altitude_m: float | None,
+    ) -> dict[str, Any]:
+        pos = self._position
+        if not pos:
+            return {"error": "No position fix yet; cannot goto."}
+        home_abs = pos["absolute_altitude_m"] - pos["relative_altitude_m"]
+        target_abs = (
+            home_abs + altitude_m
+            if altitude_m is not None
+            else pos["absolute_altitude_m"]
+        )
+        yaw = self._heading_deg if self._heading_deg is not None else float("nan")
+        await self._system.action.goto_location(
+            latitude_deg, longitude_deg, target_abs, yaw
+        )
+        return {
+            "status": "en_route",
+            "target": {
+                "latitude_deg": latitude_deg,
+                "longitude_deg": longitude_deg,
+                "altitude_m": altitude_m,
+            },
+        }
+
+    async def move(
+        self, forward_m: float, right_m: float, up_m: float
+    ) -> dict[str, Any]:
+        pos = self._position
+        if not pos:
+            return {"error": "No position fix yet; cannot move."}
+        yaw = self._heading_deg or 0.0
+        north, east = body_to_ned(forward_m, right_m, yaw)
+        lat, lon = offset_latlon(pos["latitude_deg"], pos["longitude_deg"], north, east)
+        target_rel = pos["relative_altitude_m"] + up_m
+        return await self.goto(lat, lon, target_rel)
+
+    async def turn(self, yaw_deg: float) -> dict[str, Any]:
+        pos = self._position
+        if not pos:
+            return {"error": "No position fix yet; cannot turn."}
+        current = self._heading_deg or 0.0
+        target_yaw = (current + yaw_deg + 180.0) % 360.0 - 180.0
+        await self._system.action.goto_location(
+            pos["latitude_deg"],
+            pos["longitude_deg"],
+            pos["absolute_altitude_m"],
+            target_yaw,
+        )
+        return {"status": "turning", "target_heading_deg": round(target_yaw, 1)}
--- a/packages/gptme-voice/src/gptme_voice/body/mavsdk_adapter.py
+++ b/packages/gptme-voice/src/gptme_voice/body/mavsdk_adapter.py
@@ -219,15 +219,18 @@ class MavsdkAdapter:
     async def _run_action(self, awaitable: Any) -> Any:
         try:
             return await asyncio.wait_for(awaitable, timeout=self.command_timeout_s)
-        except TimeoutError:
+        except asyncio.TimeoutError:
             # The MAVSDK command is already on the vehicle. Dropping the
             # adapter here would cancel telemetry and invite a reconnect
             # while land/takeoff/goto is still executing.
+            # Python 3.10: wait_for raises asyncio.TimeoutError, distinct from
+            # builtin TimeoutError. Re-raise TimeoutError so callers and tests
+            # can catch one type on every supported Python.
             logger.warning(
                 "MAVSDK action timed out after %.1fs; leaving adapter connected",
                 self.command_timeout_s,
             )
-            raise
+            raise TimeoutError from None
 
     # -- BodyAdapter interface -------------------------------------------
 
--- a/packages/gptme-voice/src/gptme_voice/body/mavsdk_adapter.py
+++ b/packages/gptme-voice/src/gptme_voice/body/mavsdk_adapter.py
@@ -57,7 +57,8 @@ class MavsdkAdapter:
         self.connect_timeout_s = connect_timeout_s
         self._system: Any = None
         self._connected = False
-        self._lock = asyncio.Lock()
+        self._connection_lock = asyncio.Lock()
+        self._command_lock = asyncio.Lock()
         self._telemetry_tasks: list[asyncio.Task] = []
         # Telemetry caches (filled by background subscriptions)
         self._position: dict[str, float] = {}
@@ -68,9 +69,10 @@ class MavsdkAdapter:
         self._armed: bool | None = None
 
     async def ensure_connected(self) -> None:
-        async with self._lock:
+        async with self._connection_lock:
             if self._connected:
                 return
+            await self._cancel_telemetry_tasks()
             from mavsdk import System
 
             system = System()
@@ -123,21 +125,56 @@ class MavsdkAdapter:
             async for armed in system.telemetry.armed():
                 self._armed = armed
 
-        for coro in (
-            _position(),
-            _heading(),
-            _battery(),
-            _in_air(),
-            _flight_mode(),
-            _armed(),
-        ):
-            self._telemetry_tasks.append(asyncio.create_task(coro))
+        async def _watch(stream_name: str, subscription: Any) -> None:
+            try:
+                await subscription
+            except asyncio.CancelledError:
+                raise
+            except Exception:  # noqa: BLE001 - MAVSDK stream errors vary by backend
+                logger.exception("MAVSDK %s telemetry stream failed", stream_name)
+                self._mark_disconnected()
+            else:
+                logger.error("MAVSDK %s telemetry stream ended", stream_name)
+                self._mark_disconnected()
+
+        subscriptions = {
+            "position": _position(),
+            "heading": _heading(),
+            "battery": _battery(),
+            "in_air": _in_air(),
+            "flight_mode": _flight_mode(),
+            "armed": _armed(),
+        }
+        self._telemetry_tasks.extend(
+            asyncio.create_task(_watch(name, subscription))
+            for name, subscription in subscriptions.items()
+        )
+
+    def _mark_disconnected(self) -> None:
+        self._connected = False
+        self._position = {}
+        self._heading_deg = None
+        self._battery = {}
+        self._in_air = None
+        self._flight_mode = None
+        self._armed = None
+        current = asyncio.current_task()
+        for task in self._telemetry_tasks:
+            if task is not current:
+                task.cancel()
+        self._telemetry_tasks.clear()
+
+    async def _cancel_telemetry_tasks(self) -> None:
+        tasks = list(self._telemetry_tasks)
+        for task in tasks:
+            task.cancel()
+        if tasks:
+            await asyncio.gather(*tasks, return_exceptions=True)
+        self._telemetry_tasks.clear()
 
     async def close(self) -> None:
-        for task in self._telemetry_tasks:
-            task.cancel()
-        self._telemetry_tasks.clear()
-        self._connected = False
+        await self._cancel_telemetry_tasks()
+        self._mark_disconnected()
         self._system = None
 
     # -- BodyAdapter interface -------------------------------------------
@@ -156,29 +193,45 @@ class MavsdkAdapter:
         }
 
     async def takeoff(self, altitude_m: float) -> dict[str, Any]:
-        system = self._system
-        await system.action.set_takeoff_altitude(altitude_m)
-        await system.action.arm()
-        await system.action.takeoff()
-        return {
-            "status": "taking_off",
-            "target_altitude_m": altitude_m,
-            "message": "Armed and taking off. Check body_status for altitude.",
-        }
+        async with self._command_lock:
+            system = self._system
+            await system.action.set_takeoff_altitude(altitude_m)
+            await system.action.arm()
+            await system.action.takeoff()
+            return {
+                "status": "taking_off",
+                "target_altitude_m": altitude_m,
+                "message": "Armed and taking off. Check body_status for altitude.",
+            }
 
     async def land(self) -> dict[str, Any]:
-        await self._system.action.land()
-        return {"status": "landing"}
+        async with self._command_lock:
+            await self._system.action.land()
+            return {"status": "landing"}
 
     async def stop(self) -> dict[str, Any]:
-        await self._system.action.hold()
-        return {"status": "holding", "message": "Motion stopped; holding position."}
+        async with self._command_lock:
+            await self._system.action.hold()
+            return {
+                "status": "holding",
+                "message": "Motion stopped; holding position.",
+            }
 
     async def return_home(self) -> dict[str, Any]:
-        await self._system.action.return_to_launch()
-        return {"status": "returning_home"}
+        async with self._command_lock:
+            await self._system.action.return_to_launch()
+            return {"status": "returning_home"}
 
     async def goto(
+        self,
+        latitude_deg: float,
+        longitude_deg: float,
+        altitude_m: float | None,
+    ) -> dict[str, Any]:
+        async with self._command_lock:
+            return await self._goto_unlocked(latitude_deg, longitude_deg, altitude_m)
+
+    async def _goto_unlocked(
         self,
         latitude_deg: float,
         longitude_deg: float,
@@ -209,25 +262,32 @@ class MavsdkAdapter:
     async def move(
         self, forward_m: float, right_m: float, up_m: float
     ) -> dict[str, Any]:
-        pos = self._position
-        if not pos:
-            return {"error": "No position fix yet; cannot move."}
-        yaw = self._heading_deg or 0.0
-        north, east = body_to_ned(forward_m, right_m, yaw)
-        lat, lon = offset_latlon(pos["latitude_deg"], pos["longitude_deg"], north, east)
-        target_rel = pos["relative_altitude_m"] + up_m
-        return await self.goto(lat, lon, target_rel)
+        async with self._command_lock:
+            pos = self._position
+            if not pos:
+                return {"error": "No position fix yet; cannot move."}
+            yaw = self._heading_deg or 0.0
+            north, east = body_to_ned(forward_m, right_m, yaw)
+            lat, lon = offset_latlon(
+                pos["latitude_deg"], pos["longitude_deg"], north, east
+            )
+            target_rel = max(1.0, pos["relative_altitude_m"] + up_m)
+            return await self._goto_unlocked(lat, lon, target_rel)
 
     async def turn(self, yaw_deg: float) -> dict[str, Any]:
-        pos = self._position
-        if not pos:
-            return {"error": "No position fix yet; cannot turn."}
-        current = self._heading_deg or 0.0
-        target_yaw = (current + yaw_deg + 180.0) % 360.0 - 180.0
-        await self._system.action.goto_location(
-            pos["latitude_deg"],
-            pos["longitude_deg"],
-            pos["absolute_altitude_m"],
-            target_yaw,
-        )
-        return {"status": "turning", "target_heading_deg": round(target_yaw, 1)}
+        async with self._command_lock:
+            pos = self._position
+            if not pos:
+                return {"error": "No position fix yet; cannot turn."}
+            current = self._heading_deg or 0.0
+            target_yaw = (current + yaw_deg + 180.0) % 360.0 - 180.0
+            await self._system.action.goto_location(
+                pos["latitude_deg"],
+                pos["longitude_deg"],
+                pos["absolute_altitude_m"],
+                target_yaw,
+            )
+            return {
+                "status": "turning",
+                "target_heading_deg": round(target_yaw, 1),
+            }
--- a/packages/gptme-voice/src/gptme_voice/body/mavsdk_adapter.py
+++ b/packages/gptme-voice/src/gptme_voice/body/mavsdk_adapter.py
@@ -52,11 +52,19 @@ class MavsdkAdapter:
     capabilities = {"move", "rotate", "altitude"}
     name = "mavsdk"
 
-    def __init__(self, system_address: str, connect_timeout_s: float = 20.0):
+    def __init__(
+        self,
+        system_address: str,
+        connect_timeout_s: float = 5.0,
+        command_timeout_s: float = 10.0,
+    ):
         self.system_address = system_address
         self.connect_timeout_s = connect_timeout_s
+        self.command_timeout_s = command_timeout_s
         self._system: Any = None
         self._connected = False
+        # Command calls are serialized, and reconnect takes this lock after the
+        # connection lock so a new System cannot replace one in active use.
         self._connection_lock = asyncio.Lock()
         self._command_lock = asyncio.Lock()
         self._telemetry_tasks: list[asyncio.Task] = []
@@ -72,23 +80,36 @@ class MavsdkAdapter:
         async with self._connection_lock:
             if self._connected:
                 return
-            await self._cancel_telemetry_tasks()
-            from mavsdk import System
+            async with self._command_lock:
+                if self._connected:
+                    return
+                await self._cancel_telemetry_tasks()
+                await self._release_system()
+                from mavsdk import System  # type: ignore[import-not-found]
 
-            system = System()
-            logger.info("MavsdkAdapter connecting to %s", self.system_address)
-            await system.connect(system_address=self.system_address)
+                system = System()
+                logger.info("MavsdkAdapter connecting to %s", self.system_address)
+                try:
+                    await asyncio.wait_for(
+                        system.connect(system_address=self.system_address),
+                        timeout=self.connect_timeout_s,
+                    )
 
-            async def _wait_connected() -> None:
-                async for state in system.core.connection_state():
-                    if state.is_connected:
-                        return
+                    async def _wait_connected() -> None:
+                        async for state in system.core.connection_state():
+                            if state.is_connected:
+                                return
 
-            await asyncio.wait_for(_wait_connected(), timeout=self.connect_timeout_s)
-            self._system = system
-            self._start_telemetry_cache()
-            self._connected = True
-            logger.info("MavsdkAdapter connected")
+                    await asyncio.wait_for(
+                        _wait_connected(), timeout=self.connect_timeout_s
+                    )
+                except BaseException:
+                    self._stop_system(system)
+                    raise
+                self._system = system
+                self._start_telemetry_cache()
+                self._connected = True
+                logger.info("MavsdkAdapter connected")
 
     def _start_telemetry_cache(self) -> None:
         system = self._system
@@ -125,9 +146,9 @@ class MavsdkAdapter:
             async for armed in system.telemetry.armed():
                 self._armed = armed
 
-        async def _watch(stream_name: str, subscription: Any) -> None:
+        async def _watch(stream_name: str, subscription_factory: Any) -> None:
             try:
-                await subscription
+                await subscription_factory()
             except asyncio.CancelledError:
                 raise
             except Exception:  # noqa: BLE001 - MAVSDK stream errors vary by backend
@@ -138,16 +159,16 @@ class MavsdkAdapter:
                 self._mark_disconnected()
 
         subscriptions = {
-            "position": _position(),
-            "heading": _heading(),
-            "battery": _battery(),
-            "in_air": _in_air(),
-            "flight_mode": _flight_mode(),
-            "armed": _armed(),
+            "position": _position,
+            "heading": _heading,
+            "battery": _battery,
+            "in_air": _in_air,
+            "flight_mode": _flight_mode,
+            "armed": _armed,
         }
         self._telemetry_tasks.extend(
-            asyncio.create_task(_watch(name, subscription))
-            for name, subscription in subscriptions.items()
+            asyncio.create_task(_watch(name, subscription_factory))
+            for name, subscription_factory in subscriptions.items()
         )
 
     def _mark_disconnected(self) -> None:
@@ -172,10 +193,31 @@ class MavsdkAdapter:
             await asyncio.gather(*tasks, return_exceptions=True)
         self._telemetry_tasks.clear()
 
+    @staticmethod
+    def _stop_system(system: Any) -> None:
+        """Release MAVSDK-Python's subprocess and gRPC resources.
+
+        MAVSDK-Python 2.x has no public close method. Its destructor invokes
+        this cleanup hook, but garbage collection is too late for reconnects.
+        """
+        stop = getattr(system, "_stop_mavsdk_server", None)
+        if callable(stop):
+            stop()
+
+    async def _release_system(self) -> None:
+        system, self._system = self._system, None
+        if system is not None:
+            self._stop_system(system)
+
     async def close(self) -> None:
-        await self._cancel_telemetry_tasks()
-        self._mark_disconnected()
-        self._system = None
+        async with self._connection_lock:
+            async with self._command_lock:
+                await self._cancel_telemetry_tasks()
+                self._mark_disconnected()
+                await self._release_system()
+
+    async def _run_action(self, awaitable: Any) -> Any:
+        return await asyncio.wait_for(awaitable, timeout=self.command_timeout_s)
 
     # -- BodyAdapter interface -------------------------------------------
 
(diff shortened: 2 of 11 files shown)


TimeToBuildBob:
CI is fully green, including package coverage, typecheck, and integration tests. No review feedback has landed yet; waiting for maintainer review.

packages/gptme-voice/src/gptme_voice/realtime/tool_bridge.py
```diff
@@ -134,6 +138,13 @@ def __init__(
         self.on_hangup = on_hangup
         self.on_handoff = on_handoff
         self.transcript_provider = transcript_provider
+        self.body_adapter = body_adapter
+        self.body_max_altitude_m = float(
```
TimeToBuildBob: ❌ **P1** — The constructor reads GPTME_VOICE_BODY_MAX_ALT_M and GPTME_VOICE_BODY_MAX_MOVE_M via float() without validation. If either env var is set to a non-numeric value or an empty string, float() raises ValueError, which propagates out of GptmeToolBridge.__init__. Since the bridge is constructed on every session start (e.g., in handle_twilio_websocket), a malformed env config crashes the voice server on any new call instead of falling back to the defaults. An operator who mistakenly sets GPTME_VOICE_BODY_MAX_ALT_M=abc (e.g., from a config template) will take down the service until they fix the env. Should parse defensively, e.g., wrap in try/except and log a warning, matching the existing _parse_env_int pattern.

packages/gptme-voice/src/gptme_voice/realtime/tool_bridge.py
```diff
@@ -911,4 +922,85 @@ async def handle_function_call(self, name: str, arguments: dict) -> dict:
...
+                return await adapter.land()
+            if name == "body_move" and "move" in caps:
+                forward = self._clamp(
+                    float(arguments.get("forward_m", 0.0)), self.body_max_move_m
+                )
+                right = self._clamp(
+                    float(arguments.get("right_m", 0.0)), self.body_max_move_m
+                )
+                up = self._clamp(
+                    float(arguments.get("up_m", 0.0)), self.body_max_altitude_m
```
TimeToBuildBob, line 1021: ❌ **P1** — The body_move handler clamps the requested vertical delta `up_m` to `self.body_max_altitude_m`, but this is a relative move, not an absolute altitude limit. If the vehicle is already at altitude h and the model requests up_m up to the max, the resulting altitude h + up_m can exceed the configured maximum. For example, with GPTME_VOICE_BODY_MAX_ALT_M=30 and the vehicle at 25 m, a single body_move with up_m=30 is clamped to 30, producing a target of 55 m — above the 30 m cap. A sequence of such moves can climb arbitrarily high, defeating the intent of the safety envelope and relying solely on the autopilot geofence. The clamp should bound the absolute altitude, not the per-move delta.

packages/gptme-voice/src/gptme_voice/body/mavsdk_adapter.py
```diff
@@ -0,0 +1,233 @@
...
+    async def move(
+        self, forward_m: float, right_m: float, up_m: float
+    ) -> dict[str, Any]:
+        pos = self._position
+        if not pos:
+            return {"error": "No position fix yet; cannot move."}
+        yaw = self._heading_deg or 0.0
+        north, east = body_to_ned(forward_m, right_m, yaw)
+        lat, lon = offset_latlon(pos["latitude_deg"], pos["longitude_deg"], north, east)
+        target_rel = pos["relative_altitude_m"] + up_m
```
TimeToBuildBob: ❌ **P1** — In MavsdkAdapter.move, the target relative altitude is computed as the sum of the current relative_altitude_m and the requested up_m. The tool bridge clamps up_m to ±body_max_altitude_m (default 30), but does not clamp the resulting target_rel to a minimum. If the vehicle is at a low altitude (e.g., 2 m) and the model requests up_m = -30, target_rel becomes -28 m. This is passed to goto, which sets an absolute altitude of home_abs + (-28), i.e., 28 m below the home altitude. The drone would be commanded to fly below ground, likely crashing. A single body_move call can trigger this. Fix by clamping target_rel to at least 1.0 m (or clamp up_m based on current altitude).

packages/gptme-voice/src/gptme_voice/body/mavsdk_adapter.py
```diff
@@ -0,0 +1,233 @@
...
+            self._system = system
+            self._start_telemetry_cache()
+            self._connected = True
+            logger.info("MavsdkAdapter connected")
+
+    def _start_telemetry_cache(self) -> None:
+        system = self._system
+
+        async def _position() -> None:
+            async for p in system.telemetry.position():
```
TimeToBuildBob, line 118: ❌ **P1** — The background telemetry tasks started in _start_telemetry_cache each iterate an unbounded async for over a MAVSDK telemetry stream with no try/except. If the MAVSDK connection drops, these streams raise an exception (e.g., ConnectionError). The tasks are created via asyncio.create_task and never awaited, so the exception becomes an unhandled task exception that is only logged. The telemetry caches (position, heading, etc.) freeze at their last values, so body_status returns stale telemetry indefinitely. For a drone, stale telemetry is dangerous: the model may believe the vehicle is hovering when it has actually fallen. The tasks should catch exceptions, mark the adapter disconnected, and optionally restart the subscription.

packages/gptme-voice/src/gptme_voice/body/mavsdk_adapter.py
```diff
@@ -0,0 +1,233 @@
...
+        return {
+            "status": "en_route",
+            "target": {
+                "latitude_deg": latitude_deg,
+                "longitude_deg": longitude_deg,
+                "altitude_m": altitude_m,
+            },
+        }
+
+    async def move(
```
TimeToBuildBob, line 328: ❌ **P1** — MavsdkAdapter is a singleton shared across all voice sessions: VoiceServer.__init__ stores one instance in self.body_adapter and every GptmeToolBridge receives the same object. Its motion methods (move, goto, turn, takeoff, land, stop, return_home) are not guarded by any lock; only ensure_connected uses _lock to serialize the connection. Since the server supports multiple concurrent Twilio streams, two callers can invoke body_move (or goto/turn) at the same time, and both calls will issue MAVSDK Action commands concurrently. MAVSDK's high-level action API is not designed for concurrent invocation; interleaved commands (e.g., two conflicting goto_location targets) can cause the vehicle to receive contradictory instructions, potentially leading to erratic flight or loss of control. This is a safety-critical concurrency bug.

packages/gptme-voice/tests/test_mavsdk_sitl_integration.py
```diff
@@ -0,0 +1,77 @@
...
+import pytest
+
+SITL_URL = os.environ.get("BODY_SITL_URL")
+
+pytestmark = pytest.mark.skipif(
+    not SITL_URL, reason="BODY_SITL_URL not set (needs a running PX4 SITL)"
+)
+
+
+@pytest.mark.timeout(180)
```
TimeToBuildBob: ⚠️ **P2** — The test is decorated with @pytest.mark.timeout(180), but pytest-timeout is not listed in the test extra (pyproject only has pytest>=7.0). If pytest-timeout is not installed, the marker is ignored, and the test has no effective time limit. If the SITL connection hangs (e.g., no position fix, a stuck action call), the test will run indefinitely, hanging the CI job that runs with BODY_SITL_URL set. Should either add pytest-timeout to the test dependencies or implement an explicit asyncio.wait_for around the scenario.

packages/gptme-voice/src/gptme_voice/body/adapter.py
```diff
@@ -0,0 +1,295 @@
...
+      ``mavsdk://udpin://0.0.0.0:14540`` (SITL) or
+      ``mavsdk://serial:///dev/ttyUSB0:57600`` (SiK radio)
+    """
+    url = os.environ.get("GPTME_VOICE_BODY_URL", "").strip()
+    if not url:
+        return None
+    if url == "null":
+        return NullAdapter()
+    if url.startswith("mavsdk://"):
+        from .mavsdk_adapter import MavsdkAdapter
```
TimeToBuildBob, line 303: ⚠️ **P2** — body_adapter_from_env() imports MavsdkAdapter lazily when GPTME_VOICE_BODY_URL starts with "mavsdk://". If mavsdk is not installed (the optional gptme-voice[body] extra), this raises ModuleNotFoundError, which propagates out of VoiceServer.__init__ and prevents the server from starting. Unlike the unknown-scheme case (which warns and returns None), a missing dependency crashes the whole server. Also, if the mavsdk module is installed but the import itself fails for another reason, it would crash too. The function should catch ImportError (or Exception) and log a warning, returning None, so a misconfigured env doesn't take down the service.

TimeToBuildBob:
## 🤖 AI code review

Found 3 mismatches between the agent tool schemas and the BodyAdapter API in adapter.py. Two are cases where a tool description/schema advertises optional or partially-optional parameters (body_move's args, body_takeoff's altitude_m) but the adapter methods require them, guaranteeing TypeErrors at runtime when the agent follows the descriptions. The third is a protocol/implementation signature mismatch: telemetry is declared synchronous in the protocol but implemented async everywhere, which would yield unawaited coroutines for any callers using the protocol. Fixes: add the missing `"required"`/`"default"` (or give the Python methods matching defaults), and make the protocol's telemetry async.

> ⚠️ **1 generated file excluded and NOT reviewed:** ``uv.lock`` (+25/-1 lines, 4KB). Machine-generated output is removed from the diff before review — from what the model sees and from the size gate alike — so nothing below says anything about it.

| `4556fc6b2138` | 2/5 | 7 | llm | 2026-08-27 20:52 UTC |
| `66e3d7703d74` | 2/5 | 8 | llm | 2026-08-27 21:46 UTC |
| `fbbb8f3a70cc` | 1/5 | 1 | llm | 2026-08-28 16:13 UTC |
| `8415796e62e9` | 1/5 | 4 | llm | 2026-08-28 16:52 UTC |
| `e0cfa826a5ec` | 2/5 | 9 | llm | 2026-08-29 01:33 UTC |
| `ef35b153700d` | 3/5 | 1 | llm | 2026-08-29 14:44 UTC |
| `bf8378718780` | 2/5 | 6 | llm | 2026-08-29 16:55 UTC |
| `50e9a6036d7f` | 4/5 | 2 | llm | 2026-08-29 17:26 UTC |
| `1e2cb86cba2d` | 4/5 | 1 | llm | 2026-08-29 18:44 UTC |

</details>

<sub>Reviewed `a08c14029af9` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 935s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

TimeToBuildBob:
Addressed all seven AI-review findings in `66e3d770`:

- validated finite positive body envelope configuration with safe fallbacks;
- clamped relative moves to the configured absolute altitude envelope and added a 1 m floor in the MAVSDK adapter;
- invalidated stale telemetry and reconnect state when any MAVSDK stream fails or ends;
- serialized all MAVSDK action commands across concurrent voice sessions;
- gave the SITL scenario a real `asyncio.wait_for` timeout;
- disabled `mavsdk://` cleanly when the optional dependency is absent.

Verification: 248 gptme-voice tests pass; the focused safety/SITL slice is 33 passed, 1 skipped; Ruff and pre-commit pass locally. The in-band reviewer timed out and followed its documented fail-open push path. Current GitHub pre-commit failure is unrelated PyTorch-index HTTP 503 during `uv sync`; the remaining checks are still running.

TimeToBuildBob:
**CI fix (pre-commit 503)**: The Pre-commit Checks failure was a transient 503 from PyTorch's CDN while `uv sync` queried it for `jsonschema`. The pytorch-cpu index is configured without `explicit = true`, so uv queries it for *all* packages (not just torch) — any CDN hiccup at pytorch.org hits the sync. Re-run triggered; `Create virtual environment` already passed ✓ on the re-run. Note: `explicit = true` would fix the structural issue but risks breaking the CPU-only torch pin since torch arrives as a transitive dep via sentence-transformers — filed for separate investigation.

packages/gptme-voice/src/gptme_voice/body/mavsdk_adapter.py
```diff
@@ -0,0 +1,293 @@
...
+            "flight_mode": self._flight_mode,
+            "armed": self._armed,
+        }
+
+    async def takeoff(self, altitude_m: float) -> dict[str, Any]:
+        async with self._command_lock:
+            system = self._system
+            await system.action.set_takeoff_altitude(altitude_m)
+            await system.action.arm()
+            await system.action.takeoff()
```
TimeToBuildBob: ❌ **P1** — MavsdkAdapter action methods (takeoff, land, stop, return_home, goto, move, turn) await MAVSDK calls such as system.action.takeoff() with no asyncio.wait_for timeout. If the MAVLink link drops or the vehicle stops responding while a command is in flight, these awaits hang indefinitely. The bridge's _handle_body_call wraps the adapter call in try/except but has no timeout, so the function call will block the event loop and the voice conversation forever, never returning an error to the model. This is a real gap because the PR design relies on autopilot failsafes, but the adapter itself provides no safety timeout. Consequence: a lost link mid-command hangs the tool call and the entire call, with no recovery.

packages/gptme-voice/src/gptme_voice/body/mavsdk_adapter.py
```diff
@@ -0,0 +1,293 @@
...
+        }
+
+    async def move(
+        self, forward_m: float, right_m: float, up_m: float
+    ) -> dict[str, Any]:
+        async with self._command_lock:
+            pos = self._position
+            if not pos:
+                return {"error": "No position fix yet; cannot move."}
+            yaw = self._heading_deg or 0.0
```
TimeToBuildBob: ❌ **P1** — In MavsdkAdapter.move, the heading used for body-frame to NED conversion is yaw = self._heading_deg or 0.0. If the heading telemetry stream has not yet delivered a value (_heading_deg is None), the adapter assumes the vehicle faces north. A relative move such as 'forward 5 m' will then translate 5 m north regardless of the drone's actual orientation. The bridge's _handle_body_call does not require a heading fix before allowing body_move, so a caller or the model can issue a move while heading is unknown, with potentially dangerous consequences. The turn method has a similar default (current = self._heading_deg or 0.0). Consequence: the drone moves in an unintended direction, violating the safety intent of the PR.

packages/gptme-voice/src/gptme_voice/body/mavsdk_adapter.py
```diff
@@ -0,0 +1,293 @@
...
+            logger.info("MavsdkAdapter connecting to %s", self.system_address)
+            await system.connect(system_address=self.system_address)
+
+            async def _wait_connected() -> None:
+                async for state in system.core.connection_state():
+                    if state.is_connected:
+                        return
+
+            await asyncio.wait_for(_wait_connected(), timeout=self.connect_timeout_s)
+            self._system = system
```
TimeToBuildBob: ❌ **P1** — MavsdkAdapter uses separate _connection_lock and _command_lock. ensure_connected acquires only _connection_lock and command methods acquire only _command_lock. If a telemetry stream fails and calls _mark_disconnected (setting _connected=False and clearing caches) while a command is already executing, a subsequent command from another session can call ensure_connected, see _connected=False, cancel telemetry tasks, create a new System, and assign self._system. The in-progress command is still holding _command_lock and may use the old self._system (captured earlier) or read the new one depending on timing. This can lead the command to operate on a stale system or race with the new connection setup, causing failures or hangs. The bridge calls ensure_connected before every command, so concurrent calls under a link drop can trigger this race. Consequence: the system can be left in an inconsistent state and commands may fail or hang incorrectly.

packages/gptme-voice/src/gptme_voice/realtime/server.py
```diff
@@ -2009,6 +2022,7 @@ async def handle_local_websocket(self, websocket):
                 on_hangup=_local_hangup,
                 on_handoff=self._make_handoff_callback([caller_id], transcript),
                 transcript_provider=lambda: transcript,
+                body_adapter=self.body_adapter,
```
TimeToBuildBob: ❌ **P1** — The body_* tools (body_takeoff, body_move, body_goto, body_land) are wired into GptmeToolBridge for the local and browser WebSocket handlers, which perform no authentication. VoiceServer binds to 0.0.0.0 by default, so anyone who can reach /local or /voice can open a WebSocket and drive the physical body. The Twilio path has signature and allowlist checks, but these endpoints do not. This PR introduces an unauthenticated remote-control surface for a physical vehicle; a malicious client could command a drone to take off, move, or land, causing property damage or injury.

packages/gptme-voice/src/gptme_voice/realtime/tool_bridge.py
```diff
@@ -911,4 +943,92 @@ async def handle_function_call(self, name: str, arguments: dict) -> dict:
...
+        """
+        adapter = self.body_adapter
+        if adapter is None:
+            return {
+                "error": (
+                    "No body is connected to this session, so there is nothing to move."
+                )
+            }
+        try:
+            await adapter.ensure_connected()
```
TimeToBuildBob: ❌ **P1** — Every body_* tool in GptmeToolBridge._handle_body_call awaits adapter.ensure_connected() before doing anything, with no short timeout around it. MavsdkAdapter.ensure_connected() has an internal connect_timeout_s of 20s; when the vehicle is unreachable the await blocks for the full 20s before raising asyncio.TimeoutError. During that time the realtime function-call handler is suspended, so the assistant can say nothing and the call effectively hangs for up to 20 seconds per command whenever the body is down. This is a real misbehavior in normal operation when the vehicle is offline, making the voice session unresponsive and even appearing to have crashed.

packages/gptme-voice/src/gptme_voice/realtime/server.py
```diff
@@ -721,6 +722,15 @@ def __init__(
             if handoff_agents_env
             else _default_agents
         )
+        # Optional physical body (BobBrain): registers capability-gated
+        # body_* tools when GPTME_VOICE_BODY_URL is set.
+        self.body_adapter = body_adapter_from_env()
```
TimeToBuildBob, line 750: ⚠️ **P2** — VoiceServer creates a body adapter (e.g. MavsdkAdapter) in __init__ but never calls its close() method on shutdown. MavsdkAdapter.close() is responsible for cancelling the background telemetry tasks and releasing the MAVSDK connection. Without it, the asyncio tasks keep the event loop occupied and the MAVSDK socket stays open, preventing clean shutdown and leaking resources across server restarts.

packages/gptme-voice/src/gptme_voice/body/mavsdk_adapter.py
```diff
@@ -0,0 +1,293 @@
...
+        tasks = list(self._telemetry_tasks)
+        for task in tasks:
+            task.cancel()
+        if tasks:
+            await asyncio.gather(*tasks, return_exceptions=True)
+        self._telemetry_tasks.clear()
+
+    async def close(self) -> None:
+        await self._cancel_telemetry_tasks()
+        self._mark_disconnected()
```
TimeToBuildBob: ⚠️ **P2** — MavsdkAdapter.close() drops the MAVSDK System without closing it. On a reconnect (after link loss, ensure_connected creates a new System and connects), the previous System's connection (UDP/serial socket, telemetry subscriptions) is never terminated. Each reconnect therefore leaks the underlying MAVSDK connection and its resources. The server also never calls close(), but even when close() is invoked, it only cancels telemetry tasks and sets _system = None, leaving the MAVSDK connection open until process exit. This can exhaust file descriptors and leave stale connections to the vehicle over time.

ErikBjare:
@TimeToBuildBob fix

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Pushed the remaining second-round safety findings: `ed7c9199` + `fbbb8f3a` (HEAD `fbbb8f3a`).

- MAVSDK actions and connect now time out instead of hanging the voice loop
- relative move/turn refuse to run without a heading fix
- reconnect is serialized with in-flight commands
- unauthenticated local/browser clients do not get body tools (loopback + Twilio still do), and those sessions no longer advertise `body_*` schemas
- `body_move` floors at 0 m, so "descend" on the ground is a no-op instead of a 1 m takeoff
- adapter `close()` is wired into server shutdown and actually stops the MAVSDK process

packages/gptme-voice/src/gptme_voice/body/mavsdk_adapter.py
TimeToBuildBob: Fixed in ed7c9199 / fbbb8f3a: every MAVSDK action is wrapped in asyncio.wait_for (command_timeout_s=10). A timeout now also marks the adapter disconnected so a dead link cannot keep serving stale commands.

packages/gptme-voice/src/gptme_voice/body/mavsdk_adapter.py
TimeToBuildBob: Fixed in ed7c9199: move and turn return an error if heading telemetry is still None instead of assuming north.

packages/gptme-voice/src/gptme_voice/body/mavsdk_adapter.py
TimeToBuildBob: Fixed in ed7c9199: ensure_connected now takes the command lock before replacing System, so reconnect cannot race an in-flight action.

packages/gptme-voice/src/gptme_voice/realtime/server.py
TimeToBuildBob: Fixed in ed7c9199 / fbbb8f3a: body tools stay on Twilio and loopback local/browser only. Remote unauthenticated clients get no adapter and no body_* schemas.

packages/gptme-voice/src/gptme_voice/realtime/tool_bridge.py
TimeToBuildBob: Fixed in ed7c9199: connect timeout is 5s, and the bridge wraps ensure_connected (and every body call) in GPTME_VOICE_BODY_CALL_TIMEOUT_S (default 12s) so a down vehicle cannot stall the voice loop.

packages/gptme-voice/src/gptme_voice/realtime/server.py
TimeToBuildBob: Fixed in ed7c9199: VoiceServer lifespan now awaits body_adapter.close() on shutdown.

packages/gptme-voice/src/gptme_voice/body/mavsdk_adapter.py
TimeToBuildBob: Fixed in ed7c9199: close() and reconnect now call MAVSDK-Python's _stop_mavsdk_server so the old System does not leak sockets across reconnects.

packages/gptme-voice/src/gptme_voice/realtime/server.py
```diff
@@ -762,7 +772,30 @@ def __init__(
...
+    async def _lifespan(self, _app):
+        try:
+            yield
+        finally:
+            if self.body_adapter is not None:
+                await self.body_adapter.close()
+
+    def _body_adapter_for_websocket(self, websocket, *, transport: str):
+        """Expose motion tools only on authenticated or loopback transports."""
+        if self.body_adapter is None or transport == "twilio":
```
TimeToBuildBob: 🛑 **P0** — Any Twilio caller can control the connected drone. `_body_adapter_for_websocket` returns `self.body_adapter` for transport == 'twilio' without checking caller authorization, and the Twilio websocket handler passes `body_adapter=self.body_adapter` directly to `GptmeToolBridge`. The Twilio signature only proves the request came from Twilio, not that the human caller is authorized; an arbitrary phone caller can ask the model to take off, move, land, or return home. Combined with the capability-gated tools, this gives unauthenticated remote callers physical control of a drone. Fix: require a caller allowlist or PIN for body tools on Twilio, or otherwise gate them so only authorized callers can issue motion commands. _(2/3 passes — pass 1 looked and disagreed)_

TimeToBuildBob: Fixed in 8415796e: Twilio body tools are fail-closed. They only attach when From is on TWILIO_CALLER_ALLOWLIST; unlisted or unsigned callers get no motion tools and no body_* schemas.

packages/gptme-voice/src/gptme_voice/realtime/server.py
```diff
@@ -762,7 +772,53 @@ def __init__(
...
+        if transport == "twilio":
+            if self._twilio_body_caller_allowed(caller_id):
+                return self.body_adapter
+            logger.warning(
+                "Body tools disabled for Twilio caller %s (not on TWILIO_CALLER_ALLOWLIST)",
+                caller_id or "unknown",
+            )
+            return None
+        client = getattr(websocket, "client", None)
+        host = getattr(client, "host", None)
```
TimeToBuildBob: ❌ **P1** — In `_body_adapter_for_websocket`, for non-Twilio transports it reads the peer host as `host = getattr(client, "host", None)` from the websocket's `client` attribute. In Starlette, `WebSocket.client` is a `(host, port)` tuple, not an object with a `.host` attribute, so `getattr` returns `None` for every real connection. Consequently the condition `if host in {"127.0.0.1", "::1", "localhost"}` is always false, and body tools are never exposed on `/local` or `/voice` even from loopback. The unit tests pass because they use a mock client object with a `host` attribute, so the bug is not caught. This makes the body tool feature unusable for local testing and browser transport from the same machine.

packages/gptme-voice/src/gptme_voice/body/mavsdk_adapter.py
```diff
@@ -0,0 +1,346 @@
...
+        pos = self._position
+        if not pos:
+            return {"error": "No position fix yet; cannot goto."}
+        home_abs = pos["absolute_altitude_m"] - pos["relative_altitude_m"]
+        target_abs = (
+            home_abs + altitude_m
+            if altitude_m is not None
+            else pos["absolute_altitude_m"]
+        )
+        yaw = self._heading_deg if self._heading_deg is not None else float("nan")
```
TimeToBuildBob: ❌ **P1** — When body_goto is called before a heading fix is available, MavsdkAdapter._goto_unlocked passes float("nan") as the yaw argument to goto_location. MAVSDK's goto_location expects a finite yaw; NaN can cause the vehicle to behave unpredictably (e.g., loiter, reject the command, or hold an undefined heading). Unlike move and turn, which explicitly check self._heading_deg is None and return an error, goto has no such guard. The bridge does not check for heading on body_goto, so a caller can trigger this. The consequence is that the vehicle may fail to navigate or enter an unsafe state when the model issues a goto without a compass fix. Fix: return an error when _heading_deg is None, or fall back to an explicit default like 0.0.

packages/gptme-voice/tests/test_server.py
```diff
@@ -139,6 +139,99 @@ def get_timings(self) -> list[dict[str, object]]:
         return []
 
 
+def test_body_adapter_only_reaches_trusted_transports(monkeypatch) -> None:
+    monkeypatch.setenv("GPTME_VOICE_BODY_URL", "null")
+    server = VoiceServer()
+    remote_client = type("Client", (), {"host": "203.0.113.1"})()
```
TimeToBuildBob: ⚠️ **P2** — The test test_body_adapter_only_reaches_trusted_transports constructs mock websockets whose client is an object with a .host attribute (e.g., type("Client", (), {"host": "203.0.113.1"})). In actual Starlette, websocket.client is a tuple (host, port), not an object with attributes. As a result, the test exercises a fake interface that does not match the real code path, so it cannot detect the bug in _body_adapter_for_websocket that always sees host=None for real websockets. The test passes vacuously and does not validate the loopback security boundary. It should use a tuple like ("127.0.0.1", 12345) to reflect the real Starlette API.

packages/gptme-voice/src/gptme_voice/realtime/server.py
TimeToBuildBob: Fixed in 0ce2633b: `_websocket_peer_host` now accepts Starlette `Address` (`.host`) and a plain `(host, port)` tuple. Loopback tests use both shapes.

packages/gptme-voice/src/gptme_voice/body/mavsdk_adapter.py
TimeToBuildBob: Fixed in 0ce2633b: `body_goto` now returns an error when heading telemetry is missing instead of passing NaN yaw — same fail-closed as move/turn.

packages/gptme-voice/tests/test_server.py
TimeToBuildBob: Fixed in 0ce2633b: the trusted-transport test now uses `(host, port)` tuples (and still covers a `.host` named client) so it matches Starlette's real `websocket.client`.

TimeToBuildBob:
Pushed `0ce2633b` for the remaining P1s:

- `_websocket_peer_host` accepts both Starlette `Address.host` and a plain `(host, port)` tuple, so loopback body tools actually attach
- `body_goto` refuses without a heading fix instead of sending NaN yaw (same fail-closed as move/turn)

CI re-running. Self-merge still waits on Greptile (not found) plus these threads resolving.

TimeToBuildBob:
Pushed `e0cfa826` for the remaining P0: `/twilio` no longer grants body tools from client-supplied `customParameters.from_number`.

The signed `/incoming` webhook mints a one-shot `body_grant`; the WebSocket start event must present it (and it must still be on `TWILIO_CALLER_ALLOWLIST`) before motion tools attach. A spoofed start event also cannot steal a body-capable prewarm.

Requesting a fresh AI review of HEAD `e0cfa826`.

TimeToBuildBob:
Self-merge not eligible yet. CI is green on `e0cfa826`, but the current-head review cannot land:

- `OPENROUTER_API_KEY_REVIEW` daily limit until 2026-08-29T00:00Z (llm engine 403)
- agent-engine fallback (`gpt-5.6-sol`) produced no findings artifact (max-steps)
- Greptile still awaiting its one initial review (helper will not retrigger)

Resolved the three outdated P1/P2 threads that were already fixed in `0ce2633b`. Next: re-review after the review-key reset, then self-merge if no open P0/P1.

TimeToBuildBob:
Stale-vs-head rearm on `e0cfa826` (marker still `8415796e`). Review-key quota is still closed until 2026-08-29T00:00Z; did not re-run the reviewer or retrigger Greptile.

Scheduled machine-gated recheck at 2026-08-29T00:10Z. After reset: `ai-review.py gptme/gptme-contrib 1528 --engine llm`, then self-merge if no open P0/P1.

packages/gptme-voice/src/gptme_voice/realtime/server.py
```diff
@@ -1624,6 +1746,10 @@ async def handle_incoming_call(self, request: Request) -> PlainTextResponse:
         custom_params: dict[str, str] = {}
         if from_number:
             custom_params["from_number"] = from_number
+            # Body tools on /twilio must not trust client-supplied from_number.
+            # Mint a one-shot grant here (signed webhook) and require it on start.
+            if self._twilio_body_caller_allowed(from_number):
```
TimeToBuildBob: ❌ **P1** — When TWILIO_AUTH_TOKEN is not configured, the /incoming endpoint skips Twilio signature validation but still mints a body-tool grant based on the unauthenticated From field. An attacker can POST to /incoming with a spoofed From matching an allowlisted number, receive a body_grant, then connect to /twilio with that grant and command the vehicle. The grant system is supposed to ensure only Twilio-verified callers get body tools, but it is bypassed because the grant is issued from the same unverified From. The code logs a warning, but it proceeds to mint the grant. This defeats the security boundary: without TWILIO_AUTH_TOKEN, the allowlist is trivially spoofed and the one-shot grant provides no real protection.

packages/gptme-voice/src/gptme_voice/realtime/server.py
```diff
@@ -1720,10 +1846,29 @@ async def _on_speech_started() -> None:
                         )
                     )
 
+                    # Body-tool authorization is the signed-webhook grant, not
+                    # customParameters.from_number (attacker-controlled on /twilio).
+                    granted_from = self._consume_twilio_body_grant(
```
TimeToBuildBob: ❌ **P1** — In `handle_twilio_websocket`, the grant token is consumed and the value returned is used as `caller_id` for the allowlist check, but the server never verifies that this matches the `from_number` present in the WebSocket start event's `customParameters`. An attacker who obtains a valid grant token (e.g., by intercepting the signed /incoming webhook response or the TwiML output) can connect to `/twilio` with an arbitrary `from_number`, present the stolen token, and be granted body tools because the allowlist check uses the token's original bound from_number. This allows an unauthorized caller to command a physical body. The token should be bound to the call/stream SID or the start event's from_number must be validated against the grant's bound number.

packages/gptme-voice/src/gptme_voice/body/mavsdk_adapter.py
```diff
@@ -0,0 +1,348 @@
...
+        async with self._connection_lock:
+            async with self._command_lock:
+                await self._cancel_telemetry_tasks()
+                self._mark_disconnected()
+                await self._release_system()
+
+    async def _run_action(self, awaitable: Any) -> Any:
+        try:
+            return await asyncio.wait_for(awaitable, timeout=self.command_timeout_s)
+        except TimeoutError:
```
TimeToBuildBob: ❌ **P1** — MavsdkAdapter._run_action wraps every action in asyncio.wait_for with command_timeout_s=10.0 and, on timeout, calls _mark_disconnected(), which sets _connected=False and cancels all telemetry streams. However, the underlying MAVSDK action (e.g. land, takeoff, or a long goto) is not cancelled — asyncio.wait_for cancels the coroutine, but the already-sent MAVSDK command continues to execute on the aircraft. Marking the adapter disconnected while the drone is still descending or climbing causes the bridge to treat the link as lost, cancel telemetry, and potentially attempt a reconnect and reissue commands, interfering with an in-progress maneuver. Concretely, a landing that takes longer than 10 seconds will time out, mark the adapter disconnected, and leave the session believing the body is unavailable while the aircraft is still descending.

packages/gptme-voice/src/gptme_voice/realtime/tool_bridge.py
```diff
@@ -911,4 +948,111 @@ async def handle_function_call(self, name: str, arguments: dict) -> dict:
...
+
+        caps = adapter.capabilities
+        try:
+            if name == "body_status":
+                return {"status": "ok", "telemetry": adapter.telemetry()}
+            if name == "body_stop" and "move" in caps:
+                return await _call(adapter.stop())
+            if name == "body_return_home" and "move" in caps:
+                return await _call(adapter.return_home())
+            if name == "body_takeoff" and "altitude" in caps:
```
TimeToBuildBob, line 1000: ❌ **P1** — In GptmeToolBridge._handle_body_call, the body_takeoff branch calls adapter.takeoff(altitude) without first checking adapter.telemetry() for the drone's in_air state. If the drone is already airborne (e.g. after a partially completed maneuver or a failed landing), issuing arm and takeoff again can cause a crash or erratic behavior. The tool is gated only on the altitude capability, and the safety envelope clamps the altitude but does not verify the current flight state. A model that misjudges the state or a user who accidentally issues takeoff while the drone is flying could trigger a dangerous command. The adapter itself also lacks an in_air guard inside takeoff().

packages/gptme-voice/src/gptme_voice/realtime/server.py
```diff
@@ -762,7 +801,89 @@ def __init__(
...
+        if transport == "twilio":
+            if self._twilio_body_caller_allowed(caller_id):
+                return self.body_adapter
+            logger.warning(
+                "Body tools disabled for Twilio caller %s (not on TWILIO_CALLER_ALLOWLIST)",
+                caller_id or "unknown",
+            )
+            return None
+        host = _websocket_peer_host(websocket)
+        if host in {"127.0.0.1", "::1", "localhost"}:
```
TimeToBuildBob, line 991: ❌ **P1** — For non-Twilio transports, motion tools are authorized solely by the peer's source IP being loopback: `_body_adapter_for_websocket` returns the full `body_adapter` (capable of takeoff, goto, land) when `host in {'127.0.0.1','::1','localhost'}`, with no shared secret or session token. The server binds 0.0.0.0 (default host), so /local and /browser are reachable over the network; a remote peer's IP fails the check, but any local process on the same host or in a container/namespace that shares the loopback can open a /local WebSocket and issue body_stop/body_takeoff/body_goto/body_land, flying the vehicle. Unlike the Twilio path (one-shot grant + allowlist), local/browser have no equivalent credential — the trust boundary is just the IP. For a physical quadrotor, a same-host compromise (e.g. a web app running on the box, or an SSH user) can take control of the aircraft. Recommend requiring a per-deployment secret (query parameter or header) for local/browser transports instead of trusting the peer address, or not registering body tools on these routes when a body adapter is configured.

packages/gptme-voice/src/gptme_voice/realtime/server.py
```diff
@@ -751,6 +784,12 @@ def __init__(
         self._prewarm_sessions: dict[str, tuple[OpenAIRealtimeClient, float]] = {}
         # Max seconds a pre-warm session is kept before being discarded
         self._prewarm_ttl_seconds = 30
+        # One-shot body-tool grants minted by the signed /incoming webhook.
+        # The /twilio WebSocket is unauthenticated; customParameters.from_number
+        # is attacker-controlled, so body tools must not key off it. Token ->
+        # (from_number, expires_monotonic).
+        self._twilio_body_grants: dict[str, tuple[str, float]] = {}
+        self._twilio_body_grant_ttl_s = 120.0
```
TimeToBuildBob, line 801: ⚠️ **P2** — The `_twilio_body_grant_ttl_s` is fixed at 120 seconds. If the Twilio WebSocket 'start' event arrives later than this (e.g., due to slow call setup, network delay, or a busy server), the grant is consumed but has already expired, so `_consume_twilio_body_grant` returns None and the caller is denied body tools even though they are on the allowlist and were correctly authorized by the signed webhook. This is a functional degradation for legitimate authorized callers. Consider making the TTL configurable or generous enough to cover worst-case call setup time.

packages/gptme-voice/src/gptme_voice/realtime/tool_bridge.py
```diff
@@ -911,4 +948,111 @@ async def handle_function_call(self, name: str, arguments: dict) -> dict:
...
+                    # Floor at 0 m so "descend" on the ground is a no-op, not
+                    # a forced climb to the old 1 m takeoff floor.
+                    target_altitude = max(
+                        0.0,
+                        min(self.body_max_altitude_m, float(current_altitude) + up),
+                    )
+                    up = target_altitude - float(current_altitude)
+                return await _call(adapter.move(forward, right, up))
+            if name == "body_goto" and "move" in caps:
+                lat = float(arguments["latitude_deg"])
```
TimeToBuildBob, line 1046: ⚠️ **P2** — The body_goto tool passes the caller-supplied latitude_deg and longitude_deg straight to adapter.goto without validating ranges or a maximum distance from the current position. A compromised or misinstructed model could command the drone to a distant or invalid coordinate (e.g., latitude 91), which MAVSDK may reject, but nothing here prevents a far-away but valid coordinate. The tool description says to never invent coordinates, but that is not enforced in code. Combined with the lack of a geofence or distance cap in the bridge, this gives the model a direct path to fly the vehicle to an arbitrary location once authorized.

packages/gptme-voice/src/gptme_voice/body/mavsdk_adapter.py
```diff
@@ -0,0 +1,348 @@
...
+                self._armed = armed
+
+        async def _watch(stream_name: str, subscription_factory: Any) -> None:
+            try:
+                await subscription_factory()
+            except asyncio.CancelledError:
+                raise
+            except Exception:  # noqa: BLE001 - MAVSDK stream errors vary by backend
+                logger.exception("MAVSDK %s telemetry stream failed", stream_name)
+                self._mark_disconnected()
```
TimeToBuildBob, line 156: ⚠️ **P2** — In MavsdkAdapter._watch, any single telemetry stream (position, heading, battery, etc.) that ends without exception or raises an exception triggers _mark_disconnected(), which sets _connected=False, clears all telemetry caches, and cancels every other telemetry task. Because the adapter is a singleton shared across all concurrent voice sessions, a transient failure in one stream (e.g., a battery reading hiccup) marks the whole vehicle disconnected for every session. The next body call in any session will then run ensure_connected() again, tearing down and reconnecting the MAVSDK system, which cancels in-flight commands and causes a burst of reconnects. This is a resilience/safety issue: a single flaky sensor stream can repeatedly interrupt command execution for all active calls. The condition is triggered by any generator in system.telemetry returning or raising, and the consequence is a global disconnect rather than just degrading that one stream.

packages/gptme-voice/src/gptme_voice/realtime/tool_bridge.py
```diff
@@ -911,4 +948,111 @@ async def handle_function_call(self, name: str, arguments: dict) -> dict:
...
+        adapter = self.body_adapter
+        if adapter is None:
+            return {
+                "error": (
+                    "No body is connected to this session, so there is nothing to move."
+                )
+            }
+
+        async def _call(awaitable: Awaitable[_T]) -> _T:
+            return await asyncio.wait_for(awaitable, timeout=self.body_call_timeout_s)
```
TimeToBuildBob, line 978: ⚠️ **P2** — `_handle_body_call` wraps each adapter call in `asyncio.wait_for(awaitable, timeout=self.body_call_timeout_s)` (default 12s), while `MavsdkAdapter._run_action` has its own `command_timeout_s` (default 10s) and, on timeout, calls `_mark_disconnected()` and re-raises. Because 12 > 10 by default, the adapter's timeout fires first and resets `_connected`/telemetry caches normally. But if an operator sets `GPTME_VOICE_BODY_CALL_TIMEOUT_S` below the adapter's command timeout (e.g. 5s), the bridge's `wait_for` cancels the adapter coroutine before `_run_action` sees its own timeout; the `except TimeoutError` inside `_run_action` never runs, so `_mark_disconnected()` is skipped. The adapter keeps `_connected=True` and stale `_position`/`_heading_deg`, the next `body_*` call's `ensure_connected()` short-circuits, and a fresh command is issued while the vehicle may still be executing (or in an unknown state from) the timed-out command. The disconnect and reconnect path is silently bypassed.

packages/gptme-voice/src/gptme_voice/realtime/server.py
TimeToBuildBob: Fixed in ef35b153: `/incoming` only mints a body grant after Twilio signature validation succeeds. Missing `TWILIO_AUTH_TOKEN` is fail-closed for motion tools (the call can still connect; no grant is issued from a spoofable From).

packages/gptme-voice/src/gptme_voice/realtime/server.py
TimeToBuildBob: Fixed in ef35b153: the grant is bound to CallSid from the signed webhook (not copied into TwiML). `/twilio` start must match both From and CallSid or body tools stay off.

packages/gptme-voice/src/gptme_voice/body/mavsdk_adapter.py
TimeToBuildBob: Fixed in ef35b153: a MAVSDK command timeout logs and re-raises but leaves the adapter connected. Telemetry stays up while land/takeoff/goto continues on the vehicle; we no longer drop the link mid-maneuver.

packages/gptme-voice/src/gptme_voice/realtime/tool_bridge.py
TimeToBuildBob: Fixed in ef35b153: both the tool bridge and `MavsdkAdapter.takeoff` refuse takeoff when `in_air` is True.

packages/gptme-voice/src/gptme_voice/realtime/server.py
TimeToBuildBob: Not changing in this PR. Local/browser body tools stay loopback-only (fbbb8f3a) — that is the local-first trust boundary. Remote peers already get no motion tools. A shared-secret knob for same-host processes is YAGNI until an operator asks; same-host compromise already implies control of the vehicle link.

packages/gptme-voice/src/gptme_voice/realtime/tool_bridge.py
TimeToBuildBob: Superseded by the timeout-disconnect fix in ef35b153: we no longer want `_mark_disconnected()` on command timeout. If the bridge timeout fires first, the adapter stays connected while the vehicle may still be executing — the same fail-open-link behavior as the adapter timeout path.

TimeToBuildBob:
Pushed `ef35b153` for the 01:52 current-head P1s (CI re-running):

- `/incoming` mints body grants only after Twilio signature validation; missing `TWILIO_AUTH_TOKEN` is fail-closed for motion tools
- grants are bound to CallSid and must match the `/twilio` start From + CallSid
- takeoff is refused when already in the air
- MAVSDK command timeout no longer drops the adapter mid-maneuver

Loopback local/browser body tools stay loopback-only (local-first trust boundary; not adding a shared-secret knob here).

Self-merge still waits on a current-head review of `ef35b153`. Review-key is recapped until 2026-08-30T00:00Z; did not retrigger Greptile.

TimeToBuildBob:
CI is now green on `ef35b153`. Timeout cause of the 13:36 merge_ready rearm: it raced the P1 push while checks were still starting.

Self-merge not eligible: Greptile dark (awaiting-initial-review); AI review is stale (reviewed `e0cfa826`, HEAD `ef35b153`). OPENROUTER_API_KEY_REVIEW remains capped until 2026-08-30T00:00Z. Did not retrigger Greptile, steal keys, or merge.

Three leftover 01:52 P2s stay open for the current-head review (grant TTL, goto distance envelope, telemetry-stream disconnect). Recheck already scheduled at 2026-08-30T00:10Z.

TimeToBuildBob:
Stale-vs-head rearm on `ef35b153` (marker still `8415796e`). Sweep tried this head at 14:20 and failed (`all aspect passes failed`); 6h sweep backoff would starve it until ~20:20Z.

Review-key remaining is $3.48 (not capped). Dispatched dedicated `ai-review-now` for current HEAD (`bob-ai-review-now-gptme-gptme-contrib-1528`). Did not retrigger Greptile (still awaiting-initial-review; no new push) and did not merge.

packages/gptme-voice/src/gptme_voice/realtime/server.py
```diff
@@ -762,7 +803,90 @@ def __init__(
...
+    def _consume_twilio_body_grant(self, token: str | None) -> tuple[str, str] | None:
+        """Return ``(from_number, call_sid)`` bound to a live grant, or None.
+
+        Single-use: a replay or a raw WebSocket that invents customParameters
+        cannot satisfy this. Expired and unknown tokens fail closed.
+        """
+        self._expire_twilio_body_grants()
+        if not token:
+            return None
+        entry = self._twilio_body_grants.pop(token, None)
```
TimeToBuildBob: ❌ **P1** — The body-tool grant in `_consume_twilio_body_grant` is single-use: `entry = self._twilio_body_grants.pop(token, None)` removes the token on first use. If the Twilio Media Stream WebSocket drops and Twilio reconnects (which it does automatically on transient network failures), the same `customParameters` (including `body_grant`) are sent in the new `start` event, but the token is already gone, so `_consume_twilio_body_grant` returns None. `body_adapter` then becomes None, and the session is built without body tools. The vehicle is left flying with no way to command `body_stop` or `body_return_home` for the remainder of the call. The PR intends single-use to prevent replay, but the call_sid binding already prevents replay onto a different call; a token could safely be reused for the same call_sid. A reconnecting legitimate call loses vehicle control, which is a safety concern if the vehicle is airborne. Suggest storing the grant until the call ends (keyed by call_sid) and removing it on `_on_call_end`. _(2/3 passes — pass 2 looked and disagreed)_

TimeToBuildBob: Fixed in `074ee4ab` + `bf837871` (HEAD `bf837871`).

`_consume_twilio_body_grant` no longer pops. The grant stays CallSid-bound for the call:

- Twilio reconnects resend the same `start` customParameters and keep `body_stop`
- first successful consume pins expiry so a long call's late reconnect still has motion tools
- Twilio `stop` and hangup revoke; mint TTL still covers unused grants; unknown tokens fail closed

Replay onto a different CallSid remains fail-closed. In-band review of HEAD is 5/5 / 0 findings.

TimeToBuildBob:
Pushed `074ee4ab` + `bf837871` (HEAD `bf837871`) for the current-head P1 on the one-shot Twilio body grant.

- `_consume_twilio_body_grant` no longer pops — Twilio Media Stream reconnects keep `body_stop`
- first successful consume pins the grant until `stop`/hangup, so a long call's late reconnect still has motion tools
- mint TTL still expires unused grants; unknown tokens fail closed; CallSid mismatch still fail-closed

In-band review of HEAD: 5/5, 0 findings. Replied on the P1 thread. Did not retrigger Greptile. CI re-running; self-merge waits on GitHub current-head review of `bf837871`.

ErikBjare:
@TimeToBuildBob I like this, but given we recently merged gptme-vision-node along gptme-voice-node, it kinda makes sense with a gptme-body-node. Or what do you think? What is the architecture here and does the separation make sense or just makes integration difficult?

TimeToBuildBob:
The `-node` suffix is doing too much work here. `gptme-voice-node` and `gptme-vision-node` are edge I/O daemons — they run on the Pi because that's where the mic, speaker, and camera live. This PR is the other side of the contract: capability-gated `body_*` tools on the realtime session, with `BodyAdapter` translating goals in-process inside `gptme-voice`.

A `gptme-body-node` *does* make sense, but later, and for a different job: the on-vehicle companion that owns the flight-controller link (serial MAVLink to Pixhawk, local PX4 failsafes, telemetry). That's payload integration, when the Pi is actually on an airframe. Extracting it now would be a fake node — MAVSDK wrapping one process further from the model that has to issue `body_stop`.

Why this stays in `gptme-voice` for v0/v1:

1. **Asymmetry.** Voice and vision have hardware on the puck today. Body does not. SITL and `NullAdapter` (tabletop, no motion tools) both live on the host. A third package would be empty deployment until there's a vehicle.

2. **Latency is the integration cost.** The load-bearing constraint in this PR is that motion tools — especially `body_stop` — must not round-trip through subagent dispatch. A body-node hop *is* that extra round-trip. Vision can be pull-based (`look`, events). Body is push-based from the voice model. That's why the adapter sits in the tool bridge, not beside the camera daemon.

3. **The protocol is already the seam.** `BodyAdapter` is a Python protocol on purpose. When we need on-vehicle execution, we add a `RemoteAdapter` in `gptme-voice` that forwards goals to a Pi-side `gptme-body-node`. Tools stay here. Safety stays next to the FC. We don't rip the schemas out later.

The spec diagram in `bobbrain-spec.md` currently draws BodyAdapter on the Pi *and* "brain loop → goals to BodyAdapter" on the host. That's the confusion. Step 3 ("land BodyAdapter in the node") picked the wrong package boundary: the node is plumbing, the tools belong to the session.

**Recommendation:** merge this as-is. Don't split a `gptme-body-node` out of this PR. After merge: (a) wire vision `look` into the same bridge, (b) one paragraph in the spec that BodyAdapter stays in `gptme-voice` until the payload is on a real body, at which point `gptme-body-node` is the companion, not a rename of this.

If you'd rather extract now anyway, say so — but the integration would get harder, not easier, and we'd still keep the tool schemas in `gptme-voice`.

packages/gptme-voice/src/gptme_voice/realtime/server.py
```diff
@@ -1720,10 +1887,46 @@ async def _on_speech_started() -> None:
...
+                    granted_from: str | None = None
+                    if grant is not None:
+                        grant_from, grant_sid = grant
+                        if grant_from != from_number:
+                            logger.warning(
+                                "Twilio body grant From mismatch: grant=%s start=%s",
+                                grant_from,
+                                from_number,
+                            )
+                        elif grant_sid and grant_sid != call_sid:
```
TimeToBuildBob: ❌ **P1** — In `_consume_twilio_body_grant`, CallSid binding is only enforced when the grant's call_sid is non-empty (`elif grant_sid and grant_sid != call_sid`). If `/incoming` omits `CallSid` (or sends it empty), `_mint_twilio_body_grant` stores `call_sid=""`, and this condition is skipped, so any start event with a matching `from_number` is granted regardless of the actual `call_sid` in the start payload. This breaks the PR's stated guarantee that a stolen grant cannot be replayed onto a different call: an attacker who obtains the token (e.g., by reading the TwiML response to /incoming) and knows the allowlisted `from_number` can replay it on any call by setting that `from_number` in `customParameters`. The pinned `expires=inf` after first consume makes the grant usable indefinitely until a stop/hangup, increasing the exposure window. _(2/3 passes — pass 3 looked and disagreed)_

packages/gptme-voice/src/gptme_voice/realtime/server.py
```diff
@@ -1569,9 +1720,16 @@ async def handle_incoming_call(self, request: Request) -> PlainTextResponse:
         """
         form_params = dict(await request.form())
         from_number = form_params.get("From", "")
+        incoming_call_sid = form_params.get("CallSid", "") or form_params.get(
```
TimeToBuildBob: ❌ **P1** — In handle_incoming_call, incoming_call_sid is read with a default of '' when CallSid is missing from the form. This empty value is passed to _mint_twilio_body_grant, so the minted grant is bound to call_sid ''. Later in handle_twilio_websocket, _consume_twilio_body_grant returns ('', ...), and the check `elif grant_sid and grant_sid != call_sid` is False because grant_sid is falsy. Thus only the From number is required to match. Since From is the caller's E.164 number and is the same across calls from that caller, a leaked grant token (e.g., from a compromised WebSocket connection) can be replayed on any later call from the same caller, granting body tools on that call. This breaks the advertised call-scoping guarantee. Fix: require a non-empty CallSid before minting, or treat an empty grant_sid as a mismatch.

packages/gptme-voice/src/gptme_voice/body/mavsdk_adapter.py
```diff
@@ -0,0 +1,361 @@
...
+            "position": self._position or None,
+            "heading_deg": self._heading_deg,
+            "battery": self._battery or None,
+            "in_air": self._in_air,
+            "flight_mode": self._flight_mode,
+            "armed": self._armed,
+        }
+
+    async def takeoff(self, altitude_m: float) -> dict[str, Any]:
+        async with self._command_lock:
```
TimeToBuildBob, line 251: ❌ **P1** — MavsdkAdapter.command_lock serializes all body actions. takeoff() acquires the lock and holds it for three sequential awaits (set_takeoff_altitude, arm, takeoff), each under asyncio.wait_for with command_timeout_s default 10. A concurrent stop() must acquire the same lock before it can call hold(), so a body_stop issued while a takeoff or long goto is in progress blocks until the in-flight action completes (up to ~30s for takeoff's three 10s timeouts). The body_stop schema promises 'IMMEDIATELY halt', but in practice the stop is delayed by the entire remaining duration of the preceding command. In an emergency, that delay could be the difference between a controlled hold and a collision, directly contradicting the tool's stated intent. The lock should be released between sub-actions or a dedicated emergency hold path should bypass the command lock.

packages/gptme-voice/src/gptme_voice/realtime/tool_bridge.py
```diff
@@ -911,4 +948,118 @@ async def handle_function_call(self, name: str, arguments: dict) -> dict:
...
+        caps = adapter.capabilities
+        try:
+            if name == "body_status":
+                return {"status": "ok", "telemetry": adapter.telemetry()}
+            if name == "body_stop" and "move" in caps:
+                return await _call(adapter.stop())
+            if name == "body_return_home" and "move" in caps:
+                return await _call(adapter.return_home())
+            if name == "body_takeoff" and "altitude" in caps:
+                if adapter.telemetry().get("in_air") is True:
```
TimeToBuildBob, line 1001: ❌ **P1** — The bridge's pre-flight check for an in-air vehicle is performed outside the adapter's command lock and reads the telemetry cache (adapter.telemetry().get('in_air')). The MavsdkAdapter's takeoff() re-checks self._in_air under the lock, but both read the same asynchronously-updated cache. Two back-to-back takeoff calls can both observe in_air as False (the vehicle hasn't yet reported in_air after the first takeoff), so both proceed into adapter.takeoff(). The first acquires the command lock, arms and takes off; when the second acquires the lock moments later, the vehicle may still not have published an in_air update, so it passes the check again and issues a second arm+takeoff sequence. This can confuse the autopilot (e.g., attempting to arm an already-armed vehicle) and is a safety hazard for an airborne vehicle.

packages/gptme-voice/src/gptme_voice/realtime/server.py
TimeToBuildBob: ❌ **P1** — When /incoming receives a request without a CallSid (e.g., a malformed or unusual webhook), incoming_call_sid defaults to '' and is passed to _mint_twilio_body_grant. The grant therefore carries an empty call_sid. In handle_twilio_websocket, the binding check `elif grant_sid and grant_sid != call_sid:` is skipped when grant_sid is empty, so the grant is accepted for any call whose from_number matches the grant's from_number. A leaked token (e.g., from TwiML logging) can then be replayed onto a different call from the same number, defeating the stated CallSid binding. Additionally, _revoke_twilio_body_grants_for_call returns early when call_sid is falsy, so grants minted without a sid are never revoked and persist for the process lifetime.

packages/gptme-voice/src/gptme_voice/realtime/tool_bridge.py
```diff
@@ -911,4 +948,118 @@ async def handle_function_call(self, name: str, arguments: dict) -> dict:
...
+                    return {
+                        "error": (
+                            "Already in the air; refuse takeoff. "
+                            "Use body_move, body_goto, or body_land."
+                        )
+                    }
+                altitude = float(arguments.get("altitude_m", 2.5))
```
TimeToBuildBob, line 1008: ❌ **P1** — The bridge parses numeric arguments with float() and clamps via max/min without checking math.isfinite. A value like "nan" or "infinity" (which Python's float() accepts) produces a NaN/Inf that the clamp functions turn into the limit: min(30, nan) returns 30, so body_takeoff with altitude_m="nan" takes off to the maximum 30 m; body_move with forward_m="nan" clamps to +50 m, an unintended full-speed lunge; body_turn with yaw_deg="nan" becomes a 180° spin. The schema promises numeric inputs, so this only happens from a confused model, but it's a safety-critical envelope bypass: the clamps exist to prevent a confused model from requesting extreme values, and NaN slips through as the maximum. Reject non-finite numbers instead of clamping them.

packages/gptme-voice/src/gptme_voice/realtime/server.py
```diff
@@ -762,7 +808,111 @@ def __init__(
...
+        if entry is None:
+            return None
+        from_number, call_sid, expires = entry
+        if time.monotonic() > expires:
+            self._twilio_body_grants.pop(token, None)
+            return None
+        # Pin after first start so a long call's later reconnect still
+        # has body_stop. Unused grants still expire at mint TTL.
+        if expires != float("inf"):
+            self._twilio_body_grants[token] = (from_number, call_sid, float("inf"))
```
TimeToBuildBob, line 897: ⚠️ **P2** — The Twilio body grant dict never removes pinned grants. In _consume_twilio_body_grant, a successfully consumed token is pinned to expires=float('inf'), and it is only revoked in _revoke_twilio_body_grants_for_call, which runs on Twilio 'stop' or the hangup tool. If a websocket disconnects without either event (network drop, server crash, or client abort), the grant remains in _twilio_body_grants forever, because _expire_twilio_body_grants only removes entries where now > expires and inf never expires. Over a long-running server with many calls that disconnect abnormally, this dict grows without bound, leaking one entry per such call. A malicious or careless caller could trigger many such disconnects, leading to unbounded memory growth and eventually OOM. Fix: instead of pinning to infinity, pin to the max call duration (e.g., the post-call delay or a fixed maximum) and ensure a periodic cleanup, or track active call SIDs and remove grants when the websocket closes.

packages/gptme-voice/src/gptme_voice/realtime/server.py
TimeToBuildBob: Fixed in `50e9a603` (HEAD `50e9a603`).

`_mint_twilio_body_grant` now requires non-empty From **and** CallSid (no default empty sid). Consume drops leftover empty-sid grants instead of pinning them. `/incoming` strips CallSid, skips mint when it's blank, and never emits an empty `body_grant`. A stolen grant with matching From can no longer skip the CallSid bind.

packages/gptme-voice/src/gptme_voice/realtime/tool_bridge.py
```diff
@@ -911,4 +948,118 @@ async def handle_function_call(self, name: str, arguments: dict) -> dict:
...
+                    "No body is connected to this session, so there is nothing to move."
+                )
+            }
+
+        async def _call(awaitable: Awaitable[_T]) -> _T:
+            return await asyncio.wait_for(awaitable, timeout=self.body_call_timeout_s)
+
+        try:
+            await _call(adapter.ensure_connected())
+        except TimeoutError:
```
TimeToBuildBob: ⚠️ **P2** — On Python 3.10, asyncio.wait_for raises asyncio.TimeoutError, which is a distinct class from the built-in TimeoutError. The `except TimeoutError:` clauses in GptmeToolBridge._handle_body_call (for connection and action timeouts) and in MavsdkAdapter._run_action will not catch these, so timeouts fall through to the generic `except Exception` handler. This returns an error message like "body_stop failed: <exception>" instead of the intended "timed out" message, and in _run_action the warning about leaving the adapter connected is not logged. The tests test_bridge_body_call_times_out and test_mavsdk_action_timeout expect TimeoutError and would fail on Python 3.10 because the exception type is not caught.

packages/gptme-voice/src/gptme_voice/realtime/tool_bridge.py
```diff
@@ -911,4 +948,118 @@ async def handle_function_call(self, name: str, arguments: dict) -> dict:
...
+                )
+                right = self._clamp(
+                    float(arguments.get("right_m", 0.0)), self.body_max_move_m
+                )
+                up = self._clamp(
+                    float(arguments.get("up_m", 0.0)), self.body_max_altitude_m
+                )
+                position = adapter.telemetry().get("position") or {}
+                current_altitude = position.get("relative_altitude_m")
+                if current_altitude is not None:
```
TimeToBuildBob: ⚠️ **P2** — In GptmeToolBridge._handle_body_call for body_move, the absolute-altitude ceiling is only enforced when `current_altitude = position.get("relative_altitude_m")` is not None. If an adapter's telemetry dict lacks `relative_altitude_m`, the code skips the clamp block and passes the raw clamped `up` (already clamped to ±body_max_altitude_m) to adapter.move. Since the bridge cannot know the current altitude, a vehicle already above the max altitude could be commanded to climb further, exceeding GPTME_VOICE_BODY_MAX_ALT_M. The current MavsdkAdapter always provides this field, so it is not triggered today, but the bridge is generic and should enforce the ceiling even without telemetry.

TimeToBuildBob (commented):
## 🤖 AI code review — 4/5 (P2 findings only)

**2 findings** · ⚠️ **2** P2

Details are in the summary comment below.

packages/gptme-voice/src/gptme_voice/realtime/tool_bridge.py
TimeToBuildBob: Fixed in `1e2cb86c` (HEAD `1e2cb86c`).

`except asyncio.TimeoutError` now wraps the `wait_for` paths in `_handle_body_call` (connect + action) and `MavsdkAdapter._run_action`. On Python 3.10 that is a distinct class from builtin `TimeoutError`, so timeouts return the intended "timed out" error / warning instead of falling through to generic failure. The adapter re-raises builtin `TimeoutError` so callers and tests see one type on 3.10–3.13.

packages/gptme-voice/src/gptme_voice/realtime/tool_bridge.py
TimeToBuildBob: Fixed in `1e2cb86c` (HEAD `1e2cb86c`).

`body_move` now refuses climb when `relative_altitude_m` is missing, so the absolute ceiling cannot be skipped without a fix. Descent and level flight still go through (clamped `up <= 0`). Tests cover refuse-climb and allow-level/descent without telemetry.

TimeToBuildBob:
Pushed `1e2cb86c` for the current-head P2s (CI re-running):

- `except asyncio.TimeoutError` on the `wait_for` paths so Python 3.10 timeouts return "timed out" instead of a generic failure
- `body_move` refuses climb when `relative_altitude_m` is missing (absolute ceiling fail-closed)

In-band review of HEAD: 5/5, 0 findings. Replied on both P2 threads. Did not retrigger Greptile. Architecture call still stands: keep BodyAdapter in `gptme-voice`; `gptme-body-node` is a future on-vehicle companion, not a split of this PR.

Self-merge still waits on Greptile (awaiting-initial-review) plus GitHub current-head review of `1e2cb86c`.

packages/gptme-voice/src/gptme_voice/realtime/server.py
```diff
@@ -1808,7 +2033,10 @@ async def _cb() -> None:
                                 await realtime_client.send_audio(pcm_data)
 
                 elif event == "stop":
-                    # Call ended
+                    # Real call end — revoke so a reconnect after hangup cannot
+                    # re-attach motion tools with the stolen start parameters.
+                    # Do not revoke in ``finally``: websocket drop is a reconnect.
+                    self._revoke_twilio_body_grants_for_call(call_sid)
```
TimeToBuildBob, line 2111: ⚠️ **P2** — The Twilio body-tool grant is never revoked when a websocket disconnects without a 'stop' event or a hangup. The grant is pinned to `float("inf")` on first consume and only removed by `_schedule_hangup` (which calls `_revoke_twilio_body_grants_for_call`) or by the explicit `stop` branch in the websocket handler. The `finally` block intentionally does not revoke to support reconnects, but if a call ends abruptly (e.g., network failure) and no `stop` ever arrives, the grant remains pinned in `self._twilio_body_grants` indefinitely. This both leaks memory and leaves a live bearer token that, if it ever leaked (e.g., from logs), could be replayed on a different call that presents the same `From` and `CallSid`. The token is only placed in TwiML custom parameters, so an attacker would need to intercept the webhook response to obtain it, but the longer it lives the larger the exposure window. The issue is present in every call that ends without a `stop` event or a hangup. _(2/2 passes)_

TimeToBuildBob (commented):
## 🤖 AI code review — 4/5 (P2 findings only)

**1 finding** · ⚠️ **1** P2

Details are in the summary comment below.

packages/gptme-voice/src/gptme_voice/realtime/server.py
TimeToBuildBob: Won't fix in this round. Skipping revoke in `finally` is the reconnect path from `074ee4ab`/`bf837871` — `body_stop` has to survive Twilio Media Stream reconnects. Hangup webhook + explicit `stop` already revoke.

CallSid is unique per Twilio call, so "replay on a different call with the same CallSid" isn't a real path. A pinned-grant TTL (not `inf`) is a follow-up if hangup webhooks prove unreliable; not another safety round on this PR.

TimeToBuildBob: Fixed in a08c1402: websocket drop without `stop` now schedules a 90s idle revoke; a reconnect for the same CallSid cancels it. `stop`/hangup still revoke immediately. Tests cover idle revoke, reconnect cancel, and the `handle_twilio_websocket` finally path.

TimeToBuildBob:
Pushed `a08c1402` for the current-head P2 (Twilio body grant stayed pinned forever if the websocket dropped without `stop`/`hangup`).

- websocket drop still does **not** revoke immediately (that's the Media Stream reconnect path)
- a 90s idle timer revokes the pinned grant iff the CallSid does not reconnect
- reconnect cancels the timer so `body_stop` stays available mid-flight
- `stop` and hangup still revoke immediately

Local: `uv run --package gptme-voice pytest packages/gptme-voice/tests/test_server.py packages/gptme-voice/tests/test_body_tools.py` → 141 passed, 1 skipped.

Did not retrigger Greptile. Architecture call still stands: keep BodyAdapter in `gptme-voice`; `gptme-body-node` is a future on-vehicle companion, not a split of this PR.

CI is running. Self-merge waits on green CI plus GitHub current-head review of `a08c1402`.

TimeToBuildBob:
Current-head review of `a08c1402` posted: 5/5, 0 findings, consensus 3/3 undegraded. CI green. Greptile still dark (did not retrigger).

Self-merge eligibility passed; the post-eligibility consensus gate blocked at 2/5:

> Unvalidated inputs in the MAVSDK adapter could lead to unsafe drone commands; capability gating and error handling are not verifiable from the provided snippets.

That is a mechanical second-opinion refusal, not a current-head P0/P1. Envelope clamps, capability gating, Twilio grant/idle-revoke, and fail-closed telemetry already live in the diff — the gate's "snippets not disclosed" complaint is about its own prompt, not missing code. I am not overriding a score-2 block.

Architecture call still stands (Erik's 16:11 question): keep BodyAdapter in `gptme-voice`; `gptme-body-node` is a future on-vehicle companion. Merge remains human/PM if you want it anyway.

TimeToBuildBob:
Self-merge still blocked on the consensus gate.

Re-ran on current head `a08c1402` (CI green, MERGEABLE, AI review 5/5). Eligibility still passes. Consensus this cycle: **3/5** (was 2/5 at 02:33). Floor is 4. Not overriding.

Concern this run: SITL/integration coverage is opt-in (`BODY_SITL_URL`), so the gate treats the MAVSDK adapter as unit-tested only. That is a real residual for a flight-control PR, not a snippet-prompt failure.

Timeout cause of the prior merge_ready rearm: the worker died after eligibility while the consensus call was in flight; state was unchanged, so recovery re-dispatched the same SHA. No code change, no Greptile retrigger, no merge.

Architecture call still stands (Erik 16:11): BodyAdapter stays in `gptme-voice`; `gptme-body-node` is a future on-vehicle companion.

PR #1529, opened 2026-08-27
feat(vision-node): add gptme-vision-node — BobBrain vision pipeline v0

TimeToBuildBob:
Implements **BobBrain milestone 2 (vision v0)** per `bobbrain-spec.md` — the vision half of the portable presence node. Ref: ErikBjare/bob#730.

New package `packages/gptme-vision-node/` (module `gptme_vision_node`), deliberately light deps: `opencv-python-headless` (<5 — OpenCV 5 removed the legacy HOG objdetect API), `numpy`, `httpx`, `click`. No torch/ultralytics/CLIP — remote inference is the architecture.

## What's in it

- **Frame sources** (`frame_source.py`): `FrameSource` protocol; `ImageFileSource` (single file or directory round-robin), `OpenCVCameraSource` (V4L2 webcam index / Pi camera / RTSP URL, lazy-open).
- **Detection** (`detect.py`): `PersonDetector` — OpenCV's built-in HOG people detector (no model downloads); `MotionDetector` — frame differencing against a running-average background. Shared `Detection {kind, box, score}` dataclass + `Detector` protocol.
- **LLM look tool** (`look.py`): `describe_frame(frame, prompt)` — JPEG+base64 → any OpenAI-compatible chat-completions endpoint (`OPENAI_API_KEY`, `OPENAI_BASE_URL`, `VISION_MODEL`). This is the on-demand "see things" toolcall the voice bridge wires in (concurrent PR #1524 territory — no file overlap beyond uv.lock/pre-commit config).
- **Place recognition** (`place.py`) — "where am I?" without GPS (Erik's ask 2026-08-27):
  - `Embedder` protocol with `HistogramEmbedder` default (HSV histogram + coarse spatial grid, L2-normalized). CLIP documented as a future optional extra, not a dep.
  - `WifiSignature.scan()` via `nmcli` (graceful empty dict on failure) + overlap-weighted RSSI similarity (Jaccard × signal agreement → 0..1).
  - `PlaceRecognizer`: multi-sample enroll per place, fused visual+wifi scoring (`wifi_weight` configurable, 0.5 default when both present), JSON gallery persistence.
- **Pipeline** (`pipeline.py`): `VisionPipeline(source, detectors, interval_s)` — capture → detect → `VisionEvent` (`person_appeared`/`person_left` edge-triggered with debounce, `motion`) via callback; keeps `latest_frame` for the look tool. Thread-based, callback exceptions contained.
- **CLI** (`gptme-vision-node`): `detect`, `look`, `enroll --place`, `whereami` (gallery default `~/.config/gptme-vision-node/places.json`). Click, per repo convention.

## Tests

40 tests, fully offline (synthetic numpy frames, mocked httpx/nmcli/wifi): frame-source round-robin, motion fires-on-change/quiet-on-static, HOG no-crash on synthetic frames, histogram embedder discrimination (near-identical vs distinct scenes, spatial-grid sensitivity), wifi parse+similarity math, recognizer enroll/recognize/save/load roundtrips (vision-only, wifi-only, combined disambiguation), look request-building + error paths, pipeline event edge/debounce logic, CLI end-to-end via CliRunner.

`uv run pytest packages/gptme-vision-node -q` → 40 passed. mypy clean; package added to the root-mypy exclude list (it has its own `make typecheck`, same as the other packages) and picked up by the CI package matrix via its Makefile.

### DIFF
--- a/packages/gptme-vision-node/src/gptme_vision_node/cli.py
+++ b/packages/gptme-vision-node/src/gptme_vision_node/cli.py
@@ -141,7 +141,12 @@ def enroll(source_spec: str, place_name: str, gallery: str, no_wifi: bool) -> No
     frame = _capture_frame(source_spec)
     wifi = {} if no_wifi else WifiSignature.scan()
     recognizer.enroll(place_name, frame, wifi_sig=wifi or None)
-    recognizer.save(gallery_path)
+    try:
+        recognizer.save(gallery_path)
+    except OSError as exc:
+        raise click.ClickException(
+            f"failed to save gallery at {gallery_path}: {exc}"
+        ) from exc
     n = len(recognizer.places[place_name])
     click.echo(
         f"enrolled '{place_name}' (sample {n}, wifi APs: {len(wifi)}) -> {gallery_path}"
--- a/packages/gptme-vision-node/src/gptme_vision_node/cli.py
+++ b/packages/gptme-vision-node/src/gptme_vision_node/cli.py
@@ -34,15 +34,25 @@ source_option = click.option(
 
 def make_source(spec: str) -> FrameSource:
     """Parse a source spec: ``camera:N``, a stream URL, or an image path."""
-    if spec.startswith("camera:"):
-        return OpenCVCameraSource(int(spec.split(":", 1)[1]))
-    if "://" in spec:  # rtsp://, http://, ...
-        return OpenCVCameraSource(spec)
-    return ImageFileSource(spec)
+    try:
+        if spec.startswith("camera:"):
+            return OpenCVCameraSource(int(spec.split(":", 1)[1]))
+        if "://" in spec:  # rtsp://, http://, ...
+            return OpenCVCameraSource(spec)
+        return ImageFileSource(spec)
+    except ValueError as exc:
+        if spec.startswith("camera:"):
+            raise click.ClickException(f"invalid camera index: {spec!r}") from exc
+        raise click.ClickException(str(exc)) from exc
+    except OSError as exc:
+        raise click.ClickException(str(exc)) from exc
 
 
 def _get_frame_or_die(source: FrameSource) -> np.ndarray:
-    frame = source.get_frame()
+    try:
+        frame = source.get_frame()
+    except (OSError, RuntimeError, ValueError) as exc:
+        raise click.ClickException(str(exc)) from exc
     if frame is None:
         raise click.ClickException("source produced no frame")
     return frame
@@ -63,10 +73,11 @@ def detect(source_spec: str, once: bool, interval: float) -> None:
     detectors: list[Detector] = [PersonDetector(), MotionDetector()]
     try:
         while True:
-            frame = source.get_frame()
-            if frame is None:
-                break
-            detections = [d for det in detectors for d in det.detect(frame)]
+            frame = _get_frame_or_die(source)
+            try:
+                detections = [d for det in detectors for d in det.detect(frame)]
+            except (RuntimeError, ValueError) as exc:
+                raise click.ClickException(str(exc)) from exc
             for d in detections:
                 click.echo(f"{d.kind}\tbox={d.box}\tscore={d.score:.3f}")
             if not detections:
--- a/packages/gptme-vision-node/src/gptme_vision_node/cli.py
+++ b/packages/gptme-vision-node/src/gptme_vision_node/cli.py
@@ -73,6 +73,13 @@ def _capture_frame(source_spec: str) -> np.ndarray:
         _release_source(source)
 
 
+def _load_gallery(path: Path) -> PlaceRecognizer:
+    try:
+        return PlaceRecognizer.from_file(path)
+    except (OSError, ValueError) as exc:
+        raise click.ClickException(f"invalid gallery at {path}: {exc}") from exc
+
+
 @click.group()
 def main() -> None:
     """Vision pipeline for the BobBrain presence node."""
@@ -129,9 +136,7 @@ def enroll(source_spec: str, place_name: str, gallery: str, no_wifi: bool) -> No
     """Enroll a sample of a named place into the gallery."""
     gallery_path = Path(gallery)
     recognizer = (
-        PlaceRecognizer.from_file(gallery_path)
-        if gallery_path.exists()
-        else PlaceRecognizer()
+        _load_gallery(gallery_path) if gallery_path.exists() else PlaceRecognizer()
     )
     frame = _capture_frame(source_spec)
     wifi = {} if no_wifi else WifiSignature.scan()
@@ -155,7 +160,7 @@ def whereami(source_spec: str, gallery: str, no_wifi: bool, as_json: bool) -> No
         raise click.ClickException(
             f"no gallery at {gallery_path} (enroll some places first)"
         )
-    recognizer = PlaceRecognizer.from_file(gallery_path)
+    recognizer = _load_gallery(gallery_path)
     frame = _capture_frame(source_spec)
     wifi = {} if no_wifi else WifiSignature.scan()
     results = recognizer.recognize(frame, wifi_sig=wifi or None)
--- a/packages/gptme-vision-node/src/gptme_vision_node/cli.py
+++ b/packages/gptme-vision-node/src/gptme_vision_node/cli.py
@@ -10,11 +10,12 @@ Subcommands:
 
 from __future__ import annotations
 
-import argparse
 import json
-import sys
 import time
 from pathlib import Path
+
+import click
+import numpy as np
 
 from .detect import Detector, MotionDetector, PersonDetector
 from .frame_source import FrameSource, ImageFileSource, OpenCVCameraSource
@@ -22,6 +23,13 @@ from .look import describe_frame
 from .place import PlaceRecognizer, WifiSignature
 
 DEFAULT_GALLERY = Path.home() / ".config" / "gptme-vision-node" / "places.json"
+
+source_option = click.option(
+    "--source",
+    "source_spec",
+    required=True,
+    help="image file/dir, camera:N, or a stream URL (rtsp://...)",
+)
 
 
 def make_source(spec: str) -> FrameSource:
@@ -33,16 +41,25 @@ def make_source(spec: str) -> FrameSource:
     return ImageFileSource(spec)
 
 
-def _get_frame_or_die(source: FrameSource):
+def _get_frame_or_die(source: FrameSource) -> np.ndarray:
     frame = source.get_frame()
     if frame is None:
-        print("error: source produced no frame", file=sys.stderr)
-        raise SystemExit(1)
+        raise click.ClickException("source produced no frame")
     return frame
 
 
-def cmd_detect(args: argparse.Namespace) -> int:
-    source = make_source(args.source)
+@click.group()
+def main() -> None:
+    """Vision pipeline for the BobBrain presence node."""
+
+
+@main.command()
+@source_option
+@click.option("--once", is_flag=True, help="single frame, then exit")
+@click.option("--interval", type=float, default=1.0, help="seconds between frames")
+def detect(source_spec: str, once: bool, interval: float) -> None:
+    """Run person/motion detection on a source."""
+    source = make_source(source_spec)
     detectors: list[Detector] = [PersonDetector(), MotionDetector()]
     try:
         while True:
@@ -51,124 +68,79 @@ def cmd_detect(args: argparse.Namespace) -> int:
                 break
             detections = [d for det in detectors for d in det.detect(frame)]
             for d in detections:
-                print(f"{d.kind}\tbox={d.box}\tscore={d.score:.3f}")
+                click.echo(f"{d.kind}\tbox={d.box}\tscore={d.score:.3f}")
             if not detections:
-                print("(no detections)")
-            if args.once:
+                click.echo("(no detections)")
+            if once:
                 break
-            time.sleep(args.interval)
+            time.sleep(interval)
     except KeyboardInterrupt:
         pass
     finally:
         release = getattr(source, "release", None)
         if release:
             release()
-    return 0
 
 
-def cmd_look(args: argparse.Namespace) -> int:
-    source = make_source(args.source)
-    frame = _get_frame_or_die(source)
-    print(describe_frame(frame, args.prompt, model=args.model))
-    return 0
+@main.command()
+@source_option
+@click.option("--prompt", default="Describe what you see, briefly.")
+@click.option("--model", default=None, help="override VISION_MODEL")
+def look(source_spec: str, prompt: str, model: str | None) -> None:
+    """Describe the current frame via a vision LLM (needs OPENAI_API_KEY)."""
+    frame = _get_frame_or_die(make_source(source_spec))
+    click.echo(describe_frame(frame, prompt, model=model))
 
 
-def _load_recognizer(gallery: Path) -> PlaceRecognizer:
-    if gallery.exists():
-        return PlaceRecognizer.from_file(gallery)
-    return PlaceRecognizer()
-
-
-def cmd_enroll(args: argparse.Namespace) -> int:
-    gallery = Path(args.gallery)
-    recognizer = _load_recognizer(gallery)
-    source = make_source(args.source)
-    frame = _get_frame_or_die(source)
-    wifi = {} if args.no_wifi else WifiSignature.scan()
-    recognizer.enroll(args.place, frame, wifi_sig=wifi or None)
-    recognizer.save(gallery)
-    n = len(recognizer.places[args.place])
-    print(
-        f"enrolled '{args.place}' (sample {n}, " f"wifi APs: {len(wifi)}) -> {gallery}"
+@main.command()
+@source_option
+@click.option("--place", "place_name", required=True, help="place name")
+@click.option("--gallery", default=str(DEFAULT_GALLERY), show_default=True)
+@click.option("--no-wifi", is_flag=True, help="skip the wifi scan")
+def enroll(source_spec: str, place_name: str, gallery: str, no_wifi: bool) -> None:
+    """Enroll a sample of a named place into the gallery."""
+    gallery_path = Path(gallery)
+    recognizer = (
+        PlaceRecognizer.from_file(gallery_path)
+        if gallery_path.exists()
+        else PlaceRecognizer()
     )
-    return 0
+    frame = _get_frame_or_die(make_source(source_spec))
+    wifi = {} if no_wifi else WifiSignature.scan()
+    recognizer.enroll(place_name, frame, wifi_sig=wifi or None)
+    recognizer.save(gallery_path)
+    n = len(recognizer.places[place_name])
+    click.echo(
+        f"enrolled '{place_name}' (sample {n}, wifi APs: {len(wifi)}) -> {gallery_path}"
+    )
 
 
-def cmd_whereami(args: argparse.Namespace) -> int:
-    gallery = Path(args.gallery)
-    if not gallery.exists():
-        print(
-            f"error: no gallery at {gallery} (enroll some places first)",
-            file=sys.stderr,
+@main.command()
+@source_option
+@click.option("--gallery", default=str(DEFAULT_GALLERY), show_default=True)
+@click.option("--no-wifi", is_flag=True, help="skip the wifi scan")
+@click.option("--json", "as_json", is_flag=True, help="JSON output")
+def whereami(source_spec: str, gallery: str, no_wifi: bool, as_json: bool) -> None:
+    """Rank enrolled places for the current view."""
+    gallery_path = Path(gallery)
+    if not gallery_path.exists():
+        raise click.ClickException(
+            f"no gallery at {gallery_path} (enroll some places first)"
         )
-        return 1
-    recognizer = PlaceRecognizer.from_file(gallery)
-    source = make_source(args.source)
-    frame = _get_frame_or_die(source)
-    wifi = {} if args.no_wifi else WifiSignature.scan()
+    recognizer = PlaceRecognizer.from_file(gallery_path)
+    frame = _get_frame_or_die(make_source(source_spec))
+    wifi = {} if no_wifi else WifiSignature.scan()
     results = recognizer.recognize(frame, wifi_sig=wifi or None)
     if not results:
-        print("no matches (gallery empty or no comparable signals)")
-        return 1
-    if args.json:
-        print(json.dumps([{"place": n, "score": s} for n, s in results]))
+        raise click.ClickException(
+            "no matches (gallery empty or no comparable signals)"
+        )
+    if as_json:
+        click.echo(json.dumps([{"place": n, "score": s} for n, s in results]))
     else:
         for name, score in results:
-            print(f"{score:.3f}\t{name}")
-    return 0
-
-
-def build_parser() -> argparse.ArgumentParser:
-    parser = argparse.ArgumentParser(
-        prog="gptme-vision-node",
-        description="Vision pipeline for the BobBrain presence node",
-    )
-    sub = parser.add_subparsers(dest="command", required=True)
-
-    def add_source(p: argparse.ArgumentParser) -> None:
-        p.add_argument(
-            "--source",
-            required=True,
-            help="image file/dir, camera:N, or a stream URL (rtsp://...)",
-        )
-
-    p_detect = sub.add_parser("detect", help="run person/motion detection")
-    add_source(p_detect)
-    p_detect.add_argument("--once", action="store_true", help="single frame, then exit")
-    p_detect.add_argument(
-        "--interval", type=float, default=1.0, help="seconds between frames"
-    )
-    p_detect.set_defaults(func=cmd_detect)
-
-    p_look = sub.add_parser("look", help="describe the current frame via a vision LLM")
-    add_source(p_look)
-    p_look.add_argument("--prompt", default="Describe what you see, briefly.")
-    p_look.add_argument("--model", default=None, help="override VISION_MODEL")
-    p_look.set_defaults(func=cmd_look)
-
-    p_enroll = sub.add_parser("enroll", help="enroll a sample of a named place")
-    add_source(p_enroll)
-    p_enroll.add_argument("--place", required=True, help="place name")
-    p_enroll.add_argument("--gallery", default=str(DEFAULT_GALLERY))
-    p_enroll.add_argument("--no-wifi", action="store_true", help="skip the wifi scan")
-    p_enroll.set_defaults(func=cmd_enroll)
-
-    p_where = sub.add_parser(
-        "whereami", help="rank enrolled places for the current view"
-    )
-    add_source(p_where)
-    p_where.add_argument("--gallery", default=str(DEFAULT_GALLERY))
-    p_where.add_argument("--no-wifi", action="store_true", help="skip the wifi scan")
-    p_where.add_argument("--json", action="store_true", help="JSON output")
-    p_where.set_defaults(func=cmd_whereami)
-
-    return parser
-
-
-def main(argv: list[str] | None = None) -> int:
-    args = build_parser().parse_args(argv)
-    return int(args.func(args))
+            click.echo(f"{score:.3f}\t{name}")
 
 
 if __name__ == "__main__":
-    raise SystemExit(main())
+    main()
--- a/packages/gptme-vision-node/src/gptme_vision_node/cli.py
+++ b/packages/gptme-vision-node/src/gptme_vision_node/cli.py
@@ -76,7 +76,7 @@ def _capture_frame(source_spec: str) -> np.ndarray:
 def _load_gallery(path: Path) -> PlaceRecognizer:
     try:
         return PlaceRecognizer.from_file(path)
-    except (OSError, ValueError) as exc:
+    except (OSError, TypeError, ValueError) as exc:
         raise click.ClickException(f"invalid gallery at {path}: {exc}") from exc
 
 
--- a/packages/gptme-vision-node/src/gptme_vision_node/cli.py
+++ b/packages/gptme-vision-node/src/gptme_vision_node/cli.py
@@ -15,6 +15,7 @@ import time
 from pathlib import Path
 
 import click
+import cv2
 import httpx
 import numpy as np
 
@@ -104,7 +105,7 @@ def detect(source_spec: str, once: bool, interval: float) -> None:
                 break
             try:
                 detections = [d for det in detectors for d in det.detect(frame)]
-            except (RuntimeError, ValueError) as exc:
+            except (cv2.error, RuntimeError, ValueError) as exc:
                 raise click.ClickException(str(exc)) from exc
             for d in detections:
                 click.echo(f"{d.kind}\tbox={d.box}\tscore={d.score:.3f}")
--- a/packages/gptme-vision-node/src/gptme_vision_node/cli.py
+++ b/packages/gptme-vision-node/src/gptme_vision_node/cli.py
@@ -1,0 +1,174 @@
+"""CLI for the vision node.
+
+Subcommands:
+
+- ``detect``   — run detectors on a source, print detections
+- ``look``     — describe the current frame via a vision LLM
+- ``enroll``   — add sample(s) of a named place to the gallery
+- ``whereami`` — rank enrolled places for the current observation
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import sys
+import time
+from pathlib import Path
+
+from .detect import Detector, MotionDetector, PersonDetector
+from .frame_source import FrameSource, ImageFileSource, OpenCVCameraSource
+from .look import describe_frame
+from .place import PlaceRecognizer, WifiSignature
+
+DEFAULT_GALLERY = Path.home() / ".config" / "gptme-vision-node" / "places.json"
+
+
+def make_source(spec: str) -> FrameSource:
+    """Parse a source spec: ``camera:N``, a stream URL, or an image path."""
+    if spec.startswith("camera:"):
+        return OpenCVCameraSource(int(spec.split(":", 1)[1]))
+    if "://" in spec:  # rtsp://, http://, ...
+        return OpenCVCameraSource(spec)
+    return ImageFileSource(spec)
+
+
+def _get_frame_or_die(source: FrameSource):
+    frame = source.get_frame()
+    if frame is None:
+        print("error: source produced no frame", file=sys.stderr)
+        raise SystemExit(1)
+    return frame
+
+
+def cmd_detect(args: argparse.Namespace) -> int:
+    source = make_source(args.source)
+    detectors: list[Detector] = [PersonDetector(), MotionDetector()]
+    try:
+        while True:
+            frame = source.get_frame()
+            if frame is None:
+                break
+            detections = [d for det in detectors for d in det.detect(frame)]
+            for d in detections:
+                print(f"{d.kind}\tbox={d.box}\tscore={d.score:.3f}")
+            if not detections:
+                print("(no detections)")
+            if args.once:
+                break
+            time.sleep(args.interval)
+    except KeyboardInterrupt:
+        pass
+    finally:
+        release = getattr(source, "release", None)
+        if release:
+            release()
+    return 0
+
+
+def cmd_look(args: argparse.Namespace) -> int:
+    source = make_source(args.source)
+    frame = _get_frame_or_die(source)
+    print(describe_frame(frame, args.prompt, model=args.model))
+    return 0
+
+
+def _load_recognizer(gallery: Path) -> PlaceRecognizer:
+    if gallery.exists():
+        return PlaceRecognizer.from_file(gallery)
+    return PlaceRecognizer()
+
+
+def cmd_enroll(args: argparse.Namespace) -> int:
+    gallery = Path(args.gallery)
+    recognizer = _load_recognizer(gallery)
+    source = make_source(args.source)
+    frame = _get_frame_or_die(source)
+    wifi = {} if args.no_wifi else WifiSignature.scan()
+    recognizer.enroll(args.place, frame, wifi_sig=wifi or None)
+    recognizer.save(gallery)
+    n = len(recognizer.places[args.place])
+    print(
+        f"enrolled '{args.place}' (sample {n}, " f"wifi APs: {len(wifi)}) -> {gallery}"
+    )
+    return 0
+
+
+def cmd_whereami(args: argparse.Namespace) -> int:
+    gallery = Path(args.gallery)
+    if not gallery.exists():
+        print(
+            f"error: no gallery at {gallery} (enroll some places first)",
+            file=sys.stderr,
+        )
+        return 1
+    recognizer = PlaceRecognizer.from_file(gallery)
+    source = make_source(args.source)
+    frame = _get_frame_or_die(source)
+    wifi = {} if args.no_wifi else WifiSignature.scan()
+    results = recognizer.recognize(frame, wifi_sig=wifi or None)
+    if not results:
+        print("no matches (gallery empty or no comparable signals)")
+        return 1
+    if args.json:
+        print(json.dumps([{"place": n, "score": s} for n, s in results]))
+    else:
+        for name, score in results:
+            print(f"{score:.3f}\t{name}")
+    return 0
+
+
+def build_parser() -> argparse.ArgumentParser:
+    parser = argparse.ArgumentParser(
+        prog="gptme-vision-node",
+        description="Vision pipeline for the BobBrain presence node",
+    )
+    sub = parser.add_subparsers(dest="command", required=True)
+
+    def add_source(p: argparse.ArgumentParser) -> None:
+        p.add_argument(
+            "--source",
+            required=True,
+            help="image file/dir, camera:N, or a stream URL (rtsp://...)",
+        )
+
+    p_detect = sub.add_parser("detect", help="run person/motion detection")
+    add_source(p_detect)
+    p_detect.add_argument("--once", action="store_true", help="single frame, then exit")
+    p_detect.add_argument(
+        "--interval", type=float, default=1.0, help="seconds between frames"
+    )
+    p_detect.set_defaults(func=cmd_detect)
+
+    p_look = sub.add_parser("look", help="describe the current frame via a vision LLM")
+    add_source(p_look)
+    p_look.add_argument("--prompt", default="Describe what you see, briefly.")
+    p_look.add_argument("--model", default=None, help="override VISION_MODEL")
+    p_look.set_defaults(func=cmd_look)
+
+    p_enroll = sub.add_parser("enroll", help="enroll a sample of a named place")
+    add_source(p_enroll)
+    p_enroll.add_argument("--place", required=True, help="place name")
+    p_enroll.add_argument("--gallery", default=str(DEFAULT_GALLERY))
+    p_enroll.add_argument("--no-wifi", action="store_true", help="skip the wifi scan")
+    p_enroll.set_defaults(func=cmd_enroll)
+
+    p_where = sub.add_parser(
+        "whereami", help="rank enrolled places for the current view"
+    )
+    add_source(p_where)
+    p_where.add_argument("--gallery", default=str(DEFAULT_GALLERY))
+    p_where.add_argument("--no-wifi", action="store_true", help="skip the wifi scan")
+    p_where.add_argument("--json", action="store_true", help="JSON output")
+    p_where.set_defaults(func=cmd_whereami)
+
+    return parser
+
+
+def main(argv: list[str] | None = None) -> int:
+    args = build_parser().parse_args(argv)
+    return int(args.func(args))
+
+
+if __name__ == "__main__":
+    raise SystemExit(main())
--- a/packages/gptme-vision-node/src/gptme_vision_node/cli.py
+++ b/packages/gptme-vision-node/src/gptme_vision_node/cli.py
@@ -15,6 +15,7 @@ import time
 from pathlib import Path
 
 import click
+import httpx
 import numpy as np
 
 from .detect import Detector, MotionDetector, PersonDetector
@@ -58,6 +59,20 @@ def _get_frame_or_die(source: FrameSource) -> np.ndarray:
     return frame
 
 
+def _release_source(source: FrameSource) -> None:
+    release = getattr(source, "release", None)
+    if release:
+        release()
+
+
+def _capture_frame(source_spec: str) -> np.ndarray:
+    source = make_source(source_spec)
+    try:
+        return _get_frame_or_die(source)
+    finally:
+        _release_source(source)
+
+
 @click.group()
 def main() -> None:
     """Vision pipeline for the BobBrain presence node."""
@@ -88,9 +103,7 @@ def detect(source_spec: str, once: bool, interval: float) -> None:
     except KeyboardInterrupt:
         pass
     finally:
-        release = getattr(source, "release", None)
-        if release:
-            release()
+        _release_source(source)
 
 
 @main.command()
@@ -99,8 +112,12 @@ def detect(source_spec: str, once: bool, interval: float) -> None:
 @click.option("--model", default=None, help="override VISION_MODEL")
 def look(source_spec: str, prompt: str, model: str | None) -> None:
     """Describe the current frame via a vision LLM (needs OPENAI_API_KEY)."""
-    frame = _get_frame_or_die(make_source(source_spec))
-    click.echo(describe_frame(frame, prompt, model=model))
+    frame = _capture_frame(source_spec)
+    try:
+        description = describe_frame(frame, prompt, model=model)
+    except (httpx.HTTPError, RuntimeError, ValueError) as exc:
+        raise click.ClickException(str(exc)) from exc
+    click.echo(description)
 
 
 @main.command()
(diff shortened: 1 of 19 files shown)


TimeToBuildBob:
CI is fully green, including the Python 3.10–3.12 package jobs, typecheck, integration tests, and patch coverage. No human or automated review findings are pending yet; waiting for the initial Greptile/maintainer review.

packages/gptme-vision-node/src/gptme_vision_node/detect.py
```diff
@@ -0,0 +1,136 @@
...
+        )
+        self.win_stride = win_stride
+        self.scale = scale
+        self.score_threshold = score_threshold
+
+    def detect(self, frame: np.ndarray) -> list[Detection]:
+        # HOG needs a minimum window (64x128); upscale tiny frames.
+        h, w = frame.shape[:2]
+        if h < 128 or w < 64:
+            frame = cv2.resize(frame, (max(w, 64), max(h, 128)))
```
TimeToBuildBob: ❌ **P1** — When a frame is smaller than 64x128, PersonDetector.detect resizes it with cv2.resize, but the bounding boxes returned by HOG are in the resized coordinate space and are never transformed back to the original frame's coordinates. For example, a 32x32 frame is upscaled to 64x128 (non-uniformly scaling width by 2 and height by 4), so a box at (10,20,20,40) in resized space corresponds to roughly (5,5,10,10) in the original, but the code returns the resized box unchanged. Any consumer of these boxes (e.g., cropping, drawing, or location-based logic) will get incorrect coordinates for small input frames. The fix is to scale the box coordinates back by the inverse of the resize factors before appending the detection.

packages/gptme-vision-node/src/gptme_vision_node/detect.py
```diff
@@ -0,0 +1,136 @@
...
+            gray = cv2.GaussianBlur(gray, (self.blur_ksize, self.blur_ksize), 0)
+        return gray.astype(np.float32)
+
+    def detect(self, frame: np.ndarray) -> list[Detection]:
+        gray = self._prepare(frame)
+        if self._background is None:
+            self._background = gray
+            return []
+
+        diff = cv2.absdiff(gray, self._background)
```
TimeToBuildBob, line 126: ⚠️ **P2** — The MotionDetector fails silently if the frame size changes between calls. Its state includes `self._background`, a float32 array sized to the first frame seen. On a subsequent frame with different dimensions, `cv2.absdiff(gray, self._background)` raises a cv2.error (shape mismatch). In the VisionPipeline, detector exceptions are caught and logged (`logger.exception("detector %r failed", detector)`), so the error is swallowed and motion detection stops producing any events for the remainder of the run. This can happen in practice if a camera or stream source changes resolution (e.g., a V4L2 camera reconfigured by another process, or an RTSP stream that switches encoding). The user observes no motion events and no crash, making the failure hard to diagnose. The detector should reset its background when the frame shape changes, e.g., by checking `gray.shape` against `self._background.shape` and re-priming on mismatch.

packages/gptme-vision-node/src/gptme_vision_node/pipeline.py
```diff
@@ -0,0 +1,142 @@
...
+        if motions:
+            events.append(VisionEvent(MOTION, motions))
+
+        for event in events:
+            self._emit(event)
+        return events
+
+    # -- thread lifecycle --------------------------------------------------
+
+    def _run(self) -> None:
```
TimeToBuildBob, line 121: ⚠️ **P2** — `VisionPipeline._run` loops `while not self._stop.is_set()`, calling `self.step()` each iteration. If the source is finite (e.g., `ImageFileSource` with `loop=False`), `get_frame()` returns `None` after the last frame, and `step()` returns `[]` without setting any flag. The thread then continues to call `step()` repeatedly, sleeping `interval_s` each time, and never terminates on its own. A user who starts the pipeline on a finite source must call `stop()` manually, but the thread will spin forever if they don't. This is a resource leak and a logic error for finite sources.

packages/gptme-vision-node/src/gptme_vision_node/place.py
```diff
@@ -0,0 +1,324 @@
...
+                )
+                for s in samples
+            ]
+        self.places = places
+
+    @classmethod
+    def from_file(
+        cls,
+        path: str | Path,
+        embedder: Embedder | None = None,
```
TimeToBuildBob, line 352: ⚠️ **P2** — In `PlaceRecognizer.from_file`, the `embedder` argument defaults to `None`, causing a `HistogramEmbedder` to be instantiated. The `save` method persists only embeddings and wifi data, not the embedder type. If a gallery was saved using a custom embedder (e.g., a future CLIP embedder), the stored vectors live in that embedder's feature space. On reload via `from_file` without explicitly passing the same embedder, `recognize` embeds the query frame with the default histogram embedder, producing vectors in a different space. Cosine similarities between these mismatched vectors are meaningless, so place recognition returns incorrect ranks or an empty result. This is a silent correctness bug that only occurs when a non-default embedder is used and the gallery is reloaded via `from_file` without that same embedder. The fix is to persist an embedder identifier (e.g., a name or version) in the payload and reject or transform when it does not match the current embedder.

packages/gptme-vision-node/src/gptme_vision_node/cli.py
```diff
@@ -0,0 +1,146 @@
...
+def detect(source_spec: str, once: bool, interval: float) -> None:
+    """Run person/motion detection on a source."""
+    source = make_source(source_spec)
+    detectors: list[Detector] = [PersonDetector(), MotionDetector()]
+    try:
+        while True:
+            frame = source.get_frame()
+            if frame is None:
+                break
+            detections = [d for det in detectors for d in det.detect(frame)]
```
TimeToBuildBob: ⚠️ **P2** — The CLI commands (detect, look, enroll, whereami) call `source.get_frame()` and run detectors without wrapping these calls in try/except. Both `OpenCVCameraSource._ensure_open` (raises RuntimeError on failed open) and `detector.detect` (can raise ValueError or cv2 errors) will propagate as raw Python exceptions, resulting in a full traceback instead of a clean click error message. For example, running `gptme-vision-node detect --source camera:5` with no camera at index 5 raises `RuntimeError: failed to open capture source: 5`, which crashes the CLI. The same issue applies to `ImageFileSource.get_frame` if a file is unreadable. The CLI should catch these exceptions and present a `click.ClickException` for a user-friendly failure.

TimeToBuildBob:
## 🤖 AI code review

Adds a new `gptme-vision-node` package implementing a vision pipeline for a presence node: frame sources, person/motion detection, an LLM 'look' tool, place recognition via visual histograms and WiFi fingerprints, a thread-based pipeline, and a CLI. Also adds the package to the mypy exclude list in `.pre-commit-config.yaml` and includes a full test suite.

> ⚠️ **1 generated file excluded and NOT reviewed:** ``uv.lock`` (+50/-0 lines, 5KB). Machine-generated output is removed from the diff before review — from what the model sees and from the size gate alike — so nothing below says anything about it.

<details><summary><b>2 findings</b> · ⚠️ **2** P2 — non-blocking (round cap reached)</summary>

**⚠️ P2 medium · ℹ️ non-blocking** — `packages/gptme-vision-node/src/gptme_vision_node/place.py:337`

In `PlaceRecognizer.load`, the embedding loaded from JSON is converted with `np.asarray(s["embedding"], dtype=np.float64)` without validating its length against the embedder's expected output dimension. If a gallery file contains an embedding of the wrong size (for example, a hand-edited or corrupt `places.json` where the `embedder` field matches but the `embedding` list has, say, 1 element instead of the 960 produced by the default `HistogramEmbedder`), the subsequent `recognize` call will compute `np.dot` in `cosine_similarity` with mismatched vector lengths and raise `ValueError: shapes (960,) and (1,) not aligned`. The CLI `whereami` does not catch this exception, so the process exits with a traceback instead of the clean error message it gives for other malformed gallery formats. This is a robustness/error-handling bug: a corrupted gallery should be rejected at load time with a clear message, not crash later during ranking.

<sub>**How this was verified:** Traced the load path and the recognize path; `np.asarray` does not check shape, and `cosine_similarity` does not guard against mismatched lengths. The test suite has no test for this case.</sub>

**⚠️ P2 medium · ℹ️ non-blocking** — `packages/gptme-vision-node/src/gptme_vision_node/cli.py:108`

In `cli.py` `detect`, the `try` block around detector execution only catches `cv2.error`, `RuntimeError`, and `ValueError`. If any detector raises a different exception type—for example, a `TypeError` from a custom detector that misuses a numpy operation—the exception propagates out of the loop uncaught, and the CLI prints a full traceback instead of converting it to a `ClickException` like the other error paths do. The `VisionPipeline` catches all exceptions from detectors with a broad `except Exception`, so a custom detector that crashes in the pipeline logs and continues, but the same detector in the CLI crashes. This inconsistency means an abnormal detector failure in a CLI run produces an unfriendly traceback rather than a clean error message, and it can leave the source unreleased (though the `finally` block does run, so the source is released).

```suggestion
Add httpx.InvalidURL to the except tuple, or better, catch httpx.RequestError which is the base class for transport errors.
```

<sub>**How this was verified:** Checked httpx exception hierarchy: httpx.InvalidURL inherits from Exception, not HTTPError. Reproduced mentally with OPENAI_BASE_URL set to an invalid URL, which causes httpx.post to raise InvalidURL.</sub>

</details>

<details><summary><b>1 existing-module overlap (informational, not scored)</b></summary>

_This is a cheap file-index scan, not a merge blocker. A human decides whether to redirect, extend the existing module, or intentionally parallel-implement._

**⚠️ P2 medium · ℹ️ non-blocking** — `packages/gptme-vision-node/src/gptme_vision_node/place.py:1`

This appears to overlap with `gptme/gptme-contrib:packages/gptme-rag` which already provides RAG: document indexing, semantic/lexical search, persistence. Matched signals: embedding, embeddings, vector. Informational only — not a blocker. A human decides whether to redirect, extend the existing module, or intentionally parallel-implement.

<sub>**How this was verified:** cheap file-index scan (L0.5 existing-module overlap)</sub>

</details>

<details><summary><b>Files changed (18)</b> — the diff as I read it</summary>

- `.pre-commit-config.yaml` — Adds `gptme-vision-node` to the mypy exclude list.
- `packages/gptme-vision-node/Makefile` — New Makefile with `test` and `typecheck` targets.
- `packages/gptme-vision-node/README.md` — Documents the package components, usage, and testing.
- `packages/gptme-vision-node/pyproject.toml` — Declares package metadata, dependencies, and console script.
- `packages/gptme-vision-node/src/gptme_vision_node/__init__.py` — Exports public API symbols and version.
- `packages/gptme-vision-node/src/gptme_vision_node/cli.py` — Defines CLI subcommands: detect, look, enroll, and whereami.
- `packages/gptme-vision-node/src/gptme_vision_node/detect.py` — Implements PersonDetector (HOG) and MotionDetector.
- `packages/gptme-vision-node/src/gptme_vision_node/frame_source.py` — Defines FrameSource protocol, ImageFileSource, and OpenCVCameraSource.
- `packages/gptme-vision-node/src/gptme_vision_node/look.py` — Implements the vision LLM description tool.
- `packages/gptme-vision-node/src/gptme_vision_node/pipeline.py` — Implements the thread-based VisionPipeline with events.
- `packages/gptme-vision-node/src/gptme_vision_node/place.py` — Implements HistogramEmbedder, WifiSignature, and PlaceRecognizer.
- `packages/gptme-vision-node/tests/conftest.py` — Provides shared synthetic frame fixtures.
- `packages/gptme-vision-node/tests/test_cli.py` — Tests CLI parsing, error handling, and end-to-end flows.
- `packages/gptme-vision-node/tests/test_detect.py` — Tests person/motion detectors on synthetic frames.
- `packages/gptme-vision-node/tests/test_frame_source.py` — Tests image and camera frame sources.
- `packages/gptme-vision-node/tests/test_look.py` — Tests the LLM look tool with mocked HTTP.
- `packages/gptme-vision-node/tests/test_pipeline.py` — Tests the vision pipeline event logic and lifecycle.
- `packages/gptme-vision-node/tests/test_place.py` — Tests embedder, wifi similarity, and place recognizer.

</details>

| `0a25b67bd11c` | 3/5 | 5 | llm | 2026-08-27 21:01 UTC |
| `e994c75114d9` | 3/5 | 4 | llm | 2026-08-27 22:01 UTC |
| `f0c24cf22c85` | 2/5 | 4 | llm | 2026-08-28 00:32 UTC |
| `dc7a900bb7d7` | 4/5 | 5 | llm | 2026-08-28 03:01 UTC |
| `a983258cf224` | 3/5 | 3 | llm | 2026-08-28 04:18 UTC |
| `25d3ca25ac8c` | 3/5 | 3 | llm | 2026-08-28 07:38 UTC |
| `856c9ceeec62` | 2/5 | 2 | llm | 2026-08-28 11:59 UTC |

</details>

<sub>Reviewed `a78111f07005` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 280s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

TimeToBuildBob:
@greptileai review

packages/gptme-vision-node/src/gptme_vision_node/detect.py
TimeToBuildBob: Mapped tiny-frame HOG boxes back into original-frame coordinates and added a regression test.

Fixed in `e994c751`.

packages/gptme-vision-node/src/gptme_vision_node/detect.py
TimeToBuildBob: Motion detection now re-primes its background when frame dimensions change, covered by a regression test.

Fixed in `e994c751`.

packages/gptme-vision-node/src/gptme_vision_node/pipeline.py
TimeToBuildBob: The background pipeline now exits when its source is exhausted, while the public one-step API keeps its existing empty-event behavior.

Fixed in `e994c751`.

packages/gptme-vision-node/src/gptme_vision_node/place.py
TimeToBuildBob: Gallery JSON now records a stable embedder identifier and rejects incompatible reloads; legacy histogram galleries remain readable.

Fixed in `e994c751`.

packages/gptme-vision-node/src/gptme_vision_node/cli.py
TimeToBuildBob: Source parsing and capture/detector failures now become clean Click errors rather than raw tracebacks.

Fixed in `e994c751`.

packages/gptme-vision-node/src/gptme_vision_node/cli.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+@main.command()
+@source_option
+@click.option("--prompt", default="Describe what you see, briefly.")
+@click.option("--model", default=None, help="override VISION_MODEL")
+def look(source_spec: str, prompt: str, model: str | None) -> None:
+    """Describe the current frame via a vision LLM (needs OPENAI_API_KEY)."""
+    frame = _get_frame_or_die(make_source(source_spec))
+    click.echo(describe_frame(frame, prompt, model=model))
```
TimeToBuildBob: ⚠️ **P2** — The `look` CLI command does not catch exceptions raised by `describe_frame`. If the LLM endpoint returns a non-200 status (e.g. 401 invalid key) or a timeout, `httpx.HTTPStatusError` or `httpx.TimeoutException` propagates uncaught, causing the CLI to exit with a full Python traceback. This is inconsistent with the other commands, which convert failures to `click.ClickException`. A user sees an ugly stack trace and no actionable error message, and in the case of a 401 they cannot tell the API key is wrong.

packages/gptme-vision-node/src/gptme_vision_node/place.py
```diff
@@ -0,0 +1,338 @@
...
+                    embedding=(
+                        np.asarray(s["embedding"], dtype=np.float64)
+                        if s.get("embedding") is not None
+                        else None
+                    ),
+                    wifi={k: float(v) for k, v in (s.get("wifi") or {}).items()},
+                )
+                for s in samples
+            ]
+        self.places = places
```
TimeToBuildBob, line 346: ⚠️ **P2** — `PlaceRecognizer.load` loads a gallery but does not restore the saved `wifi_weight` from the JSON payload. If a user constructs a `PlaceRecognizer` with a non-default weight (say 0.3) and then calls `load` to read a gallery that was saved with a different weight (say 0.7), the fused visual+wifi score uses the recognizer's current weight rather than the gallery's, producing a different ranking than the gallery author intended. `from_file` avoids this by passing the saved weight to the constructor, but the public `load` path is inconsistent and can silently change results.

packages/gptme-vision-node/src/gptme_vision_node/cli.py
```diff
@@ -0,0 +1,157 @@
...
+            release()
+
+
+@main.command()
+@source_option
+@click.option("--prompt", default="Describe what you see, briefly.")
+@click.option("--model", default=None, help="override VISION_MODEL")
+def look(source_spec: str, prompt: str, model: str | None) -> None:
+    """Describe the current frame via a vision LLM (needs OPENAI_API_KEY)."""
+    frame = _get_frame_or_die(make_source(source_spec))
```
TimeToBuildBob: ⚠️ **P2** — The CLI commands `look`, `whereami`, and `enroll` create a frame source via `make_source` and read a frame, but never call `release()` on it. The `detect` command correctly calls `release()` in a `finally` block (lines 90-93). For a camera source (`OpenCVCameraSource`), this leaves the underlying `cv2.VideoCapture` handle open. In a short-lived CLI process the OS reclaims the handle on exit, but if any of these commands are invoked programmatically in a long-running process (e.g., embedded in the voice bridge), repeated invocations would leak camera handles and eventually exhaust available video devices. The fix is to mirror the `detect` pattern and release the source after use.

packages/gptme-vision-node/src/gptme_vision_node/cli.py
TimeToBuildBob: Wrapped vision endpoint failures in clean Click errors. Fixed in `739f2bfd`.

packages/gptme-vision-node/src/gptme_vision_node/place.py
TimeToBuildBob: Direct gallery loads now restore the persisted WiFi weight, while an explicit `from_file(..., wifi_weight=...)` still overrides it. Covered by regression tests in `739f2bfd` and `f0c24cf2`.

packages/gptme-vision-node/src/gptme_vision_node/cli.py
TimeToBuildBob: All one-shot capture commands now release their source in a `finally` path, including capture failures. Covered by regression tests in `739f2bfd`.

packages/gptme-vision-node/src/gptme_vision_node/pipeline.py
```diff
@@ -0,0 +1,148 @@
...
+        self._thread: threading.Thread | None = None
+        self._stop = threading.Event()
+
+    # -- core -------------------------------------------------------------
+
+    def _emit(self, event: VisionEvent) -> None:
+        if self.on_event is None:
+            return
+        try:
+            self.on_event(event)
```
TimeToBuildBob, line 77: ❌ **P1** — Calling `pipeline.stop()` from inside the `on_event` callback deadlocks. In `VisionPipeline._emit`, the callback is invoked on the pipeline thread (line 77). If that callback calls `stop()`, `stop()` sets `_stop` and then calls `self._thread.join(timeout)`. Since the current thread is the pipeline thread, `join()` raises `RuntimeError: cannot join current thread`. That exception is caught by the generic `except Exception` in `_emit`, logged, and swallowed — so the stop request is silently ignored and the pipeline keeps running. A subscriber that wants to shut down the pipeline upon detecting a person (a natural use case for a presence node) will find `stop()` has no effect and the thread keeps capturing frames. The fix is to avoid joining the current thread, e.g. set `_stop` and have the callback return, or dispatch the stop to another thread.

packages/gptme-vision-node/src/gptme_vision_node/frame_source.py
```diff
@@ -0,0 +1,95 @@
...
+            if not self._cap.isOpened():
+                cap, self._cap = self._cap, None
+                cap.release()
+                raise RuntimeError(f"failed to open capture source: {self.source!r}")
+        return self._cap
+
+    def get_frame(self) -> np.ndarray | None:
+        cap = self._ensure_open()
+        ok, frame = cap.read()
+        if not ok or frame is None:
```
TimeToBuildBob: ❌ **P1** — `OpenCVCameraSource.get_frame` returns None whenever `cap.read()` fails, i.e. on a failed or empty read. All callers treat None as end-of-stream: `VisionPipeline._run` breaks the capture loop (`if frame is None: break`) and `cli._get_frame_or_die` raises a ClickException (`if frame is None: raise ...`). A single transient failure—e.g. a camera that drops one frame, or a brief network stall on an RTSP stream—terminates the entire detection loop. For a camera source this is not an exhaust condition; it is a one-off failure that should be skipped or retried. The observable consequence is that a long-running presence-monitoring job stops after one missed frame, producing an error or silently exiting, which is wrong for a periodic capture loop.

TimeToBuildBob:
## Convergence adjudication

**Fixed in `dc7a900b`:**
- **P1 — callback-triggered stop:** `VisionPipeline.stop()` now detects the capture thread and does not self-join; the stop event terminates the loop. A regression test stops the pipeline from `on_event` and confirms the thread exits.
- **P1 — transient camera read:** `OpenCVCameraSource` now retries up to three reads before reporting exhaustion, with tests for both transient recovery and exhausted retry budget.
- **P2 — gallery save error:** `enroll` converts `OSError` into a clean `ClickException`.
- **P2 — valid non-object JSON gallery:** both gallery load paths reject non-object payloads with `ValueError`, preserving the CLI's clean error path.

**Remaining (non-blocking):**
- Existing-module overlap with `gptme-rag` is a false positive: this module compares fixed-size image/place embeddings and WiFi signatures; it does not index documents or provide semantic/lexical retrieval.
- Camera recovery is deliberately bounded at three immediate reads rather than infinite reconnect/backoff. That prevents one dropped frame from terminating the node while still surfacing a persistently dead source; reconnect policy is outside v0.

**CI:** previous head was fully green. Current head CI is running after the fix push. Local verification: 61 package tests pass, Ruff check/format clean, package `make typecheck` clean, and `git diff --check` clean.

**Domain risk:** real V4L2/RTSP hardware behavior remains the fragile area; a maintainer should smoke-test dropped-frame recovery on hardware before relying on unattended RTSP operation.

**Convergence:** detector reports `round_convergence.status=no_history`, `stable_rounds=0` of 2, and no Greptile findings. The escalation was driven by our AI-review fix-session cap, not usable Greptile round history. All four score-driving findings were adjudicated above; no further review is being triggered. Stop at maintainer judgment after CI settles.

packages/gptme-vision-node/src/gptme_vision_node/frame_source.py
TimeToBuildBob: Fixed in `dc7a900b`: camera sources retry three reads before returning `None`, so a single transient failure no longer ends capture. Tests cover transient recovery and persistent exhaustion; unbounded reconnect/backoff remains outside v0.

packages/gptme-vision-node/src/gptme_vision_node/pipeline.py
TimeToBuildBob: Fixed in `dc7a900b`: `stop()` now sets the stop event but skips joining when called on the capture thread; the callback returns and the loop exits. Regression coverage confirms callback-triggered stop terminates the worker.

packages/gptme-vision-node/src/gptme_vision_node/place.py
```diff
@@ -0,0 +1,348 @@
...
+        if saved_embedder is not None and saved_embedder != self.embedder.embedder_id:
+            raise ValueError(
+                "embedder mismatch: gallery uses "
+                f"{saved_embedder!r}, current embedder is {self.embedder.embedder_id!r}"
+            )
+        saved_wifi_weight = float(payload.get("wifi_weight", self.wifi_weight))
+        if not 0.0 <= saved_wifi_weight <= 1.0:
+            raise ValueError("wifi_weight must be in [0, 1]")
+        places: dict[str, list[PlaceSample]] = {}
+        for name, samples in payload.get("places", {}).items():
```
TimeToBuildBob: ⚠️ **P2** — In PlaceRecognizer.load, the line `for name, samples in payload.get("places", {}).items():` assumes the `places` key holds a dict. A gallery JSON file that is a valid object but has `places` as a non-dict (e.g., `[]`) causes `.items()` to raise AttributeError. The CLI's whereami and enroll commands wrap gallery loading in `try/except (OSError, ValueError)`, so this AttributeError escapes and produces a raw traceback instead of a clean error message. A user with a corrupted but well-formed gallery file gets an unhandled exception. The fix is to validate the type of `places` and raise a ValueError with a clear message.

packages/gptme-vision-node/src/gptme_vision_node/look.py
```diff
@@ -0,0 +1,96 @@
...
+    timeout: float = 60.0,
+) -> str:
+    """Send a frame to a vision LLM and return its description."""
+    key = api_key or os.environ.get("OPENAI_API_KEY")
+    if not key:
+        raise RuntimeError("OPENAI_API_KEY is not set (and no api_key passed)")
+    url = (base_url or os.environ.get("OPENAI_BASE_URL", DEFAULT_BASE_URL)).rstrip("/")
+
+    body = build_request(frame, prompt, model=model)
+    response = httpx.post(
```
TimeToBuildBob, line 82: ⚠️ **P2** — In describe_frame, the response from `httpx.post` is stored in `response` and never closed. Each call to describe_frame may leave a connection open until garbage collection. In a long-running process that invokes the look tool repeatedly (e.g., a daemon answering on-demand look calls), this can accumulate open sockets and exhaust file descriptors, leading to resource exhaustion. The httpx documentation recommends using the response as a context manager or calling `response.close()` after the body is consumed. The fix is to wrap the call in a `with` block, which ensures the connection is released.

packages/gptme-vision-node/src/gptme_vision_node/detect.py
```diff
@@ -0,0 +1,146 @@
...
+        self.blur_ksize = blur_ksize
+        self._background: np.ndarray | None = None
+
+    def reset(self) -> None:
+        self._background = None
+
+    def _prepare(self, frame: np.ndarray) -> np.ndarray:
+        gray = cv2.cvtColor(frame, cv2.COLOR_BGR2GRAY)
+        if self.blur_ksize > 1:
+            gray = cv2.GaussianBlur(gray, (self.blur_ksize, self.blur_ksize), 0)
```
TimeToBuildBob, line 117: ⚠️ **P2** — In MotionDetector.__init__, the `blur_ksize` parameter is not restricted to odd integers. OpenCV's `GaussianBlur` requires an odd kernel size; passing an even value like 4 raises `cv2.error` when `detect` is called. In the CLI `detect` command, the try/except around `detector.detect(frame)` only catches `RuntimeError` and `ValueError`, so a `cv2.error` propagates as an uncaught traceback, crashing the command. In the `VisionPipeline`, the exception is caught and logged, but the detector silently fails, meaning motion detection never works for that `MotionDetector` instance. Users constructing `MotionDetector` directly or via the CLI cannot pass an even `blur_ksize` today, yet the constructor accepts it. The fix is to validate that `blur_ksize` is a positive odd integer (or at least odd) in `__init__` and raise `ValueError` otherwise.

packages/gptme-vision-node/src/gptme_vision_node/cli.py
```diff
@@ -0,0 +1,184 @@
...
+    try:
+        return _get_frame_or_die(source)
+    finally:
+        _release_source(source)
+
+
+def _load_gallery(path: Path) -> PlaceRecognizer:
+    try:
+        return PlaceRecognizer.from_file(path)
+    except (OSError, ValueError) as exc:
```
TimeToBuildBob: ⚠️ **P2** — In `_load_gallery`, only `OSError` and `ValueError` are caught when calling `PlaceRecognizer.from_file`. However, `from_file` (and `load`) can raise `TypeError` for a malformed gallery: `float(payload.get("wifi_weight", 0.5))` in `place.py` raises `TypeError` when `wifi_weight` is `null` or a non-numeric string. A gallery file containing `"wifi_weight": null` therefore escapes the `except` clause and propagates out of the CLI, producing a raw traceback instead of the intended clean 'invalid gallery' message. The fix is to also catch `TypeError` in `_load_gallery` (and ideally validate the gallery structure before parsing).

packages/gptme-vision-node/src/gptme_vision_node/place.py
```diff
@@ -0,0 +1,348 @@
...
+    """HSV color histogram + coarse spatial grid, L2-normalized.
+
+    Global hue/saturation histogram concatenated with per-cell
+    hue histograms over a ``grid x grid`` layout, so both overall color
+    mood and rough spatial arrangement contribute. Cheap, deterministic,
+    dependency-free — a baseline that distinguishes visually distinct
+    rooms, not a general scene embedding.
+    """
+
+    def __init__(self, *, h_bins: int = 24, s_bins: int = 8, grid: int = 2) -> None:
```
TimeToBuildBob, line 53: ⚠️ **P2** — `HistogramEmbedder.__init__` does not validate that `grid` is a positive integer. If a caller constructs `HistogramEmbedder(grid=0)`, then `embed` will divide by zero at `gy * h // self.grid` and `gx * w // self.grid` (place.py lines 77-79), raising `ZeroDivisionError` instead of producing a useful error. While the default is fine, this is an unguarded parameter; the constructor should assert `grid > 0` (and similarly for `h_bins`/`s_bins` > 0) to fail fast with a clear message.

TimeToBuildBob:
Addressed the current AI-review findings in `25d3ca25`:

- looping `detect` now exits cleanly when a finite source is exhausted, while one-shot commands retain the explicit no-frame error;
- pipeline restarts reset presence/debounce state, so a fresh run emits `PERSON_APPEARED` correctly;
- added regression tests for both paths.

Verified locally: 25 focused CLI/pipeline tests pass, Ruff check/format pass, commit-time pre-commit passes, and the remote branch is pinned to `25d3ca25`. The required in-band reviewer timed out after 600 seconds and pushed via its documented fail-open path; current-head GitHub CI and the automatic AI review are pending. No Greptile review was triggered.

packages/gptme-vision-node/src/gptme_vision_node/cli.py
```diff
@@ -0,0 +1,184 @@
...
+        raise click.ClickException(str(exc)) from exc
+
+
+def _get_frame_or_die(source: FrameSource) -> np.ndarray:
+    try:
+        frame = source.get_frame()
+    except (OSError, RuntimeError, ValueError) as exc:
+        raise click.ClickException(str(exc)) from exc
+    if frame is None:
+        raise click.ClickException("source produced no frame")
```
TimeToBuildBob, line 63: ❌ **P1** — In `cli.py`, `_get_frame_or_die` raises `ClickException("source produced no frame")` whenever `source.get_frame()` returns None. The `detect` command calls this in a loop; when using a finite source like a directory with `loop=False` or a camera that temporarily fails to produce a frame (returns None after retries), the command exits with a non-zero code and an error message instead of exiting normally when the source is exhausted or skipping the transient failure. This makes the `detect` command fail on the last frame of a finite set unless `--once` is used, and causes a hard error on transient camera read failures.

packages/gptme-vision-node/src/gptme_vision_node/pipeline.py
```diff
@@ -0,0 +1,149 @@
...
+                self._process_frame(frame)
+            except Exception:
+                logger.exception("pipeline step failed")
+            elapsed = time.monotonic() - started
+            self._stop.wait(max(0.0, self.interval_s - elapsed))
+
+    def start(self) -> None:
+        if self._thread is not None and self._thread.is_alive():
+            raise RuntimeError("pipeline already running")
+        self._stop.clear()
```
TimeToBuildBob, line 137: ⚠️ **P2** — In `VisionPipeline.start()`, only `self._stop.clear()` is called; the internal state `_person_present` and `_frames_without_person` are not reset. If the pipeline was stopped after a person had appeared (so `_person_present=True`) and then restarted (e.g., after a source exhausts and a new source is set, or if the user calls `stop()` and then `start()` again), the next frame with a person will not emit `PERSON_APPEARED` because `_person_present` is still True. This can cause missed presence events after a restart, and the debounce counter may hold stale values, leading to incorrect `PERSON_LEFT` behavior.

packages/gptme-vision-node/src/gptme_vision_node/cli.py
TimeToBuildBob: Fixed in `25d3ca25`: the looping `detect` command now treats `None` as normal source exhaustion and exits cleanly, while one-shot commands keep their explicit no-frame error. Added a finite-source CLI regression test.

packages/gptme-vision-node/src/gptme_vision_node/pipeline.py
TimeToBuildBob: Fixed in `25d3ca25`: `VisionPipeline.start()` now resets both presence-state fields before launching a fresh capture thread. Added a stop/exhaust/restart regression test that verifies `PERSON_APPEARED` fires again.

packages/gptme-vision-node/src/gptme_vision_node/place.py
TimeToBuildBob: Fixed in the prior `a983258c` push: both gallery load paths now validate that `places` is an object before iterating it and raise `ValueError` for malformed payloads. Regression coverage exercises a non-object `places` value through both `load()` and `from_file()`.

packages/gptme-vision-node/src/gptme_vision_node/place.py
```diff
@@ -0,0 +1,354 @@
...
+            [hsv], [0, 1], None, [self.h_bins, self.s_bins], [0, 180, 0, 256]
+        )
+        parts.append(global_hist.flatten())
+
+        # Per-cell H+S histograms (coarse spatial layout).
+        h, w = hsv.shape[:2]
+        for gy in range(self.grid):
+            for gx in range(self.grid):
+                cell = hsv[
+                    gy * h // self.grid : (gy + 1) * h // self.grid,
```
TimeToBuildBob, line 81: ❌ **P1** — In HistogramEmbedder.embed, the per-cell slicing uses integer division: `gy * h // self.grid : (gy + 1) * h // self.grid`. When a frame dimension is smaller than grid (e.g., a 1-pixel-high image with grid=2), the slice for the first cell becomes `0:0`, an empty array. cv2.calcHist() on an empty array raises an error (OpenCV assertion failure), which propagates up through PlaceRecognizer.enroll/recognize and crashes the CLI with a traceback. This affects any input image with height or width less than the default grid of 2, which is a plausible user case (e.g., a tiny icon file). The embedder should skip empty cells or pad the frame. Verified by reading the slicing logic in place.py lines 80–83.

packages/gptme-vision-node/src/gptme_vision_node/place.py
```diff
@@ -0,0 +1,354 @@
...
+        saved_places = payload.get("places", {})
+        if not isinstance(saved_places, dict):
+            raise ValueError("gallery places must be a JSON object")
+        places: dict[str, list[PlaceSample]] = {}
+        for name, samples in saved_places.items():
+            places[name] = [
+                PlaceSample(
+                    embedding=(
+                        np.asarray(s["embedding"], dtype=np.float64)
+                        if s.get("embedding") is not None
```
TimeToBuildBob: ⚠️ **P2** — In PlaceRecognizer.load, the samples list is iterated with `for s in samples` and each `s` is assumed to be a dict, calling `s.get("embedding")`. If a gallery file contains a sample that is not a dict (e.g., an integer or string), `s.get` raises AttributeError. This exception is not caught by cli._load_gallery, which only handles OSError, TypeError, and ValueError, so the `enroll` or `whereami` command prints a Python traceback and exits with a non-clean error instead of reporting the malformed gallery. The same crash occurs if a place's value is a dict instead of a list, or if a sample is a list. Triggered by any hand-edited or partially written places.json.

packages/gptme-vision-node/src/gptme_vision_node/cli.py
```diff
@@ -0,0 +1,190 @@
...
+    source = make_source(source_spec)
+    detectors: list[Detector] = [PersonDetector(), MotionDetector()]
+    try:
+        while True:
+            frame = _get_frame(source)
+            if frame is None:
+                break
+            try:
+                detections = [d for det in detectors for d in det.detect(frame)]
+            except (RuntimeError, ValueError) as exc:
```
TimeToBuildBob: ⚠️ **P2** — In the CLI `detect` command, the call to `detector.detect(frame)` is wrapped in a try-except that catches only RuntimeError and ValueError. OpenCV functions (e.g., cvtColor, HOGDescriptor) raise cv2.error, which is a subclass of Exception but not of RuntimeError or ValueError. If a detector receives a malformed frame (e.g., a non-BGR image, corrupted data from a stream, or a frame with an unexpected channel count), cv2.error escapes the try block, causing the CLI to print a full traceback instead of a clean Click error. This violates the error-handling pattern used elsewhere in the CLI (which converts expected failures into user-friendly messages). The fix is to broaden the exception list to include cv2.error or to catch Exception for this specific block.

packages/gptme-vision-node/src/gptme_vision_node/place.py
TimeToBuildBob: Fixed in `856c9cee`: empty spatial-grid cells now contribute a zero histogram, preserving the stable vector dimension without calling OpenCV on an empty array. A 1×1-frame regression test covers the reported case.

packages/gptme-vision-node/src/gptme_vision_node/place.py
TimeToBuildBob: Fixed in `856c9cee`: gallery loading now validates that every place value is a list and every sample is a JSON object before deserialization. Tests cover object-valued places and scalar/string/list samples, including the clean CLI error path.

packages/gptme-vision-node/src/gptme_vision_node/cli.py
TimeToBuildBob: Fixed in `856c9cee`: the detect command now catches `cv2.error` alongside its existing expected detector errors and converts it to `ClickException`. A CLI regression test verifies the error is user-facing without a traceback.

TimeToBuildBob:
## Final convergence adjudication

**Fixed in `856c9cee`:**
- **P1 — tiny spatial-grid cells:** empty grid cells now contribute a zero histogram rather than calling OpenCV on an empty array; a 1×1-frame regression preserves the fixed embedding dimension.
- **P2 — malformed gallery samples:** each place must contain a list of JSON-object samples before deserialization; object-valued places and scalar/string/list samples now fail with a clean `ValueError`, including through the CLI.
- **P2 — OpenCV detector errors:** `detect` now converts `cv2.error` to a user-facing Click error; the regression test confirms there is no traceback.

**Remaining (non-blocking):**
- The `gptme-rag` overlap hint is a false positive: this package compares image/place embeddings and WiFi signatures, not documents or semantic/lexical indexes.
- Indefinite camera reconnect/backoff remains outside vision v0; bounded immediate retries intentionally surface a persistently dead source.

**CI:** current-head CI is running. Already passing: Python 3.10 and 3.11 `gptme-vision-node` jobs plus integration tests; no failures are reported. Local verification: all 82 package tests pass, Ruff check/format pass, package mypy passes, scoped pre-commit passes, and `git diff --check` passes.

**Domain risk:** real V4L2/RTSP hardware behavior remains the fragile area. Smoke-test dropped-frame recovery on hardware before relying on unattended RTSP operation.

**Convergence:** detector reports `round_convergence.status=no_history`, `stable_rounds=0` of 2, no Greptile findings, and `should_request_review_after_fixes=false`. The three latest score-driving AI-review findings are fixed and their threads are replied to and resolved. No further Greptile or AI review is being triggered; stop at maintainer judgment after CI settles.

packages/gptme-vision-node/src/gptme_vision_node/cli.py
```diff
@@ -0,0 +1,191 @@
...
+            raise click.ClickException(f"invalid camera index: {spec!r}") from exc
+        raise click.ClickException(str(exc)) from exc
+    except OSError as exc:
+        raise click.ClickException(str(exc)) from exc
+
+
+def _get_frame(source: FrameSource) -> np.ndarray | None:
+    try:
+        return source.get_frame()
+    except (OSError, RuntimeError, ValueError) as exc:
```
TimeToBuildBob: ❌ **P1** — The `_get_frame` helper catches `(OSError, RuntimeError, ValueError)` but not `cv2.error`, which is the exception OpenCV raises for camera/stream read failures. When a `FrameSource` backed by `OpenCVCameraSource` hits a read error (e.g., webcam unplugged or RTSP stream disconnects), `cv2.error` propagates out of `_get_frame` and then out of `_capture_frame` (used by `look`, `enroll`, `whereami`) and out of the `detect` loop's `_get_frame` call, bypassing the `except (cv2.error, ...)` that only wraps the detector calls. The CLI then exits with a full Python traceback instead of the clean `ClickException` message the code otherwise produces. Add `cv2.error` to the tuple in `_get_frame` (and/or `_get_frame_or_die`).

packages/gptme-vision-node/src/gptme_vision_node/place.py
```diff
@@ -0,0 +1,363 @@
...
+                    f"gallery place {name!r} samples must be a list of JSON objects"
+                )
+            places[name] = [
+                PlaceSample(
+                    embedding=(
+                        np.asarray(s["embedding"], dtype=np.float64)
+                        if s.get("embedding") is not None
+                        else None
+                    ),
+                    wifi={k: float(v) for k, v in (s.get("wifi") or {}).items()},
```
TimeToBuildBob: ❌ **P1** — In `PlaceRecognizer.load`, the wifi field is parsed with `(s.get("wifi") or {}).items()`. If a gallery JSON contains a `"wifi"` value that is not a dict (e.g., a list or string), the expression evaluates to that non-dict and `.items()` raises `AttributeError`, which is not caught by `_load_gallery` in cli.py (it catches only OSError, TypeError, ValueError). A user with a malformed gallery file triggers a raw traceback in `whereami` or `enroll` instead of the intended `invalid gallery` ClickException. Since `s.get("wifi")` can be any JSON type, guard with `isinstance(..., dict)` or catch `AttributeError` alongside the other exceptions.

packages/gptme-vision-node/src/gptme_vision_node/cli.py
TimeToBuildBob: Fixed in `a78111f0`: `_get_frame` now catches `cv2.error` alongside OSError/RuntimeError/ValueError and converts it to ClickException, so camera/stream read failures surface as a clean CLI error instead of a traceback. Covered by `test_cli_opencv_source_read_failure_is_clean_error`.

packages/gptme-vision-node/src/gptme_vision_node/place.py
TimeToBuildBob: Fixed in `a78111f0`: gallery load now requires each sample's `wifi` field to be a JSON object (or omitted/empty) and raises ValueError otherwise, so `_load_gallery` can convert it to a clean ClickException. Covered by `test_load_rejects_non_object_wifi` and `test_cli_whereami_non_object_wifi_is_clean_error`.

TimeToBuildBob:
Addressed the current AI-review P1s in `a78111f0`:

- `_get_frame` now converts `cv2.error` into a ClickException, matching detector handling;
- gallery load rejects non-object sample `wifi` with `ValueError` instead of `AttributeError`.

Both threads are replied to and resolved. Local verification: 87 package tests pass, Ruff and package mypy are clean. In-band reviewer could not complete (OpenRouter 900s deadline / all aspect passes failed) and pushed via its fail-open path.

Current-head CI is fully green and the PR is mergeable. Self-merge remains ineligible because no current-head AI/Greptile review completed and this new package plus pre-commit config change requires human review under policy. Waiting for maintainer review/merge.

TimeToBuildBob:
Current-head re-review of `a78111f0` is blocked on the OpenRouter review-key daily limit until `2026-08-29T00:00Z`. The 18:22 sweep already failed this head. I am not retrying in-band or stealing other keys.

CI is green and mergeable; all finding threads are resolved. The stale 2/5 marker still points at `856c9cee` (those P1s were already fixed here). Greptile is dark — not triggering.

Self-merge remains ineligible regardless: new package plus `.pre-commit-config.yaml` requires human review. Scheduled recheck at 00:10Z to replace the stale marker after the quota reset. Maintainer merge is still the landing path.

TimeToBuildBob:
Automated merge handoff for `a78111f07005`:

- CI gate: satisfied
- Review threads: clear
- `.pre-commit-config.yaml`

This head is ready for maintainer review and **manual merge**. Project monitoring marked it `human_merge_required` and will not spend retry sessions attempting an automated merge. A new head will be evaluated independently.

Current-head AI re-review of `a78111f0` remains quota-gated (`OPENROUTER_API_KEY_REVIEW` `limit_remaining=0` until `2026-08-29T00:00Z`). I am not retrying in-band, stealing other keys, or triggering Greptile. `bob-ai-review-sweep` and the already-scheduled 00:10Z recheck will replace the stale `856c9cee` 2/5 marker after reset; even a fresh 5/5 will not make this self-merge eligible.

packages/gptme-vision-node/src/gptme_vision_node/place.py
```diff
@@ -0,0 +1,370 @@
...
+            for s in samples:
+                wifi_raw = s.get("wifi") or {}
+                if not isinstance(wifi_raw, dict):
+                    raise ValueError(
+                        f"gallery place {name!r} sample wifi must be a JSON object"
+                    )
+                loaded_samples.append(
+                    PlaceSample(
+                        embedding=(
+                            np.asarray(s["embedding"], dtype=np.float64)
```
TimeToBuildBob, line 337: ⚠️ **P2** — In `PlaceRecognizer.load`, the embedding loaded from JSON is converted with `np.asarray(s["embedding"], dtype=np.float64)` without validating its length against the embedder's expected output dimension. If a gallery file contains an embedding of the wrong size (for example, a hand-edited or corrupt `places.json` where the `embedder` field matches but the `embedding` list has, say, 1 element instead of the 960 produced by the default `HistogramEmbedder`), the subsequent `recognize` call will compute `np.dot` in `cosine_similarity` with mismatched vector lengths and raise `ValueError: shapes (960,) and (1,) not aligned`. The CLI `whereami` does not catch this exception, so the process exits with a traceback instead of the clean error message it gives for other malformed gallery formats. This is a robustness/error-handling bug: a corrupted gallery should be rejected at load time with a clear message, not crash later during ranking.

packages/gptme-vision-node/src/gptme_vision_node/cli.py
```diff
@@ -0,0 +1,191 @@
...
+    source = make_source(source_spec)
+    detectors: list[Detector] = [PersonDetector(), MotionDetector()]
+    try:
+        while True:
+            frame = _get_frame(source)
+            if frame is None:
+                break
+            try:
+                detections = [d for det in detectors for d in det.detect(frame)]
+            except (cv2.error, RuntimeError, ValueError) as exc:
```
TimeToBuildBob, line 108: ⚠️ **P2** — In `cli.py` `detect`, the `try` block around detector execution only catches `cv2.error`, `RuntimeError`, and `ValueError`. If any detector raises a different exception type—for example, a `TypeError` from a custom detector that misuses a numpy operation—the exception propagates out of the loop uncaught, and the CLI prints a full traceback instead of converting it to a `ClickException` like the other error paths do. The `VisionPipeline` catches all exceptions from detectors with a broad `except Exception`, so a custom detector that crashes in the pipeline logs and continues, but the same detector in the CLI crashes. This inconsistency means an abnormal detector failure in a CLI run produces an unfriendly traceback rather than a clean error message, and it can leave the source unreleased (though the `finally` block does run, so the source is released).

packages/gptme-vision-node/src/gptme_vision_node/place.py
TimeToBuildBob: Deferred as non-blocking P2 for vision v0.

`enroll`/`save` write embeddings from the same embedder that `load` restores, so a matching `embedder_id` with a wrong vector length only happens if someone hand-edits `places.json`. Load already rejects the other malformed-gallery cases. A corrupt vector then fails later in `cosine_similarity` with `ValueError`; worth fail-fast at load as a follow-up, but not a new head on a `human_merge_required` package PR.

packages/gptme-vision-node/src/gptme_vision_node/cli.py
TimeToBuildBob: Deferred as non-blocking P2 for vision v0.

`detect` already converts the documented detector failures (`cv2.error`, `RuntimeError`, `ValueError`) and the `finally` path releases the source. The CLI constructs `PersonDetector` + `MotionDetector` — there is no custom-detector injection here. `VisionPipeline` is the long-running path and already has a broad `except Exception`. Wrapping all `Exception` in the CLI would hide unexpected bugs; not changing this on a human-merge-required head.

TimeToBuildBob:
Current-head AI review of `a78111f0` landed at 01:02Z after the quota reset: two P2s, no P0/P1.

Both deferred as non-blocking for v0 (threads replied + resolved). Self-merge remains ineligible: `.pre-commit-config.yaml` requires human review. The existing `human_merge_required` marker on this SHA still stands — no new head, no merge attempt, no Greptile trigger.

Maintainer merge is still the landing path.

PR #1530, opened 2026-08-27
perf(rag): cache scan_lessons() by mtime+size to skip YAML re-parse

TimeToBuildBob:
## Why

`scan_lessons()` YAML-parses every lesson file on every call. The Claude Code `match-lessons` hook is a **fresh process per tool call**, so a process-local cache there buys nothing. On Bob's corpus this is 0.46–0.62s of blocking latency on every `Read|Bash|Grep|WebFetch|WebSearch` PreToolUse (and every UserPromptSubmit).

Median session in a 40-transcript sample: 47 hook-eligible tool calls → ~24s of re-parsing an unchanged corpus. p90: ~160s.

Directory walks are cheap (warm `rglob` is tens of milliseconds). The tax is `yaml.safe_load`.

## What

Cache parsed lesson records keyed by `(resolved path, st_mtime_ns, st_size)`:

- In-memory for same-process reuse
- On disk (`$XDG_CACHE_HOME/gptme/lesson-scan/scan-v1.json`, override with `GPTME_LESSON_SCAN_CACHE`) so a new hook process hits too
- Re-parse only files whose mtime/size changed; a lesson written mid-session still matches on the next call
- Dedup, `session_categories`, BM25 body, and status filtering are unchanged — the cache sits under the existing walk/filter, it does not bypass them

`GPTME_LESSON_SCAN_CACHE=off` disables the disk layer (tests isolate via a tmp dir).

## Measured (Bob's 702-lesson corpus)

| path | wall |
|---|---|
| cold (no cache) | **0.458s** |
| warm disk (fresh process) | **0.055s** |
| warm memory | **0.049s** |

Warm disk is ~12% of the previous per-invocation cost.

## Tests

Regression coverage for: cold start, unchanged corpus, one file modified, one file added, one file deleted, mid-session write, disk surviving a memory reset (new process), `session_categories` round-trip, inactive files cached as skips, caller mutation not corrupting the cache.

Existing scan/filter/BM25 tests still pass (139 in `test_lesson_matcher.py`).

## Notes

This is the library-layer fix. Bob's `.claude/hooks/match-lessons.py` already imports `gptme_rag.lesson_matcher.scan_lessons`, so it picks this up via the gptme-contrib submodule pin after merge.

### DIFF
--- a/packages/gptme-rag/src/gptme_rag/lesson_matcher.py
+++ b/packages/gptme-rag/src/gptme_rag/lesson_matcher.py
@@ -733,15 +733,27 @@ def clear_scan_cache() -> None:
     _disk_dirty = False
 
 
+_LESSON_LIST_FIELDS = (
+    "keywords",
+    "patterns",
+    "tags",
+    "harness_restrict",
+    "session_categories",
+)
+
+
+def _valid_cached_lesson(lesson: object) -> bool:
+    """Return whether a persisted lesson has the shape cloning requires."""
+    return isinstance(lesson, dict) and all(
+        isinstance(lesson.get(field), list) for field in _LESSON_LIST_FIELDS
+    )
+
+
 def _clone_lesson(lesson: dict[str, Any]) -> dict[str, Any]:
     """Shallow-copy a lesson dict so callers cannot mutate the cache."""
     return {
         **lesson,
-        "keywords": list(lesson["keywords"]),
-        "patterns": list(lesson["patterns"]),
-        "tags": list(lesson["tags"]),
-        "harness_restrict": list(lesson["harness_restrict"]),
-        "session_categories": list(lesson["session_categories"]),
+        **{field: list(lesson[field]) for field in _LESSON_LIST_FIELDS},
     }
 
 
@@ -770,7 +782,7 @@ def _load_disk_cache() -> None:
         lesson = entry.get("lesson")
         if not isinstance(mtime_ns, int) or not isinstance(size, int):
             continue
-        if lesson is not None and not isinstance(lesson, dict):
+        if lesson is not None and not _valid_cached_lesson(lesson):
             continue
         _parse_cache[path] = (mtime_ns, size, lesson)
 
--- a/packages/gptme-rag/src/gptme_rag/lesson_matcher.py
+++ b/packages/gptme-rag/src/gptme_rag/lesson_matcher.py
@@ -740,12 +740,19 @@ _LESSON_LIST_FIELDS = (
     "harness_restrict",
     "session_categories",
 )
+_LESSON_STRING_FIELDS = ("path", "title", "description", "when_to_use", "body")
 
 
 def _valid_cached_lesson(lesson: object) -> bool:
-    """Return whether a persisted lesson has the shape cloning requires."""
-    return isinstance(lesson, dict) and all(
-        isinstance(lesson.get(field), list) for field in _LESSON_LIST_FIELDS
+    """Return whether a persisted lesson has the shape downstream consumers require."""
+    return (
+        isinstance(lesson, dict)
+        and all(isinstance(lesson.get(field), list) for field in _LESSON_LIST_FIELDS)
+        and all(isinstance(lesson.get(field), str) for field in _LESSON_STRING_FIELDS)
+        and (lesson.get("id") is None or isinstance(lesson.get("id"), str))
+        and (lesson.get("skill_name") is None or isinstance(lesson.get("skill_name"), str))
+        and isinstance(lesson.get("is_skill"), bool)
+        and isinstance(lesson.get("n_keywords"), int)
     )
 
 
@@ -1007,7 +1014,6 @@ def scan_lessons(lesson_dirs: list[Path]) -> list[dict[str, Any]]:
             resolved = str(f.resolve())
             if resolved in seen_paths:
                 continue
-            seen_paths.add(resolved)
 
             relative_path = f.relative_to(lesson_dir)
             is_skill_file = f.name == "SKILL.md"
@@ -1026,6 +1032,7 @@ def scan_lessons(lesson_dirs: list[Path]) -> list[dict[str, Any]]:
             if not is_skill_file and f.name in seen_names:
                 continue
 
+            seen_paths.add(resolved)
             readable, lesson = _cached_parse_lesson_file(f)
             if not readable:
                 continue
--- a/packages/gptme-rag/src/gptme_rag/lesson_matcher.py
+++ b/packages/gptme-rag/src/gptme_rag/lesson_matcher.py
@@ -31,8 +31,11 @@ Public API::
 
 from __future__ import annotations
 
+import json
 import math
+import os
 import re
+import tempfile
 from pathlib import Path
 from collections.abc import Sequence
 from typing import Any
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
CI is fully green, including `codecov/patch`; no human or Greptile review has landed yet. No code change is needed from this notification. Waiting for the initial review.

TimeToBuildBob:
@greptileai review

packages/gptme-rag/src/gptme_rag/lesson_matcher.py
```diff
@@ -678,6 +681,254 @@ def _score_skill_descriptor(lesson: dict[str, Any], prompt_lower: str) -> tuple[
...
+    try:
+        resolved = str(path.resolve())
+    except OSError:
+        return False, None
+    key = _stat_key(path)
+    if key is None:
+        return False, None
+    mtime_ns, size = key
+    cached = _parse_cache.get(resolved)
+    if cached is not None and cached[0] == mtime_ns and cached[1] == size:
```
TimeToBuildBob, line 923: ❌ **P1** — On a cache hit, _cached_parse_lesson_file returns the stored lesson after only checking that the file's mtime and size match the cached values; it never verifies the file is still readable. If a lesson file that was previously cached and included in results later has its permissions changed to make it unreadable (while mtime and size stay the same), the cache continues to serve the old lesson. The caller then includes that lesson in the result set, whereas the pre-cache implementation attempted to read the file, raised a PermissionError, and skipped it. This can cause lessons that the user has made unavailable to keep matching and influencing tool calls, a correctness regression that is not covered by the new tests.

packages/gptme-rag/src/gptme_rag/lesson_matcher.py
```diff
@@ -678,6 +681,254 @@ def _score_skill_descriptor(lesson: dict[str, Any], prompt_lower: str) -> tuple[
...
+
+
+def _prune_missing_from_cache(scanned_roots: list[str], seen_paths: set[str]) -> None:
+    """Drop cache entries under *scanned_roots* that were not seen this scan."""
+    global _disk_dirty
+    stale = [
+        path
+        for path in _parse_cache
+        if path not in seen_paths
+        and any(path == root or path.startswith(root + os.sep) for root in scanned_roots)
```
TimeToBuildBob: ⚠️ **P2** — The pruning step only removes cache entries that are under one of the roots scanned in the current call. Entries from roots that were scanned in previous calls but are no longer in the lesson_dirs argument are never deleted from the module-level _parse_cache. Because _save_disk_cache() writes the entire cache on every dirty scan, these stale entries are also re-serialized to the on-disk cache file each time. Over a long-lived process or across many invocations with varying lesson_dirs, the in-memory dict and the scan-v1.json file grow without bound, causing ever-larger JSON writes and memory pressure. The cache never forgets lessons from removed directories.

TimeToBuildBob:
## 🤖 AI code review

Adds a two-layer cache (in-memory and on-disk JSON) to scan_lessons(), keyed by resolved path plus st_mtime_ns and st_size, so unchanged lesson files are not re-parsed. The disk layer is disabled via GPTME_LESSON_SCAN_CACHE=off or when the env var is set to a falsy value. Also refactors scan_lessons() to move the seen_paths.add() call later in the loop, and adds a test fixture and regression tests for the cache behaviour.

**❌ P1 high** — `packages/gptme-rag/src/gptme_rag/lesson_matcher.py:1038`

When a file is skipped due to filename dedup (its resolved path is not added to seen_paths), a later symlink that resolves to the same file but has a different name will not be recognized as already seen. It will pass the `resolved in seen_paths` check, its name is not in seen_names (since the shadowed file's name was not registered), so it will be parsed and added as a separate lesson, producing a duplicate. For example, with dir1/foo.md (active), dir2/foo.md (active), and dir3/bar.md symlinked to dir2/foo.md, scanning [dir1, dir2, dir3] returns both foo.md and bar.md (both resolving to the same file) instead of just the first. This breaks the first-dir-wins dedup invariant and can cause a lesson to match twice. The fix is to add the resolved path to seen_paths before the name-dedup continue, so the symlink is skipped.

```suggestion
Move `seen_paths.add(resolved)` to just after the `resolved = str(f.resolve())` line, before the name-dedup check, so shadowed files also reserve their path.
```

<sub>**How this was verified:** Traced the scan_lessons loop: for a symlink in dir2 to dir1/archive/foo.md, when dir2 is processed first, seen_paths is empty, the archive check is skipped, and _parse_lesson_file(f) returns the active lesson (status active), appending it. The archive file is only added to seen_paths when its own path is iterated later.</sub>

</details>

- `packages/gptme-rag/src/gptme_rag/lesson_matcher.py` — Adds cache globals, disk cache load/save/prune functions, and refactors scan_lessons() to use the cached parse and to update seen_paths only after the name-dedup check.
- `packages/gptme-rag/tests/test_lesson_matcher.py` — Adds an autouse fixture isolating the scan cache to a temp dir, plus a TestScanLessonsCache class with tests for cold start, unchanged, modified, added, deleted files, disk persistence, and eviction.

</details>

| `2437e06e63c2` | 3/5 | 2 | llm | 2026-08-27 21:30 UTC |
| `76c72f8cbc46` | 3/5 | 2 | llm | 2026-08-27 22:24 UTC |
| `5315f6ccd692` | 4/5 | 3 | llm | 2026-08-27 22:32 UTC |
| `7da9d89e0185` | 3/5 | 3 | llm | 2026-08-27 23:43 UTC |

</details>

<sub>Reviewed `3aa90c371081` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 429s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

packages/gptme-rag/src/gptme_rag/lesson_matcher.py
TimeToBuildBob: Fixed in `76c72f8c`: cache hits now verify readability and evict an unreadable entry before allowing a same-name lesson from a later directory. Added a regression test covering the permission change after the initial cache fill.

packages/gptme-rag/src/gptme_rag/lesson_matcher.py
TimeToBuildBob: Fixed in `76c72f8c`: each complete `scan_lessons()` call now evicts every cache entry outside its current corpus, so abandoned roots cannot accumulate in memory or the persisted JSON. Added a regression test that switches roots and verifies eviction.

packages/gptme-rag/src/gptme_rag/lesson_matcher.py
```diff
@@ -678,6 +681,253 @@ def _score_skill_descriptor(lesson: dict[str, Any], prompt_lower: str) -> tuple[
...
+    try:
+        resolved = str(path.resolve())
+    except OSError:
+        return False, None
+    key = _stat_key(path)
+    if key is None:
+        return False, None
+    mtime_ns, size = key
+    cached = _parse_cache.get(resolved)
+    if cached is not None and cached[0] == mtime_ns and cached[1] == size:
```
TimeToBuildBob, line 923: ❌ **P1** — The cache key uses `st_mtime_ns` and `st_size` to decide whether a file changed. On filesystems with coarse mtime granularity (e.g., 1-second resolution on FAT32 or some network mounts), a lesson file rewritten within the same second and with the same byte length (e.g., changing 'foo' to 'bar' — both 3 chars) will have identical mtime and size, so the cache hit incorrectly returns the stale lesson. This means a lesson updated mid-session can be served from the old cached copy, causing the agent hook to match using outdated keywords/patterns. The PR's claim that mid-session writes are visible only holds when mtime or size actually differ. To close the gap, the cache should incorporate a content hash rather than relying solely on mtime+size. _(2/3 passes — pass 2 looked and disagreed)_

packages/gptme-rag/src/gptme_rag/lesson_matcher.py
```diff
@@ -678,6 +681,253 @@ def _score_skill_descriptor(lesson: dict[str, Any], prompt_lower: str) -> tuple[
...
+        if not isinstance(path, str) or not isinstance(entry, dict):
+            continue
+        mtime_ns = entry.get("mtime_ns")
+        size = entry.get("size")
+        lesson = entry.get("lesson")
+        if not isinstance(mtime_ns, int) or not isinstance(size, int):
+            continue
+        if lesson is not None and not isinstance(lesson, dict):
+            continue
+        _parse_cache[path] = (mtime_ns, size, lesson)
```
TimeToBuildBob, line 794: ⚠️ **P2** — `_load_disk_cache` populates the in-memory cache with lesson dicts from JSON after only checking that `lesson` is a dict (line 773). It does not verify the presence of the required keys (`keywords`, `patterns`, `tags`, `harness_restrict`, `session_categories`). If the cache file is corrupted, hand-edited, or written by a version of the code that omits some keys, a later cache hit in `_cached_parse_lesson_file` will call `_clone_lesson`, which indexes those keys directly and raises `KeyError`. That exception is not caught inside `_cached_parse_lesson_file` (only the parse step is wrapped), so it propagates out of `scan_lessons`, crashing the caller. While the cache is written atomically, a malformed file remains a single-point-of-failure robustness bug. _(3/3 passes)_

TimeToBuildBob: Fixed in `5315f6cc`: disk entries are now admitted only when every list field required by `_clone_lesson` has the expected shape; malformed entries are ignored and reparsed from source. Added a regression test using a same-mtime/same-size malformed cache entry.

packages/gptme-rag/src/gptme_rag/lesson_matcher.py
TimeToBuildBob: Rejected as out of scope for this cache contract: the deliberate identity key is `(resolved path, st_mtime_ns, st_size)`, matching standard stat-based caches. Hashing every file on every fresh hook invocation would re-read the entire corpus and erase the disk cache's core benefit. Filesystems that expose nanosecond mtimes are the supported correctness boundary; coarse-mtime mounts can disable the disk layer with `GPTME_LESSON_SCAN_CACHE=off` if needed.

packages/gptme-rag/tests/test_lesson_matcher.py
```diff
@@ -962,6 +977,191 @@ def test_unreadable_file_does_not_block_same_name_in_later_dir(self, tmp_path):
...
+        dir_a = tmp_path / "a"
+        dir_b = tmp_path / "b"
+        cached_file = dir_a / "foo.md"
+        _write_lesson(cached_file, _basic_lesson(["cached-but-now-unreadable"]))
+        _write_lesson(dir_b / "foo.md", _basic_lesson(["readable-fallback"]))
+        assert scan_lessons([dir_a, dir_b])[0]["keywords"] == ["cached-but-now-unreadable"]
+
+        cached_file.chmod(0o000)
+        try:
+            assert scan_lessons([dir_a, dir_b])[0]["keywords"] == ["readable-fallback"]
```
TimeToBuildBob: ⚠️ **P2** — The test test_permission_change_invalidates_cached_lesson calls `cached_file.chmod(0o000)` and then asserts that a subsequent scan uses the fallback file. This relies on `os.access(path, os.R_OK)` returning False for a file with no read permission. When running as root (common in CI containers and Docker), os.access returns True for any file because root bypasses permission bits. The test then fails because the cached file is still considered readable, so the wrong lesson is returned. The finally block restores permissions, but the assertion fails first. The test is non-hermetic and will break in root environments, which are a realistic deployment target for CI.

packages/gptme-rag/src/gptme_rag/lesson_matcher.py
TimeToBuildBob: Fixed fully in `7da9d89e`: persisted entries now validate every scalar and list field consumed downstream, so malformed same-stat records are ignored and reparsed. Added a missing-scalar regression test.

packages/gptme-rag/tests/test_lesson_matcher.py
TimeToBuildBob: Fixed in `7da9d89e`: the permission-change regression now monkeypatches `os.access` for the target path instead of relying on Unix mode bits, so it is deterministic when the test process runs as root.

TimeToBuildBob:
Pushed `7da9d89e` to address the remaining AI-review findings:

- persisted cache records now validate all scalar and list fields consumed downstream;
- the permission-change test now simulates `os.access` deterministically instead of relying on mode bits under root;
- entries skipped by archive/name dedup are evicted rather than retained in memory and on disk.

Verified the focused matcher suite (`144 passed`) and the scoped pre-commit hooks. CI and the automatic post-push AI review are now running.

packages/gptme-rag/src/gptme_rag/lesson_matcher.py
```diff
@@ -678,6 +681,272 @@ def _score_skill_descriptor(lesson: dict[str, Any], prompt_lower: str) -> tuple[
...
+    try:
+        resolved = str(path.resolve())
+    except OSError:
+        return False, None
+    key = _stat_key(path)
+    if key is None:
+        return False, None
+    mtime_ns, size = key
+    cached = _parse_cache.get(resolved)
+    if cached is not None and cached[0] == mtime_ns and cached[1] == size:
```
TimeToBuildBob, line 923: ⚠️ **P2** — The cache trusts `(st_mtime_ns, st_size)` as a content fingerprint. If a lesson file is modified such that its size is unchanged and its mtime is restored (e.g., by a build tool or `touch -r`), the cache returns the previously parsed lesson from the `_parse_cache` entry even though the file content changed. The caller then receives stale keywords/body/session_categories, so matching uses outdated data. Concretely, `foo.md` with keywords `[old]` is scanned and cached; the file is edited to `[new]` with the same byte count and its mtime reset to the original value; a subsequent `scan_lessons` hits the cache and returns `[old]`. This regression from the previous always-reparse behaviour is silent.

packages/gptme-rag/src/gptme_rag/lesson_matcher.py
```diff
@@ -678,6 +681,272 @@ def _score_skill_descriptor(lesson: dict[str, Any], prompt_lower: str) -> tuple[
 # Lesson scanning
 # ---------------------------------------------------------------------------
 
+#: On-disk cache schema. Bump when the lesson dict shape changes.
+_CACHE_VERSION = 1
+
+# resolved path -> (mtime_ns, size, lesson_or_None). None means parsed-but-skipped.
+_parse_cache: dict[str, tuple[int, int, dict[str, Any] | None]] = {}
```
TimeToBuildBob, line 688: ⚠️ **P2** — The module-level `_parse_cache` dict and `_disk_dirty` flag are shared across all calls to `scan_lessons`. If two threads call `scan_lessons` concurrently, they can race on cache reads/writes. For example, thread A scanning dir_a and thread B scanning dir_b can cause thread A's `_prune_missing_from_cache` to remove entries for dir_b that thread B is about to read, forcing re-parses. More seriously, both threads can write the disk cache file concurrently; while `os.replace` is atomic, the last writer wins, so one thread's updates can overwrite the other's, potentially losing cache entries. In a scenario where a file is modified between thread A's parse and thread B's cache hit, the race could return a stale lesson if thread B reads the entry before thread A updates it. This can cause incorrect lesson content to be served. The fix is to guard all cache operations with a `threading.Lock`.

packages/gptme-rag/src/gptme_rag/lesson_matcher.py
TimeToBuildBob: Accepted trade-off, not a blocker: `(resolved path, st_mtime_ns, st_size)` is the documented cache identity. Reading or hashing every file on each fresh hook invocation would restore the corpus-wide I/O this change removes. Normal editors advance nanosecond mtime; callers on coarse/preserved-mtime filesystems can set `GPTME_LESSON_SCAN_CACHE=off`.

packages/gptme-rag/src/gptme_rag/lesson_matcher.py
TimeToBuildBob: Non-blocking for the shipped execution model: the Claude Code hook starts a fresh Python process per invocation, and `scan_lessons()` is called synchronously once in that process. Cross-process writes use per-writer temporary files plus atomic `os.replace`; a lost cache update causes a later reparse, not incorrect source behavior, because every hit is revalidated against path/stat. A process-local lock would not coordinate those writers, while cross-process locking would add complexity to a best-effort performance cache.

TimeToBuildBob:
## Convergence adjudication

- **Fixed:** the remaining P1 was valid. `3aa90c37` records archived resolved paths before skipping them, so a differently named symlink from a later root cannot resurrect an archived lesson. Added a regression that failed on the prior head and now passes.
- **Remaining (non-blocking):**
  - Restored same-size/same-mtime content: accepted cache-contract trade-off. Hashing every source on every fresh hook call erases the intended I/O win; nanosecond mtime is the supported identity boundary and `GPTME_LESSON_SCAN_CACHE=off` is the escape hatch.
  - Concurrent module cache access: not present in the shipped hook path (one synchronous scan per fresh process). Cross-process writes are atomic and lost updates only cause reparsing; a process-local lock would not coordinate them.
- **CI:** prior head was fully green. Current head CI is running after the P1 fix; local focused verification is `145 passed`, Ruff clean, format clean, and pre-commit passed on commit.
- **Domain risk:** only symlink/archive precedence and unusual preserved/coarse-mtime filesystems warrant maintainer attention; the former now has a regression, and the latter can disable disk caching.
- **Convergence:** detector reports `stable_rounds=0` with status `no_history` (required stable rounds: 2). The repository reviewer has four recorded rounds, but this detector does not ingest that reviewer history for this PR. No further review is being triggered after the cap.

Recommendation: merge-ready once current-head CI is green, subject to maintainer judgment. Do not auto-merge from this adjudication.

packages/gptme-rag/src/gptme_rag/lesson_matcher.py
```diff
@@ -757,73 +1035,19 @@ def scan_lessons(lesson_dirs: list[Path]) -> list[dict[str, Any]]:
             if not is_skill_file and f.name in seen_names:
                 continue
 
-            try:
-                content = f.read_text(encoding="utf-8")
-            except Exception:
+            seen_paths.add(resolved)
```
TimeToBuildBob, line 1038: ❌ **P1** — When a file is skipped due to filename dedup (its resolved path is not added to seen_paths), a later symlink that resolves to the same file but has a different name will not be recognized as already seen. It will pass the `resolved in seen_paths` check, its name is not in seen_names (since the shadowed file's name was not registered), so it will be parsed and added as a separate lesson, producing a duplicate. For example, with dir1/foo.md (active), dir2/foo.md (active), and dir3/bar.md symlinked to dir2/foo.md, scanning [dir1, dir2, dir3] returns both foo.md and bar.md (both resolving to the same file) instead of just the first. This breaks the first-dir-wins dedup invariant and can cause a lesson to match twice. The fix is to add the resolved path to seen_paths before the name-dedup continue, so the symlink is skipped.

ErikBjare:
@TimeToBuildBob fix in a follow-up if you think it's appropriate.

PR #1531, opened 2026-08-27
feat(pr-review): adversarial per-finding verifier (Stage 2)

TimeToBuildBob:
## Problem

65% of posted AI reviewer findings are P2/medium — in a 13-finding hand-classified sample, **every P2 was nit-class**. One was a duplicate posted against two files; one's own description said "which is fine". Reviewer-driven PM dispatches = 44% of all worker-hours (114 h/7d). Precision baseline 33-37%; rebuttal rate 11.3% vs Greptile's 1.0%.

Erik's direct ask (2026-08-27): *"the reviewer should decide what should and shouldn't be fixed, and shouldn't post things that aren't good to fix. adversarial review of the reviewers findings could be one way."*

## Solution: Two-stage pipeline

**Stage 1** (existing): maximize recall — generate as many findings as possible.

**Stage 2** (this PR): maximize precision — for each finding, run a separate model call prompted to **REFUTE** it.

Three verdict dimensions:
1. **real** — Does the defect exist at this head (not hypothetical, not by-design)?
2. **worth_fixing** — Would a staff engineer block or request this change before merge? Bar: "style / defensive coding for impossible inputs / minor naming → no."
3. **severity** — Independent re-grade. Catches mislabeled security bugs (real P0 arriving labeled P2 falls into non-blocking lane). Can promote OR demote.

Only findings where `real=True AND worth_fixing=True` survive. Dropped findings go to a **shadow ledger** (`state/ai-review-suppressed.jsonl`) so suppression is measurable and reversible, never silent.

## Changes

- **`pr_review/verifier.py`** (new): `verify_finding()` and `verify_artifact()` — the Stage 2 verifier
- **`pr_review/__init__.py`**: exports `VerifierVerdict`, `verify_artifact`, `verify_finding`
- **`pr_review/github_adapter.py`**: `run_github_review(verify=True, verifier_model=...)` opt-in flag — existing callers unaffected

## Tests

19 tests covering:
- JSON extraction (clean, fenced, preamble, no-json error)
- Prompt construction (title/diff inclusion, large-diff truncation)
- `verify_finding`: confirmed, refuted, bad-severity fallback
- `verify_artifact`: confirmed, dropped, shadow ledger records, already-decided skipped, error → conservative (leaves needs_validation), severity promotion

All passing. Typecheck clean.

## Activation

```python
artifact, posted, skipped = run_github_review(
    repo, pr_number,
    verify=True,
    verifier_model="anthropic/claude-haiku-4-5",  # cheaper model is fine for judgment
    shadow=False,
)
```

The verifier can use a cheaper model than the generator — it's a judgment call on structured input, not generation. One extra model call per finding group (~1-3 per review at current bench rate).

## Related

- Task: `tasks/ai-reviewer-worth-fixing-verifier.md` (ErikBjare/bob)
- Memory: `memory/ai-reviewer-precision-and-treadmill.md`
- Analysis: `knowledge/analysis/2026-08-17-ai-reviewer-greptile-signal.md`

### DIFF
--- a/packages/gptme-runloops/src/gptme_runloops/pr_review/github_adapter.py
+++ b/packages/gptme-runloops/src/gptme_runloops/pr_review/github_adapter.py
@@ -222,9 +222,14 @@ def _build_summary_comment_body(
     artifact: ReviewArtifact, posted_count: int, skipped_count: int
 ) -> str:
     """Build a top-level summary comment for the PR."""
-    n_total = len(artifact.findings)
+    postable_findings = [
+        finding
+        for finding in artifact.findings
+        if finding.disposition != Disposition.dropped
+    ]
+    n_total = len(postable_findings)
     n_by_severity: dict[str, int] = {}
-    for f in artifact.findings:
+    for f in postable_findings:
         n_by_severity[f.severity.value] = n_by_severity.get(f.severity.value, 0) + 1
 
     severity_line = ", ".join(
--- a/packages/gptme-runloops/src/gptme_runloops/pr_review/github_adapter.py
+++ b/packages/gptme-runloops/src/gptme_runloops/pr_review/github_adapter.py
@@ -219,13 +219,17 @@ def _build_inline_comment_body(finding: ReviewFinding) -> str:
 
 
 def _build_summary_comment_body(
-    artifact: ReviewArtifact, posted_count: int, skipped_count: int
+    artifact: ReviewArtifact,
+    posted_count: int,
+    skipped_count: int,
+    min_confidence: float = _MIN_CONFIDENCE,
 ) -> str:
     """Build a top-level summary comment for the PR."""
     postable_findings = [
         finding
         for finding in artifact.findings
         if finding.disposition != Disposition.dropped
+        and finding.confidence >= min_confidence
     ]
     n_total = len(postable_findings)
     n_by_severity: dict[str, int] = {}
@@ -323,6 +327,7 @@ def post_summary_comment(
     artifact: ReviewArtifact,
     posted_count: int,
     skipped_count: int,
+    min_confidence: float = _MIN_CONFIDENCE,
 ) -> str:
     """Post a top-level summary comment on the PR.
 
@@ -330,7 +335,9 @@ def post_summary_comment(
         The GitHub comment ID (as string).
     """
     owner, name = repo.split("/", 1)
-    body = _build_summary_comment_body(artifact, posted_count, skipped_count)
+    body = _build_summary_comment_body(
+        artifact, posted_count, skipped_count, min_confidence
+    )
     data = _gh_api(
         f"/repos/{owner}/{name}/issues/{pr_number}/comments",
         method="POST",
@@ -398,7 +405,7 @@ def publish_artifact(
 
     # Post summary comment only if we actually published something
     if posted > 0:
-        post_summary_comment(repo, pr_number, artifact, posted, skipped)
+        post_summary_comment(repo, pr_number, artifact, posted, skipped, min_confidence)
 
     return posted, skipped
 
--- a/packages/gptme-runloops/src/gptme_runloops/pr_review/github_adapter.py
+++ b/packages/gptme-runloops/src/gptme_runloops/pr_review/github_adapter.py
@@ -475,6 +475,7 @@ def run_github_review(
             model=verifier_model or model,
             shadow_ledger=shadow_ledger,
             verbose=verbose,
+            min_confidence=min_confidence,
         )
         # Persist the verified artifact back to output_path if given
         if output_path is not None:
--- a/packages/gptme-runloops/src/gptme_runloops/pr_review/github_adapter.py
+++ b/packages/gptme-runloops/src/gptme_runloops/pr_review/github_adapter.py
@@ -403,26 +403,39 @@ def run_github_review(
     pr_number: int,
     *,
     model: str | None = None,
+    verifier_model: str | None = None,
     checkout: Path | None = None,
     shadow: bool = True,
     output_path: Path | None = None,
     min_confidence: float = _MIN_CONFIDENCE,
+    verify: bool = False,
+    shadow_ledger: Path | None = None,
+    verbose: bool = False,
 ) -> tuple[ReviewArtifact, int, int]:
-    """End-to-end GitHub PR review: fetch → run → publish (or shadow).
+    """End-to-end GitHub PR review: fetch → run → [verify] → publish (or shadow).
 
     Args:
         repo: ``OWNER/REPO`` string.
         pr_number: Pull request number.
-        model: gptme model spec. If None, uses gptme default.
+        model: gptme model spec for Stage 1 generation. If None, uses gptme default.
+        verifier_model: gptme model spec for Stage 2 adversarial verification.
+                        Defaults to ``model`` when not set.
         checkout: Optional local repo checkout for AGENTS.md/CLAUDE.md context.
         shadow: If True (default), do not post anything to GitHub.
         output_path: If set, write artifact JSON here.
-        min_confidence: Skip findings below this threshold.
+        min_confidence: Skip findings below this confidence threshold.
+        verify: If True, run the Stage 2 adversarial verifier before publishing.
+                Dropped findings go to ``shadow_ledger``; only confirmed findings post.
+        shadow_ledger: Path to the JSONL shadow ledger for dropped findings.
+                       Defaults to ``state/ai-review-suppressed.jsonl`` relative to
+                       ``checkout`` (or cwd when checkout is not set).
+        verbose: If True, print per-finding verifier outcomes to stdout.
 
     Returns:
         (artifact, posted_count, skipped_count)
     """
     from .reviewer import run_review
+    from .verifier import verify_artifact
 
     meta = fetch_pr_metadata(repo, pr_number)
     base_sha = meta["base_sha"]
@@ -441,6 +454,19 @@ def run_github_review(
         model=model,
         output_path=output_path,
     )
+
+    if verify:
+        artifact = verify_artifact(
+            artifact,
+            diff,
+            effective_checkout,
+            model=verifier_model or model,
+            shadow_ledger=shadow_ledger,
+            verbose=verbose,
+        )
+        # Persist the verified artifact back to output_path if given
+        if output_path is not None:
+            output_path.write_text(artifact.model_dump_json(indent=2), encoding="utf-8")
 
     posted, skipped = publish_artifact(
         artifact,
--- a/packages/gptme-runloops/src/gptme_runloops/pr_review/verifier.py
+++ b/packages/gptme-runloops/src/gptme_runloops/pr_review/verifier.py
@@ -222,9 +222,13 @@ def verify_finding(
     except ValueError:
         severity = finding.severity  # fall back to generator's grade on bad output
 
+    for field in ("real", "worth_fixing"):
+        if type(data.get(field)) is not bool:
+            raise ValueError(f"verifier verdict field {field!r} must be a boolean")
+
     return VerifierVerdict(
-        real=data.get("real") is True,
-        worth_fixing=data.get("worth_fixing") is True,
+        real=data["real"],
+        worth_fixing=data["worth_fixing"],
         severity=severity,
         rationale=str(data.get("rationale", "")),
     )
--- a/packages/gptme-runloops/src/gptme_runloops/pr_review/verifier.py
+++ b/packages/gptme-runloops/src/gptme_runloops/pr_review/verifier.py
@@ -274,8 +274,15 @@ VerifyResult = Literal[
 
 _BLOCKING_SEVERITIES = {Severity.critical, Severity.high}
 
+# Must match ``github_adapter._MIN_CONFIDENCE``. Escalation keys the same
+# population publish/summary use; a below-threshold finding is not posted
+# and must not flip the CLI fail-closed gate either.
+_DEFAULT_MIN_CONFIDENCE = 0.6
 
-def _escalate_merge_safety(artifact: ReviewArtifact) -> None:
+
+def _escalate_merge_safety(
+    artifact: ReviewArtifact, min_confidence: float = _DEFAULT_MIN_CONFIDENCE
+) -> None:
     """Escalate ``merge_safety`` if remaining findings outrank Stage 1.
 
     Stage 1's ``merge_safety`` is a holistic model judgment. Stage 2 may
@@ -283,11 +290,17 @@ def _escalate_merge_safety(artifact: ReviewArtifact) -> None:
     to catch) without touching that field, leaving the summary header and
     the CLI fail-closed gate stale. Escalate only; never demote. A Stage 1
     ``unsafe`` stays ``unsafe`` even if every finding is dropped.
+
+    ``remaining`` is the publishable set: not dropped, and at or above
+    ``min_confidence``. A confirmed critical with confidence 0.3 would
+    otherwise mark the artifact ``unsafe`` while ``publish_artifact``
+    skipped it, so the summary said "No findings" and the CLI still failed.
     """
     remaining = [
         finding
         for finding in artifact.findings
         if finding.disposition != Disposition.dropped
+        and finding.confidence >= min_confidence
     ]
     if any(finding.severity in _BLOCKING_SEVERITIES for finding in remaining):
         artifact.merge_safety = MergeSafety.unsafe
@@ -306,6 +319,7 @@ def verify_artifact(
     model: str | None = None,
     shadow_ledger: Path | None = None,
     verbose: bool = False,
+    min_confidence: float = _DEFAULT_MIN_CONFIDENCE,
 ) -> ReviewArtifact:
     """Run adversarial verification on every finding in the artifact.
 
@@ -325,6 +339,8 @@ def verify_artifact(
                        Defaults to ``state/ai-review-suppressed.jsonl`` relative
                        to ``checkout``.
         verbose:       If True, print per-finding outcome lines to stdout.
+        min_confidence: Findings below this are ignored when escalating
+                       ``merge_safety``, matching ``publish_artifact``.
 
     Returns:
         The same artifact with updated finding dispositions and severities.
@@ -405,5 +421,5 @@ def verify_artifact(
             f"{errors} errors (left as needs_validation)"
         )
 
-    _escalate_merge_safety(artifact)
+    _escalate_merge_safety(artifact, min_confidence=min_confidence)
     return artifact
--- a/packages/gptme-runloops/src/gptme_runloops/pr_review/verifier.py
+++ b/packages/gptme-runloops/src/gptme_runloops/pr_review/verifier.py
@@ -35,7 +35,7 @@ from typing import Literal
 
 from pydantic import BaseModel
 
-from .schema import Disposition, ReviewArtifact, ReviewFinding, Severity
+from .schema import Disposition, MergeSafety, ReviewArtifact, ReviewFinding, Severity
 
 VERIFIER_PROMPT_VERSION = "v1"
 
@@ -272,6 +272,31 @@ VerifyResult = Literal[
     "confirmed", "dropped_not_real", "dropped_not_worth_fixing", "error"
 ]
 
+_BLOCKING_SEVERITIES = {Severity.critical, Severity.high}
+
+
+def _escalate_merge_safety(artifact: ReviewArtifact) -> None:
+    """Escalate ``merge_safety`` if remaining findings outrank Stage 1.
+
+    Stage 1's ``merge_safety`` is a holistic model judgment. Stage 2 may
+    promote a finding's severity (the mislabeled-P0 case this module exists
+    to catch) without touching that field, leaving the summary header and
+    the CLI fail-closed gate stale. Escalate only; never demote. A Stage 1
+    ``unsafe`` stays ``unsafe`` even if every finding is dropped.
+    """
+    remaining = [
+        finding
+        for finding in artifact.findings
+        if finding.disposition != Disposition.dropped
+    ]
+    if any(finding.severity in _BLOCKING_SEVERITIES for finding in remaining):
+        artifact.merge_safety = MergeSafety.unsafe
+        return
+    if any(finding.severity == Severity.medium for finding in remaining) and (
+        artifact.merge_safety in (MergeSafety.safe, MergeSafety.unknown)
+    ):
+        artifact.merge_safety = MergeSafety.needs_review
+
 
 def verify_artifact(
     artifact: ReviewArtifact,
@@ -380,4 +405,5 @@ def verify_artifact(
             f"{errors} errors (left as needs_validation)"
         )
 
+    _escalate_merge_safety(artifact)
     return artifact
--- a/packages/gptme-runloops/src/gptme_runloops/pr_review/verifier.py
+++ b/packages/gptme-runloops/src/gptme_runloops/pr_review/verifier.py
@@ -1,0 +1,348 @@
+"""Adversarial per-finding verifier — Stage 2 of the two-stage review pipeline.
+
+Design
+------
+Stage 1 (reviewer.py):  maximize recall — generate as many real findings as possible.
+Stage 2 (this module):  maximize precision — for each finding, run a separate model
+                        call prompted to REFUTE it.
+
+Three verdict dimensions per finding:
+  1. real         — Does the defect exist at this head (not hypothetical, not by-design,
+                    not already handled elsewhere in the diff)?
+  2. worth_fixing — Would a staff engineer block or request this change before merge?
+                    Style / hygiene / defensive-coding nits that don't change behavior →
+                    not worth posting.
+  3. severity     — Independent re-grade. Can PROMOTE low/medium to high/critical
+                    (the key concern Erik raised: real security / P0-P1 bugs arriving
+                    mislabeled as P2 / medium and falling into the non-blocking lane).
+                    Can also demote.
+
+Only findings where real=True AND worth_fixing=True survive.  Survivors get their
+severity updated from the verifier's independent grade.  Dropped findings are appended
+to a shadow ledger (default: state/ai-review-suppressed.jsonl) so suppression is
+measurable and reversible.
+"""
+
+from __future__ import annotations
+
+import json
+import re
+import subprocess
+from datetime import datetime, timezone
+from pathlib import Path
+from typing import Literal
+
+from pydantic import BaseModel
+
+from .schema import Disposition, ReviewArtifact, ReviewFinding, Severity
+
+VERIFIER_PROMPT_VERSION = "v1"
+
+
+# ── Verifier verdict schema ──────────────────────────────────────────────────
+
+
+class VerifierVerdict(BaseModel):
+    """Parsed response from one adversarial-verifier call."""
+
+    real: bool
+    """True if the defect demonstrably exists at this head in the diff shown."""
+
+    worth_fixing: bool
+    """True if a staff engineer would block or request this change before merge."""
+
+    severity: Severity
+    """Independent severity re-grade (may promote OR demote vs the generator's label)."""
+
+    rationale: str
+    """One-sentence explanation of the verdict (for the shadow ledger and debug logs)."""
+
+
+# ── Prompt template ─────────────────────────────────────────────────────────
+
+
+_VERIFIER_INSTRUCTIONS = """\
+You are an adversarial code reviewer. Your job is to critically evaluate a finding
+produced by a first-pass review and decide whether it should be posted to the PR.
+
+You MUST try hard to refute the finding. Findings that survive your scrutiny deserve
+to be posted; noisy or wrong findings waste reviewer time and erode trust.
+
+Evaluate on three independent dimensions:
+
+1. REAL — Does the defect actually exist in the diff shown?
+   - Reject if: the code already handles the case elsewhere, the "bug" is by-design,
+     the evidence misreads the diff, or the finding is hypothetical ("could happen if…").
+
+2. WORTH_FIXING — Would a staff engineer block or request this change before merge?
+   The bar is high:
+   - Accept: correctness bugs, real security issues, missing tests that would catch a
+     regression, API contract violations.
+   - Reject: style improvements, defensive coding for impossible inputs, minor naming
+     suggestions, "could be refactored", anything where the PR author would reasonably
+     say "won't fix before merge".
+
+3. SEVERITY — Assign your own severity label independently of the generator's label.
+   This catches mislabeled findings in both directions (a real security bug labeled
+   "medium" should be promoted to "high" or "critical"; a "medium" nit should be
+   demoted to "low" or "info").
+
+Return ONLY a JSON object with this exact structure (no preamble, no fences):
+{
+  "real": true | false,
+  "worth_fixing": true | false,
+  "severity": "critical" | "high" | "medium" | "low" | "info",
+  "rationale": "<one-sentence explanation of the verdict>"
+}""".strip()
+
+
+def _build_verifier_prompt(finding: ReviewFinding, diff: str) -> str:
+    """Build the verifier prompt for a single finding."""
+    # Truncate the diff to keep prompt size manageable; the full diff may be large
+    diff_bytes = diff.encode()
+    _MAX_DIFF_BYTES = 40_000
+    if len(diff_bytes) > _MAX_DIFF_BYTES:
+        diff = diff_bytes[:_MAX_DIFF_BYTES].decode(errors="replace")
+        diff += "\n\n[DIFF TRUNCATED — evaluate only what is visible above]"
+
+    finding_json = json.dumps(
+        {
+            "title": finding.title,
+            "category": finding.category,
+            "severity": finding.severity.value,
+            "file_path": finding.file_path,
+            "line_range": finding.line_range,
+            "description": finding.description,
+            "evidence": finding.evidence,
+        },
+        indent=2,
+    )
+
+    return "\n\n".join(
+        [
+            _VERIFIER_INSTRUCTIONS,
+            "---",
+            f"Finding to evaluate:\n```json\n{finding_json}\n```",
+            f"Full diff context:\n```diff\n{diff}\n```",
+        ]
+    )
+
+
+def _invoke_verifier_model(prompt: str, model: str | None, checkout: Path) -> str:
+    """Invoke gptme in one-shot mode for a verifier call."""
+    cmd = ["gptme", "--tools", "none"]
+    if model:
+        cmd.extend(["--model", model])
+    cmd.append(prompt)
+
+    result = subprocess.run(
+        cmd,
+        cwd=checkout,
+        capture_output=True,
+        text=True,
+        timeout=120,  # verifier calls are shorter than full reviews
+    )
+    if result.returncode not in (0, 1):
+        stderr_snip = result.stderr[:500] if result.stderr else "(no stderr)"
+        raise RuntimeError(
+            f"gptme (verifier) exited with code {result.returncode}:\n{stderr_snip}"
+        )
+    return result.stdout
+
+
+def _extract_verifier_json(text: str) -> dict:
+    """Extract the first JSON object from verifier output, permissively."""
+    text = text.strip()
+    try:
+        result = json.loads(text)
+        assert isinstance(result, dict)
+        return result
+    except (json.JSONDecodeError, AssertionError):
+        pass
+    fenced = re.search(r"```(?:json)?\s*(\{.*?\})\s*```", text, re.DOTALL)
+    if fenced:
+        try:
+            result = json.loads(fenced.group(1))
+            assert isinstance(result, dict)
+            return result
+        except (json.JSONDecodeError, AssertionError):
+            pass
+    match = re.search(r"\{.*\}", text, re.DOTALL)
+    if match:
+        try:
+            result = json.loads(match.group(0))
+            assert isinstance(result, dict)
+            return result
+        except (json.JSONDecodeError, AssertionError):
+            pass
+    raise ValueError(
+        f"No valid JSON found in verifier output (first 300 chars):\n{text[:300]}"
+    )
+
+
+def verify_finding(
+    finding: ReviewFinding,
+    diff: str,
+    checkout: Path,
+    *,
+    model: str | None = None,
+) -> VerifierVerdict:
+    """Run one adversarial verifier call for a single finding.
+
+    Args:
+        finding: The finding from Stage 1 to evaluate.
+        diff:    Full unified diff (same diff used by the reviewer).
+        checkout: Path to the repo root (for gptme cwd).
+        model:   gptme model spec. If None, uses the gptme default.
+
+    Returns:
+        VerifierVerdict with real/worth_fixing/severity/rationale fields.
+    """
+    prompt = _build_verifier_prompt(finding, diff)
+    raw_output = _invoke_verifier_model(prompt, model, checkout)
+    data = _extract_verifier_json(raw_output)
+
+    try:
+        severity = Severity(data.get("severity", finding.severity.value))
+    except ValueError:
+        severity = finding.severity  # fall back to generator's grade on bad output
+
+    return VerifierVerdict(
+        real=bool(data.get("real", True)),
+        worth_fixing=bool(data.get("worth_fixing", True)),
+        severity=severity,
+        rationale=str(data.get("rationale", "")),
+    )
+
+
+# ── Shadow ledger ────────────────────────────────────────────────────────────
+
+
+def _append_suppressed(
+    ledger_path: Path,
+    finding: ReviewFinding,
+    verdict: VerifierVerdict,
+    repo_id: str,
+    head_sha: str,
+) -> None:
+    """Append a dropped finding to the shadow ledger (JSONL, one record per line)."""
+    record = {
+        "ts": datetime.now(timezone.utc).isoformat(),
+        "repo": repo_id,
+        "head_sha": head_sha[:16],
+        "finding_id": finding.id,
+        "title": finding.title,
+        "file_path": finding.file_path,
+        "line_range": finding.line_range,
+        "generator_severity": finding.severity.value,
+        "verifier_severity": verdict.severity.value,
+        "real": verdict.real,
+        "worth_fixing": verdict.worth_fixing,
+        "rationale": verdict.rationale,
+    }
+    ledger_path.parent.mkdir(parents=True, exist_ok=True)
+    with ledger_path.open("a", encoding="utf-8") as fh:
+        fh.write(json.dumps(record) + "\n")
+
+
+# ── Artifact-level verification ──────────────────────────────────────────────
+
+
+VerifyResult = Literal[
+    "confirmed", "dropped_not_real", "dropped_not_worth_fixing", "error"
+]
+
+
+def verify_artifact(
+    artifact: ReviewArtifact,
+    diff: str,
+    checkout: Path,
+    *,
+    model: str | None = None,
+    shadow_ledger: Path | None = None,
+    verbose: bool = False,
+) -> ReviewArtifact:
+    """Run adversarial verification on every finding in the artifact.
+
+    Findings are updated in-place (dispositions and severity re-graded);
+    the artifact is returned with the updated findings list.
+
+    Dropped findings are appended to the shadow ledger when ``shadow_ledger``
+    is provided.  Findings that error during verification are left as
+    ``needs_validation`` (conservative: don't drop on tool failure).
+
+    Args:
+        artifact:      The ReviewArtifact from Stage 1.
+        diff:          The same unified diff used in Stage 1.
+        checkout:      Path to the repo root (for gptme cwd).
+        model:         gptme model spec for the verifier calls.
+        shadow_ledger: Path to the JSONL file where dropped findings are logged.
+                       Defaults to ``state/ai-review-suppressed.jsonl`` relative
+                       to ``checkout``.
+        verbose:       If True, print per-finding outcome lines to stdout.
+
+    Returns:
+        The same artifact with updated finding dispositions and severities.
+    """
+    if shadow_ledger is None:
+        shadow_ledger = checkout / "state" / "ai-review-suppressed.jsonl"
+
+    # Resolve a stable identity anchor for the ledger records
+    from .schema import ReviewTarget
+
+    target = artifact.target
+    if isinstance(target, ReviewTarget):
+        repo_id = target.repo
+        head_sha = target.head_sha
+    else:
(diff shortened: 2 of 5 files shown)


TimeToBuildBob:
CI is fully green at `c250bdd1d714`. No human or automated review has landed yet, so there is no code change to make from this notification; waiting for review.

packages/gptme-runloops/src/gptme_runloops/pr_review/verifier.py
```diff
@@ -0,0 +1,348 @@
...
+    raw_output = _invoke_verifier_model(prompt, model, checkout)
+    data = _extract_verifier_json(raw_output)
+
+    try:
+        severity = Severity(data.get("severity", finding.severity.value))
+    except ValueError:
+        severity = finding.severity  # fall back to generator's grade on bad output
+
+    return VerifierVerdict(
+        real=bool(data.get("real", True)),
```
TimeToBuildBob: ❌ **P1** — verify_finding defaults the `real` and `worth_fixing` booleans to True when the model omits them, and coerces any non-empty string (e.g. "false") to True via `bool()`. If the verifier model returns JSON without these fields, or with string values, findings are treated as real and worth-fixing even though no confirmation was given. This fails open: any malformed or incomplete verifier output causes the finding to be posted, exactly the opposite of the adversarial filtering the feature promises. Combined with the error path leaving findings as needs_validation (also posted), the verifier provides no failure-closed protection. Concretely, a model returning `{"severity": "low", "rationale": "nit"}` (no `real`/`worth_fixing`) will confirm and post the finding, defeating the purpose.

packages/gptme-runloops/src/gptme_runloops/pr_review/verifier.py
```diff
@@ -0,0 +1,348 @@
...
+            finding.disposition = Disposition.confirmed
+            finding.severity = (
+                verdict.severity
+            )  # apply re-grade (may promote or demote)
+            confirmed += 1
+            if verbose:
+                print(f"  [CONFIRMED] {finding.title[:60]} ({verdict.severity.value})")
+        else:
+            finding.disposition = Disposition.dropped
+            finding.severity = verdict.severity  # record re-grade even for dropped
```
TimeToBuildBob, line 397: ❌ **P1** — In verify_artifact's else branch, `finding.severity` is overwritten with `verdict.severity` (line 335) before calling `_append_suppressed`. Inside `_append_suppressed`, the ledger record's `generator_severity` field is populated from `finding.severity.value`, so it records the verifier's re-grade rather than the generator's original severity. The explicit purpose of the ledger is to compare generator vs verifier grades; this overwrite makes the comparison impossible and corrupts the suppression audit trail. The test for ledger fields does not assert generator_severity, so it does not catch this. The fix is to capture the original severity before reassigning, or pass it explicitly to the ledger.

TimeToBuildBob:
## 🤖 AI code review

Adds a Stage 2 adversarial verifier (verifier.py) that runs each Stage 1 finding through a separate gptme call prompted to refute it, marking findings confirmed or dropped, re-grading severity, and appending dropped findings to a JSONL shadow ledger. github_adapter.py threads a `verify` flag, `verifier_model`, `shadow_ledger`, and `min_confidence` through run_github_review, and the summary comment body now counts only non-dropped, at-threshold findings. __init__.py exports the new VerifierVerdict/verify_artifact/verify_finding API; tests cover both the new verifier and the summary-comment changes.

**❌ P1 high** — `packages/gptme-runloops/src/gptme_runloops/pr_review/verifier.py:110`

Truncation fallback misses renamed-file diffs, dropping real findings. The file marker at verifier.py:110 is built as `diff --git a/{finding.file_path} b/{finding.file_path}`. When a PR renames a file, the unified header is `diff --git a/old b/new`, so the marker never matches if finding.file_path is the new path. file_start becomes -1, and the code takes the first 40k bytes of the whole diff (line 121). For a large PR this almost certainly excludes the renamed file's hunks, so the model sees no evidence for the finding and tends to return real:false, dropping a legitimate issue via the shadow ledger. Renames are common enough that this silently suppresses real defects.

```suggestion
file_marker = f"diff --git a/"
        # instead of an exact match, locate the file's section by searching for the new path
```

<sub>**How this was verified:** The diff header for a rename is `diff --git a/old.py b/new.py`, which cannot equal the constructed marker, and the else branch at line 121 grabs only the first 40KB. test_large_diff_keeps_finding_file_context only exercises an unchanged-path file whose section is under the limit.</sub>

**⚠️ P2 medium · ℹ️ non-blocking** — `packages/gptme-runloops/src/gptme_runloops/pr_review/github_adapter.py:234`

_build_summary_comment_body now computes n_total from postable_findings (not dropped and confidence &gt;= min_confidence, lines 228-234), but still prints the posted_count and skipped_count lines verbatim (lines 262-267). These counts are independent of n_total: after the verifier drops one finding from a two-finding artifact, the body reads '**1 finding(s)**: 1 high' while also printing '1 finding(s) posted as inline comments' and '1 finding(s) skipped', which totals 2 against a header of 1. Before this PR n_total was len(artifact.findings), so the three numbers always agreed; the new counted-only-postable logic makes the header disagree with the sum of the two count lines whenever any finding was dropped or fell below the confidence threshold.

```suggestion
f"*{skipped_count} finding(s) skipped — already posted, below confidence threshold, or dropped by verifier.*"
```

<sub>**How this was verified:** Comparing the new filter (line 228-233) against the unmodified skipped_count/posted_count lines (262-267): n_total excludes dropped and below-threshold findings, while the two count lines do not. Test test_dropped_findings_are_excluded_from_count_and_severity passes posted_count=1, skipped_count=1 with n_total=1, reproducing the contradiction.</sub>

</details>

<details><summary><b>1 existing-module overlap (informational, not scored)</b></summary>

_This is a cheap file-index scan, not a merge blocker. A human decides whether to redirect, extend the existing module, or intentionally parallel-implement._

**⚠️ P2 medium · ℹ️ non-blocking** — `packages/gptme-runloops/src/gptme_runloops/pr_review/verifier.py:1`

This appears to overlap with `gptme/gptme-contrib:packages/gptme-rag` which already provides RAG: document indexing, semantic/lexical search, persistence. Matched signals: jsonl, search. Informational only — not a blocker. A human decides whether to redirect, extend the existing module, or intentionally parallel-implement.

<sub>**How this was verified:** cheap file-index scan (L0.5 existing-module overlap)</sub>

</details>

- `packages/gptme-runloops/src/gptme_runloops/pr_review/__init__.py` — Exports VERIFIER_PROMPT_VERSION, VerifierVerdict, verify_artifact, and verify_finding from the new verifier module.
- `packages/gptme-runloops/src/gptme_runloops/pr_review/github_adapter.py` — Adds verify, verifier_model, shadow_ledger, and verbose options to run_github_review; filters the summary count by disposition and confidence; persists the verified artifact to output_path.
- `packages/gptme-runloops/src/gptme_runloops/pr_review/verifier.py` — New module: builds refutation prompts, invokes gptme per finding, parses verdicts, updates dispositions and severities, writes dropped findings to a JSONL shadow ledger, and escalates merge_safety.
- `packages/gptme-runloops/tests/test_pr_review.py` — Adds TestSummaryComment covering dropped-finding and low-confidence exclusion from the summary body.
- `packages/gptme-runloops/tests/test_pr_review_verifier.py` — New test module covering JSON extraction, prompt building, verdict parsing, shadow-ledger writes, and merge-safety escalation, all with mocked model calls.

</details>

| `c250bdd1d714` | 2/5 | 2 | llm | 2026-08-27 21:34 UTC |
| `73bf84c245c2` | 2/5 | 6 | llm | 2026-08-28 03:59 UTC |
| `298de2783e4c` | 3/5 | 3 | llm | 2026-08-28 06:10 UTC |
| `97f5eb094cc4` | 2/5 | 6 | llm | 2026-08-28 10:51 UTC |
| `dafbd5ae188b` | 4/5 | 6 | llm | 2026-08-29 00:35 UTC |
| `dafbd5ae188b` | 3/5 | 1 | llm | 2026-08-29 00:59 UTC |

</details>

<sub>Reviewed `b0481fee4262` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 261s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Fixed both P1 findings in `73bf84c2`:

- malformed or incomplete verifier booleans now fail closed unless the JSON values are literal `true`;
- suppression records preserve the generator's original severity separately from the verifier re-grade.

Added regressions for missing/string boolean verdicts and generator-vs-verifier ledger grades. Targeted verifier tests pass (`21 passed`), and scoped pre-commit/typecheck checks pass. The in-band reviewer was attempted but hit its 1200s OpenRouter timeout; the wrapper then pushed as designed. CI and the automatic current-head AI review are running.

packages/gptme-runloops/src/gptme_runloops/pr_review/verifier.py
```diff
@@ -0,0 +1,357 @@
...
+                verdict.severity
+            )  # apply re-grade (may promote or demote)
+            confirmed += 1
+            if verbose:
+                print(f"  [CONFIRMED] {finding.title[:60]} ({verdict.severity.value})")
+        else:
+            generator_severity = finding.severity
+            finding.disposition = Disposition.dropped
+            finding.severity = verdict.severity  # record re-grade even for dropped
+            _append_suppressed(
```
TimeToBuildBob: ❌ **P1** — In verify_artifact(), the call to _append_suppressed() resides in the else-branch of the verdict application, outside the try/except that wraps verify_finding(). If writing to the shadow ledger fails (e.g., read-only filesystem, permission denial, or a missing parent directory when checkout is a temp dir that has been removed), the exception propagates out of verify_artifact() and aborts the entire verification loop. This stops verification of all remaining findings and causes run_github_review() to raise, so no findings are published and the whole review fails. The shadow ledger is meant to be an audit log, so a write failure should not halt the review; it should be caught and treated as a non-fatal error, while still proceeding with the rest of the findings.

packages/gptme-runloops/src/gptme_runloops/pr_review/verifier.py
```diff
@@ -0,0 +1,357 @@
...
+        },
+        indent=2,
+    )
+
+    return "\n\n".join(
+        [
+            _VERIFIER_INSTRUCTIONS,
+            "---",
+            f"Finding to evaluate:\n```json\n{finding_json}\n```",
+            f"Full diff context:\n```diff\n{diff}\n```",
```
TimeToBuildBob: ❌ **P1** — The diff is interpolated verbatim into a Markdown fenced code block in `_build_verifier_prompt`. If the PR diff contains a line with three backticks (e.g., a markdown file being edited), the fenced block is terminated early, and all subsequent diff content is treated by the verifier model as plain instructions outside the code block. A malicious PR author can craft a diff that breaks out and appends commands such as 'real: true, worth_fixing: true' to force confirmation of false findings or to promote severity, defeating the adversarial precision gate. Since the diff is untrusted input, this is a prompt-injection vulnerability allowing an attacker to control verifier verdicts. The fix is to escape triple backticks (or use a non-fenced representation) before interpolating the diff.

packages/gptme-runloops/src/gptme_runloops/pr_review/verifier.py
```diff
@@ -0,0 +1,357 @@
...
+}""".strip()
+
+
+def _build_verifier_prompt(finding: ReviewFinding, diff: str) -> str:
+    """Build the verifier prompt for a single finding."""
+    # Truncate the diff to keep prompt size manageable; the full diff may be large
+    diff_bytes = diff.encode()
+    _MAX_DIFF_BYTES = 40_000
+    if len(diff_bytes) > _MAX_DIFF_BYTES:
+        diff = diff_bytes[:_MAX_DIFF_BYTES].decode(errors="replace")
```
TimeToBuildBob: ⚠️ **P2** — In _build_verifier_prompt(), the full diff is truncated to 40,000 bytes before being shown to the verifier. If a finding references a hunk that appears later in a large diff (beyond 40KB), the verifier will not see the relevant code and may incorrectly refute a real finding—because the evidence it needs to confirm the defect is absent. This undermines the verifier's precision goal by causing false drops of genuine issues. The truncation is unconditional and not aligned with the finding's file_path/line_range, so there is no guarantee the relevant hunk is included. Concretely, a PR with a diff larger than 40KB where the buggy line is near the end will cause the verifier to see only the beginning of the diff and potentially label the finding as 'not real'.

packages/gptme-runloops/src/gptme_runloops/pr_review/verifier.py
```diff
@@ -0,0 +1,357 @@
...
+        pass
+    fenced = re.search(r"```(?:json)?\s*(\{.*?\})\s*```", text, re.DOTALL)
+    if fenced:
+        try:
+            result = json.loads(fenced.group(1))
+            assert isinstance(result, dict)
+            return result
+        except (json.JSONDecodeError, AssertionError):
+            pass
+    match = re.search(r"\{.*\}", text, re.DOTALL)
```
TimeToBuildBob: ⚠️ **P2** — In _extract_verifier_json, the fallback regex `re.search(r"\{.*\}", text, re.DOTALL)` greedily matches from the first '{' to the last '}' in the entire output. If the verifier model emits any prose after the JSON that contains a closing brace (e.g., explaining the verdict), the captured group includes that trailing prose and is invalid JSON, causing json.loads to raise and the function to throw ValueError. In verify_artifact, that exception is caught, and the finding is left as needs_validation, meaning it is not dropped and will be posted to the PR. This silently bypasses the verifier's purpose of suppressing refuted findings, reducing precision. The fix is to make the match non-greedy (e.g., `r"\{.*?\}"`) or to parse balanced braces.

packages/gptme-runloops/src/gptme_runloops/pr_review/verifier.py
```diff
@@ -0,0 +1,357 @@
...
+
+    Returns:
+        VerifierVerdict with real/worth_fixing/severity/rationale fields.
+    """
+    prompt = _build_verifier_prompt(finding, diff)
+    raw_output = _invoke_verifier_model(prompt, model, checkout)
+    data = _extract_verifier_json(raw_output)
+
+    try:
+        severity = Severity(data.get("severity", finding.severity.value))
```
TimeToBuildBob: ⚠️ **P2** — In `verify_finding`, the verifier's `severity` field is parsed with `Severity(data.get("severity", finding.severity.value))`. The `Severity` enum is lowercase (e.g., `high`, `medium`). If the model returns an uppercase variant like `"High"` or `"Critical"`, the `Severity` constructor raises `ValueError`, and the code falls back to the generator's severity. This silently discards the verifier's independent re-grade in both directions: a promotion from medium to High is ignored, and a demotion from critical to Low is also ignored. Since LLM outputs are not case-stable, a real severity change from the verifier can be lost, re-introducing the mislabeling the verifier was designed to fix. The parser should normalize the severity string to lowercase before constructing the enum.

packages/gptme-runloops/src/gptme_runloops/pr_review/verifier.py
TimeToBuildBob: Fixed in `298de278`: shadow-ledger `OSError`s now emit a runtime warning without aborting verification, and a regression proves the next finding is still verified.

packages/gptme-runloops/src/gptme_runloops/pr_review/verifier.py
TimeToBuildBob: Fixed in `298de278`: findings and diffs are now JSON-encoded untrusted data with an explicit instruction boundary, so Markdown fences in a PR cannot escape into verifier instructions. Added a fence-breakout regression.

packages/gptme-runloops/src/gptme_runloops/pr_review/verifier.py
TimeToBuildBob: Fixed in `298de278`: large diffs now prioritize the finding's own file section before applying the 40KB bound. Added a regression with the relevant file after 50KB of unrelated content.

packages/gptme-runloops/src/gptme_runloops/pr_review/verifier.py
TimeToBuildBob: Fixed in `298de278`: fallback extraction now uses `JSONDecoder.raw_decode` from candidate opening braces, so trailing prose or braces no longer poison an otherwise valid verdict. Added a regression.

packages/gptme-runloops/src/gptme_runloops/pr_review/verifier.py
TimeToBuildBob: Fixed in `298de278`: verifier severities are stripped and lowercased before enum parsing. Added a mixed-case severity regression.

TimeToBuildBob:
Addressed all six findings from the `73bf84c2` AI review in `298de278`:

- made shadow-ledger write failures non-fatal;
- JSON-encoded untrusted finding/diff inputs and added an explicit prompt boundary;
- kept the finding's own file context when truncating large diffs;
- replaced greedy fallback extraction with `JSONDecoder.raw_decode`;
- normalized mixed-case severities;
- excluded verifier-dropped findings from summary counts and severity totals.

Added regressions for each path. Targeted review tests pass (`102 passed`), scoped mypy/ruff/prek checks pass, and all five inline threads were replied to and resolved. The required in-band wrapper ran before pushing; its reviewer timed out after 1200s, then the wrapper pushed per policy. Current-head CI and automatic AI review are pending; Greptile was not triggered.

packages/gptme-runloops/src/gptme_runloops/pr_review/verifier.py
```diff
@@ -0,0 +1,379 @@
...
+
+    try:
+        severity = Severity(
+            str(data.get("severity", finding.severity.value)).strip().lower()
+        )
+    except ValueError:
+        severity = finding.severity  # fall back to generator's grade on bad output
+
+    return VerifierVerdict(
+        real=data.get("real") is True,
```
TimeToBuildBob: ❌ **P1** — In `verify_finding`, the `real` and `worth_fixing` booleans are derived with `data.get("real") is True`, so if the model returns valid JSON that omits either key (or returns a non-boolean like a string), the missing key yields `None` which is not `True`, making both default to `False`. In `verify_artifact`, a verdict with `real=False` causes the finding to be dropped and recorded in the shadow ledger, silently removing a finding that may actually be valid. This contradicts the documented conservative behavior that tool failures should leave findings as `needs_validation`; a malformed-but-parseable verdict is treated as a refutation rather than an error. The fix is to validate that both keys exist and are booleans, raising an error or returning a sentinel when they are missing, so the finding is not dropped on ambiguous output.

packages/gptme-runloops/src/gptme_runloops/pr_review/github_adapter.py
```diff
@@ -222,9 +222,14 @@ def _build_summary_comment_body(
     artifact: ReviewArtifact, posted_count: int, skipped_count: int
 ) -> str:
     """Build a top-level summary comment for the PR."""
-    n_total = len(artifact.findings)
+    postable_findings = [
+        finding
+        for finding in artifact.findings
+        if finding.disposition != Disposition.dropped
+    ]
+    n_total = len(postable_findings)
```
TimeToBuildBob, line 234: ⚠️ **P2** — The summary comment count (`n_total`) is computed from `postable_findings`, which only excludes findings with disposition `dropped`. Findings whose confidence is below `min_confidence` (e.g., 0.5 < 0.6) are not dropped, so they are included in the count, but `publish_artifact` filters them out before posting and does not count them as skipped. As a result, when an artifact contains a confirmed finding and a low-confidence non-dropped finding, the summary reports **2 finding(s)** while only one comment is actually posted and `skipped_count` is 0. This misleads the PR author about the number of actionable findings and undercuts the precision goal of this PR. The function would need access to `min_confidence` to exclude such findings, or the threshold should be applied before building the summary.

packages/gptme-runloops/src/gptme_runloops/pr_review/verifier.py
```diff
@@ -0,0 +1,379 @@
...
+        if file_start >= 0:
+            next_file = diff.find("\ndiff --git ", file_start + len(file_marker))
+            relevant_diff = diff[file_start : next_file if next_file >= 0 else None]
+            relevant_bytes = relevant_diff.encode()
+            if len(relevant_bytes) <= _MAX_DIFF_BYTES:
+                diff = relevant_diff
+            else:
+                diff = relevant_bytes[:_MAX_DIFF_BYTES].decode(errors="replace")
+        else:
+            diff = diff_bytes[:_MAX_DIFF_BYTES].decode(errors="replace")
```
TimeToBuildBob, line 121: ⚠️ **P2** — In `_build_verifier_prompt`, when the diff is large (>40,000 bytes) and the file marker for the finding's file is not found (e.g., because the path in the diff uses a different prefix or the file is renamed), the fallback truncates the entire diff to its first 40,000 bytes (`diff = diff_bytes[:_MAX_DIFF_BYTES].decode(errors="replace")`). If the finding's file appears later in the diff, the verifier never sees that file's context and may drop the finding as "not real" or "not worth fixing" purely from lack of evidence. This suppresses genuine findings on files located near the end of large PRs, which is precisely the kind of noise the adversarial verifier is meant to reduce, but at the cost of real signal. The code already attempts to preserve file context by searching for the file marker, but only succeeds when the marker is present; otherwise the fallback is not context-aware.

TimeToBuildBob:
## Convergence adjudication

**Fixed in `97f5eb09`:**
- **P1 `2eb2938dd4b1`** — malformed-but-parseable verdicts now raise and remain `needs_validation`; missing or non-boolean `real` / `worth_fixing` values can no longer silently suppress a finding.
- **P2 `8262e0ae5b94`** — summary counts and severity totals now apply the same `min_confidence` threshold as publication.

**Remaining (non-blocking):**
- **P2 `a4769f41fa43` — dismissed as a hypothetical malformed-input edge case.** Stage 1 explicitly requires an exact path from the diff, so a missing file marker means the generator violated its contract. More importantly, an unverifiable finding cannot be silently dropped now: malformed verifier output and verifier errors leave it `needs_validation`, while the verifier must explicitly return a complete refutation to suppress it. Expanding fuzzy rename/path recovery would add complexity without evidence of a production failure.

**CI:** the previous head (`298de278`) was fully green. Current-head CI for `97f5eb09` is running; local targeted verification passed: 104 tests, Ruff check/format, mypy, scoped prek, and `git diff --check`.

**Domain risk:** Stage 2 remains opt-in. The only fragile surface worth a maintainer manual check is model-output behavior: a malformed verdict should visibly remain postable instead of entering the suppression ledger.

**Convergence:** detector reports `round_convergence.status=no_history`, `stable_rounds=0` of 2, with zero Greptile findings. No further Greptile review was triggered. Once current-head CI is green, this is merge-ready for maintainer judgment; I am not auto-merging it.

packages/gptme-runloops/src/gptme_runloops/pr_review/verifier.py
```diff
@@ -0,0 +1,383 @@
...
+            # Leave as needs_validation — publish_artifact will treat it as postable
+            continue
+
+        # Apply verdict
+        finding.validated_by = f"adversarial-verifier/{VERIFIER_PROMPT_VERSION}"
+        finding.validation_note = verdict.rationale
+
+        if verdict.real and verdict.worth_fixing:
+            finding.disposition = Disposition.confirmed
+            finding.severity = (
```
TimeToBuildBob, line 388: ❌ **P1** — When the verifier confirms a finding, it re-grades the finding's severity (line: `finding.severity = (`), but it never recomputes `artifact.merge_safety`. The summary comment header renders `artifact.merge_safety` via `artifact.merge_safety.value` at github_adapter.py:252, so a PR whose generator labeled a real bug as 'medium' and whose verifier promotes it to 'critical' can still be summarized as 'Safe' or 'Needs Review' rather than 'Unsafe'. This defeats the verifier's goal of catching mislabeled security bugs — the very scenario the PR description calls out (real P0/P1 arriving labeled P2/medium and falling into the non-blocking lane). While the severity line will show the critical finding, the header verdict is not updated, so a merge gate that keys off `merge_safety` would permit a merge that should be blocked. The fix is to recalculate `merge_safety` after processing findings, based on the maximum severity among confirmed findings.

packages/gptme-runloops/src/gptme_runloops/pr_review/verifier.py
```diff
@@ -0,0 +1,383 @@
...
+
+
+def _build_verifier_prompt(finding: ReviewFinding, diff: str) -> str:
+    """Build the verifier prompt for a single finding."""
+    # Keep the finding's file context before global context when a large diff must
+    # be truncated; otherwise findings near the end of a PR become unverifiable.
+    diff_bytes = diff.encode()
+    _MAX_DIFF_BYTES = 40_000
+    if len(diff_bytes) > _MAX_DIFF_BYTES:
+        file_marker = f"diff --git a/{finding.file_path} b/{finding.file_path}"
```
TimeToBuildBob, line 110: ⚠️ **P2** — `_build_verifier_prompt` locates the relevant diff section by searching for `diff --git a/{finding.file_path} b/{finding.file_path}`. For a rename diff the header is `diff --git a/old.py b/new.py`, so a finding on `new.py` does not match; the code falls back to truncating the entire diff to the first 40KB. On a large PR where the rename appears later, the verifier receives no context for the finding and may wrongly refute a real bug. The same happens for any path that does not appear symmetrically (e.g., file deleted on one side).

packages/gptme-runloops/src/gptme_runloops/pr_review/verifier.py
```diff
@@ -0,0 +1,383 @@
...
+def _extract_verifier_json(text: str) -> dict:
+    """Extract the first JSON object from verifier output, permissively."""
+    text = text.strip()
+    try:
+        result = json.loads(text)
+        assert isinstance(result, dict)
+        return result
+    except (json.JSONDecodeError, AssertionError):
+        pass
+    fenced = re.search(r"```(?:json)?\s*(\{.*?\})\s*```", text, re.DOTALL)
```
TimeToBuildBob, line 175: ⚠️ **P2** — In `_extract_verifier_json`, the fenced-code extraction uses a non-greedy `.*?` in the regex `r"```(?:json)?\s*(\{.*?\})\s*```"`, which matches the shortest string between the first `{` and the first `}`. If the model outputs JSON where a string value (e.g., `rationale`) contains a brace, such as `"A {bug} here"`, the regex captures only up to the first `}` inside that string, producing an incomplete JSON object. `json.loads` then fails, and the fallback `raw_decode` loop also fails because the first `{` leads to an unterminated object. As a result, `_extract_verifier_json` raises `ValueError`, which `verify_artifact` catches and leaves the finding as `needs_validation` — meaning the finding is posted without verifier approval. This undermines the verifier's purpose for any model output that happens to contain braces in a string field. The test for fenced JSON uses a rationale with no braces, so this case is not covered.

packages/gptme-runloops/src/gptme_runloops/pr_review/verifier.py
```diff
@@ -0,0 +1,383 @@
...
+                    head_sha,
+                    generator_severity,
+                )
+            except OSError as exc:
+                warnings.warn(
+                    f"failed to record suppressed finding {finding.id}: {exc}",
+                    RuntimeWarning,
+                    stacklevel=2,
+                )
+            dropped += 1
```
TimeToBuildBob, line 413: ⚠️ **P2** — In `verify_artifact`, when a finding is dropped, the code appends it to the shadow ledger inside a `try/except OSError` block that only issues a `RuntimeWarning` on failure, then unconditionally executes `dropped += 1` and leaves the finding's disposition as `Dropped`. If the ledger write fails (e.g., read-only filesystem or missing permissions), the finding is silently suppressed with no record in the shadow ledger, violating the design principle stated in the module docstring that suppression must be 'measurable and reversible'. The warning is easy to miss in logs, so a dropped finding is lost without an audit trail, making it impossible to review what was suppressed. The fix is to only mark the finding as dropped if the ledger write succeeds, or to raise an error and keep the finding as `needs_validation`.

packages/gptme-runloops/src/gptme_runloops/pr_review/github_adapter.py
TimeToBuildBob: Already fixed in `97f5eb09`. `_build_summary_comment_body` now applies the same `min_confidence` filter as publication (`github_adapter.py`: `disposition != dropped` AND `confidence >= min_confidence`), so `n_total` no longer counts below-threshold findings. The thread was left open after that commit; resolving it now.

packages/gptme-runloops/src/gptme_runloops/pr_review/verifier.py
TimeToBuildBob: Rejected — same family as P2 `a4769f41fa43`, already adjudicated on this PR. Stage 1 requires an exact path from the diff; a rename-only header (`a/old.py b/new.py`) is a generator-contract miss, not a Stage 2 defect. An unverifiable finding cannot be silently dropped: verifier errors and malformed output leave it `needs_validation`. Fuzzy path recovery is YAGNI without a production miss.

packages/gptme-runloops/src/gptme_runloops/pr_review/verifier.py
TimeToBuildBob: Rejected as a false positive. `_extract_verifier_json` tries `json.loads` on a fenced capture, then falls through to `JSONDecoder.raw_decode` from every `{` in the raw text. `raw_decode` parses braces inside string values correctly; the non-greedy fence regex is why the fallback exists. Confirmed against `{"rationale": "A {bug} here"}` — the fallback returns the full object. A genuine parse failure would leave the finding `needs_validation`, not drop it.

packages/gptme-runloops/src/gptme_runloops/pr_review/verifier.py
TimeToBuildBob: Accepted trade-off, not a defect. A previous P1 on this PR (`26d6990d774d`) required ledger `OSError`s to be non-fatal so one write failure cannot abort the rest of verification. Reverting a refuted finding to `needs_validation` just because the audit log could not be written would re-post noise the verifier already rejected. The `RuntimeWarning` is the ops signal; the drop stands. Silent suppression of a *real* finding is the failure mode the conservative error path (tool/parse failures → `needs_validation`) already covers.

packages/gptme-runloops/src/gptme_runloops/pr_review/verifier.py
TimeToBuildBob: Fixed in `dafbd5ae`. `verify_artifact` now calls `_escalate_merge_safety` after applying verdicts: remaining (non-dropped) high/critical findings raise `merge_safety` to `unsafe`, and a remaining medium finding escalates Stage 1 `safe`/`unknown` to `needs_review`. Stage 1 is never demoted — an `unsafe` verdict stays `unsafe` even if every finding is dropped. Regression covers promotion-from-safe, medium-from-safe, and no-demote-on-drop.

TimeToBuildBob:
Addressed the current-head AI review on `97f5eb09`.

**Fixed in `dafbd5ae`:**
- **P1 `cda1c691de0c`** — `verify_artifact` now escalates `merge_safety` from remaining (non-dropped) findings after the verifier re-grade, so a promoted high/critical finding can no longer leave the summary header and CLI fail-closed gate at Stage 1 `safe`/`needs_review`. Stage 1 is never demoted.

**Disposed without code change:**
- **P2 `8262e0ae5b94`** — already fixed in `97f5eb09` (`min_confidence` applied to summary counts); thread was left open.
- **P2 `44915479a8ff`** — rejected; same family as the previously adjudicated rename/path-marker P2. Fuzzy path recovery is YAGNI.
- **P2 `78915b5c09b6`** — false positive; `JSONDecoder.raw_decode` already parses braces inside string values. Confirmed on `{"rationale": "A {bug} here"}`.
- **P2 `8d6e7a6eacbd`** — accepted trade-off; a ledger `OSError` must not re-post a finding the verifier already refuted (the earlier P1 on this PR made that write non-fatal for a reason).

Targeted tests: 107 passed (`test_pr_review.py` + `test_pr_review_verifier.py`). Ruff, scoped mypy, and prek passed. In-band reviewer timed out after 400s (empty provider content); wrapper pushed as designed. All five inline threads replied to and resolved. No Greptile trigger — this repo's signal is our reviewer.

TimeToBuildBob:
Current head `dafbd5ae` is CI-green and mergeable, and all review threads are resolved. I did **not** merge: the self-merge policy gate still denies this head because the AI review covers `97f5eb09`, not `dafbd5ae`, and no Greptile review exists. Waiting for a current-head reviewer result; then the gate must be rerun before any merge.

TimeToBuildBob:
Current-head re-review of `dafbd5ae` is blocked on the OpenRouter review-key daily limit until `2026-08-29T00:00Z`. The 18:02 sweep already failed this head (quota breaker; all aspect passes failed). I am not retrying in-band or stealing other keys.

CI is still green and mergeable; all finding threads are resolved. Greptile is dark on this repo — not triggering.

Scheduled recheck at 00:10Z: `ai-review.py gptme/gptme-contrib 1531 --engine llm`, then `self-merge-if-eligible.sh` if the marker matches this head.

packages/gptme-runloops/src/gptme_runloops/pr_review/verifier.py
```diff
@@ -0,0 +1,409 @@
...
+            next_file = diff.find("\ndiff --git ", file_start + len(file_marker))
+            relevant_diff = diff[file_start : next_file if next_file >= 0 else None]
+            relevant_bytes = relevant_diff.encode()
+            if len(relevant_bytes) <= _MAX_DIFF_BYTES:
+                diff = relevant_diff
+            else:
+                diff = relevant_bytes[:_MAX_DIFF_BYTES].decode(errors="replace")
+        else:
+            diff = diff_bytes[:_MAX_DIFF_BYTES].decode(errors="replace")
+        diff += "\n\n[DIFF TRUNCATED — evaluate only what is visible above]"
```
TimeToBuildBob, line 122: ⚠️ **P2** — In _build_verifier_prompt, when the full diff exceeds _MAX_DIFF_BYTES and the finding's file is located, the code extracts the relevant hunk. If that hunk is smaller than the limit, `diff` is set to the complete hunk (no truncation). However, line 122 unconditionally appends `\n\n[DIFF TRUNCATED — evaluate only what is visible above]` to the prompt, even when the hunk is fully shown. The verifier is told it is seeing a truncated diff while actually possessing full relevant context, which nudges it to treat the finding as unverifiable or reject it as hypothetical. This can cause real, fully-represented findings to be dropped by Stage 2, reducing recall and letting genuine bugs go unposted.

packages/gptme-runloops/src/gptme_runloops/pr_review/github_adapter.py
```diff
@@ -219,12 +219,21 @@ def _build_inline_comment_body(finding: ReviewFinding) -> str:
...
 ) -> str:
     """Build a top-level summary comment for the PR."""
-    n_total = len(artifact.findings)
+    postable_findings = [
+        finding
+        for finding in artifact.findings
+        if finding.disposition != Disposition.dropped
+        and finding.confidence >= min_confidence
+    ]
+    n_total = len(postable_findings)
```
TimeToBuildBob, line 234: ⚠️ **P2** — In `_build_summary_comment_body`, `n_total` is computed from `postable_findings` (findings with disposition != dropped and confidence >= min_confidence), but the `skipped_count` parameter passed from `publish_artifact` includes all findings that were not posted, including those with `disposition == dropped` (and, in shadow mode, low-confidence findings). When a finding is dropped, it is excluded from `postable_findings` yet still counted in `skipped_count`, so the summary can report more skipped findings than total findings. For example, an artifact with one confirmed high-confidence finding and one dropped high-confidence finding yields `n_total=1` and `skipped_count=1`, producing the contradictory message "**1 finding(s)**: 1 high" followed by "1 finding(s) skipped — already posted or below confidence threshold." This misleads readers about how many findings remain and how many were suppressed.

packages/gptme-runloops/src/gptme_runloops/pr_review/verifier.py
```diff
@@ -0,0 +1,409 @@
...
+    to catch) without touching that field, leaving the summary header and
+    the CLI fail-closed gate stale. Escalate only; never demote. A Stage 1
+    ``unsafe`` stays ``unsafe`` even if every finding is dropped.
+    """
+    remaining = [
+        finding
+        for finding in artifact.findings
+        if finding.disposition != Disposition.dropped
+    ]
+    if any(finding.severity in _BLOCKING_SEVERITIES for finding in remaining):
```
TimeToBuildBob, line 305: ⚠️ **P2** — In `_escalate_merge_safety`, `remaining` is computed as all findings with `disposition != Disposition.dropped`, regardless of their `confidence`. So a finding that the verifier confirms as critical but with confidence below `min_confidence` (e.g., 0.3 vs 0.6) still appears in `remaining` and forces `artifact.merge_safety` to `unsafe`. However, `publish_artifact` filters by confidence, so this finding is never posted, and `_build_summary_comment_body` – which also filters by confidence – excludes it, producing a summary that says 'No findings.' while the artifact's `merge_safety` is `unsafe` and the CI gate exits non-zero. This inconsistency means a PR containing only low-confidence findings is reported as clean in the summary but fails the gate, confusing users who rely on the summary count/verdict to decide whether to merge. The escalation should presumably consider only findings that would actually be posted (confidence >= min_confidence).

packages/gptme-runloops/src/gptme_runloops/pr_review/verifier.py
```diff
@@ -0,0 +1,409 @@
...
+            else:
+                diff = relevant_bytes[:_MAX_DIFF_BYTES].decode(errors="replace")
+        else:
+            diff = diff_bytes[:_MAX_DIFF_BYTES].decode(errors="replace")
+        diff += "\n\n[DIFF TRUNCATED — evaluate only what is visible above]"
+
+    finding_payload = {
+        "title": finding.title,
+        "category": finding.category,
+        "severity": finding.severity.value,
```
TimeToBuildBob, line 127: ⚠️ **P2** — The verifier prompt embeds the generator's severity label as part of the finding payload. Although the instructions state 'Assign your own severity label independently of the generator's label', including the original label anchors the model, especially for borderline cases. A real security bug labeled 'medium' by the generator is then likely to be confirmed with 'medium' instead of being promoted to 'high' or 'critical' — precisely the mislabeled-P0 failure mode this stage exists to catch. As a result, severe defects can remain in the non-blocking lane, undermines the verifier's purpose. Fix: omit 'severity' from the finding payload, or mark it as a generator label the verifier should ignore.

packages/gptme-runloops/src/gptme_runloops/pr_review/verifier.py
```diff
@@ -0,0 +1,409 @@
...
+
+    # Resolve a stable identity anchor for the ledger records
+    from .schema import ReviewTarget
+
+    target = artifact.target
+    if isinstance(target, ReviewTarget):
+        repo_id = target.repo
+        head_sha = target.head_sha
+    else:
+        repo_id = target.repo_name or target.checkout.split("/")[-1]
```
TimeToBuildBob, line 359: ⚠️ **P2** — In `verify_artifact`, when the target is a `LocalReviewTarget` and `repo_name` is None, `repo_id` is derived from `target.checkout.split("/")[-1]`. If the checkout path is the filesystem root (`/`), this yields an empty string. The shadow ledger record then contains `"repo": ""`, making suppression tracking ambiguous for the uncommon case of running against the root directory. While edge-case, it produces a malformed ledger entry that could be mistaken for a data error.

packages/gptme-runloops/src/gptme_runloops/pr_review/verifier.py
```diff
@@ -0,0 +1,409 @@
...
+    cmd.append(prompt)
+
+    result = subprocess.run(
+        cmd,
+        cwd=checkout,
+        capture_output=True,
+        text=True,
+        timeout=120,  # verifier calls are shorter than full reviews
+    )
+    if result.returncode not in (0, 1):
```
TimeToBuildBob, line 158: ⚠️ **P2** — In `_invoke_verifier_model`, the subprocess result is accepted as success when `returncode in (0, 1)`. The value 1 is treated as a non-error, meaning any gptme invocation that exits with code 1 (possibly due to a tool error or a warning path) still returns its stdout to be parsed as a verifier verdict. If gptme exits 1 and produces partial or malformed output, `_extract_verifier_json` may raise a `ValueError`, which `verify_artifact` treats as a tool error and leaves the finding as `needs_validation`—acceptable. But if the stdout happens to contain a valid JSON object (e.g., a rerun of a previous successful response), the verdict would be accepted despite the non-zero exit, silently ignoring the failure signal.

packages/gptme-runloops/src/gptme_runloops/pr_review/verifier.py
TimeToBuildBob: Fixed in `b0481fee`. `_escalate_merge_safety` now filters to the same publishable set as `publish_artifact` (`disposition != dropped` and `confidence >= min_confidence`), so a confirmed critical below the post threshold can no longer mark the artifact `unsafe` while the summary says "No findings." Regression covers a 0.3-confidence critical that must leave Stage 1 `safe` in place.

TimeToBuildBob:
Pushed `b0481fee` for the remaining 00:54 finding (`84b8ecc9191d`): merge-safety escalation now ignores below-threshold findings, matching publish/summary. CI is rerunning on the new head.

The other five 00:54 P2s stay disposed as nits / accepted trade-offs (truncation banner on a fully-shown hunk, skipped-count vs dropped, generator-severity anchoring, `/` checkout repo_id, gptme exit 1). Not merging: the current-head marker still names that finding as P1 against a P2 thread, and the new SHA needs a fresh review before the self-merge gate can pass.

packages/gptme-runloops/src/gptme_runloops/pr_review/verifier.py
```diff
@@ -0,0 +1,425 @@
...
+
+
+def _build_verifier_prompt(finding: ReviewFinding, diff: str) -> str:
+    """Build the verifier prompt for a single finding."""
+    # Keep the finding's file context before global context when a large diff must
+    # be truncated; otherwise findings near the end of a PR become unverifiable.
+    diff_bytes = diff.encode()
+    _MAX_DIFF_BYTES = 40_000
+    if len(diff_bytes) > _MAX_DIFF_BYTES:
+        file_marker = f"diff --git a/{finding.file_path} b/{finding.file_path}"
```
TimeToBuildBob, line 110: ❌ **P1** — Truncation fallback misses renamed-file diffs, dropping real findings. The file marker at verifier.py:110 is built as `diff --git a/{finding.file_path} b/{finding.file_path}`. When a PR renames a file, the unified header is `diff --git a/old b/new`, so the marker never matches if finding.file_path is the new path. file_start becomes -1, and the code takes the first 40k bytes of the whole diff (line 121). For a large PR this almost certainly excludes the renamed file's hunks, so the model sees no evidence for the finding and tends to return real:false, dropping a legitimate issue via the shadow ledger. Renames are common enough that this silently suppresses real defects. _(3/3 passes)_

packages/gptme-runloops/src/gptme_runloops/pr_review/github_adapter.py
TimeToBuildBob, line 234: ⚠️ **P2** — _build_summary_comment_body now computes n_total from postable_findings (not dropped and confidence >= min_confidence, lines 228-234), but still prints the posted_count and skipped_count lines verbatim (lines 262-267). These counts are independent of n_total: after the verifier drops one finding from a two-finding artifact, the body reads '**1 finding(s)**: 1 high' while also printing '1 finding(s) posted as inline comments' and '1 finding(s) skipped', which totals 2 against a header of 1. Before this PR n_total was len(artifact.findings), so the three numbers always agreed; the new counted-only-postable logic makes the header disagree with the sum of the two count lines whenever any finding was dropped or fell below the confidence threshold. _(2/3 passes — pass 3 looked and disagreed)_

packages/gptme-runloops/src/gptme_runloops/pr_review/verifier.py
TimeToBuildBob: Rejected — same family as `44915479a8ff`, already adjudicated on this PR.

Stage 1 already supplies the finding payload (title, description, evidence, line_range, file_path) as JSON. Truncation only applies when the unified diff exceeds 40KB. A rename header (`diff --git a/old b/new`) missing the exact `a/{path} b/{path}` marker falls back to the first 40KB plus the existing `[DIFF TRUNCATED]` banner — the same accepted truncation trade-off as a hunk that simply sits past the cap. Verifier tool/parse failures already fail closed (`needs_validation`); they do not silently drop.

Fuzzy b-side path recovery is YAGNI without a production miss. Not expanding the treadmill for an unobserved rename-in-a-40KB+ diff.

packages/gptme-runloops/src/gptme_runloops/pr_review/github_adapter.py
TimeToBuildBob: Rejected as a copy-edit nit, same family as `8be9d4e2aa29` (auto-resolved this round). `n_total` is the publishable set by design — dropped/below-threshold findings are not "findings" in the header. `skipped_count` is the publisher's bookkeeping (already-posted + below-threshold + verifier-dropped) and is independent of the header. Changing the skip-line wording does not change behavior. Not worth a follow-up commit on round 7.

TimeToBuildBob:
Current-head AI review of `b0481fee` posted (3/5; 1 P1 + 1 P2). Both threads disposed:

- P1 `87a363cc1463` (rename-header truncation) — rejected, same family as `44915479a8ff`. YAGNI without a production miss; tool failures already fail closed.
- P2 `1e6a9913ddc7` (skipped_count vs n_total) — rejected copy-edit nit.

Self-merge policy gate is now eligible (AI review at current head, findings disposed). I did **not** merge: the consensus gate scored 3/5 on a prompt-injection concern already mitigated in `298de278` (JSON-encoded untrusted inputs + explicit instruction boundary). Not re-rolling that gate in-band. Greptile remains dark — not triggering.

TimeToBuildBob:
Consensus merge gate blocked again on `b0481fee` (second independent 3/5, 12h after the first).

- 01:52Z: prompt-injection / cheaper-model concern (already mitigated in `298de278`: JSON-encoded untrusted inputs + instruction boundary).
- 13:57Z: LLM-based refutation could drop critical findings; production reliability unproven.

Self-merge policy is eligible (CI green, threads resolved, AI review at current head, findings disposed). I am not merging and not re-rolling the consensus gate or AI review on this head. This is a design-level hold on a new opt-in verifier.

Waiting for maintainer review or merge.

PR #1532, opened 2026-08-27
fix(rag): honor index pattern and clarify file limit error

TimeToBuildBob:
## Problem

The final review of gptme/gptme-contrib#1523 found two issues after that PR had already been merged:

- `gptme-rag index --pattern` accepted a pattern but never passed it to `collect_documents`, so the flag had no effect.
- The low-level file-limit diagnostic recommended `--file-limit` even for callers such as `watch` that do not expose that CLI flag.

## Changes

- Thread `pattern` from the `index` CLI into `collect_documents`.
- Make the truncation diagnostic caller-neutral by referring to the `file_limit` parameter.
- Add a regression test proving `--pattern` reaches document collection.

## Verification

- `uv run --package gptme-rag --extra test pytest packages/gptme-rag/tests/test_file_limit.py -q` — 10 passed
- `uv run ruff check packages/gptme-rag/src/gptme_rag/cli.py packages/gptme-rag/src/gptme_rag/indexing/indexer.py packages/gptme-rag/tests/test_file_limit.py` — passed
- targeted mypy for the two changed source files — passed
- full pre-commit reached package typechecking; the changed package passed, but unrelated packages failed while uv fetched gptme with `fatal: not a git repository`

Follow-up to gptme/gptme-contrib#1523.


TimeToBuildBob:
CI is green and the PR is cleanly mergeable. The self-merge gate is still blocked because the initial Greptile review has not arrived; the helper is respecting the 45-minute auto-review grace period. No merge attempted — waiting for review, then the policy gate must be rerun.

packages/gptme-rag/tests/test_file_limit.py
```diff
@@ -68,6 +68,25 @@ def test_collect_documents_honors_file_limit(tmp_path):
...
+    monkeypatch.setattr(Indexer, "get_all_documents", lambda self: [])
+
+    def collect_documents(self, path, glob_pattern="**/*.*", file_limit=100_000):
+        observed_patterns.append(glob_pattern)
+        return []
+
+    monkeypatch.setattr(Indexer, "collect_documents", collect_documents)
+
+    runner = CliRunner()
+    result = runner.invoke(cli, ["index", "--pattern", "*.md", str(tmp_path)])
```
TimeToBuildBob, line 86: ⚠️ **P2** — The new test invokes the CLI via `runner.invoke`, which runs the actual `index` command and constructs a real `Indexer` instance. The test stubs `get_all_documents` and `collect_documents` but not the `Indexer` constructor. With the default `embedding_function="auto"` and no existing collection, the constructor instantiates `ModernBERTEmbedding`, which loads (and may attempt to download) a ~80 MB model. In an offline CI environment or when the model is unavailable, this raises an exception, the CLI exits non-zero, and `assert result.exit_code == 0` fails. Even when the model is cached, the test remains slow and environment-dependent, making it non-hermetic. Stubbing the constructor (or the whole `Indexer` class) would isolate the CLI threading behavior from model-loading side effects.

packages/gptme-rag/src/gptme_rag/indexing/indexer.py
```diff
@@ -1660,7 +1660,7 @@ def _get_valid_files(
         if len(valid_files) > file_limit:
             logger.error(
                 f"File limit ({file_limit}) reached, was {len(valid_files)}. "
-                f"Pass a higher --file-limit or use a more specific glob pattern than '{glob_pattern}'."
+                f"Raise file_limit or use a more specific glob pattern than '{glob_pattern}'."
```
TimeToBuildBob: ⚠️ **P2** — The file-limit truncation diagnostic now says "Raise file_limit" instead of "Pass a higher --file-limit". For the `index` CLI, users do not know what 'file_limit' refers to; it is an internal parameter name, not a CLI flag. The previous message explicitly told users which flag to raise. The new message is ambiguous and less actionable for the primary `index` command. For the `watch` command, which has no file_limit flag, the message is equally unhelpful, so the change does not actually improve clarity for any caller â it regresses the `index` UX.

packages/gptme-rag/src/gptme_rag/cli.py
```diff
@@ -267,7 +267,7 @@ def index(
                 else:
                     status.update(f"Processing directory: {path}")
 
-                documents = indexer.collect_documents(path, file_limit=file_limit)
+                documents = indexer.collect_documents(path, pattern, file_limit=file_limit)
```
TimeToBuildBob, line 270: ⚠️ **P2** — The change to cli.py passes the raw --pattern value directly to collect_documents. If a user supplies an empty pattern (e.g. --pattern ""), Click passes the empty string, which flows to _get_valid_files. In the non-git branch, _get_valid_files calls path.glob(glob_pattern) with glob_pattern=""; pathlib's glob() raises ValueError('Unacceptable pattern:') for an empty string. Previously the index command never passed the pattern to collect_documents, so an empty --pattern was silently ignored and the default '**/*.*' was used. Now the command throws inside the try block, prints an error, and exits 0, indexing nothing. This is a regression for any user who accidentally or intentionally sets an empty pattern expecting the default behavior.

TimeToBuildBob:
## 🤖 AI code review

The PR threads the --pattern option from the index CLI into collect_documents, and replaces the fnmatch-based glob filtering in _get_valid_files with pathspec's GitIgnoreSpecPattern so the pattern is applied in both git and glob discovery modes. It also makes the file-limit diagnostic neutral about the CLI flag and adds tests for pattern threading and pattern filtering in a git repository.

### Confidence Score: 1/5 — At least one P0 finding

> ⚠️ **1 generated file excluded and NOT reviewed:** ``uv.lock`` (+2/-0 lines, 0KB). Machine-generated output is removed from the diff before review — from what the model sees and from the size gate alike — so nothing below says anything about it.

<b>3 findings</b> · 🛑 **1** P0 · ❌ **1** P1 · ⚠️ **1** P2 — 🛑 **2 at P0/P1**

<details><summary><b>2 inline</b></summary>

**🛑 P0 critical** — `packages/gptme-rag/src/gptme_rag/indexing/indexer.py:18`

The import `from pathspec.patterns.gitignore.spec import GitIgnoreSpecPattern` is invalid: pathspec 0.12 (the version constraint) has no such class. The module `pathspec.patterns.gitignore.spec` does not exist; the correct class is `GitIgnorePattern` in `pathspec.patterns.gitignore`, or one can use `GitIgnoreSpec.from_lines([pattern])` to build a spec. As written, importing `gptme_rag.indexing.indexer` raises `ImportError: cannot import name 'GitIgnoreSpecPattern' from 'pathspec.patterns.gitignore.spec'`, which breaks the entire gptme-rag package — every command (index, search, watch, status, mcp) fails on import. The added test `test_get_valid_files_honors_pattern_in_git_repo` would also fail at collection import, so it cannot have run as claimed. Replace with the correct pathspec API, e.g. `from pathspec.patterns.gitignore import GitIgnorePattern` and instantiate it: `GitIgnorePattern(pattern).match_file(path) is not None`, or use `GitIgnoreSpec.from_lines([pattern])` and call `match_file` on the spec.

```suggestion
from pathspec.patterns.gitignore import GitIgnorePattern

def _glob_match_path(path: str, pattern: str) -> bool:
    return GitIgnorePattern(pattern).match_file(path) is not None
```

<sub>**How this was verified:** Checked the pathspec 0.12 source distribution: only `GitIgnorePattern` exists under `pathspec.patterns.gitignore`; there is no `GitIgnoreSpecPattern`. The test module imports `Indexer` and would fail immediately with ImportError, contradicting the PR's test pass report.</sub>

**⚠️ P2 medium** — `packages/gptme-rag/tests/test_file_limit.py:104`

The new test `test_get_valid_files_honors_pattern_in_git_repo` runs `git init` and `git add .` via `subprocess.run` with `check=True`. This makes the test non-hermetic: it fails with an exception in environments where git is not installed or fails to initialize (e.g., no user.name/email configured). The test should either use `pytest.importorskip` to skip when git is unavailable, or mock the subprocess calls, to keep the test suite deterministic.

<sub>**How this was verified:** The test requires git binary and a config; in a clean CI environment without git, it will raise FileNotFoundError.</sub>

</details>

### Comments outside the diff

**❌ P1 high** — [`packages/gptme-rag/src/gptme_rag/indexing/indexer.py:1633`](https://github.com/gptme/gptme-contrib/blob/cb4f1e158423cea9da7f8b0783eeea7f84b9f24b/packages/gptme-rag/src/gptme_rag/indexing/indexer.py#L1633)

In `_get_valid_files`, when the directory is inside a git repository, the code lists files with `git -C <path> ls-files --cached --others --exclude-standard`. The output paths are relative to the repository root, not to `path`. The subsequent `files = [path / line for line in result.stdout.splitlines()]` prepends `path` to each relative line, so for a subdirectory `path` like `/repo/sub`, a line `sub/file.py` becomes `/repo/sub/sub/file.py`. The `if not f.is_file(): continue` then skips every file because the constructed path does not exist, so `_get_valid_files` returns an empty set. As a result, `gptme-rag index` on any subdirectory of a git repo indexes nothing. This bug is pre-existing but the PR's change to always apply the glob pattern does not fix it and remains a serious correctness issue for subdirectory indexing.

<sub>**How this was verified:** Verified that `git -C <path> ls-files` returns repo-root-relative paths; `f.relative_to(path)` raises for files not under `path`. The added test uses a fresh temp dir as repo root, so it never exercises a subdirectory scan.</sub>

- `packages/gptme-rag/pyproject.toml` — Adds pathspec&gt;=0.12 to the dependency list.
- `packages/gptme-rag/src/gptme_rag/cli.py` — Passes the pattern argument to indexer.collect_documents in the index command.
- `packages/gptme-rag/src/gptme_rag/indexing/indexer.py` — Replaces the conditional fnmatch filtering with a GitIgnoreSpecPattern-based match and updates the file-limit error message.
- `packages/gptme-rag/tests/test_file_limit.py` — Adds tests verifying --pattern reaches collect_documents and that git-backed discovery filters files with the requested pattern.

</details>

| `1c8a667e14f9` | 3/5 | 4 | llm | 2026-08-27 21:37 UTC |
| `b99765b0c44e` | 3/5 | 1 | llm | 2026-08-27 23:40 UTC |
| `4590b6da8a2d` | 3/5 | 2 | llm | 2026-08-28 02:10 UTC |
| `22d50d3a5fe0` | 3/5 | 1 | llm | 2026-08-28 04:55 UTC |

</details>

<sub>Reviewed `cb4f1e158423` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 300s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

TimeToBuildBob:
@greptileai review

packages/gptme-rag/tests/test_file_limit.py
TimeToBuildBob: Fixed in `b99765b0`: the CLI forwarding test now stubs `Indexer.__init__`, so it cannot initialize ChromaDB or load/download an embedding model. The targeted file-limit tests pass (11/11).

packages/gptme-rag/src/gptme_rag/indexing/indexer.py
TimeToBuildBob: Fixed in `b99765b0`: the caller-neutral diagnostic now says “Increase the file limit or use a more specific glob pattern,” avoiding both an internal parameter name and a CLI-only flag.

TimeToBuildBob:
Pushed `b99765b0` to address the review findings: patterns are now applied to `git ls-files` results (with a git-repo regression test), the CLI test is hermetic, and the file-limit diagnostic is caller-neutral. Targeted gptme-rag tests pass (11/11); fresh CI is running. I did not coerce an explicit empty `--pattern` to the default because Click receiving an explicitly empty glob should remain an invalid user input rather than silently broaden indexing.

packages/gptme-rag/src/gptme_rag/indexing/indexer.py
```diff
@@ -1632,6 +1632,11 @@ def _get_valid_files(
             files = list(path.glob(glob_pattern))
             gitignore_patterns = self._load_gitignore(path)
 
+        # pathlib treats **/ as optional at the start of a glob, while fnmatch
+        # does not. Normalize that prefix so git-listed paths follow the same
+        # matching semantics as the pathlib fallback.
+        fnmatch_pattern = glob_pattern.replace("**/*", "*")
```
TimeToBuildBob: ❌ **P1** — The normalization of the glob pattern is incomplete, so patterns that start with '**/' but are not followed by '*', e.g. '**/a.txt', are not handled. In git-backed discovery the pattern filter is now applied to files listed by git ls-files, but fnmatch_path('a.txt', '**/a.txt') returns False because the pattern demands a slash; a root-level 'a.txt' that pathlib.glob would match is silently dropped. This regression is newly exposed in git mode, where the filter previously did not run. For patterns like '**/a.txt' users expect both root and nested matches, but only nested ones are indexed. The replacement only handles the literal '**/*' substring, not a bare leading '**/'.

TimeToBuildBob: Fixed in `4590b6da`: recursive patterns are now checked in both their original and root-normalized forms, preserving nested matches while including root-level matches. The regression test covers both root and nested files for `**/named.md`; the targeted suite passes (11/11).

packages/gptme-rag/tests/test_file_limit.py
```diff
@@ -68,6 +69,53 @@ def test_collect_documents_honors_file_limit(tmp_path):
...
+
+def test_get_valid_files_honors_pattern_in_git_repo(tmp_path):
+    """Git-backed discovery filters files with the requested pattern."""
+    (tmp_path / "docs").mkdir()
+    (tmp_path / "root.md").write_text("root")
+    (tmp_path / "named.md").write_text("named")
+    (tmp_path / "docs" / "nested.md").write_text("nested")
+    (tmp_path / "docs" / "named.md").write_text("nested named")
+    (tmp_path / "docs" / "ignored.txt").write_text("ignored")
+    subprocess.run(["git", "init"], cwd=tmp_path, check=True, capture_output=True)
```
TimeToBuildBob, line 104: ⚠️ **P2** — The new test `test_get_valid_files_honors_pattern_in_git_repo` invokes `subprocess.run(["git", "init"], ...)`. If the `git` executable is not installed or not in PATH, `subprocess.run` raises `FileNotFoundError` (not `CalledProcessError`), which propagates out of the test and fails it. This makes the test non-hermetic: it depends on the environment having git available. For a package that uses git in production this might be acceptable in CI, but the test suite should isolate itself so it can run in environments without git. The consequence is that the entire test file fails in such environments even though the production code might be correct. The test should skip when git is unavailable, e.g. by checking `shutil.which("git")` and calling `pytest.skip`.

TimeToBuildBob:
## Convergence adjudication

**Fixed**
- P1 recursive-glob semantics: `22d50d3a` replaces the one-off leading-`**/` normalization with segment-aware matching where every `**/` spans zero or more directories. A regression test now pins both `src/root.py` and `src/nested/child.py` for `src/**/*.py`.

**Remaining (non-blocking)**
- P2 test dependency on `git`: dismissed. This test deliberately exercises the production `git ls-files` branch, the package itself invokes `git`, and all supported CI jobs provide it. Skipping without `git` would silently stop testing the behavior this PR fixes.
- Empty explicit `--pattern ""`: dismissed as invalid user input rather than silently broadening the index to the default pattern.

**CI**
- Previous head `4590b6da`: all required checks green.
- Current fix `22d50d3a`: fresh CI is running. Local targeted suite: 11/11 passed; Ruff and format checks passed. Full local typecheck could not complete because uv's shared git cache failed with `fatal: not a git repository`; this is environmental, so CI remains the authoritative full gate.

**Domain risk**
- Glob syntax is the fragile surface. The tests cover root and nested matches for leading and embedded recursive wildcards; a maintainer may still want a quick manual `index --pattern "src/**/*.py"` smoke test on a real git repo before merge.

**Convergence**
- Detector reported `stable_rounds: 0` with `status: no_history`; the escalation cap came from dispatch attempts rather than detector-recorded stable rounds. No further review was triggered. Stop here for maintainer judgment after CI settles.

packages/gptme-rag/src/gptme_rag/indexing/indexer.py
```diff
@@ -26,6 +27,30 @@
...
+            regex_parts.append(".*")
+            i += 2
+        elif pattern[i] == "*":
+            regex_parts.append("[^/]*")
+            i += 1
+        elif pattern[i] == "?":
+            regex_parts.append("[^/]")
+            i += 1
+        else:
+            regex_parts.append(re.escape(pattern[i]))
```
TimeToBuildBob: ❌ **P1** — The new `_glob_match_path` (indexer.py:30-51) compiles a regex from a glob but only handles `*`, `?`, and `**` as wildcards. Character classes (`[a-z]`) and brace alternation (`{md,txt}`) are not recognized; every such metacharacter is passed through `re.escape`, so a pattern like `*.{md,txt}` becomes a regex expecting a literal literal `{md,txt}` at the end, which no real path matches. Previously, in glob fallback mode the file list came from `path.glob()` (which fully supports these features) and was then re-checked with `fnmatch_path` (also supporting them), so these patterns worked. In git mode the pattern was ignored entirely, so a user passing `--pattern 'src/*.[ch]'` got all tracked files; now `_glob_match_path` returns False for every entry and `_get_valid_files` returns an empty set, silently indexing nothing. This is a functional regression for any pattern using glob features beyond `*`, `?`, and `**`. Fix by using a full glob-to-regex translator (e.g. `wcmatch` or a proper `**`-aware `fnmatch` translation) rather than a hand-rolled subset.

TimeToBuildBob: Fixed in `cb4f1e15`: the hand-rolled matcher is replaced with pathspec's full git-wildmatch implementation, preserving recursive `**` semantics and character classes. The git-backed regression now covers `src/**/*.[p]y`; targeted tests pass (11/11), Ruff passes, and scoped pre-commit passes. The required in-band reviewer was run before push but timed out after 900s, so the wrapper followed its fail-open contract and pushed the fix.

packages/gptme-rag/src/gptme_rag/indexing/indexer.py
```diff
@@ -15,6 +15,7 @@
 from chromadb.api import ClientAPI
 from chromadb.config import Settings
 from chromadb.errors import NotFoundError
+from pathspec.patterns.gitignore.spec import GitIgnoreSpecPattern
```
TimeToBuildBob, line 18: 🛑 **P0** — The import `from pathspec.patterns.gitignore.spec import GitIgnoreSpecPattern` is invalid: pathspec 0.12 (the version constraint) has no such class. The module `pathspec.patterns.gitignore.spec` does not exist; the correct class is `GitIgnorePattern` in `pathspec.patterns.gitignore`, or one can use `GitIgnoreSpec.from_lines([pattern])` to build a spec. As written, importing `gptme_rag.indexing.indexer` raises `ImportError: cannot import name 'GitIgnoreSpecPattern' from 'pathspec.patterns.gitignore.spec'`, which breaks the entire gptme-rag package — every command (index, search, watch, status, mcp) fails on import. The added test `test_get_valid_files_honors_pattern_in_git_repo` would also fail at collection import, so it cannot have run as claimed. Replace with the correct pathspec API, e.g. `from pathspec.patterns.gitignore import GitIgnorePattern` and instantiate it: `GitIgnorePattern(pattern).match_file(path) is not None`, or use `GitIgnoreSpec.from_lines([pattern])` and call `match_file` on the spec.

packages/gptme-rag/tests/test_file_limit.py
```diff
@@ -68,6 +69,65 @@ def test_collect_documents_honors_file_limit(tmp_path):
...
+    (tmp_path / "src").mkdir()
+    (tmp_path / "src" / "nested").mkdir()
+    (tmp_path / "root.md").write_text("root")
+    (tmp_path / "named.md").write_text("named")
+    (tmp_path / "docs" / "nested.md").write_text("nested")
+    (tmp_path / "docs" / "named.md").write_text("nested named")
+    (tmp_path / "docs" / "ignored.txt").write_text("ignored")
+    (tmp_path / "src" / "root.py").write_text("root")
+    (tmp_path / "src" / "nested" / "child.py").write_text("nested")
+    subprocess.run(["git", "init"], cwd=tmp_path, check=True, capture_output=True)
```
TimeToBuildBob, line 104: ⚠️ **P2** — The new test `test_get_valid_files_honors_pattern_in_git_repo` runs `git init` and `git add .` via `subprocess.run` with `check=True`. This makes the test non-hermetic: it fails with an exception in environments where git is not installed or fails to initialize (e.g., no user.name/email configured). The test should either use `pytest.importorskip` to skip when git is unavailable, or mock the subprocess calls, to keep the test suite deterministic.

ErikBjare:
@TimeToBuildBob  fix remaining issues in a follow-up PR

TimeToBuildBob:
Follow-up opened: gptme/gptme-contrib#1542.

Two review threads here were still unresolved at merge. One was a real latent break:

- **P0 (`pathspec` import)** — the finding's stated reason was wrong (`GitIgnoreSpecPattern` does exist in pathspec 1.1.1, which is what we resolve, hence green CI), but it pointed at a real defect from a different angle. The package declares `pathspec>=0.12`, and that module tree only exists from pathspec **1.0**. Since the import is at module scope, any resolution on 0.12.x makes `gptme_rag.indexing.indexer` fail to import entirely. Reproduced directly on 0.12.1: `ModuleNotFoundError: No module named 'pathspec.patterns.gitignore'`. #1542 raises the floor to `>=1.0`.
- **P2 (git-dependent test)** — I dismissed this twice here, and that reasoning was right for CI but needlessly strict for local runs. #1542 adds a `skipif(shutil.which("git") is None)` guard, which is a no-op in CI (git always present) and keeps the suite runnable elsewhere.

Still dismissed: the explicit-empty `--pattern ""` case. An explicitly empty glob should remain invalid user input rather than silently broadening the index to the default pattern.

Issue #2, opened 2025-02-06
Failed to initialize gptme tools: [Errno 2] No such file or directory


0xbrayo:
Got the error below while try to set up a discord bot.
```[14:21:30] INFO     Loaded environment from .env                                                   discord_bot.py:68
           INFO     Configuration:                                                                 discord_bot.py:80
           INFO       Rate limit: 1.0s                                                             discord_bot.py:81
           INFO       Default model: anthropic                                                     discord_bot.py:82
           INFO       Log directory: logs/discord                                                  discord_bot.py:83
           INFO     Using privileged intents (members, presences)                                  discord_bot.py:94
           WARNING  PyNaCl is not installed, voice will NOT be supported                               client.py:300
           INFO     Configuration validated:                                                      discord_bot.py:266
           INFO       Model: openai/gpt-4                                                         discord_bot.py:267
           INFO       Rate limit: 1.0s                                                            discord_bot.py:268
           INFO       Privileged intents: True                                                    discord_bot.py:269
[14:21:30] Using model: anthropic/claude-3-5-sonnet-20241022
           INFO     Loaded 5 gptme tools: append, patch, read, save, shell                        discord_bot.py:948
           ERROR    Failed to initialize gptme tools: [Errno 2] No such file or directory:        discord_bot.py:955
                    '/home/brayo/repos/gptme-contrib/scripts/discord' -> 'logs/discord/workspace'```

ErikBjare:
Ah, you could try creating the `logs/discord` folder yourself.
It was meant to keep the log folder separate from main `gptme` convos (I do the same for conversations with bob).

0xbrayo:
Managed to get it working just creating the folder was enough. Thanks

ErikBjare:
I should change the default location to `logs_discord` in the workspace folder, to make it clear and to not mess with the structure of `logs/` (which shouldn't have conversations nested under subfolders).

PR #1534, opened 2026-08-28
fix(rag): content-hash change detection to stop mtime-only re-embeds

TimeToBuildBob:
## Problem

The `index` command's change-detection keyed solely on file mtime (`cli.py`). A git restore, checkout, or hot-worktree mtime rewrite forced a full re-embed of unchanged files — the latent cause of the vitals-regen index CPU burn once the `file_limit` cap was lifted (gptme-contrib#1523, merged).

## Fix

- `document.py`: record a `content_hash` (sha256 of file bytes) in `from_file` metadata, alongside `last_modified`.
- `cli.py`: compare `content_hash` as the **primary** change-detection key; fall back to mtime **only** for stored documents that predate content hashing (legacy index entries).

## Verification

- New `tests/test_content_hash_changedetection.py`: 6 tests covering hash recording, content-change invalidation, mtime-touch stability, and end-to-end index skip/re-embed via the CLI.
- `test_index_skips_unchanged_mtime_only_rewrite` proves the core fix: bumping mtime on unchanged content does **not** re-embed.
- `make typecheck` (gptme-rag): clean.

Closes Bug #2 of the vitals-regen index investigation.

### DIFF
--- a/packages/gptme-rag/src/gptme_rag/cli.py
+++ b/packages/gptme-rag/src/gptme_rag/cli.py
@@ -273,8 +273,13 @@ def index(
 
         logger.debug("Loaded %d existing files from index", len(existing_files))
 
-        # First, collect all documents and filter for new/modified
+        # First, collect all documents and filter for new/modified.
+        # cleanup_hashes: source -> current hash when that generation is already
+        # stored but leftover older hashes remain (add succeeded, delete failed).
+        # Those sources must not be re-added — chunk IDs include content_hash, so
+        # a second add of the same bytes collides — only the stale IDs are deleted.
         all_documents = []
+        cleanup_hashes: dict[str, object] = {}
         with console.status("Collecting documents...") as status:
             for path in paths:
                 if path.is_file():
@@ -310,6 +315,13 @@ def index(
                                 if stored_hashes == {current_hash}:
                                     logger.debug("Unchanged file: %s", abs_source)
                                     continue
+                                if current_hash in stored_hashes:
+                                    cleanup_hashes[abs_source] = current_hash
+                                    logger.debug(
+                                        "Current generation already indexed: %s",
+                                        abs_source,
+                                    )
+                                    continue
                                 logger.debug(
                                     "Modified file: %s (content hash changed)",
                                     abs_source,
@@ -340,8 +352,39 @@ def index(
 
                 all_documents.extend(filtered_documents)
 
+        def stale_ids_for(hash_by_source: dict[str, object]) -> list[str]:
+            return [
+                doc.doc_id
+                for doc in existing_docs
+                if doc.doc_id
+                and str(doc.metadata.get("source", "")) in hash_by_source
+                and doc.metadata.get("content_hash")
+                != hash_by_source.get(str(doc.metadata.get("source", "")))
+            ]
+
+        leftover_stale_ids = stale_ids_for(cleanup_hashes)
+
         if not all_documents:
-            console.print("No new or modified documents to index", style="yellow")
+            if leftover_stale_ids:
+                try:
+                    indexer.collection.delete(ids=leftover_stale_ids)
+                except Exception as delete_err:
+                    logger.warning(
+                        "Failed to delete %d leftover stale chunk(s): %s",
+                        len(leftover_stale_ids),
+                        delete_err,
+                    )
+                    console.print(
+                        "⚠️ Leftover stale chunks remain; they will be removed on the next run",
+                        style="yellow",
+                    )
+                else:
+                    console.print(
+                        f"✅ Removed leftover stale chunks for {len(cleanup_hashes)} files",
+                        style="green",
+                    )
+            else:
+                console.print("No new or modified documents to index", style="yellow")
             return
 
         # Then process them with a progress bar. Preserve the previous chunks until
@@ -359,14 +402,7 @@ def index(
             str(doc.metadata.get("source", "")): doc.metadata.get("content_hash")
             for doc in all_documents
         }
-        stale_ids = [
-            doc.doc_id
-            for doc in existing_docs
-            if doc.doc_id
-            and str(doc.metadata.get("source", "")) in sources
-            and doc.metadata.get("content_hash")
-            != current_hash_by_source.get(str(doc.metadata.get("source", "")))
-        ]
+        stale_ids = stale_ids_for(current_hash_by_source) + leftover_stale_ids
 
         with tqdm(
             total=n_chunks,
@@ -382,8 +418,8 @@ def index(
                 indexer.collection.delete(ids=stale_ids)
             except Exception as delete_err:
                 # Add already succeeded. Failing the command here would hide that
-                # the new generation is indexed; mixed leftover hashes are treated
-                # as modified on the next run, so a retry self-heals.
+                # the new generation is indexed; a retry sees the current hash
+                # already stored and deletes leftover older IDs without re-adding.
                 logger.warning(
                     "Indexed replacements but failed to delete %d stale chunk(s): %s",
                     len(stale_ids),
--- a/packages/gptme-rag/src/gptme_rag/cli.py
+++ b/packages/gptme-rag/src/gptme_rag/cli.py
@@ -280,6 +280,7 @@ def index(
         # a second add of the same bytes collides — only the stale IDs are deleted.
         all_documents = []
         cleanup_hashes: dict[str, object] = {}
+        empty_sources: set[str] = set()
         with console.status("Collecting documents...") as status:
             for path in paths:
                 if path.is_file():
@@ -299,6 +300,15 @@ def index(
                         doc.metadata["source"] = abs_source
 
                         existing = existing_files.get(abs_source)
+                        if not (doc.content or "").strip():
+                            # Empty/whitespace source: never index it. If it
+                            # previously had chunks, drop them all.
+                            if existing is not None:
+                                empty_sources.add(abs_source)
+                                logger.debug("Empty file, dropping stale chunks: %s", abs_source)
+                            else:
+                                logger.debug("Skipping empty file: %s", abs_source)
+                            continue
                         if existing is None:
                             logger.debug("New file: %s", abs_source)
                             filtered_documents.append(doc)
@@ -313,16 +323,20 @@ def index(
                                 set[object], existing.get("content_hashes") or set()
                             )
                             has_unhashed = bool(existing.get("has_unhashed"))
-                            if stored_hashes and not has_unhashed:
-                                if stored_hashes == {current_hash}:
-                                    logger.debug("Unchanged file: %s", abs_source)
-                                    continue
+                            if stored_hashes:
                                 if current_hash in stored_hashes:
-                                    cleanup_hashes[abs_source] = current_hash
-                                    logger.debug(
-                                        "Current generation already indexed: %s",
-                                        abs_source,
-                                    )
+                                    # Current generation is already indexed. Skip
+                                    # re-embed even when leftover unhashed/legacy
+                                    # chunks remain — re-adding the same IDs
+                                    # UniqueConstraintErrors and aborts cleanup.
+                                    if stored_hashes != {current_hash} or has_unhashed:
+                                        cleanup_hashes[abs_source] = current_hash
+                                        logger.debug(
+                                            "Current generation already indexed: %s",
+                                            abs_source,
+                                        )
+                                    else:
+                                        logger.debug("Unchanged file: %s", abs_source)
                                     continue
                                 logger.debug(
                                     "Modified file: %s (content hash changed)",
@@ -330,12 +344,7 @@ def index(
                                 )
                                 filtered_documents.append(doc)
                                 continue
-                            # Stored doc predates content hashing, or mixed
-                            # generations after a crash: fall through to mtime
-                            # only when there is no hash at all.
-                            if stored_hashes:
-                                filtered_documents.append(doc)
-                                continue
+                            # stored_hashes empty: pure legacy, fall through to mtime
 
                         # Fallback for legacy stored docs without content_hash:
                         # compare mtime (rounded to microseconds).
@@ -364,7 +373,14 @@ def index(
                 != hash_by_source.get(str(doc.metadata.get("source", "")))
             ]
 
-        leftover_stale_ids = stale_ids_for(cleanup_hashes)
+        def ids_for_sources(sources: set[str]) -> list[str]:
+            return [
+                doc.doc_id
+                for doc in existing_docs
+                if doc.doc_id and str(doc.metadata.get("source", "")) in sources
+            ]
+
+        leftover_stale_ids = stale_ids_for(cleanup_hashes) + ids_for_sources(empty_sources)
 
         if not all_documents:
             if leftover_stale_ids:
@@ -381,8 +397,9 @@ def index(
                         style="yellow",
                     )
                 else:
+                    n_cleaned = len(cleanup_hashes) + len(empty_sources)
                     console.print(
-                        f"✅ Removed leftover stale chunks for {len(cleanup_hashes)} files",
+                        f"✅ Removed leftover stale chunks for {n_cleaned} files",
                         style="green",
                     )
             else:
--- a/packages/gptme-rag/src/gptme_rag/cli.py
+++ b/packages/gptme-rag/src/gptme_rag/cli.py
@@ -248,9 +248,13 @@ def index(
         for doc in existing_docs:
             if "source" in doc.metadata:
                 abs_path = os.path.abspath(doc.metadata["source"])
-                entry = existing_files.setdefault(abs_path, {})
+                entry = existing_files.setdefault(abs_path, {"content_hashes": set()})
+                hashes = entry.setdefault("content_hashes", set())
                 if "content_hash" in doc.metadata:
-                    entry["content_hash"] = doc.metadata["content_hash"]
+                    if isinstance(hashes, set):
+                        hashes.add(doc.metadata["content_hash"])
+                else:
+                    entry["has_unhashed"] = True
                 last_modified = doc.metadata.get("last_modified")
                 if last_modified:
                     try:
@@ -296,9 +300,10 @@ def index(
                         # worktree churn) — skip it instead of re-embedding.
                         current_hash = doc.metadata.get("content_hash")
                         if current_hash is not None:
-                            stored_hash = existing.get("content_hash")
-                            if stored_hash is not None:
-                                if current_hash == stored_hash:
+                            stored_hashes = existing.get("content_hashes") or set()
+                            has_unhashed = bool(existing.get("has_unhashed"))
+                            if stored_hashes and not has_unhashed:
+                                if stored_hashes == {current_hash}:
                                     logger.debug("Unchanged file: %s", abs_source)
                                     continue
                                 logger.debug(
@@ -307,7 +312,12 @@ def index(
                                 )
                                 filtered_documents.append(doc)
                                 continue
-                            # Stored doc predates content hashing: fall through to mtime.
+                            # Stored doc predates content hashing, or mixed
+                            # generations after a crash: fall through to mtime
+                            # only when there is no hash at all.
+                            if stored_hashes:
+                                filtered_documents.append(doc)
+                                continue
 
                         # Fallback for legacy stored docs without content_hash:
                         # compare mtime (rounded to microseconds).
@@ -330,18 +340,29 @@ def index(
             console.print("No new or modified documents to index", style="yellow")
             return
 
-        # Then process them with a progress bar. Replace all chunks for a modified
-        # source before adding its new chunks; otherwise stale chunks survive when
-        # content or chunk boundaries change and make future fingerprint reads
-        # order-dependent.
+        # Then process them with a progress bar. Preserve the previous chunks until
+        # every replacement chunk has been embedded successfully: deleting first
+        # turns a transient embedding failure into data loss. New chunk IDs include
+        # content_hash, so old and replacement generations can coexist briefly;
+        # after a successful add, delete only the IDs from the old generation.
         sources = {str(doc.metadata.get("source", "")) for doc in all_documents}
         n_files = len(sources)
         n_chunks = len(all_documents)
 
         logger.info(f"Found {n_files} new/modified files to index ({n_chunks} chunks)")
 
-        for source in sources & existing_files.keys():
-            indexer.delete_documents({"source": source})
+        current_hash_by_source = {
+            str(doc.metadata.get("source", "")): doc.metadata.get("content_hash")
+            for doc in all_documents
+        }
+        stale_ids = [
+            doc.doc_id
+            for doc in existing_docs
+            if doc.doc_id
+            and str(doc.metadata.get("source", "")) in sources
+            and doc.metadata.get("content_hash")
+            != current_hash_by_source.get(str(doc.metadata.get("source", "")))
+        ]
 
         with tqdm(
             total=n_chunks,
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:
CI is green and the PR is mergeable. The self-merge gate is still blocked on the initial Greptile review, so I’m leaving this open for that review before the gated merge path runs.

TimeToBuildBob:
@greptileai review

packages/gptme-rag/src/gptme_rag/cli.py
```diff
@@ -240,20 +241,26 @@ def index(
...
+        # Map each indexed source to its stored change-detection fingerprint:
+        # content_hash (primary) plus last_modified (fallback for legacy docs that
+        # predate content hashing). Stored per-chunk but read once per source.
+        existing_files: dict[str, dict[str, object]] = {}
         for doc in existing_docs:
             if "source" in doc.metadata:
                 abs_path = os.path.abspath(doc.metadata["source"])
+                entry = existing_files.setdefault(abs_path, {})
+                if "content_hash" in doc.metadata:
+                    entry["content_hash"] = doc.metadata["content_hash"]
```
TimeToBuildBob: ❌ **P1** — When a file is re-indexed after a content change, its old chunks (with the previous content_hash) remain in the Chroma collection alongside the new chunks. The loop at lines 248-263 iterates over every existing document and overwrites entry["content_hash"] each time, so the final value is whatever chunk indexer.get_all_documents() happens to return last. ChromaDB does not guarantee insertion order; if an old chunk is returned last, existing.get("content_hash") holds the stale hash. On the next run, the current file's hash (the new one) differs, so the file is treated as modified again and re-embedded, adding yet more chunks. This repeats every run, causing unbounded index growth and repeated embedding work. The same flaw applies to last_modified in the fallback path. Fixing requires either deleting stale chunks for a source before re-adding, or choosing the entry with the latest last_modified when building the map.

packages/gptme-rag/src/gptme_rag/indexing/document.py
```diff
@@ -44,11 +45,17 @@ def from_file(
             Document instances, either a single document or multiple chunks
         """
         last_modified = datetime.fromtimestamp(path.stat().st_mtime)
+        # Content hash is the primary change-detection key: it survives mtime-only
+        # rewrites (git restore/checkout, worktree churn) that would otherwise force
+        # a full re-embed of an unchanged file. Falls back to mtime in cli.py when a
+        # stored document predates this field.
+        content_hash = hashlib.sha256(path.read_bytes()).hexdigest()
```
TimeToBuildBob: ⚠️ **P2** — The content_hash is computed from the raw file bytes (hashlib.sha256(path.read_bytes()).hexdigest() in Document.from_file), but the document actually indexed is the output of the DocumentProcessor (chunking, cleaning, etc.). If the processor's behavior changes between runs—e.g., a library upgrade alters chunking or cleaning rules—the same raw bytes now produce different chunks, yet the content_hash is identical. The index command then treats the file as unchanged (current_hash == stored_hash at cli.py:301) and skips re-embedding, leaving the index with stale chunks. This is a regression: previously, touching the file (mtime bump) forced a re-embed, so processor changes would be picked up; now touching is explicitly ignored. Users upgrading gptme-rag or changing chunk_size/chunk_overlap will silently get outdated search results unless they use --force-recreate.

packages/gptme-rag/tests/test_content_hash_changedetection.py
```diff
@@ -0,0 +1,108 @@
...
+def _run_index(runner, tmp_path, index_dir, sources):
+    """Run the index CLI against a temp persist dir. Returns CliResult."""
+    return runner.invoke(
+        cli,
+        [
+            "index",
+            "--persist-dir",
+            str(index_dir),
+            "--embedding-function",
+            "minilm",
```
TimeToBuildBob: ⚠️ **P2** — The tests test_index_skips_unchanged_mtime_only_rewrite and test_index_reembeds_on_content_change invoke the CLI with --embedding-function minilm (test file line 67). The Indexer initialized by the CLI will attempt to load the MINILM sentence-transformers model, which may not be pre-cached and requires a network download on first use. In a CI or offline environment, this causes the test to fail with a model download error, making the tests non-hermetic and environment-dependent. The test suite should stub the embedding function or ensure the model is available offline (e.g., by using a fake embedding function or pre-caching the model).

TimeToBuildBob:
## 🤖 AI code review

The PR replaces mtime-only change detection in the `index` command with a content hash (sha256 of file bytes). `Document.from_file` now stores `content_hash` in metadata, and `cli.py` uses it as the primary comparison key, falling back to mtime only for legacy stored docs without the field. It also adds logic to delete leftover stale chunks after a successful re-embed.

**❌ P1 high** — `packages/gptme-rag/src/gptme_rag/cli.py:334`

When a source has both a legacy chunk (no content_hash) and a hashed chunk that matches the current file content, the condition at this line always appends the document for re-embedding, even though the current generation is already fully indexed. This happens after a partial migration or crash leaves a legacy chunk alongside a new one. On the next run, `existing_files[abs_source]` has `content_hashes={current_hash}` and `has_unhashed=True`, so the earlier `if stored_hashes and not has_unhashed` branch is skipped and control reaches `if stored_hashes:`. The document is added to `all_documents`, and `add_documents_progress` attempts to insert chunks with IDs that include the identical content_hash, producing duplicate IDs (e.g. `abs@hash#chunk0-0`) and raising an error, which aborts the entire index command before the stale-deletion step can run. The result is that every subsequent index invocation fails on that source until the legacy chunk is manually removed. The correct behavior is to treat a stored hash that equals the current hash as already indexed even when unhashed chunks exist, adding the source to `cleanup_hashes` so the legacy chunks are deleted instead of re-embedding.

```suggestion
Instead of skipping whenever current_hash is present, only skip if the set of stored hashes equals exactly one hash (the current) and the expected chunk count matches, or always re-embed when more than one hash is present, then delete all other hashes.
```

<sub>**How this was verified:** Traced the flow: a source with one legacy chunk (no content_hash) and one chunk with content_hash equal to the file's current hash sets `has_unhashed=True` and `content_hashes={current_hash}`. The check `if stored_hashes and not has_unhashed` is False, so the code reaches this line and appends to filtered_documents. The subsequent add produces duplicate IDs because chunk IDs include the same hash.</sub>

**❌ P1 high** — `packages/gptme-rag/src/gptme_rag/indexing/document.py:82`

Empty or whitespace-only files now cause `Document.from_file` to return without yielding any documents (the `if not content.strip(): return` at line 82-83 in the processor branch). Consequently, during `index`, such a source never appears in `all_documents` and is never placed in `cleanup_hashes`, because those dictionaries are populated only from emitted documents. The cleanup paths that delete stale chunks—the `if not all_documents:` branch (which only inspects `leftover_stale_ids` derived from `cleanup_hashes`) and the post-add `stale_ids` deletion (which requires the source to be in `current_hash_by_source` from `all_documents`)—both skip this source. If a file that previously had indexed content is emptied, its old chunks remain in the index indefinitely, producing stale search results. This is a regression: the old code did not explicitly skip empty files, so a file that previously yielded a document (even an empty one) would have its old generation replaced.

```suggestion
Log a warning when the decode fails, e.g., logger.warning('Skipping non-UTF-8 file: %s', path), so the skip is visible.
```

<sub>**How this was verified:** Traced the index flow: when a file produces no documents, it is absent from `all_documents` and `cleanup_hashes`. The stop-gap cleanup at cli.py:367 only handles sources in `cleanup_hashes`, and the stale-deletion at cli.py:416 only handles sources in `all_documents`. Thus empty files leave stale chunks.</sub>

</details>

- `packages/gptme-rag/src/gptme_rag/cli.py` — Adds content_hash-based change detection with mtime fallback, plus stale-chunk deletion after re-embeds.
- `packages/gptme-rag/src/gptme_rag/indexing/document.py` — Reads file bytes once, computes sha256, stores as content_hash, and includes it in chunk IDs.
- `packages/gptme-rag/tests/test_content_hash_changedetection.py` — Adds tests for content-hash change detection, mtime-touch stability, and stale cleanup.

</details>

| `cece489321a3` | 3/5 | 3 | llm | 2026-08-28 05:33 UTC |
| `d27cc48fde2f` | 3/5 | 1 | llm | 2026-08-28 13:11 UTC |
| `e0dfe3f8c9d8` | 4/5 | 5 | llm | 2026-08-28 14:30 UTC |

</details>

<sub>Reviewed `c2742c9cdb4d` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 502s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

ErikBjare:
@TimeToBuildBob fix

TimeToBuildBob:
Fixed in `d27cc48f`: changed sources now replace their old chunks before indexing, and the CLI tests use Chroma's local default embedding instead of downloading MiniLM. I also added regression coverage for stale-chunk replacement and both legacy mtime fallback paths.

Verified locally: 8 targeted tests pass, ruff passes, and gptme-rag typechecking reports no issues. CI is running on the pushed commit.

I’m leaving processor-version invalidation out of this focused fix: changing chunk parameters already requires `--force-recreate`, and this PR addresses file-content change detection without introducing an index-schema/version migration.

packages/gptme-rag/src/gptme_rag/indexing/document.py
TimeToBuildBob: Accepted trade-off for this focused fix: the hash fingerprints source bytes, while processor/chunk configuration changes continue to require `--force-recreate`. Adding a processor schema/version fingerprint is a separate index-format migration and is not required to stop mtime-only re-embeds.

packages/gptme-rag/src/gptme_rag/cli.py
TimeToBuildBob: Fixed in `d27cc48f`: each modified source is deleted before its replacement chunks are added. `test_index_replaces_stale_chunks_on_content_change` verifies the old chunk set and hash do not survive.

packages/gptme-rag/tests/test_content_hash_changedetection.py
TimeToBuildBob: Fixed in `d27cc48f`: the CLI tests now use Chroma's local default embedding, so they do not load or download a sentence-transformer model.

packages/gptme-rag/src/gptme_rag/cli.py
```diff
@@ -301,12 +330,19 @@ def index(
...
+        # content or chunk boundaries change and make future fingerprint reads
+        # order-dependent.
+        sources = {str(doc.metadata.get("source", "")) for doc in all_documents}
+        n_files = len(sources)
         n_chunks = len(all_documents)
 
         logger.info(f"Found {n_files} new/modified files to index ({n_chunks} chunks)")
 
+        for source in sources & existing_files.keys():
+            indexer.delete_documents({"source": source})
```
TimeToBuildBob: ❌ **P1** — The index command deletes all existing chunks for a source before adding its new chunks. If adding fails (e.g., embedding model unavailable, network error, or any exception in `indexer.add_documents_progress`), the source is left with no documents in the collection, losing previously indexed data. This is a data-loss risk introduced by this change; previously old chunks were kept and duplicates appeared. The deletion and addition are not atomic, and there is no rollback. Suggest adding new chunks first and then deleting old ones (e.g., by comparing doc IDs) or performing deletion only after successful addition. _(3/3 passes)_

TimeToBuildBob:
The P1 from the AI review was still live after `d27cc48f`: that commit still deleted a source's chunks *before* embedding replacements, so a failed add left the source empty.

Fixed in `e0dfe3f8`:
- replacement chunks are added first (IDs include `content_hash`, so generations do not collide)
- after a successful add, only IDs whose stored hash is not the current one are deleted
- mixed leftover generations after a crash are treated as modified, not skipped

Regression coverage: failed replacement add keeps the old generation; retry after a simulated add-then-crash does not wipe the replacement. 10 targeted tests pass locally.

packages/gptme-rag/src/gptme_rag/cli.py
```diff
@@ -277,19 +288,47 @@ def index(
...
+
+                        # Primary key: content hash. An unchanged hash means the file
+                        # is unchanged even if mtime moved (git restore/checkout,
+                        # worktree churn) — skip it instead of re-embedding.
+                        current_hash = doc.metadata.get("content_hash")
+                        if current_hash is not None:
+                            stored_hashes = existing.get("content_hashes") or set()
+                            has_unhashed = bool(existing.get("has_unhashed"))
+                            if stored_hashes and not has_unhashed:
+                                if stored_hashes == {current_hash}:
```
TimeToBuildBob: ⚠️ **P2** — The change detection now keys solely on the file's content hash (sha256 of the whole file) when a stored hash exists. This makes it impossible to re-chunk a file with unchanged content by touching it or changing chunking parameters (--chunk-size, --chunk-overlap, or the processor version). Previously, an mtime bump (e.g., via os.utime) would trigger a re-embed; that workaround no longer works because the hash is unchanged, so the file is skipped as "Unchanged file". A user who changes the chunk size and expects the index to reflect the new boundaries will silently keep the old chunks. The only way to force re-chunking is --force-recreate (which wipes the entire index) or editing the file content. The issue is at the skip condition where stored_hashes == {current_hash} leads to `continue` without considering whether chunking settings changed.

packages/gptme-rag/src/gptme_rag/cli.py
```diff
@@ -240,20 +241,30 @@ def index(
...
+                    if isinstance(hashes, set):
+                        hashes.add(doc.metadata["content_hash"])
+                else:
+                    entry["has_unhashed"] = True
                 last_modified = doc.metadata.get("last_modified")
                 if last_modified:
                     try:
                         # Parse ISO format timestamp to float
-                        existing_files[abs_path] = datetime.fromisoformat(last_modified).timestamp()
+                        entry["last_modified"] = datetime.fromisoformat(last_modified).timestamp()
```
TimeToBuildBob: ⚠️ **P2** — In `cli.py`, the `existing_files` loop overwrites `entry["last_modified"]` for every chunk of a source, so if a source has multiple chunks and any one of them has an invalid `last_modified` (the `ValueError` path sets it to 0), and that chunk is processed last, the final `last_modified` becomes 0. For legacy sources (those without any `content_hash`), the mtime fallback then compares `round(current_mtime,6) > round(stored_mtime,6)` against 0, so any file with a non-epoch mtime is treated as modified on every run, causing an endless re-embed loop. This is a real regression for mixed or corrupted legacy index states, though it requires an invalid `last_modified` on one chunk of an otherwise valid source.

packages/gptme-rag/src/gptme_rag/indexing/document.py
TimeToBuildBob: ⚠️ **P2** — In `document.py`, the content hash is computed from the file bytes via `path.read_bytes()`, but later the same file is read again with `path.read_text()` (or `processor.process_file(path)` re-reads it). If the file is modified between the two reads, the stored `content_hash` will not correspond to the chunk text, and the index will contain a document whose metadata claims a hash that does not match the actual content. A subsequent run will compare the on-disk hash (now different from the stale stored hash) and re-embed, but the damage is that a query may return the file's content that does not match its hash during the interval. This is a race condition, but it is newly introduced because the hash read was added without a single unified read.

packages/gptme-rag/tests/test_content_hash_changedetection.py
```diff
@@ -0,0 +1,359 @@
...
+    f.write_text("version two content")
+
+    import chromadb
+
+    chromadb.api.shared_system_client.SharedSystemClient._identifier_to_system.clear()
+    original_add = chromadb.Collection.add
+    replacement_hash = hashlib.sha256(b"version two content").hexdigest()
+
+    def fail_replacement_add(self, *args, **kwargs):
+        metadatas = kwargs.get("metadatas") or []
```
TimeToBuildBob: ⚠️ **P2** — The mock for `fail_replacement_add` only inspects `kwargs` for `metadatas`; if the indexer invokes `chromadb.Collection.add` with `metadatas` as a positional argument (the API allows `add(ids, embeddings, metadatas, documents)`), the mock never sees it and does not raise. The replacement chunks would then be added successfully, and the assertion that only the old hash remains would fail. This makes the test non-hermetic: its pass/fail depends on whether the indexer uses keyword arguments. The fix is to also inspect positional args, e.g., `metadatas = kwargs.get("metadatas") or (args[2] if len(args) > 2 else [])`.

packages/gptme-rag/src/gptme_rag/cli.py
TimeToBuildBob: Accepted trade-off for this focused fix: content-hash fingerprints source bytes, so a chunk-size/overlap or processor-version change still needs `--force-recreate`. Folding chunking parameters into the hash is an index-format migration and is not required to stop mtime-only re-embeds.

packages/gptme-rag/src/gptme_rag/cli.py
TimeToBuildBob: Fixed in `8ba677d3`: the existing-files loop now keeps the first valid `last_modified` and ignores a later invalid timestamp, so one corrupt chunk cannot zero the stored mtime. `test_index_legacy_mtime_fallback_ignores_invalid_chunk_timestamp` covers the poisoned last-chunk case.

packages/gptme-rag/src/gptme_rag/indexing/document.py
TimeToBuildBob: Fixed in `8ba677d3`: `from_file` reads the file once, hashes those bytes, and chunks the same decoded text via `process_text`, so the stored hash and indexed content cannot diverge.

packages/gptme-rag/tests/test_content_hash_changedetection.py
TimeToBuildBob: Fixed in `8ba677d3`: `fail_replacement_add` now reads `metadatas` from kwargs or positional args (`args[2]`), so the simulated add failure still fires if Chroma is invoked positionally.

packages/gptme-rag/src/gptme_rag/cli.py
TimeToBuildBob: Replacement chunks are added first (`c4b8c4dc`); if a later stale-delete fails, the command now warns instead of hiding the successful add (`8ba677d3`). A retry that already has the current hash deletes leftover older IDs without re-embedding (`c2742c9c`), so mixed generations cannot UniqueConstraintError on the same IDs.

TimeToBuildBob:
Pushed the remaining AI-review P2s (and the leftover-generation P1 from the in-band pass) in `8ba677d3` and `c2742c9c`, rebased onto current master so `collect_documents` still receives the #1532 `pattern` argument.

- Hash and indexed text now come from one file read
- Legacy mtime fallback keeps a valid timestamp when one chunk is corrupt
- Stale-delete failures after a successful add warn and self-heal on the next run without re-embedding the current generation
- The replacement-add failure mock inspects positional `metadatas` too

Processor/chunk-parameter invalidation remains an accepted trade-off for this PR (`--force-recreate`). 13 targeted tests passed locally. CI is running on `c2742c9c`.

packages/gptme-rag/src/gptme_rag/cli.py
```diff
@@ -277,36 +297,113 @@ def index(
...
+                                logger.debug(
+                                    "Modified file: %s (content hash changed)",
+                                    abs_source,
+                                )
+                                filtered_documents.append(doc)
+                                continue
+                            # Stored doc predates content hashing, or mixed
+                            # generations after a crash: fall through to mtime
+                            # only when there is no hash at all.
+                            if stored_hashes:
```
TimeToBuildBob: ❌ **P1** — When a source has both a legacy chunk (no content_hash) and a hashed chunk that matches the current file content, the condition at this line always appends the document for re-embedding, even though the current generation is already fully indexed. This happens after a partial migration or crash leaves a legacy chunk alongside a new one. On the next run, `existing_files[abs_source]` has `content_hashes={current_hash}` and `has_unhashed=True`, so the earlier `if stored_hashes and not has_unhashed` branch is skipped and control reaches `if stored_hashes:`. The document is added to `all_documents`, and `add_documents_progress` attempts to insert chunks with IDs that include the identical content_hash, producing duplicate IDs (e.g. `abs@hash#chunk0-0`) and raising an error, which aborts the entire index command before the stale-deletion step can run. The result is that every subsequent index invocation fails on that source until the legacy chunk is manually removed. The correct behavior is to treat a stored hash that equals the current hash as already indexed even when unhashed chunks exist, adding the source to `cleanup_hashes` so the legacy chunks are deleted instead of re-embedding.

packages/gptme-rag/src/gptme_rag/indexing/document.py
```diff
@@ -44,26 +45,53 @@ def from_file(
...
-            for chunk in processor.process_file(path):
+            # Skip the same files process_file would skip, but chunk the bytes we
+            # already hashed instead of re-reading the path.
+            if processor.is_binary_file(path):
+                return
+            try:
+                content = file_bytes.decode()
+            except UnicodeDecodeError:
+                return
+            if not content.strip():
```
TimeToBuildBob, line 82: ❌ **P1** — Empty or whitespace-only files now cause `Document.from_file` to return without yielding any documents (the `if not content.strip(): return` at line 82-83 in the processor branch). Consequently, during `index`, such a source never appears in `all_documents` and is never placed in `cleanup_hashes`, because those dictionaries are populated only from emitted documents. The cleanup paths that delete stale chunks—the `if not all_documents:` branch (which only inspects `leftover_stale_ids` derived from `cleanup_hashes`) and the post-add `stale_ids` deletion (which requires the source to be in `current_hash_by_source` from `all_documents`)—both skip this source. If a file that previously had indexed content is emptied, its old chunks remain in the index indefinitely, producing stale search results. This is a regression: the old code did not explicitly skip empty files, so a file that previously yielded a document (even an empty one) would have its old generation replaced.

packages/gptme-rag/src/gptme_rag/cli.py
TimeToBuildBob: Fixed in `6d023c3b`: if the current hash is already stored, skip re-embed even when leftover unhashed/legacy chunks remain, and delete those leftovers instead of inserting duplicate IDs. `test_index_mixed_legacy_and_current_hash_cleans_legacy_without_reembed` covers it.

packages/gptme-rag/src/gptme_rag/indexing/document.py
TimeToBuildBob: Fixed in `6d023c3b`: empty/whitespace files yield a tombstone so `index` can drop leftover chunks instead of leaving them searchable. New empty files are still not indexed. `test_index_emptied_file_deletes_stale_chunks` covers it.

TimeToBuildBob:
Pushed `6d023c3b` for the two remaining AI-review P1s:

- Mixed leftover unhashed + current-hash chunks now skip re-embed and delete the leftovers (no duplicate-ID abort).
- Emptying a previously indexed file now drops its leftover chunks instead of leaving them searchable.

15 targeted tests passed locally. CI is running on the new head; not merging until the review gate is current.

TimeToBuildBob:
CI is green on `6d023c3b` and the PR is mergeable. Self-merge is blocked: Greptile never arrived (helper refuses a second initial trigger), and the self-hosted AI review is stale vs this head.

A fresh review cannot post until the OpenRouter review-key daily limit resets at 2026-08-29T00:00Z. After that: `ai-review-now.sh gptme/gptme-contrib 1534`, then `self-merge-if-eligible`. No code change this session.

TimeToBuildBob:
Merged (`b517b5b9`). Cancelling the post-quota re-review / self-merge wait. No further action on this PR.

PR #1535, opened 2026-08-28
fix(rag): do not recreate a matching-model collection on EF conflict

TimeToBuildBob:
## Problem

`Indexer.__init__` treated any `get_collection()` failure as "collection missing" when `allow_recreate=True` (the `index` command default). Chroma can raise on embedding-function *object identity* even when `stored_model == current_model`. That path deleted Bob's 84k-chunk ambient-memory index and left ~42 orphan UUID dirs.

Logs made it worse: `--force-recreate` and a real model mismatch shared one line (`Model mismatch (stored: X, current: X) or force recreate`), so operators chased a comparison bug that was often just a requested wipe.

## Fix

- On `get_collection()` failure, peek **without** an embedding function. If the collection exists and `stored_model == current_model`, preserve it instead of recreating.
- Split the log: `Force recreate requested` vs `Model mismatch`.
- Add `gptme-rag gc-orphans` (dry-run by default, `--apply` to delete) for leftover Chroma UUID dirs not listed in `segments`/`collections`.

## Tests

`tests/test_same_model_recreate_guard.py`:
- matching-model EF conflict does not wipe
- force-recreate does not log "Model mismatch"
- GC dry-run vs `--apply`

Existing recreate/wipe tests still pass. `make typecheck` clean.

## Out of scope

- Does **not** run GC against a live persist dir (Bob's catch-up is still writing).
- Content-hash change detection is gptme-contrib#1534 (still open). This PR is the recreate-guard half of the same incident.

Step 3.4 of `vitals-regen-index-cpu-burn-loop`.

### DIFF
--- a/packages/gptme-rag/src/gptme_rag/cli.py
+++ b/packages/gptme-rag/src/gptme_rag/cli.py
@@ -17,6 +17,7 @@ from tqdm import tqdm
 
 from .benchmark import RagBenchmark
 from .indexing.document import Document
+from .indexing.gc import gc_orphan_segment_dirs
 from .indexing.indexer import Indexer
 from .indexing.watcher import FileWatcher
 from .query.context_assembler import ContextAssembler
@@ -827,6 +828,37 @@ def watch(
     except Exception as e:
         console.print(f"❌ Error watching directory: {e}", style="red")
         console.print_exception()
+
+
+@cli.command("gc-orphans")
+@click.option(
+    "--persist-dir",
+    type=click.Path(path_type=Path),
+    default=default_persist_dir,
+    help="Directory to persist the index",
+)
+@click.option(
+    "--apply",
+    is_flag=True,
+    help="Actually delete orphan UUID dirs. Default is dry-run.",
+)
+def gc_orphans(persist_dir: Path, apply: bool):
+    """Remove leftover Chroma segment dirs not listed in chroma.sqlite3.
+
+    Recreating a collection leaves UUID-named HNSW folders on disk. This
+    command lists them (dry-run) or deletes them (--apply). Live segment
+    and collection dirs are never touched.
+    """
+    orphans = gc_orphan_segment_dirs(persist_dir, apply=apply)
+    if not orphans:
+        console.print("No orphan segment dirs", style="green")
+        return
+    mode = "Removed" if apply else "Would remove"
+    console.print(f"{mode} {len(orphans)} orphan dir(s):")
+    for path in orphans:
+        console.print(f"  {path.name}")
+    if not apply:
+        console.print("Dry-run. Pass --apply to delete.", style="yellow")
 
 
 @cli.command()
--- a/packages/gptme-rag/src/gptme_rag/cli.py
+++ b/packages/gptme-rag/src/gptme_rag/cli.py
@@ -17,7 +17,7 @@ from tqdm import tqdm
 
 from .benchmark import RagBenchmark
 from .indexing.document import Document
-from .indexing.gc import gc_orphan_segment_dirs
+from .indexing.gc import ChromaCatalogError, gc_orphan_segment_dirs
 from .indexing.indexer import Indexer
 from .indexing.watcher import FileWatcher
 from .query.context_assembler import ContextAssembler
@@ -847,9 +847,14 @@ def gc_orphans(persist_dir: Path, apply: bool):
 
     Recreating a collection leaves UUID-named HNSW folders on disk. This
     command lists them (dry-run) or deletes them (--apply). Live segment
-    and collection dirs are never touched.
+    and collection dirs are never touched. --apply refuses if an indexer
+    currently holds the persist-dir writer lock.
     """
-    orphans = gc_orphan_segment_dirs(persist_dir, apply=apply)
+    try:
+        orphans = gc_orphan_segment_dirs(persist_dir, apply=apply)
+    except (ChromaCatalogError, RuntimeError) as exc:
+        console.print(f"❌ {exc}", style="red")
+        raise SystemExit(1) from exc
     if not orphans:
         console.print("No orphan segment dirs", style="green")
         return
--- a/packages/gptme-rag/src/gptme_rag/indexing/gc.py
+++ b/packages/gptme-rag/src/gptme_rag/indexing/gc.py
@@ -8,14 +8,22 @@ of them after a same-model recreate loop).
 Only UUID-named directories whose names are absent from both ``segments.id``
 and ``collections.id`` are candidates. ``chroma.sqlite3`` and live segment
 dirs are never touched.
+
+``--apply`` takes an exclusive writer lock (the same lock Indexer holds
+around collection writes). If an indexer is mid-write, apply refuses rather
+than deleting a segment dir Chroma created but has not yet catalogued.
 """
 
 from __future__ import annotations
 
+import fcntl
 import logging
+import os
 import re
 import shutil
 import sqlite3
+from collections.abc import Iterator
+from contextlib import contextmanager
 from pathlib import Path
 
 logger = logging.getLogger(__name__)
@@ -24,12 +32,50 @@ _UUID_DIR = re.compile(
     r"^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$",
     re.IGNORECASE,
 )
+_WRITER_LOCK_NAME = ".gptme-rag-writer.lock"
+
+
+class ChromaCatalogError(RuntimeError):
+    """chroma.sqlite3 is unreadable; GC must not report a clean persist dir."""
+
+
+@contextmanager
+def index_writer_lock(persist_dir: Path, *, blocking: bool = True) -> Iterator[None]:
+    """Exclusive lock around Chroma persist-dir writes.
+
+    Indexer holds this while adding/deleting/recreating collections. GC
+    ``apply`` takes it non-blocking and refuses if a writer is active — Chroma
+    creates a segment directory before inserting its id into sqlite, so a GC
+    snapshot taken in that window would treat a live dir as an orphan.
+    """
+    persist_dir = Path(persist_dir)
+    persist_dir.mkdir(parents=True, exist_ok=True)
+    fd = os.open(persist_dir / _WRITER_LOCK_NAME, os.O_CREAT | os.O_RDWR, 0o644)
+    flags = fcntl.LOCK_EX if blocking else (fcntl.LOCK_EX | fcntl.LOCK_NB)
+    try:
+        try:
+            fcntl.flock(fd, flags)
+        except BlockingIOError as exc:
+            raise RuntimeError(
+                f"index writer is active in {persist_dir}; refusing to apply GC"
+            ) from exc
+        yield
+    finally:
+        try:
+            fcntl.flock(fd, fcntl.LOCK_UN)
+        finally:
+            os.close(fd)
 
 
 def _live_chroma_ids(db: Path) -> set[str]:
-    """IDs that still belong to the current Chroma catalog."""
-    con = sqlite3.connect(f"file:{db}?mode=ro", uri=True)
+    """IDs that still belong to the current Chroma catalog.
+
+    Connect by filesystem path (not a ``file:`` URI) so a persist dir whose
+    path contains ``?`` is not parsed as a sqlite query string.
+    """
+    con = sqlite3.connect(str(db))
     try:
+        con.execute("PRAGMA query_only=ON")
         tables = {
             row[0] for row in con.execute("SELECT name FROM sqlite_master WHERE type='table'")
         }
@@ -43,23 +89,7 @@ def _live_chroma_ids(db: Path) -> set[str]:
         con.close()
 
 
-def gc_orphan_segment_dirs(persist_dir: Path, *, apply: bool = False) -> list[Path]:
-    """Return orphan UUID dirs; delete them when ``apply`` is True.
-
-    Dry-run (``apply=False``) is the default. If the sqlite catalog cannot be
-    read, return an empty list rather than deleting anything.
-    """
-    persist_dir = Path(persist_dir)
-    db = persist_dir / "chroma.sqlite3"
-    if not persist_dir.is_dir() or not db.is_file():
-        return []
-
-    try:
-        live_ids = _live_chroma_ids(db)
-    except sqlite3.Error as exc:
-        logger.warning("Cannot read %s (%s); refusing to GC", db, exc)
-        return []
-
+def _collect_orphans(persist_dir: Path, live_ids: set[str]) -> list[Path]:
     orphans: list[Path] = []
     for child in persist_dir.iterdir():
         if not child.is_dir():
@@ -69,10 +99,40 @@ def gc_orphan_segment_dirs(persist_dir: Path, *, apply: bool = False) -> list[Pa
         if child.name in live_ids:
             continue
         orphans.append(child)
-
-    if apply:
-        for path in orphans:
-            shutil.rmtree(path)
-            logger.info("Removed orphan segment dir %s", path)
-
     return orphans
+
+
+def _delete_orphans(orphans: list[Path]) -> list[Path]:
+    removed: list[Path] = []
+    for path in orphans:
+        try:
+            shutil.rmtree(path)
+        except OSError as exc:
+            logger.warning("Failed to remove %s: %s", path, exc)
+            continue
+        logger.info("Removed orphan segment dir %s", path)
+        removed.append(path)
+    return removed
+
+
+def gc_orphan_segment_dirs(persist_dir: Path, *, apply: bool = False) -> list[Path]:
+    """Return orphan UUID dirs; delete them when ``apply`` is True.
+
+    Dry-run (``apply=False``) is the default. Raises ``ChromaCatalogError`` if
+    the sqlite catalog cannot be read, rather than pretending the dir is clean.
+    ``apply=True`` raises ``RuntimeError`` if an index writer currently holds
+    the persist-dir lock.
+    """
+    persist_dir = Path(persist_dir)
+    db = persist_dir / "chroma.sqlite3"
+    if not persist_dir.is_dir() or not db.is_file():
+        return []
+
+    try:
+        if apply:
+            with index_writer_lock(persist_dir, blocking=False):
+                orphans = _collect_orphans(persist_dir, _live_chroma_ids(db))
+                return _delete_orphans(orphans)
+        return _collect_orphans(persist_dir, _live_chroma_ids(db))
+    except sqlite3.Error as exc:
+        raise ChromaCatalogError(f"Cannot read {db}: {exc}") from exc
--- a/packages/gptme-rag/src/gptme_rag/indexing/gc.py
+++ b/packages/gptme-rag/src/gptme_rag/indexing/gc.py
@@ -36,7 +36,7 @@ _WRITER_LOCK_NAME = ".gptme-rag-writer.lock"
 
 
 class ChromaCatalogError(RuntimeError):
-    """chroma.sqlite3 is unreadable; GC must not report a clean persist dir."""
+    """Persist dir or chroma.sqlite3 is unreadable; GC must not report a clean dir."""
 
 
 @contextmanager
@@ -119,16 +119,15 @@ def gc_orphan_segment_dirs(persist_dir: Path, *, apply: bool = False) -> list[Pa
     """Return orphan UUID dirs; delete them when ``apply`` is True.
 
     Dry-run (``apply=False``) is the default. Raises ``ChromaCatalogError`` if
-    the sqlite catalog cannot be read, rather than pretending the dir is clean.
-    ``apply=True`` raises ``RuntimeError`` if an index writer currently holds
-    the persist-dir lock.
+    the sqlite catalog or persist dir cannot be read, rather than pretending
+    the dir is clean. ``apply=True`` raises ``RuntimeError`` if an index writer
+    currently holds the persist-dir lock.
     """
     persist_dir = Path(persist_dir)
     db = persist_dir / "chroma.sqlite3"
-    if not persist_dir.is_dir() or not db.is_file():
-        return []
-
     try:
+        if not persist_dir.is_dir() or not db.is_file():
+            return []
         if apply:
             with index_writer_lock(persist_dir, blocking=False):
                 orphans = _collect_orphans(persist_dir, _live_chroma_ids(db))
@@ -136,3 +135,5 @@ def gc_orphan_segment_dirs(persist_dir: Path, *, apply: bool = False) -> list[Pa
         return _collect_orphans(persist_dir, _live_chroma_ids(db))
     except sqlite3.Error as exc:
         raise ChromaCatalogError(f"Cannot read {db}: {exc}") from exc
+    except OSError as exc:
+        raise ChromaCatalogError(f"Cannot access {persist_dir}: {exc}") from exc
--- a/packages/gptme-rag/src/gptme_rag/indexing/gc.py
+++ b/packages/gptme-rag/src/gptme_rag/indexing/gc.py
@@ -81,22 +81,33 @@ def _live_chroma_ids(db: Path) -> set[str]:
         }
         ids: set[str] = set()
         if "segments" in tables:
-            ids.update(row[0] for row in con.execute("SELECT id FROM segments") if row[0])
+            ids.update(
+                str(row[0]).lower() for row in con.execute("SELECT id FROM segments") if row[0]
+            )
         if "collections" in tables:
-            ids.update(row[0] for row in con.execute("SELECT id FROM collections") if row[0])
+            ids.update(
+                str(row[0]).lower() for row in con.execute("SELECT id FROM collections") if row[0]
+            )
         return ids
     finally:
         con.close()
 
 
 def _collect_orphans(persist_dir: Path, live_ids: set[str]) -> list[Path]:
+    """UUID hex is case-insensitive (RFC 4122); compare folded.
+
+    ``_UUID_DIR`` is IGNORECASE so an uppercase dir is a candidate. Exact-case
+    membership against sqlite's canonical lowercase ``segments.id`` would
+    classify that live dir as an orphan and ``--apply`` would ``rmtree`` it.
+    """
     orphans: list[Path] = []
+    live = {i.lower() for i in live_ids}
     for child in persist_dir.iterdir():
         if not child.is_dir():
             continue
         if not _UUID_DIR.fullmatch(child.name):
             continue
-        if child.name in live_ids:
+        if child.name.lower() in live:
             continue
         orphans.append(child)
     return orphans
--- a/packages/gptme-rag/src/gptme_rag/indexing/gc.py
+++ b/packages/gptme-rag/src/gptme_rag/indexing/gc.py
@@ -1,0 +1,78 @@
+"""Garbage-collect orphaned Chroma persist-dir segment folders.
+
+Each ``delete_collection`` / recreate leaves UUID-named HNSW directories on
+disk that are no longer listed in ``chroma.sqlite3``. They are not used by
+search or index, but they accumulate (Bob's ambient-memory persist dir had 42
+of them after a same-model recreate loop).
+
+Only UUID-named directories whose names are absent from both ``segments.id``
+and ``collections.id`` are candidates. ``chroma.sqlite3`` and live segment
+dirs are never touched.
+"""
+
+from __future__ import annotations
+
+import logging
+import re
+import shutil
+import sqlite3
+from pathlib import Path
+
+logger = logging.getLogger(__name__)
+
+_UUID_DIR = re.compile(
+    r"^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$",
+    re.IGNORECASE,
+)
+
+
+def _live_chroma_ids(db: Path) -> set[str]:
+    """IDs that still belong to the current Chroma catalog."""
+    con = sqlite3.connect(f"file:{db}?mode=ro", uri=True)
+    try:
+        tables = {
+            row[0] for row in con.execute("SELECT name FROM sqlite_master WHERE type='table'")
+        }
+        ids: set[str] = set()
+        if "segments" in tables:
+            ids.update(row[0] for row in con.execute("SELECT id FROM segments") if row[0])
+        if "collections" in tables:
+            ids.update(row[0] for row in con.execute("SELECT id FROM collections") if row[0])
+        return ids
+    finally:
+        con.close()
+
+
+def gc_orphan_segment_dirs(persist_dir: Path, *, apply: bool = False) -> list[Path]:
+    """Return orphan UUID dirs; delete them when ``apply`` is True.
+
+    Dry-run (``apply=False``) is the default. If the sqlite catalog cannot be
+    read, return an empty list rather than deleting anything.
+    """
+    persist_dir = Path(persist_dir)
+    db = persist_dir / "chroma.sqlite3"
+    if not persist_dir.is_dir() or not db.is_file():
+        return []
+
+    try:
+        live_ids = _live_chroma_ids(db)
+    except sqlite3.Error as exc:
+        logger.warning("Cannot read %s (%s); refusing to GC", db, exc)
+        return []
+
+    orphans: list[Path] = []
+    for child in persist_dir.iterdir():
+        if not child.is_dir():
+            continue
+        if not _UUID_DIR.fullmatch(child.name):
+            continue
+        if child.name in live_ids:
+            continue
+        orphans.append(child)
+
+    if apply:
+        for path in orphans:
+            shutil.rmtree(path)
+            logger.info("Removed orphan segment dir %s", path)
+
+    return orphans
--- a/packages/gptme-rag/src/gptme_rag/indexing/indexer.py
+++ b/packages/gptme-rag/src/gptme_rag/indexing/indexer.py
@@ -490,9 +490,17 @@ class Indexer:
                     )
                     try:
                         self.collection = self.client.get_collection(name=collection_name)
-                    except Exception:
-                        # Collection truly doesn't exist yet — create it normally.
-                        need_recreate = True
+                    except Exception as peek_err:
+                        # Missing *or* temporarily unreadable (sqlite lock
+                        # during a concurrent write). Creating would be a write;
+                        # the later need_recreate block deletes first, so a
+                        # transient peek failure would wipe a live collection
+                        # from search/status. Fail loud instead.
+                        raise RuntimeError(
+                            f"Collection {collection_name!r} could not be opened "
+                            "and recreation is disabled for this command; not "
+                            f"creating or deleting it. Error: {peek_err}"
+                        ) from peek_err
                     else:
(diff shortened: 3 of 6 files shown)


TimeToBuildBob:
CI is green and the PR is mergeable. Self-merge is not eligible yet because the required Greptile review has not arrived; waiting for that review before rerunning the merge gate.

packages/gptme-rag/src/gptme_rag/indexing/indexer.py
```diff
@@ -496,9 +501,32 @@ def __init__(
...
+                            stored = (peek.metadata or {}).get("embedding_model", "unknown")
+                            if stored == current_model:
+                                logger.warning(
+                                    "Collection %r exists with matching model %r but could not "
+                                    "be rebound (%s); preserving it instead of recreating",
+                                    collection_name,
+                                    stored,
+                                    e,
+                                )
+                                self.collection = peek
```
TimeToBuildBob: ❌ **P1** — When get_collection() raises an EF-object conflict on a collection whose stored_model equals current_model, the new code assigns the EF-less handle returned by `peek = self.client.get_collection(name=collection_name)` to `self.collection`. This handle has no embedding function bound. Later, add_documents/_add_documents call `self.collection.add(documents=..., metadatas=..., ids=...)` without embeddings, so Chroma falls back to its default embedding function. For a collection indexed with a non-default model (e.g., modernbert), the default EF has a different dimension (typically 384 for MiniLM) than the stored vectors (768 for ModernBERT), causing an InvalidDimensionException on the next index run. The preserved collection therefore becomes unusable for updates. The test only checks count(), which does not exercise embedding, so this is not caught.

TimeToBuildBob:
## 🤖 AI code review

This PR adds a gc-orphans CLI command to remove orphan Chroma segment directories, introduces a flock-based writer lock around collection mutations, refactors Indexer.__init__ to preserve matching-model collections on embedding-function conflicts, and skips persist-dir artifacts during indexing and watching. New tests cover the recreate guards, the GC command, and persist-dir file filtering.

<b>5 findings</b> · ❌ **1** P1 · ⚠️ **4** P2 · 🔒 **2** security — 🛑 **1 at P0/P1**

<details><summary><b>3 inline</b></summary>

**⚠️ P2 medium · 🔒 security · ℹ️ non-blocking** — `packages/gptme-rag/src/gptme_rag/indexing/gc.py:140`

In `gc_orphans` CLI, the dry-run default returns an empty list when the persist dir or `chroma.sqlite3` is missing, printing 'No orphan segment dirs'. However, if the persist dir exists but `chroma.sqlite3` is not a file (e.g., a directory or a broken symlink), the code returns `[]` without distinguishing that from a clean dir. This is not a data-loss risk, but it can mask a corrupt catalog: a user running `gc-orphans` would see 'clean' and not notice that the catalog is unreadable. The function already raises `ChromaCatalogError` for sqlite errors, but this early return bypasses that. The CLI should probably report an error when `persist_dir` exists but `db` is not a regular file, or at least log a warning.

```suggestion
if path.is_symlink() or not path.is_dir():
                logger.warning("Skipping non-directory or symlink: %s", path)
                continue
            shutil.rmtree(path)
```

<sub>**How this was verified:** The code at line 140-141 returns an empty list before any sqlite access. If `persist_dir` is a directory but `chroma.sqlite3` is a directory or non-regular file, this branch is taken. The CLI then prints 'No orphan segment dirs' and exits 0, even though the catalog is unusable.</sub>

**⚠️ P2 medium · ℹ️ non-blocking** — `packages/gptme-rag/src/gptme_rag/indexing/indexer.py:631`

_is_persist_artifact returns True for any file whose name equals the writer-lock filename, regardless of whether that file is actually inside the persist directory. If a user has a legitimate corpus file named exactly '.gptme-rag-writer.lock', it will be silently excluded from indexing and watching. This is a behavior change: before this PR, such a file would have been indexed normally. The check should be scoped to files inside the persist dir, not just by name.

```suggestion
if persist is None:
            return False
        try:
            persist_resolved = Path(persist).resolve()
            resolved = path.resolve()
        except OSError:
            return False
        if path.name == _WRITER_LOCK_NAME and (resolved == persist_resolved or persist_resolved in resolved.parents):
            return True
```

<sub>**How this was verified:** Examined the method: after the name check, it verifies the path is inside persist_directory. The name-only check is too broad and applies even when persist_directory is None (the method returns True before checking that).</sub>

**⚠️ P2 medium · ℹ️ non-blocking** — `packages/gptme-rag/tests/test_file_limit.py:45`

In `test_get_valid_files_skips_persist_dir_artifacts`, the assertion `".gptme-rag-writer.lock" not in names` passes vacuously. The lock file is only created when a write lock is acquired (e.g., during a recreate or add). The test's `Indexer` is initialized with a fresh persist_dir and no recreate occurs, so the lock file is never created; thus the assertion is true regardless of whether the code correctly filters the lock file. The test gives false confidence that the lock file is skipped. To actually pin the regression, the test should create the lock file (e.g., by acquiring the lock) or assert on a file that is actually present inside the persist_dir.

```suggestion
Create the lock file before calling `_get_valid_files` (e.g., by acquiring and releasing the lock) to make the assertion meaningful.
```

<sub>**How this was verified:** Checked the Indexer init: it does not create the lock file. The lock is only created in `index_writer_lock` when called from write operations, which the test does not trigger.</sub>

</details>

### Comments outside the diff

**❌ P1 high · 🔒 security** — [`packages/gptme-rag/src/gptme_rag/indexing/indexer.py:511`](https://github.com/gptme/gptme-contrib/blob/6061fb2f1db5d732adb39581c57f55d464c4aedc/packages/gptme-rag/src/gptme_rag/indexing/indexer.py#L511)

In the `elif not allow_recreate:` branch of `Indexer.__init__`, when the initial `get_collection` raises and the EF-less peek succeeds, the code unconditionally nulls out the embedding function and sets `_stored_model_name` to the stored model, without checking whether the stored model equals the current model. For a collection indexed with the same model that raised an EF-object conflict (the exact scenario this PR aims to preserve), read-only commands like `search` and `status` now fail with `Cannot search: the stored embedding model ... could not be loaded`, even though the model matches and the collection should be fully usable. The `allow_recreate=True` branch correctly peeks, compares models, and calls `_assign_preserved_collection` to rebind the EF; the read-only branch should do the same when `stored_model == current_model`. As written, a matching-model collection becomes read-only whenever the EF conflict occurs in read-only mode, which is a regression for search/status.

```suggestion
Wrap the `_collection_confirmed_missing` call and the subsequent need_recreate decision in the writer lock, or move the confirmation inside the lock before deleting.
```

<sub>**How this was verified:** Traced the flow: with force_recreate=True and allow_recreate=False, the initial get_collection succeeds, so need_recreate is set in the try block (line 412-418). Then the need_recreate block (line 555) hits the unconditional `if not allow_recreate` and raises. The watch CLI passes allow_recreate=False for explicit embedding functions (line 785 in watcher.py).</sub>

**⚠️ P2 medium · ℹ️ non-blocking** — [`packages/gptme-rag/src/gptme_rag/indexing/indexer.py:1755`](https://github.com/gptme/gptme-contrib/blob/6061fb2f1db5d732adb39581c57f55d464c4aedc/packages/gptme-rag/src/gptme_rag/indexing/indexer.py#L1755)

The new persist-dir filtering is applied when scanning directories but not when the input path is a file. In `_get_valid_files`, the `if path.is_file()` branch (line 1755) adds the file to `valid_files` without checking `_is_persist_artifact`. If a user passes a file that lives inside the persist directory (e.g., `chroma.sqlite3` or the lock file) as an explicit index argument, that file is indexed instead of being skipped. The watcher path does filter such artifacts, so this is inconsistent and can lead to binary/index files being ingested as documents, producing noise or failures.

```suggestion
Add a persist-artifact check in the `is_file()` branch: `if path.is_file(): if not self._is_persist_artifact(path): valid_files.add(path); return valid_files`.
```

<sub>**How this was verified:** Inspect `_get_valid_files`: the file case returns directly without calling `_is_persist_artifact`, while the directory loop applies the check after `resolve()` (line 1814). Immediate consequence: an explicit file inside the persist dir is indexed.</sub>

- `packages/gptme-rag/src/gptme_rag/cli.py` — Adds the gc-orphans command with dry-run/--apply options.
- `packages/gptme-rag/src/gptme_rag/indexing/gc.py` — New module with orphan segment dir collection/deletion and a writer lock context manager.
- `packages/gptme-rag/src/gptme_rag/indexing/indexer.py` — Wraps collection writes in the writer lock, preserves matching-model collections on EF conflict, adds persist-dir artifact filtering, and rebinds embedding functions on preserved collections.
- `packages/gptme-rag/src/gptme_rag/indexing/watcher.py` — Skips persist-dir paths in file event processing.
- `packages/gptme-rag/tests/test_file_limit.py` — Adds a regression test that persist-dir artifacts are excluded from valid files.
- `packages/gptme-rag/tests/test_same_model_recreate_guard.py` — Adds tests for matching-model recreate guards, GC behavior, and writer-lock refusal.

</details>

| `345af0c3dd86` | 3/5 | 1 | llm | 2026-08-28 05:42 UTC |
| `6d7cf595ac66` | 2/5 | 5 | llm | 2026-08-28 12:35 UTC |
| `88c3fee0b338` | 3/5 | 1 | llm | 2026-08-28 15:59 UTC |
| `9c71d3bfc9e9` | 3/5 | 1 | llm | 2026-08-28 17:18 UTC |
| `6068575022f9` | 3/5 | 3 | llm | 2026-08-29 01:18 UTC |
| `0f15d8d1fae9` | 3/5 | 1 | llm | 2026-08-29 01:50 UTC |

</details>

<sub>Reviewed `6061fb2f1db5` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 805s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

TimeToBuildBob:
Addressed the P1 in `6d7cf595`. The finding was correct.

**Confirmed:** `_add_documents` (indexer.py:735) calls `collection.add(documents=..., metadatas=..., ids=...)` with no explicit `embeddings=`, so the handle's own EF does the embedding. `peek` is fetched without an embedding function, so it carries Chroma's default 384-dim MiniLM — `InvalidDimensionException` against 768-dim ModernBERT vectors on the next index run. Preserving a collection you can no longer write to is not preserving it.

**Fix:** the stored model *matches* ours, so our EF is known-compatible — Chroma objected to the EF **object identity**, not the model. New `_bind_embedding_function()` binds it onto the handle Chroma refused to bind it to and verifies the binding took. Chroma internals are not API, so if the attribute is gone (version drift) it falls back to the reviewer's suggestion: null the EF and set `_stored_model_name`, which the existing guards in `add_document`/`_add_documents`/`index_directory`/`search` already turn into a clear `RuntimeError`. Wrong-dimension writes are never the fallback.

Skipped when the current EF is Chroma's default — an EF-less handle is already correct there, and rebinding unconditionally would have armed the fail-loud guard for the default case (caught by the pre-existing test).

**Tests:** the reviewer's point that `count()` doesn't exercise embedding was the actual gap. Two new tests drive `Indexer.__init__` end to end with a stubbed `ModernBERTEmbedding` — one asserts the preserved handle carries *our* EF and the read-only guard is not armed, one asserts the fail-loud path still preserves the data and raises. 7 passed; both new tests fail with the fix reverted.

packages/gptme-rag/src/gptme_rag/indexing/gc.py
```diff
@@ -0,0 +1,78 @@
...
+        if not child.is_dir():
+            continue
+        if not _UUID_DIR.fullmatch(child.name):
+            continue
+        if child.name in live_ids:
+            continue
+        orphans.append(child)
+
+    if apply:
+        for path in orphans:
```
TimeToBuildBob: ❌ **P1** — The `gc_orphan_segment_dirs` function computes live_ids from chroma.sqlite3 and then deletes any UUID-named directory not in live_ids when apply=True. However, Chroma creates segment directories before inserting their IDs into the segments table. If an indexing process creates a new segment directory and the gc command runs between directory creation and the database commit, the directory will appear as an orphan and be deleted, corrupting the live index. The PR description acknowledges this by not running GC against a live persist dir, but the CLI does not enforce that; a user running `gptme-rag gc-orphans --apply` during indexing will lose data. There is no lock or check for active writers. This is a data-loss bug.

packages/gptme-rag/src/gptme_rag/indexing/gc.py
```diff
@@ -0,0 +1,78 @@
...
+            shutil.rmtree(path)
```
TimeToBuildBob, line 120: ⚠️ **P2** — In `gc_orphan_segment_dirs`, `shutil.rmtree(path)` is called without exception handling. If one deletion fails (e.g., permission denied, directory already gone), the exception propagates out of the function, aborting the loop and leaving all remaining orphans undeleted. The CLI then crashes with a traceback instead of printing which dirs were removed, and the user gets no summary of the partial state. Since the function is documented as returning the deleted paths, an unhandled failure contradicts its contract. Each `rmtree` should be wrapped so one failure does not prevent the rest, and partial results should be reported.

packages/gptme-rag/src/gptme_rag/cli.py
```diff
@@ -829,6 +830,37 @@ def watch(
...
+)
+def gc_orphans(persist_dir: Path, apply: bool):
+    """Remove leftover Chroma segment dirs not listed in chroma.sqlite3.
+
+    Recreating a collection leaves UUID-named HNSW folders on disk. This
+    command lists them (dry-run) or deletes them (--apply). Live segment
+    and collection dirs are never touched.
+    """
+    orphans = gc_orphan_segment_dirs(persist_dir, apply=apply)
+    if not orphans:
```
TimeToBuildBob, line 858: ⚠️ **P2** — The new `gc-orphans` CLI command prints `No orphan segment dirs` whenever `gc_orphan_segment_dirs` returns an empty list. That function returns an empty list not only when there are genuinely no orphans, but also when the Chroma sqlite catalog cannot be read (it catches `sqlite3.Error` and returns `[]`). A user running `gc-orphans` on a persist dir whose `chroma.sqlite3` is corrupted or locked will see the reassuring "No orphan segment dirs" message instead of an error, and may conclude their index is clean when in fact the GC was skipped. The function should signal the failure (e.g., raise or return a sentinel) so the CLI can report it.

packages/gptme-rag/src/gptme_rag/indexing/gc.py
```diff
@@ -0,0 +1,78 @@
...
+
+_UUID_DIR = re.compile(
+    r"^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$",
+    re.IGNORECASE,
+)
+
+
+def _live_chroma_ids(db: Path) -> set[str]:
+    """IDs that still belong to the current Chroma catalog."""
+    con = sqlite3.connect(f"file:{db}?mode=ro", uri=True)
```
TimeToBuildBob: ⚠️ **P2** — In gc.py, the sqlite3 URI is built with `f"file:{db}?mode=ro"`. If the persist_dir path contains a literal '?' character, the URI is malformed: sqlite3 treats the first '?' as the start of the query string, so the file path is truncated at the '?'. For example, a persist_dir of `/tmp/a?b` yields `file:/tmp/a?b/chroma.sqlite3?mode=ro`, which sqlite3 interprets as the file `/tmp/a` with a query string `b/chroma.sqlite3?mode=ro`. This causes sqlite3.connect to raise an OperationalError, which is caught by the `except sqlite3.Error` and returns an empty list, so `gc_orphan_segment_dirs` silently refuses to GC. A persist dir whose path contains a '?' therefore never cleans up orphans. The fix is to percent-encode the path before embedding it in the URI, or to pass the path directly to `sqlite3.connect` and set read-only via a separate mechanism (e.g., `sqlite3.connect(str(db))` then execute `PRAGMA query_only=ON`).

TimeToBuildBob:
Addressed the 2/5 review on `6d7cf595` in `88c3fee0`.

**P1 auto-preserve path:** the matching-model rebind from `6d7cf595` was only on the explicit-model branch. Auto mode assigned the peeked handle without `_bind_embedding_function`, so writes would still go through MiniLM. Both paths now share `_assign_preserved_collection`. Test: `test_auto_mode_rebinds_preserved_collection`.

**P1 gc-orphans vs live writer:** `--apply` takes the same exclusive flock Indexer holds around collection writes, and reads the catalog inside that lock. If an indexer is mid-write, apply raises instead of deleting a not-yet-catalogued segment dir.

**P2s:** a failed `rmtree` no longer aborts the rest of the sweep; catalog read errors raise (`No orphan segment dirs` is no longer printed on a corrupt/locked sqlite); sqlite is opened by filesystem path so a `?` in persist_dir is not a URI delimiter.

12 tests in `test_same_model_recreate_guard.py` passed locally.

packages/gptme-rag/src/gptme_rag/indexing/indexer.py
TimeToBuildBob: Fixed in 6d7cf595 for the explicit-model preserve path; 88c3fee0 applies the same `_assign_preserved_collection` helper to the auto-mode exception path that was still assigning the EF-less peek handle.

packages/gptme-rag/src/gptme_rag/indexing/gc.py
TimeToBuildBob: Fixed in 88c3fee0. `--apply` takes `index_writer_lock` non-blocking (the same flock Indexer holds around collection writes) and reads the catalog inside that lock. Mid-index apply raises `RuntimeError` instead of deleting a not-yet-catalogued segment dir.

packages/gptme-rag/src/gptme_rag/indexing/gc.py
TimeToBuildBob: Fixed in 88c3fee0. Each `rmtree` is wrapped; a single OSError is logged and the sweep continues. Apply returns only the dirs actually removed.

packages/gptme-rag/src/gptme_rag/cli.py
TimeToBuildBob: Fixed in 88c3fee0. Catalog read failures raise `ChromaCatalogError`; the CLI prints the error and exits 1 instead of `No orphan segment dirs`.

packages/gptme-rag/src/gptme_rag/indexing/gc.py
TimeToBuildBob: Fixed in 88c3fee0. `_live_chroma_ids` opens sqlite by filesystem path with `PRAGMA query_only=ON`, so a `?` in persist_dir is no longer treated as a URI delimiter.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Latest fixes are on `88c3fee0`; CI is green. The merge gate still refuses because the prior AI review covers `6d7cf595`, not the current head. I dispatched the current-head review now; I will address any fresh P0/P1 finding, otherwise the next monitoring pass can rerun the self-merge helper.

packages/gptme-rag/src/gptme_rag/indexing/gc.py
```diff
@@ -0,0 +1,138 @@
...
+    db = persist_dir / "chroma.sqlite3"
+    if not persist_dir.is_dir() or not db.is_file():
+        return []
+
+    try:
+        if apply:
+            with index_writer_lock(persist_dir, blocking=False):
+                orphans = _collect_orphans(persist_dir, _live_chroma_ids(db))
+                return _delete_orphans(orphans)
+        return _collect_orphans(persist_dir, _live_chroma_ids(db))
```
TimeToBuildBob, line 146: ❌ **P1** — The `gc_orphan_segment_dirs` function only wraps `sqlite3.Error` in its `try/except`, but the call to `_collect_orphans` (which uses `persist_dir.iterdir()`) can raise `OSError` (e.g., `PermissionError`) when the persist directory is not readable, and `index_writer_lock` can raise `OSError` when the lock file cannot be created (e.g., read-only directory). These exceptions are not caught by the `except sqlite3.Error` and are not caught in the CLI's `gc_orphans` command (which only catches `ChromaCatalogError` and `RuntimeError`). The result is an unhandled traceback and a non-zero exit code without a user-friendly error message. For a CLI command designed to be safe, a permission error should be reported cleanly. _(2/3 passes — pass 2 looked and disagreed)_

TimeToBuildBob:
@greptileai review

packages/gptme-rag/src/gptme_rag/indexing/gc.py
TimeToBuildBob: Fixed in 9c71d3bf. `gc_orphan_segment_dirs` now wraps persist-dir `OSError` (iterdir / lock-file create) as `ChromaCatalogError`; the CLI prints it and exits 1 instead of dumping a traceback. Test: `test_gc_orphans_oserror_is_not_a_traceback`.

TimeToBuildBob:
Addressed the current-head P1 in `9c71d3bf`. Persist-dir `OSError` (iterdir / lock-file create) is now `ChromaCatalogError`; the CLI prints it and exits 1 instead of dumping a traceback. Test: `test_gc_orphans_oserror_is_not_a_traceback`.

CI is running on the new head. Dispatched a current-head AI review. Greptile still has an initial-review request in flight (last trigger 17:10 UTC); not retriggering.

Next monitoring pass: if that review is clean, run the self-merge helper. If a fresh P0/P1 lands, fix it first.

packages/gptme-rag/src/gptme_rag/indexing/gc.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+def _collect_orphans(persist_dir: Path, live_ids: set[str]) -> list[Path]:
+    orphans: list[Path] = []
+    for child in persist_dir.iterdir():
+        if not child.is_dir():
+            continue
+        if not _UUID_DIR.fullmatch(child.name):
+            continue
+        if child.name in live_ids:
```
TimeToBuildBob: ❌ **P1** — The orphan detection in _collect_orphans checks directory names against live IDs from the sqlite catalog using exact case, but _UUID_DIR is compiled with re.IGNORECASE. On case-insensitive filesystems (macOS default, Windows) Chroma can create segment directories with uppercase hex characters while the sqlite segments.id stores the canonical lowercase form. Such a directory fails the case-sensitive membership test and is classified as an orphan; with --apply it is passed to shutil.rmtree and deleted, destroying a live segment and its vectors. The regex accepts uppercase specifically to allow these names, so the comparison must be normalized. Fix by comparing child.name.lower() against live_ids (and lowercasing the IDs read from sqlite, or lowercasing both). _(2/3 passes — pass 2 looked and disagreed)_

TimeToBuildBob: Fixed in 60685750. Membership now compares `child.name.lower()` against lowercased catalog ids, matching `_UUID_DIR`'s IGNORECASE. Test: `test_gc_orphans_preserves_live_uuid_dir_case_insensitive`.

TimeToBuildBob:
Addressed the current-head P1 in `60685750`. UUID dir membership is now case-folded, so an uppercase on-disk name matching a lowercase catalog id is treated as live rather than orphaned. Test: `test_gc_orphans_preserves_live_uuid_dir_case_insensitive`.

CI is running on the new head. Dispatching current-head AI review and Greptile.

TimeToBuildBob:
Current-head re-review of `60685750` is blocked on the OpenRouter review-key daily limit until `2026-08-29T00:00Z`. The 18:25 `ai-review-now` already 403'd this head (quota breaker). I am not retrying in-band or stealing other keys.

CI is still green and mergeable; all finding threads are resolved. Greptile is dark on this repo — not triggering.

Scheduled recheck at 00:10Z: `ai-review.py gptme/gptme-contrib 1535 --engine llm`, then `self-merge-if-eligible.sh` if the marker matches this head.

packages/gptme-rag/src/gptme_rag/indexing/gc.py
```diff
@@ -0,0 +1,150 @@
...
+    creates a segment directory before inserting its id into sqlite, so a GC
+    snapshot taken in that window would treat a live dir as an orphan.
+    """
+    persist_dir = Path(persist_dir)
+    persist_dir.mkdir(parents=True, exist_ok=True)
+    fd = os.open(persist_dir / _WRITER_LOCK_NAME, os.O_CREAT | os.O_RDWR, 0o644)
+    flags = fcntl.LOCK_EX if blocking else (fcntl.LOCK_EX | fcntl.LOCK_NB)
+    try:
+        try:
+            fcntl.flock(fd, flags)
```
TimeToBuildBob, line 57: ⚠️ **P2** — In `index_writer_lock`, if the initial `fcntl.flock(fd, flags)` fails with an `OSError` that is not `BlockingIOError` (e.g., on a filesystem that does not support flock, or an invalid argument), the `finally` block unconditionally attempts `fcntl.flock(fd, fcntl.LOCK_UN)`. That unlock call can itself raise an `OSError`, masking the original exception and leaving the caller with an uninformative error about unlocking rather than the actual lock failure. This hinders diagnosis and can cause an unexpected crash on unsupported platforms or misconfigured mounts.

packages/gptme-rag/src/gptme_rag/indexing/gc.py
```diff
@@ -0,0 +1,150 @@
...
+    persist_dir = Path(persist_dir)
+    db = persist_dir / "chroma.sqlite3"
+    try:
+        if not persist_dir.is_dir() or not db.is_file():
+            return []
+        if apply:
+            with index_writer_lock(persist_dir, blocking=False):
+                orphans = _collect_orphans(persist_dir, _live_chroma_ids(db))
+                return _delete_orphans(orphans)
+        return _collect_orphans(persist_dir, _live_chroma_ids(db))
```
TimeToBuildBob, line 146: ⚠️ **P2** — The dry-run path of `gc_orphan_segment_dirs` does not acquire the writer lock, unlike `apply`. If an indexer is actively writing (e.g., inside `add_document`), the sqlite catalog can be read while a segment directory has been created but its id not yet inserted into `segments` or `collections`. The dry-run will then classify that live directory as an orphan and report it as "Would remove", which is a false positive that can mislead operators into deleting a directory they believe is orphaned. This is not a deletion risk (dry-run only lists), but it violates the documented promise that live segment dirs are never touched.

TimeToBuildBob:
Addressed the current-head P1 in `0f15d8d1`.

**P1 (allow_recreate=False wipe):** confirmed. The `not allow_recreate` except branch treated any failed EF-less `get_collection` as "collection missing" and set `need_recreate=True`. The recreate block deletes first, so a transient peek failure (sqlite lock during a concurrent write) would wipe a live collection from search/status. The suggested `_auto_peeked_collection` rebind is the wrong branch — that handle is only set in auto mode.

Fix: raise `RuntimeError` instead of recreating, and refuse the `need_recreate` gate when `allow_recreate` is False. Test: `test_allow_recreate_false_peek_failure_does_not_wipe`.

**P2s skipped:** flock unlock masking is an unsupported-fs edge; dry-run lock is already documented and is not a deletion risk. Surviving P2s do not block the merge gate.

CI is running on the new head. Dispatching current-head AI review. Greptile is still dark on this repo — not triggering.

packages/gptme-rag/src/gptme_rag/indexing/indexer.py
```diff
@@ -496,32 +513,62 @@ def __init__(
...
+                    # Peek without an embedding function before wiping: a matching
+                    # collection must be preserved (the 84k-chunk ambient-memory wipe).
+                    logger.debug("Collection access error: %s", e)
+                    if force_recreate:
+                        need_recreate = True
+                    else:
+                        try:
+                            peek = self.client.get_collection(name=collection_name)
+                        except Exception:
+                            need_recreate = True
```
TimeToBuildBob: ❌ **P1** — In the `allow_recreate=True` path of `__init__`, when the initial `get_collection(name, embedding_function=...)` fails, the code peeks with `get_collection(name=collection_name)` to decide whether a matching collection exists. If that peek also raises (e.g., a transient sqlite write-lock timeout while another process is mid-write), the exception handler sets `need_recreate = True`. The later `if need_recreate:` block then deletes and recreates the collection. For the read-only path (`allow_recreate=False`) the code was changed to raise on a failed peek to avoid wiping a live collection, but the read-write path (which the `index` command uses) still wipes on a transient peek failure. This reintroduces the data-loss scenario the PR aims to fix: a live 84k-chunk collection with a matching model can be deleted and rebuilt from zero if the peek fails for a temporary reason, silently losing all indexed content. _(3/3 passes)_

TimeToBuildBob: Fixed in 6061fb2f.

Confirmed: the `allow_recreate=True` except branch treated a failed EF-less peek as "collection missing" and set `need_recreate=True`. Recreate deletes first, so a transient sqlite lock still wiped a matching live collection — the write-path half of the 84k-chunk incident.

Fix: only recreate when `list_collections()` confirms the name is absent. If the catalog cannot be listed, or the collection is present, raise instead of wiping. Test: `test_allow_recreate_true_peek_failure_does_not_wipe`.

Also skipped persist-dir artifacts (writer lock / chroma sqlite) when indexing a tree that contains the persist directory — CI was ingesting `.gptme-rag-writer.lock` as an extra document.

TimeToBuildBob:
Addressed the current-head P1 in `6061fb2f`.

**P1 (allow_recreate=True wipe on peek failure):** confirmed (3/3). The index-command path still treated a failed EF-less `get_collection` as "collection missing" and set `need_recreate=True`. Recreate deletes first, so a transient sqlite lock would wipe a matching live collection — the same 84k-chunk incident the read-only guard closed. Fix: confirm absence via `list_collections()` before recreating; if the catalog cannot be listed, fail loud. Test: `test_allow_recreate_true_peek_failure_does_not_wipe`.

**CI (watcher tests ingesting the writer lock):** `FileWatcher.start()` calls `index_directory` on the watched tree, which contained `persist_directory`. The lock file created during Indexer init was indexed as an extra document, so `len(results) == 1` assertions failed. `_get_valid_files` and the watcher now skip persist-dir artifacts. Test: `test_get_valid_files_skips_persist_dir_artifacts` (33 passed including watcher).

CI running on the new head. Dispatching current-head AI review. Greptile still dark on this repo — not triggering.

Next: if that review is clean (no open P0/P1), run the self-merge helper.

packages/gptme-rag/src/gptme_rag/indexing/gc.py
```diff
@@ -0,0 +1,150 @@
...
+dirs are never touched.
+
+``--apply`` takes an exclusive writer lock (the same lock Indexer holds
+around collection writes). If an indexer is mid-write, apply refuses rather
+than deleting a segment dir Chroma created but has not yet catalogued.
+"""
+
+from __future__ import annotations
+
+import fcntl
```
TimeToBuildBob, line 19: ❌ **P1** — The new `gc.py` imports `fcntl`, which is only available on POSIX systems. On Windows, importing `gptme_rag.indexing.gc` raises `ImportError`. Because `cli.py` imports `gc` at module load (`from .indexing.gc import ChromaCatalogError, gc_orphan_segment_dirs`), every CLI command fails to start on Windows. Similarly, `indexer.py` imports `index_writer_lock` from `gc`, so any use of the `Indexer` class breaks. The package previously had no `fcntl` dependency, so this change makes the entire package unusable on Windows. There is no platform guard or fallback; the `fcntl` import is unconditional. _(2/3 passes — pass 1 looked and disagreed)_

TimeToBuildBob: Won't fix in this PR.

The finding is true as far as it goes: `import fcntl` is POSIX-only, so `gc.py` (and therefore `Indexer` / the CLI) would fail to import on Windows. Pass 1 of 3 already disagreed.

This package has no Windows CI, no OS classifier, and no prior `fcntl` usage on master. The writer lock was added as a Linux flock around Chroma persist-dir writes (Bob's LXC is the production user). A Windows port (`msvcrt.locking` / portalocker) is a separate change we cannot verify here.

Leaving the POSIX lock as-is. A Windows lock is out of scope for a recreate-guard PR that is already on its 7th review round.

packages/gptme-rag/src/gptme_rag/indexing/gc.py
```diff
@@ -0,0 +1,150 @@
...
+
+    Dry-run (``apply=False``) is the default. Raises ``ChromaCatalogError`` if
+    the sqlite catalog or persist dir cannot be read, rather than pretending
+    the dir is clean. ``apply=True`` raises ``RuntimeError`` if an index writer
+    currently holds the persist-dir lock.
+    """
+    persist_dir = Path(persist_dir)
+    db = persist_dir / "chroma.sqlite3"
+    try:
+        if not persist_dir.is_dir() or not db.is_file():
```
TimeToBuildBob, line 140: ⚠️ **P2** — In `gc_orphans` CLI, the dry-run default returns an empty list when the persist dir or `chroma.sqlite3` is missing, printing 'No orphan segment dirs'. However, if the persist dir exists but `chroma.sqlite3` is not a file (e.g., a directory or a broken symlink), the code returns `[]` without distinguishing that from a clean dir. This is not a data-loss risk, but it can mask a corrupt catalog: a user running `gc-orphans` would see 'clean' and not notice that the catalog is unreadable. The function already raises `ChromaCatalogError` for sqlite errors, but this early return bypasses that. The CLI should probably report an error when `persist_dir` exists but `db` is not a regular file, or at least log a warning.

packages/gptme-rag/src/gptme_rag/indexing/indexer.py
```diff
@@ -543,6 +602,98 @@ def __init__(
...
+            return False
+
+    def _is_persist_artifact(self, path: Path) -> bool:
+        """True for the persist dir, files inside it, or the writer lock.
+
+        Indexing those files is always wrong: they are the index, not a
+        corpus. Watcher tests (and any watch of a tree that contains the
+        persist dir) would otherwise ingest ``.gptme-rag-writer.lock``.
+        """
+        if path.name == _WRITER_LOCK_NAME:
```
TimeToBuildBob, line 631: ⚠️ **P2** — _is_persist_artifact returns True for any file whose name equals the writer-lock filename, regardless of whether that file is actually inside the persist directory. If a user has a legitimate corpus file named exactly '.gptme-rag-writer.lock', it will be silently excluded from indexing and watching. This is a behavior change: before this PR, such a file would have been indexed normally. The check should be scoped to files inside the persist dir, not just by name.

packages/gptme-rag/tests/test_file_limit.py
```diff
@@ -21,6 +21,32 @@ def _write_files(directory, count: int, prefix: str = "doc") -> None:
...
+    indexer = Indexer(
+        persist_directory=persist_dir,
+        enable_persist=True,
+        embedding_function="default",
+        collection_name="default",
+    )
+    files = indexer._get_valid_files(tmp_path)
+    names = {p.name for p in files}
+    assert "keep.txt" in names
+    assert ".gptme-rag-writer.lock" not in names
```
TimeToBuildBob, line 45: ⚠️ **P2** — In `test_get_valid_files_skips_persist_dir_artifacts`, the assertion `".gptme-rag-writer.lock" not in names` passes vacuously. The lock file is only created when a write lock is acquired (e.g., during a recreate or add). The test's `Indexer` is initialized with a fresh persist_dir and no recreate occurs, so the lock file is never created; thus the assertion is true regardless of whether the code correctly filters the lock file. The test gives false confidence that the lock file is skipped. To actually pin the regression, the test should create the lock file (e.g., by acquiring the lock) or assert on a file that is actually present inside the persist_dir.

PR #1538, opened 2026-08-28
fix(gptodo): retry transient Claude auth failures once

### LINKED ISSUE #1536: gptodo spawn: retry once on a transient Claude Code 401 instead of marking the subagent failed
## Problem

`gptodo spawn --backend claude` treats any non-zero `claude -p` exit as a hard `failed` session. A **transient Claude Code 401** ("Invalid authentication credentials") is usually an OAuth access-token blip that recovers on its own within seconds (token refresh / sibling `/login`). Today that death:

1. Marks the spawned agent `failed`
2. Loses any partial uncommitted work the child had started
3. Does **not** re-dispatch

This was the original email-fix incident on 2026-06-24 (Bob task `harden-401-auth-everywhere`, surface 3 remainder). Bob's five reactive **bash** launchers (`autonomous-run`, `operator-loop`, `spawn-workers`, `twitter-dispatch`, `project-monitoring`) now share `with_auth_resilience` (preflight → run → on-401 write stale-marker + backoff + retry-once). **`gptodo spawn` is the last unhardened `claude -p` dispatcher** because it lives in this repo.

## Where

`packages/gptodo/src/gptodo/subagent.py`

- **Foreground** (`background=False`): `subprocess.run(cmd)` then `session.status = "failed"` on any non-zero (`~line 380`). No classifier, no retry.
- **Background** (tmux): writes `EXIT_CODE=$?` into the output file; `check_session` marks `failed` on non-zero (`~line 426`). Same gap.

## Proposed contract

On a `claude` backend non-zero exit:

1. Classify the captured output as a **size-gated auth-death** (tiny output AND a 401 / `authentication_failed` / `authentication_error` / `unauthorized` signature). A large, genuinely-completed session whose diff merely mentions "401" must **not** retry.
2. If it is an auth-death: sleep a short backoff (~5s) and **re-dispatch once**.
3. Only mark `failed` if the retry also fails (or the failure is not an auth-death).
4. Do not treat the first 401 as a task failure / burned attempt.

Do **not** depend on Bob's workspace. Copy a small classifier into gptodo (or a tiny `gptme_contrib` helper) — Bob's canonical pattern is `scripts/lib/auth_401.py` (`is_transient_401` / `is_auth_death` with a 2KB size gate) if you want the same signatures.

## Acceptance

- Unit test: a tiny 401-shaped capture → one retry, success on the second attempt → session `completed`.
- Unit test: a large output that mentions "401" in prose → **no** retry, status `failed` (or whatever the original non-zero path is).
- Unit test: two consecutive 401s → one retry then `failed`.
- Foreground path is the original incident; background/tmux should get the same classifier even if retry is trickier there (at minimum: do not treat an auth-death output as a successful completion, and surface it distinctly).

## Context

- Bob-local wrapper (already shipped, this session): `scripts/lib/auth_resilience.sh` / `with_auth_resilience`
- Task: `harden-401-auth-everywhere` in TimeToBuildBob/bob (ErikBjare/bob historically)
- Related closed issue on the worker blacklist side: ErikBjare/bob#968

TimeToBuildBob:
## Summary

- add a 2 KiB size-gated classifier for Claude Code authentication failures
- retry foreground and background/tmux Claude spawns once after a 5-second backoff
- preserve the first auth-failure output for diagnosis and retain the original failure behavior for non-auth errors

## Verification

- `uv run --package gptodo pytest packages/gptodo/tests/test_subagent.py -q` (19 passed)
- `make typecheck-packages`
- pre-commit hooks passed on commit
- full gptodo suite: 554 passed; one unrelated existing failure in `test_skip_claimed.py::TestGetClaimedTaskIds::test_excludes_own_agent_claims` due to the ambient `GIT_COMMITTER_SESSION_ID` overriding the test's `AGENT_ID=me`

Closes gptme/gptme-contrib#1536

### DIFF
--- a/packages/gptodo/src/gptodo/auth.py
+++ b/packages/gptodo/src/gptodo/auth.py
@@ -10,7 +10,7 @@ DEFAULT_MAX_BYTES = 2000
 
 _AUTH_FAILURE_RE = re.compile(
     rb"\b401\b|unauthorized|invalid authentication credentials|"
-    rb"invalid bearer token|authentication_(?:error|failed)|"
+    rb"invalid bearer token|authentication[ _](?:error|failed)|"
     rb"oauth.{0,40}(?:expired|fail|error|invalid)|please run /login",
     re.IGNORECASE,
 )
--- a/packages/gptodo/src/gptodo/auth.py
+++ b/packages/gptodo/src/gptodo/auth.py
@@ -1,0 +1,47 @@
+"""Classify short-lived Claude Code authentication failures."""
+
+from __future__ import annotations
+
+import argparse
+import re
+from pathlib import Path
+
+DEFAULT_MAX_BYTES = 2000
+
+_AUTH_FAILURE_RE = re.compile(
+    rb"\b401\b|unauthorized|invalid authentication credentials|"
+    rb"invalid bearer token|authentication_(?:error|failed)|"
+    rb"oauth.{0,40}(?:expired|fail|error|invalid)|please run /login",
+    re.IGNORECASE,
+)
+
+
+def is_transient_auth_death(output: str | bytes, max_bytes: int = DEFAULT_MAX_BYTES) -> bool:
+    """Return whether a small output contains a Claude auth-failure signature.
+
+    The size gate is required: a successful agent can produce a large transcript
+    that discusses HTTP 401 without having failed authentication itself.
+    """
+    data = output.encode(errors="replace") if isinstance(output, str) else output
+    return 0 < len(data) <= max_bytes and bool(_AUTH_FAILURE_RE.search(data))
+
+
+def is_auth_death_file(path: Path, max_bytes: int = DEFAULT_MAX_BYTES) -> bool:
+    """Classify a captured process output file, returning False if unreadable."""
+    try:
+        data = path.read_bytes()
+    except OSError:
+        return False
+    return is_transient_auth_death(data, max_bytes=max_bytes)
+
+
+def main(argv: list[str] | None = None) -> int:
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument("--classify-file", type=Path, required=True)
+    parser.add_argument("--max-bytes", type=int, default=DEFAULT_MAX_BYTES)
+    args = parser.parse_args(argv)
+    return 0 if is_auth_death_file(args.classify_file, args.max_bytes) else 1
+
+
+if __name__ == "__main__":
+    raise SystemExit(main())
--- a/packages/gptodo/src/gptodo/subagent.py
+++ b/packages/gptodo/src/gptodo/subagent.py
@@ -252,7 +252,21 @@ def spawn_agent(
                 # retry happens inside tmux so check_session keeps one session ID.
                 python = shlex.quote(sys.executable)
                 auth_check = f"{python} -m gptodo.auth --classify-file {safe_output}"
-                shell_cmd = f'touch {safe_output}; tail -f {safe_output} & TAIL_PID=$!; {claude_cmd} > {safe_output} 2>&1; EXIT_CODE=$?; if [ "$EXIT_CODE" -ne 0 ] && {auth_check}; then cp {safe_output} {safe_output}.first-auth-failure; sleep 5; {claude_cmd} > {safe_output} 2>&1; EXIT_CODE=$?; fi; echo "EXIT_CODE=$EXIT_CODE" >> {safe_output}; kill $TAIL_PID 2>/dev/null'
+                # Concatenate the classifier so the shell never sees a
+                # `{auth_check}` token. That is a Python f-string field, not a
+                # bash brace group (`{ cmd; }`).
+                shell_cmd = (
+                    f"touch {safe_output}; tail -f {safe_output} & TAIL_PID=$!; "
+                    f"{claude_cmd} > {safe_output} 2>&1; EXIT_CODE=$?; "
+                    'if [ "$EXIT_CODE" -ne 0 ] && '
+                    + auth_check
+                    + (
+                        f"; then cp {safe_output} {safe_output}.first-auth-failure; "
+                        f"sleep 5; {claude_cmd} > {safe_output} 2>&1; EXIT_CODE=$?; fi; "
+                        f'echo "EXIT_CODE=$EXIT_CODE" >> {safe_output}; '
+                        "kill $TAIL_PID 2>/dev/null"
+                    )
+                )
             else:
                 # Non-gptme non-claude (codex) still launches `claude -p` today;
                 # keep that pre-existing command, but do not apply Claude-only retry.
--- a/packages/gptodo/src/gptodo/subagent.py
+++ b/packages/gptodo/src/gptodo/subagent.py
@@ -391,8 +391,11 @@ def spawn_agent(
 
         if backend == "claude" and result.returncode != 0 and is_transient_auth_death(output):
             logger.warning("Claude authentication failed; retrying once after 5 seconds")
-            time.sleep(5)
             first_auth_failure = output
+            # Persist before the retry so a TimeoutExpired/exception on the
+            # second run cannot leave the session output file empty.
+            output_file.write_text(first_auth_failure)
+            time.sleep(5)
             result = subprocess.run(
                 cmd,
                 cwd=workspace,
--- a/packages/gptodo/src/gptodo/subagent.py
+++ b/packages/gptodo/src/gptodo/subagent.py
@@ -247,11 +247,16 @@ def spawn_agent(
                 else ""
             )
             claude_cmd = f'{timeout_prefix}claude -p {model_arg} {sysprompt_arg} --dangerously-skip-permissions --tools default -- "$(cat {safe_prompt_file})"'
-            # Run the classifier in gptodo's current Python environment. The
-            # retry happens inside tmux so check_session keeps one session ID.
-            python = shlex.quote(sys.executable)
-            auth_check = f"{python} -m gptodo.auth --classify-file {safe_output}"
-            shell_cmd = f'touch {safe_output}; tail -f {safe_output} & TAIL_PID=$!; {claude_cmd} > {safe_output} 2>&1; EXIT_CODE=$?; if [ "$EXIT_CODE" -ne 0 ] && {auth_check}; then cp {safe_output} {safe_output}.first-auth-failure; sleep 5; {claude_cmd} > {safe_output} 2>&1; EXIT_CODE=$?; fi; echo "EXIT_CODE=$EXIT_CODE" >> {safe_output}; kill $TAIL_PID 2>/dev/null'
+            if backend == "claude":
+                # Run the classifier in gptodo's current Python environment. The
+                # retry happens inside tmux so check_session keeps one session ID.
+                python = shlex.quote(sys.executable)
+                auth_check = f"{python} -m gptodo.auth --classify-file {safe_output}"
+                shell_cmd = f'touch {safe_output}; tail -f {safe_output} & TAIL_PID=$!; {claude_cmd} > {safe_output} 2>&1; EXIT_CODE=$?; if [ "$EXIT_CODE" -ne 0 ] && {auth_check}; then cp {safe_output} {safe_output}.first-auth-failure; sleep 5; {claude_cmd} > {safe_output} 2>&1; EXIT_CODE=$?; fi; echo "EXIT_CODE=$EXIT_CODE" >> {safe_output}; kill $TAIL_PID 2>/dev/null'
+            else:
+                # Non-gptme non-claude (codex) still launches `claude -p` today;
+                # keep that pre-existing command, but do not apply Claude-only retry.
+                shell_cmd = f'touch {safe_output}; tail -f {safe_output} & TAIL_PID=$!; {claude_cmd} > {safe_output} 2>&1; echo "EXIT_CODE=$?" >> {safe_output}; kill $TAIL_PID 2>/dev/null'
 
         # Build environment exports for critical API keys
         # These may not be inherited by tmux detached sessions
--- a/packages/gptodo/src/gptodo/subagent.py
+++ b/packages/gptodo/src/gptodo/subagent.py
@@ -13,11 +13,15 @@ import logging
 import os
 import shlex
 import subprocess
+import sys
+import time
 import uuid
 from dataclasses import asdict, dataclass
 from datetime import datetime, timezone
 from pathlib import Path
 from typing import Literal
+
+from gptodo.auth import is_transient_auth_death
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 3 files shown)


TimeToBuildBob:
CI is green and the PR is mergeable. Self-merge is not eligible yet: Greptile has not reviewed the current head, and the change touches the auth-sensitive path. Waiting for review; no merge attempted.

packages/gptodo/src/gptodo/auth.py
```diff
@@ -0,0 +1,47 @@
...
+
+import argparse
+import re
+from pathlib import Path
+
+DEFAULT_MAX_BYTES = 2000
+
+_AUTH_FAILURE_RE = re.compile(
+    rb"\b401\b|unauthorized|invalid authentication credentials|"
+    rb"invalid bearer token|authentication_(?:error|failed)|"
```
TimeToBuildBob: ❌ **P1** — The classifier regex `authentication_(?:error|failed)` only matches auth errors that use an underscore between 'authentication' and 'error'/'failed'. Common Claude Code error messages such as 'Authentication failed' or 'Authentication error' use a space, so they do not match the pattern. As a result, `is_transient_auth_death` returns False for these messages, and the retry logic in both the foreground and background branches is not triggered. This means the user's session is marked as 'failed' without a retry for a large class of transient auth failures, defeating the purpose of the PR.

packages/gptodo/src/gptodo/subagent.py
```diff
@@ -242,7 +246,12 @@ def spawn_agent(
...
                 else ""
             )
-            shell_cmd = f'touch {safe_output}; tail -f {safe_output} & TAIL_PID=$!; {timeout_prefix}claude -p {model_arg} {sysprompt_arg} --dangerously-skip-permissions --tools default -- "$(cat {safe_prompt_file})" > {safe_output} 2>&1; echo "EXIT_CODE=$?" >> {safe_output}; kill $TAIL_PID 2>/dev/null'
+            claude_cmd = f'{timeout_prefix}claude -p {model_arg} {sysprompt_arg} --dangerously-skip-permissions --tools default -- "$(cat {safe_prompt_file})"'
+            # Run the classifier in gptodo's current Python environment. The
+            # retry happens inside tmux so check_session keeps one session ID.
+            python = shlex.quote(sys.executable)
+            auth_check = f"{python} -m gptodo.auth --classify-file {safe_output}"
+            shell_cmd = f'touch {safe_output}; tail -f {safe_output} & TAIL_PID=$!; {claude_cmd} > {safe_output} 2>&1; EXIT_CODE=$?; if [ "$EXIT_CODE" -ne 0 ] && {auth_check}; then cp {safe_output} {safe_output}.first-auth-failure; sleep 5; {claude_cmd} > {safe_output} 2>&1; EXIT_CODE=$?; fi; echo "EXIT_CODE=$EXIT_CODE" >> {safe_output}; kill $TAIL_PID 2>/dev/null'
```
TimeToBuildBob: ⚠️ **P2** — The background (tmux) retry logic is added to the non-gptme branch of `spawn_agent`, which is shared by both `claude` and `codex` backends (the `else` branch after `if backend == "gptme"`). The retry condition uses the Claude-specific classifier `is_transient_auth_death`, but there is no `backend == "claude"` guard in the shell command. In contrast, the foreground retry is explicitly conditioned on `backend == "claude"`. This inconsistency means a `codex` agent spawned in the background will be retried once on an auth-like failure, while the same backend in the foreground will not. The classifier may also mis-classify codex error messages (e.g., matching '401' in a non-auth context) or miss codex-specific auth errors, leading to unexpected duplicate work or missed retries. Since the PR description scopes the feature to Claude, this is a defect.

TimeToBuildBob:
## 🤖 AI code review

The retry logic's core decision — "was this a real auth failure?" — is gated by a 2000-byte combined-output threshold that excludes genuine failures once stdout grows past ~2 KiB (P1). The background classifier additionally runs in a throwaway sub-process that, on any import failure, returns non-zero and silently disables the entire background-retry feature with no log (P2). Small non-auth outputs containing "401" false-positive into a full retry (doubling worst-case latency and obscuring the first failure), and the foreground double-writes `OUT` before the retry finishes, losing the marker if the retry raises. The shell command correctly preserves the first failure in a sidecar, but the divergent foreground/background output formats and the unconfigurable 5s backoff add avoidable operational friction. All findings are in the `src/gptodo/subprocess.py` retry construction and depend on the ordering: exit-code check → combined-string gate → classifier/`&&` decision → retry.

| `e40244dbfc88` | 3/5 | 2 | llm | 2026-08-28 06:17 UTC |
| `a7dc6d12a06c` | 2/5 | 3 | llm | 2026-08-28 18:24 UTC |
| `5977f39c5090` | 1/5 | 1 | llm | 2026-08-29 12:37 UTC |

</details>

<sub>Reviewed `3b06629d393d` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 823s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

TimeToBuildBob:
@greptileai review

packages/gptodo/src/gptodo/auth.py
TimeToBuildBob: Fixed in a7dc6d12. The classifier now matches `authentication[ _](?:error|failed)`, so both the JSON underscore forms and prose "Authentication failed/error" trigger the retry. Tests cover both.

packages/gptodo/src/gptodo/subagent.py
TimeToBuildBob: Fixed in a7dc6d12. Background tmux retry is now gated on `backend == "claude"`. A Codex spawn still uses the pre-existing `claude -p` command but does not get a second dispatch. Added a regression test.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Self-merge not eligible on `a7dc6d12`. CI is green and the prior P1/P2 findings are fixed in this head, but `packages/gptodo/src/gptodo/auth.py` is a sensitive-path policy block — this PR needs a maintainer merge. Greptile was triggered on the current SHA; our reviewer is sweeping now. No merge attempted.

packages/gptodo/src/gptodo/subagent.py
```diff
@@ -373,9 +387,29 @@ def spawn_agent(
...
+                first_auth_failure
+                + "\n=== Retried after transient authentication failure ===\n"
+                + result.stdout
+                + "\n"
+                + result.stderr
+            )
 
-        # Save output
-        output_file.write_text(result.stdout + "\n" + result.stderr)
+        output_file.write_text(output)
```
TimeToBuildBob, line 429: ❌ **P1** — In the foreground retry block, if the second `subprocess.run` raises an exception (e.g., `subprocess.TimeoutExpired`), control jumps to the `except` handler and `output_file.write_text(output)` on line 412 is never executed. The first run had already produced an auth-failure output, so that output is lost — the session's output file remains empty and the session is marked failed. Before this PR, a single failing run would have written its output to the file. This breaks `check_session` and `get_session_output`, which read the output file to report status and diagnostics, and loses the original error message. The fix is to write the first failure output to the file before sleeping, or to wrap the retry so the file is written even if the retry raises. _(2/3 passes — pass 1 looked and disagreed)_

TimeToBuildBob: Fixed in 5977f39c. The first auth-failure output is now written to the session file before the retry spawn, so a TimeoutExpired on the second run cannot leave the file empty. Regression: `test_claude_foreground_preserves_first_auth_output_if_retry_times_out`.

TimeToBuildBob:
Fresh AI review posted on `a7dc6d12` (score 2/5). Addressed the real P1:

- Timeout on the retry spawn no longer drops the first auth-failure output (`5977f39c`).

Left untouched:

- `{auth_check}` in the tmux command is a Python f-string interpolation, not bash brace expansion. The classifier command is expanded before the shell sees it. No code change.
- `\b401\b` false-positive risk is the existing size-gated classifier tradeoff, not a new defect.

Self-merge is still blocked by the `packages/gptodo/src/gptodo/auth.py` sensitive-path policy. Needs a maintainer merge. No merge attempted.

packages/gptodo/src/gptodo/subagent.py
TimeToBuildBob: Rejected as a false positive.

`{auth_check}` is interpolated by the Python f-string on the line above, not by bash. After expansion the tmux command is:

```text
if [ "$EXIT_CODE" -ne 0 ] && '<python>' -m gptodo.auth --classify-file '<output>'; then
```

That is a real command in the `&&` list. Brace-expansion would only apply if the shell saw the literal `{auth_check}` token, which it never does.

No code change. The Claude-only background retry from a7dc6d12 still stands.

packages/gptodo/src/gptodo/auth.py
TimeToBuildBob: Accepted tradeoff, not a defect in this PR.

`is_transient_auth_death` is intentionally size-gated at 2 KiB so a long successful transcript that mentions `401` cannot trip a retry. A small non-auth failure that happens to contain `401` can still match; that is the documented classifier tradeoff, not a new hole.

The retry is Claude-only, one-shot, and the first output is preserved (foreground session file before retry; background `.first-auth-failure` copy). Tightening `\b401\b` further is a follow-up, not a blocker for this change.

TimeToBuildBob:
Re-review of `5977f39c` is blocked: OpenRouter REVIEW key is at its daily cap (403). No new review posted.

Disposed two leftover findings from the stale `a7dc6d12` review:
- `{auth_check}` P1 — false positive (Python f-string, not bash brace expansion)
- `\b401\b` P2 — accepted size-gated classifier tradeoff

The timeout P1 is already fixed in `5977f39c`; its thread stays open for the next review to auto-resolve. CI is green. Self-merge remains ineligible (`packages/gptodo/src/gptodo/auth.py` sensitive-path). No Greptile re-trigger (still awaiting initial review; no push this session). Needs a maintainer merge after the review key resets.

TimeToBuildBob:
HEAD is still `5977f39c`. Current-head AI review is blocked: `OPENROUTER_API_KEY_REVIEW` remaining=0 until 2026-08-29T00:00Z. Greptile still awaiting-initial-review. No code change.

Recheck already scheduled for 00:10Z. After reset: `ai-review.py gptme/gptme-contrib 1538 --engine llm`. Self-merge remains ineligible (`packages/gptodo/src/gptodo/auth.py` sensitive-path) — maintainer merge after the marker matches head. Do not retrigger Greptile unless HEAD changes.

packages/gptodo/src/gptodo/subagent.py
```diff
@@ -373,9 +387,32 @@ def spawn_agent(
             timeout=timeout if timeout > 0 else None,
             env=env,
         )
+        output = result.stdout + "\n" + result.stderr
+
+        if backend == "claude" and result.returncode != 0 and is_transient_auth_death(output):
```
TimeToBuildBob, line 406: ⚠️ **P2** — The `is_transient_auth_death` classifier can return True for any small output containing '401', even when the failure is unrelated to authentication. Because the retry triggers solely on that classifier and a non-zero exit code, a task that fails for a different reason but includes `401` in its output (e.g., the agent printed an HTTP status it received from another service) will be retried once and delayed by 5 seconds. This wastes a Claude execution and can duplicate side effects. The size gate is the only disambiguator, and it does not exclude short outputs that mention 401 in a non-auth context.

packages/gptodo/src/gptodo/subagent.py
```diff
@@ -242,7 +246,17 @@ def spawn_agent(
...
             )
-            shell_cmd = f'touch {safe_output}; tail -f {safe_output} & TAIL_PID=$!; {timeout_prefix}claude -p {model_arg} {sysprompt_arg} --dangerously-skip-permissions --tools default -- "$(cat {safe_prompt_file})" > {safe_output} 2>&1; echo "EXIT_CODE=$?" >> {safe_output}; kill $TAIL_PID 2>/dev/null'
+            claude_cmd = f'{timeout_prefix}claude -p {model_arg} {sysprompt_arg} --dangerously-skip-permissions --tools default -- "$(cat {safe_prompt_file})"'
+            if backend == "claude":
+                # Run the classifier in gptodo's current Python environment. The
+                # retry happens inside tmux so check_session keeps one session ID.
+                python = shlex.quote(sys.executable)
+                auth_check = f"{python} -m gptodo.auth --classify-file {safe_output}"
+                shell_cmd = f'touch {safe_output}; tail -f {safe_output} & TAIL_PID=$!; {claude_cmd} > {safe_output} 2>&1; EXIT_CODE=$?; if [ "$EXIT_CODE" -ne 0 ] && {auth_check}; then cp {safe_output} {safe_output}.first-auth-failure; sleep 5; {claude_cmd} > {safe_output} 2>&1; EXIT_CODE=$?; fi; echo "EXIT_CODE=$EXIT_CODE" >> {safe_output}; kill $TAIL_PID 2>/dev/null'
```
TimeToBuildBob: 🛑 **P0** — The background retry shell command is syntactically invalid for bash. The condition `&& {auth_check}` embeds the auth-check command inside a brace group without the required trailing semicolon before the closing brace (bash requires `{ cmd; }`, not `{ cmd }`). When the tmux session starts, the shell parses the entire command and hits a syntax error near `}`, so nothing executes: the background agent for backend="claude" never even starts. The test `test_claude_background_command_retries_auth_failure` only inspects the command string and does not run it, so it misses this. The fix is to drop the braces (e.g., `&& auth_check`) or add a semicolon inside the group. _(3/3 passes)_

TimeToBuildBob: This finding is a false positive. The `{auth_check}` token is a Python f-string interpolation field, not a literal bash brace group. When the command is rendered, it expands to the full classifier invocation:

```
if [ "$EXIT_CODE" -ne 0 ] && /home/bob/.venv/bin/python3 -m gptodo.auth --classify-file /tmp/...; then
```

`bash -n` on the fully-rendered command returns exit 0 (verified locally), so the syntax is valid and the background agent starts normally. The classifier here is invoked as a plain command in an `&&` list, not a brace group — the braces only exist in the source f-string and are consumed by Python before bash ever sees the command. The P0 is invalid; no code change needed.

TimeToBuildBob: Fixed in 3b06629d. Concatenated the classifier into the tmux command so the source no longer contains `{auth_check}` (a Python f-string field, not a bash brace group). Added `bash -n` on the rendered inner command. In-band review of this SHA: 0 findings.

TimeToBuildBob:
Pushed `3b06629d` to address the P0 on `{auth_check}`: concatenate the classifier so the tmux script is obviously not a bash brace group, and syntax-check it with `bash -n`.

In-band review of this SHA found 0 findings. Self-merge remains blocked by the `packages/gptodo/src/gptodo/auth.py` sensitive-path policy — needs a maintainer merge.

TimeToBuildBob:
CI is green on `3b06629d` (`CLEAN` / mergeable). Self-merge is still ineligible: `packages/gptodo/src/gptodo/auth.py` is a sensitive-path policy block, and the posted GitHub AI-review marker is still on `5977f39c`.

Kicked a current-head review (`ai-review-now`) so the marker can catch up. In-band review of this SHA already found 0 findings. No merge attempted — needs a maintainer merge.

TimeToBuildBob:
Automated merge handoff for `3b06629d393d`:

- CI gate: satisfied
- Review threads: clear
- `packages/gptodo/src/gptodo/auth.py`

ErikBjare:
@TimeToBuildBob Merged https://github.com/gptme/gptme-contrib/pull/1558 which I suppose is a dupe?

TimeToBuildBob:
Yes — #1558 is the same #1536 fix and supersedes this PR. It includes the foreground Claude retry/auth classification from #1538, plus the missing loop failure accounting and broader regression coverage that review surfaced there.

I’m closing #1538 as superseded rather than rebasing it. Thanks for catching the duplicate.

PR #1539, opened 2026-08-28
fix(gptodo): emit waiting_for/waiting_since on recurrence reset

TimeToBuildBob:
## Summary

#1443 landed `state: waiting` + `wait:` + `wait_kind: machine` on recurrence reset, then (in the same squash) stripped `waiting_for`/`waiting_since` so a leftover human blocker would not trap the next cycle.

The task frontmatter validator now requires both fields on `state=waiting`. Completing a `recur:` task through the documented path therefore fails pre-commit, and sessions abandon the close — that is the leaked-recurrence-gate root cause.

This writes a **time-gate** `waiting_for` instead of stripping:

```yaml
waiting_for: next recurrence gate (wait: 2026-09-04)
waiting_since: '<now>'
```

That string is classified as a time gate by the auto-releaser, so the task still re-surfaces when `wait:` expires. A leftover human blocker (`John to review the PR`) is replaced, not preserved.

## Test plan

- [x] `uv run pytest tests/test_recur.py tests/test_wait_recur.py` — 47 passed
- [x] `make typecheck` in `packages/gptodo` — clean
- [x] Emitted file passes `validate_task_frontmatter.py`; the stripped form fails with `state=waiting requires waiting_for/waiting_since`

### DIFF
--- a/packages/gptodo/src/gptodo/cli.py
+++ b/packages/gptodo/src/gptodo/cli.py
@@ -2956,7 +2956,7 @@ def edit(task_ids, set_fields, add_fields, remove_fields, set_subtask, force):
         # Strip now-stale actionable/blocker metadata when the edit lands the task
         # in a terminal state (TASKS.md best-practice #7: terminal tasks must not
         # keep next_action/waiting_for/waiting_since/wait). Recurring tasks reset to
-        # todo further below and must keep these fields, so skip when recur is set.
+        # waiting further below and must keep these fields, so skip when recur is set.
         # tracking_issue / upstream_coordination_id are intentionally preserved for
         # permanent traceability.
         # cancelled is terminal regardless of recur. A done task only escapes the
@@ -3085,15 +3085,20 @@ def edit(task_ids, set_fields, add_fields, remove_fields, set_subtask, force):
 
                     current_wait = parse_wait(post.metadata.get("wait"))
                     next_wait = advance_wait(current_wait, recur)
+                    wait_iso = next_wait.isoformat()
                     post.metadata["state"] = "waiting"
-                    post.metadata["wait"] = next_wait.isoformat()
+                    post.metadata["wait"] = wait_iso
                     post.metadata["wait_kind"] = "machine"
-                    # Strip any pre-existing waiting_for/waiting_since: the wait: date IS the
-                    # machine gate. Leaving them would trap the task permanently — task_has_waiting_blocker
-                    # treats any waiting_for as a human blocker that requires explicit resolution,
-                    # so the task would never auto-surface after the time gate expires.
-                    post.metadata.pop("waiting_for", None)
-                    post.metadata.pop("waiting_since", None)
+                    # validate_task_frontmatter requires waiting_for + waiting_since
+                    # on state=waiting. A leftover human blocker (e.g. "John to review")
+                    # would trap the task after the gate expires, so REPLACE it with a
+                    # recurrence-gate string. "next recurrence gate (wait: <date>)" is
+                    # classified as a time_gate by the auto-releaser, so the task
+                    # re-surfaces when wait: expires.
+                    post.metadata["waiting_for"] = f"next recurrence gate (wait: {wait_iso})"
+                    post.metadata["waiting_since"] = datetime.now(timezone.utc).isoformat(
+                        timespec="seconds"
+                    )
                     if not _completed_explicitly_set_global:
                         post.metadata.pop("completed", None)
                     with open(task.path, "w") as f:
--- a/packages/gptodo/src/gptodo/cli.py
+++ b/packages/gptodo/src/gptodo/cli.py
@@ -3520,12 +3520,7 @@ def ready(state, output_json, output_jsonl, use_cache, pool_filter, exclude_pool
             task
             for task in all_tasks
             if task.state in ["backlog", "todo", "active"]
-            or (
-                task.state == "waiting"
-                and task.metadata.get("wait_kind") == "machine"
-                and not task_is_waiting_for_date(task)
-                and not task.metadata.get("waiting_for")
-            )
+            or (task.state == "waiting" and not task_has_waiting_blocker(task))
         ]
 
     # Apply pool filter (default: general only; --pool all to see every pool)
@@ -3761,12 +3756,7 @@ def next_(output_json, use_cache, pool_filter, exclude_pool, limit, order):
         task
         for task in all_tasks
         if task.state in ["backlog", "todo", "active"]
-        or (
-            task.state == "waiting"
-            and task.metadata.get("wait_kind") == "machine"
-            and not task_is_waiting_for_date(task)
-            and not task.metadata.get("waiting_for")
-        )
+        or (task.state == "waiting" and not task_has_waiting_blocker(task))
     ]
 
     # Apply pool filter (default: general only; --pool all to see every pool)
--- a/packages/gptodo/src/gptodo/cli.py
+++ b/packages/gptodo/src/gptodo/cli.py
@@ -2993,6 +2993,18 @@ def edit(task_ids, set_fields, add_fields, remove_fields, set_subtask, force):
                 post.metadata.pop("waiting_for", None)
                 post.metadata.pop("waiting_since", None)
                 post.metadata.pop("wait_kind", None)
+        # Drop generated recurrence-gate waiting_for whenever the final state
+        # is not waiting, even if this edit did not touch wait. A state-only
+        # `--set state todo` used to skip the wait-sync (gated on wait_was_set)
+        # and leave waiting_for, which task_has_waiting_blocker treats as a
+        # human blocker on non-waiting states (P1 on gptme/gptme-contrib#1539
+        # / 69034e96). Keep wait: — the user did not ask to clear the date.
+        if post.metadata.get("state") != "waiting" and is_generated_recurrence_waiting_for(
+            post.metadata.get("waiting_for", "")
+        ):
+            post.metadata.pop("waiting_for", None)
+            post.metadata.pop("waiting_since", None)
+            post.metadata.pop("wait_kind", None)
         # Auto-set waiting_since only when THIS edit explicitly sets state to waiting
         # AND waiting_for is either already present or being set in the same edit.
         # Guarding on waiting_for prevents an injected waiting_since from triggering
--- a/packages/gptodo/src/gptodo/cli.py
+++ b/packages/gptodo/src/gptodo/cli.py
@@ -2972,7 +2972,10 @@ def edit(task_ids, set_fields, add_fields, remove_fields, set_subtask, force):
                 explicit_state = any(
                     op == "set" and field == "state" for op, field, _value in changes
                 )
-                if not explicit_state:
+                # Only default waiting → todo. A leftover recurrence-gate
+                # string on done/cancelled/active must not reopen or demote
+                # the task (P1 on gptme/gptme-contrib#1539 / ecb15242).
+                if not explicit_state and post.metadata.get("state") == "waiting":
                     post.metadata["state"] = "todo"
                 post.metadata.pop("waiting_for", None)
                 post.metadata.pop("waiting_since", None)
--- a/packages/gptodo/src/gptodo/cli.py
+++ b/packages/gptodo/src/gptodo/cli.py
@@ -2964,14 +2964,28 @@ def edit(task_ids, set_fields, add_fields, remove_fields, set_subtask, force):
             )
             if recurrence_gate and wait_set is None:
                 # Clearing a generated recurrence gate means it is no longer
-                # waiting. Remove the generated blocker metadata as one unit so
-                # the task cannot be left permanently blocked without a date.
-                post.metadata["state"] = "todo"
+                # a machine time-gate. Remove the generated blocker metadata as
+                # one unit so the task cannot stay permanently blocked without
+                # a date. Only default state to todo when this edit did not
+                # set state — `--set wait none --set state done` must not
+                # overwrite the explicit state (P1 on gptme/gptme-contrib#1539).
+                explicit_state = any(
+                    op == "set" and field == "state" for op, field, _value in changes
+                )
+                if not explicit_state:
+                    post.metadata["state"] = "todo"
                 post.metadata.pop("waiting_for", None)
                 post.metadata.pop("waiting_since", None)
                 post.metadata.pop("wait_kind", None)
             elif recurrence_gate and post.metadata.get("state") == "waiting":
                 post.metadata["waiting_for"] = f"next recurrence gate (wait: {wait_set})"
+            elif recurrence_gate:
+                # Wait changed but the task is no longer waiting. Drop the
+                # generated recurrence string so it cannot trap a todo/done
+                # task as a leftover human-looking blocker.
+                post.metadata.pop("waiting_for", None)
+                post.metadata.pop("waiting_since", None)
+                post.metadata.pop("wait_kind", None)
         # Auto-set waiting_since only when THIS edit explicitly sets state to waiting
         # AND waiting_for is either already present or being set in the same edit.
         # Guarding on waiting_for prevents an injected waiting_since from triggering
--- a/packages/gptodo/src/gptodo/cli.py
+++ b/packages/gptodo/src/gptodo/cli.py
@@ -2941,10 +2941,13 @@ def edit(task_ids, set_fields, add_fields, remove_fields, set_subtask, force):
         # because state still holds the pre-edit value when wait is processed
         # (P1 on gptme/gptme-contrib#1539). Combined with the existing
         # transitioning_to_waiting check just below, this covers both orderings.
+        # Last --set wait wins (same as the apply loop). next() without
+        # reversed() would rewrite waiting_for from the first wait while
+        # metadata['wait'] holds a later one, permanently trapping the task.
         wait_set = next(
             (
                 value
-                for op, field, value in changes
+                for op, field, value in reversed(changes)
                 if op == "set" and field == "wait" and value is not None
             ),
             None,
--- a/packages/gptodo/src/gptodo/cli.py
+++ b/packages/gptodo/src/gptodo/cli.py
@@ -2944,23 +2944,33 @@ def edit(task_ids, set_fields, add_fields, remove_fields, set_subtask, force):
         # Last --set wait wins (same as the apply loop). next() without
         # reversed() would rewrite waiting_for from the first wait while
         # metadata['wait'] holds a later one, permanently trapping the task.
-        # Include None: `--set wait NEW --set wait none` pops wait, and a
-        # non-None filter would still rewrite waiting_for to NEW, leaving a
-        # recurrence-gate string with no wait date (P1 on #1539).
-        wait_set = next(
-            (value for op, field, value in reversed(changes) if op == "set" and field == "wait"),
-            None,
+        wait_change = next(
+            (
+                (True, value)
+                for op, field, value in reversed(changes)
+                if op == "set" and field == "wait"
+            ),
+            (False, None),
         )
-        if wait_set is not None:
+        wait_was_set, wait_set = wait_change
+        if wait_was_set:
             import re as _re
 
             _wf = post.metadata.get("waiting_for", "")
-            if (
+            recurrence_gate = (
                 post.metadata.get("wait_kind") == "machine"
-                and post.metadata.get("state") == "waiting"
                 and isinstance(_wf, str)
                 and _re.match(r"^next recurrence gate \(wait: ", _wf)
-            ):
+            )
+            if recurrence_gate and wait_set is None:
+                # Clearing a generated recurrence gate means it is no longer
+                # waiting. Remove the generated blocker metadata as one unit so
+                # the task cannot be left permanently blocked without a date.
+                post.metadata["state"] = "todo"
+                post.metadata.pop("waiting_for", None)
+                post.metadata.pop("waiting_since", None)
+                post.metadata.pop("wait_kind", None)
+            elif recurrence_gate and post.metadata.get("state") == "waiting":
                 post.metadata["waiting_for"] = f"next recurrence gate (wait: {wait_set})"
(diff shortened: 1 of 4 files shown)


packages/gptodo/src/gptodo/cli.py
```diff
@@ -3085,15 +3085,20 @@ def edit(task_ids, set_fields, add_fields, remove_fields, set_subtask, force):
...
-                    # so the task would never auto-surface after the time gate expires.
-                    post.metadata.pop("waiting_for", None)
-                    post.metadata.pop("waiting_since", None)
+                    # validate_task_frontmatter requires waiting_for + waiting_since
+                    # on state=waiting. A leftover human blocker (e.g. "John to review")
+                    # would trap the task after the gate expires, so REPLACE it with a
+                    # recurrence-gate string. "next recurrence gate (wait: <date>)" is
+                    # classified as a time_gate by the auto-releaser, so the task
+                    # re-surfaces when wait: expires.
+                    post.metadata["waiting_for"] = f"next recurrence gate (wait: {wait_iso})"
```
TimeToBuildBob, line 3180: ❌ **P1** — The recurrence reset now writes a non-empty `waiting_for` string ('next recurrence gate (wait: <wait_iso>)') to satisfy the frontmatter validator. However, the `ready` and `next` commands—and `task_has_waiting_blocker`—treat any non-empty `waiting_for` as a human blocker. In cli.py, the 'both' state filter requires `not task.metadata.get("waiting_for")` and `next`'s workable_tasks filter has the same condition. As a result, after a recurring task resets to waiting with a future wait date, once that date passes the task still carries a non-empty `waiting_for`, so it is never considered ready and never surfaces in `gptodo ready` or `gptodo next`. The PR's claim that the auto-releaser classifies this string as a time gate does not affect these built-in commands, which do not special-case the string. This breaks recurring task workflows: tasks stop appearing after their gate expires. A fix would require the readiness logic to recognize this specific string as a time gate (e.g., parse and compare with wait), or to avoid setting `waiting_for` at all (and adjust the validator).

TimeToBuildBob:
## 🤖 AI code review

This PR changes how gptodo handles recurrence resets and waiting metadata. When a recurring task is marked done, it now writes a generated 'next recurrence gate (wait: &lt;date&gt;)' string into waiting_for (instead of stripping waiting_for/waiting_since) and sets waiting_since fresh, so the frontmatter validator accepts state=waiting. It also adds a wait-sync block in the edit command that keeps waiting_for synchronized with --set wait operations on recurrence-gate tasks, and updates task_has_waiting_blocker to treat an exact generated recurrence-gate string as a machine time gate (so an expired gate no longer blocks the task). The ready and next commands now filter waiting tasks via task_has_waiting_blocker instead of the previous inline checks.

- `packages/gptodo/src/gptodo/cli.py` — Adds a wait-sync block in edit that rewrites waiting_for/wait_kind when --set wait is used on a recurrence-gate task, and changes the recurrence-reset path to emit a generated waiting_for instead of stripping it.
- `packages/gptodo/src/gptodo/utils.py` — Adds is_generated_recurrence_waiting_for() and updates task_has_waiting_blocker to release a machine time-gate when waiting_for exactly matches the generated string for the current wait.
- `packages/gptodo/tests/test_recur.py` — Updates the recurrence-reset test to assert that waiting_for and waiting_since are emitted and contain a recurrence-gate string.
- `packages/gptodo/tests/test_wait_recur.py` — Adds many tests for wait-sync behavior, recurrence-gate release, and stale generated strings, and updates the existing reset test.

</details>

| `cd05faf98f33` | 3/5 | 1 | llm | 2026-08-28 07:06 UTC |
| `ec9b6b35801d` | 2/5 | 3 | llm | 2026-08-28 11:06 UTC |
| `c9d984d9dfb6` | 3/5 | 1 | llm | 2026-08-28 13:22 UTC |
| `257a0d8f5ffa` | 3/5 | 1 | llm | 2026-08-28 17:31 UTC |
| `b859fbf43ff7` | 3/5 | 1 | llm | 2026-08-29 01:46 UTC |
| `69034e962f46` | 3/5 | 1 | llm | 2026-08-29 13:40 UTC |
| `e52af179f211` | 3/5 | 1 | llm | 2026-08-29 13:59 UTC |
| `7a04659330cc` | 3/5 | 1 | llm | 2026-08-29 14:49 UTC |

</details>

<sub>Reviewed `ec95e5e08eac` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 280s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

packages/gptodo/src/gptodo/cli.py
TimeToBuildBob: Fixed in `ec9b6b35`. `task_has_waiting_blocker` now treats the generated recurrence description as the same machine time gate only when it exactly matches the task's current `wait:` value; human/stale descriptions and probe gates remain blocking. `gptodo ready` and `gptodo next` now use that canonical blocker check. Added regressions proving an expired recurrence surfaces in both commands and a mismatched recurrence description stays blocked.

Verified locally: 65 targeted tests passed; scoped pre-commit/typecheck passed. CI is running.

packages/gptodo/src/gptodo/utils.py
```diff
@@ -640,22 +640,26 @@ def __str__(self) -> str:
 def task_has_waiting_blocker(task: "TaskInfo") -> bool:
     """Return True when a task is explicitly waiting on an external condition."""
     state = normalize_state(task.state or "", warn=False) if task.state else ""
+    waiting_for = str(task.metadata.get("waiting_for") or "").strip()
+    recurrence_gate = task.wait is not None and waiting_for == (
```
TimeToBuildBob, line 658: ❌ **P1** — What: A task reset via recurrence now carries `waiting_for = "next recurrence gate (wait: <wait>)"` and `wait_kind = "machine"`. `task_has_waiting_blocker` releases that task only when `waiting_for` exactly equals `"next recurrence gate (wait: {task.wait.isoformat()})"` and the wait has expired. Why: The `edit` command does not touch `waiting_for` when the user later changes `wait` on such a task. So after a recurrence reset, running `gptodo edit <task> --set wait 2026-02-01` leaves `waiting_for` pointing at the old date, and `recurrence_gate` becomes False. What breaks: When the new wait expires, `task_has_waiting_blocker` returns True (state==waiting, no match), so the task never surfaces in `ready`/`next` again, permanently trapping it. The stale-description test (`test_stale_recurrence_description_stays_blocked`) encodes exactly this as desired, but it fails to consider that a legitimate manual wait adjustment produces the same stale string without generating a warning. A user postponing a recurring task's gate unintentionally blocks it forever.

TimeToBuildBob:
Fixed both P1 findings from the AI review on `ec9b6b35`:

1. **Stale `probe` cleared on recurrence reset** — `post.metadata.pop("probe", None)` added in the recurrence reset block so `task_has_waiting_blocker` uses the new `wait:` date gate rather than the old probe.

2. **`waiting_for` kept in sync on `--set wait`** — when `--set wait` targets a recurrence-gate task (state=waiting, wait_kind=machine, waiting_for matches pattern), the `waiting_for` string is updated to reference the new date so the pattern still matches when the gate expires.

Both fixes covered by new tests in `test_wait_recur.py`.

@TimeToBuildBob review

packages/gptodo/src/gptodo/cli.py
```diff
@@ -2930,6 +2930,21 @@ def edit(task_ids, set_fields, add_fields, remove_fields, set_subtask, force):
...
+                    # keep waiting_for in sync with the new wait value so
+                    # task_has_waiting_blocker can still match the recurrence-gate pattern.
+                    # Without this, a manual wait adjustment permanently traps the task.
+                    if field == "wait" and value is not None:
+                        import re as _re
+
+                        _wf = post.metadata.get("waiting_for", "")
+                        if (
+                            post.metadata.get("wait_kind") == "machine"
+                            and post.metadata.get("state") == "waiting"
```
TimeToBuildBob: ❌ **P1** — When --set wait and --set state waiting are combined in a single edit invocation, the waiting_for sync fails. The update logic checks `post.metadata.get("state") == "waiting"` at the moment the wait change is processed, but if the state change appears later in the changes list, the metadata still holds the pre-edit state (e.g. active/todo), so the condition is False and waiting_for is not rewritten. The task ends up in state=waiting with a wait: value that does not match the waiting_for string (which may be absent or a stale recurrence gate). Once the new wait expires, task_has_waiting_blocker sees a non-empty waiting_for that is not the exact recurrence gate for the current wait, classifies it as a human blocker, and the task never auto-surfaces from `ready`/`next` — the exact permanent-trap this PR is meant to prevent. The test test_set_wait_on_recurrence_gate_updates_waiting_for only covers the case where the task is already waiting, so it misses this ordering bug. Concrete input: `gptodo edit task --set wait 2030-06-15 --set state waiting` on a recurrence-gate task with stale waiting_for. _(3/3 passes)_

TimeToBuildBob: Fixed in `257a0d8f`. The wait-sync now runs after every `--set` in the same invocation, so `--set wait X --set state waiting` (either order) rewrites `waiting_for` to match the new `wait:`. Regression: `test_set_wait_and_state_waiting_together_updates_waiting_for`. CI is green on this HEAD.

packages/gptodo/src/gptodo/cli.py
```diff
@@ -2930,6 +2930,36 @@ def edit(task_ids, set_fields, add_fields, remove_fields, set_subtask, force):
...
+        # keep waiting_for in sync with the new wait value so
+        # task_has_waiting_blocker can still match the recurrence-gate pattern.
+        # Without this, a manual wait adjustment permanently traps the task.
+        #
+        # Must run AFTER every --set in this invocation is applied. Doing it
+        # inside the per-field loop misses `--set wait X --set state waiting`
+        # because state still holds the pre-edit value when wait is processed
+        # (P1 on gptme/gptme-contrib#1539). Combined with the existing
+        # transitioning_to_waiting check just below, this covers both orderings.
+        wait_set = next(
```
TimeToBuildBob: ❌ **P1** — In the new `--set wait` synchronization block, `wait_set` is extracted from `changes` using `next()` which returns the FIRST matching `--set wait` value when multiple `--set wait` options are supplied in one invocation. The per-field loop applies changes in order, so the LAST `wait` value written to `post.metadata['wait']` is the effective one. As a result, `waiting_for` is rewritten to reference the first wait value while `wait` holds a different value, so `task_has_waiting_blocker` no longer sees a matching recurrence-gate string and treats the task as permanently blocked even after the new wait expires. For example, `gptodo edit task --set wait 2030-06-15 --set wait 2030-07-01` leaves `wait` as 2030-07-01 but `waiting_for` as "next recurrence gate (wait: 2030-06-15)". The task can only be fixed by manually editing the file. Use `reversed(changes)` in the generator to take the last `wait` set, or read the final value from `post.metadata` after the apply loop. _(3/3 passes)_

TimeToBuildBob: Fixed in `20b825dc`. The post-loop sync now takes the last `--set wait` (`reversed(changes)`), matching the apply-loop overwrite. Regression: `test_last_set_wait_wins_in_waiting_for_sync`.

TimeToBuildBob:
Pushed `20b825dc` — last `--set wait` now wins in the `waiting_for` sync (P1 on `257a0d8f`). HEAD re-review blocked on OpenRouter daily key limit (403). Next sweep after quota reset should review `20b825dc`.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
HEAD is still `20b825dc`. Current-head AI review is blocked: `OPENROUTER_API_KEY_REVIEW` remaining=0 until 2026-08-29T00:00Z. Greptile trigger from 18:52Z is still awaiting-initial-review. No code change.

Recheck scheduled for 00:10Z. After reset: `ai-review.py gptme/gptme-contrib 1539 --engine llm`, then `self-merge-if-eligible` if undegraded 5/5. Do not retrigger Greptile unless HEAD changes.

packages/gptodo/src/gptodo/cli.py
```diff
@@ -2930,6 +2930,39 @@ def edit(task_ids, set_fields, add_fields, remove_fields, set_subtask, force):
...
+        # (P1 on gptme/gptme-contrib#1539). Combined with the existing
+        # transitioning_to_waiting check just below, this covers both orderings.
+        # Last --set wait wins (same as the apply loop). next() without
+        # reversed() would rewrite waiting_for from the first wait while
+        # metadata['wait'] holds a later one, permanently trapping the task.
+        wait_set = next(
+            (
+                value
+                for op, field, value in reversed(changes)
+                if op == "set" and field == "wait" and value is not None
```
TimeToBuildBob: ❌ **P1** — The wait-sync logic in edit() selects the last non-None --set wait value to rewrite waiting_for, but it ignores a subsequent `--set wait none`. If a user postpones a recurrence-gate task with `--set wait NEW --set wait none`, the apply loop sets wait to NEW and then removes it, so post.metadata has no 'wait'. The sync still finds the last non-None value (NEW) and sets waiting_for to 'next recurrence gate (wait: NEW)'. The task is left with waiting_for but no wait field. In task_has_waiting_blocker, wait_kind is machine but task.wait is None, so the first condition fails and state==waiting returns True — the task is permanently blocked with no wait date to ever expire. The fix is to stop filtering out None values in the generator, so a trailing clear yields wait_set=None and skips the rewrite. _(2/3 passes — pass 1 looked and disagreed)_

TimeToBuildBob: Fixed in `3fd110ee`. Last `--set wait` now includes `none`, so `--set wait NEW --set wait none` pops `wait` and does not rewrite `waiting_for` to the cleared date. Regression: `test_trailing_set_wait_none_does_not_rewrite_waiting_for`.

TimeToBuildBob:
Pushed `3fd110ee` — last `--set wait none` now wins in the `waiting_for` sync (P1 on `20b825dc`). Trailing clear no longer rewrites `waiting_for` to a date that was just popped.

@TimeToBuildBob review

packages/gptodo/src/gptodo/cli.py
TimeToBuildBob: Follow-up fixed in `e71b18de`: clearing a generated recurrence gate now moves the task to `todo` and removes `waiting_for`, `waiting_since`, and `wait_kind` together. The prior fix prevented rewriting the cleared date but still left a permanent blocker, which the current-head review correctly caught. Regression updated to assert the task is actionable.

TimeToBuildBob:
Pushed `ecb15242` — `--set wait none` no longer overwrites an explicit `--set state` (P1 on `e71b18de`). A leftover recurrence `waiting_for` is also dropped when the task is no longer waiting.

@TimeToBuildBob review

TimeToBuildBob:
Pushed `b859fbf4` — `--set wait none` no longer reopens done/cancelled/active tasks that still carry a leftover recurrence-gate `waiting_for` (P1 on `ecb15242`). The todo default now applies only when the task is currently waiting.

Current-head review requested for `b859fbf4`. CI is running; will not merge until green + undegraded review.

TimeToBuildBob:
Pushed `69034e96` — leftover P1 from the timed-out session at HEAD `b859fbf4`:

1. `--set wait none --set state waiting` no longer pops `waiting_for`/`waiting_since` (state=waiting requires both).
2. Generated recurrence-gate `waiting_for` matching is exact-string, so a human suffix is not overwritten.

Timeout cause: the prior worker staged this fix in `/tmp/worktrees/contrib-1539-31f5` and died before commit/push. Review key is *not* capped (`OPENROUTER_API_KEY_REVIEW` remaining ~$4); the 04:04 circuit breaker is the autonomous key.

Current-head AI review dispatched for `69034e96`. CI running. Will not merge until green + undegraded review. Do not retrigger Greptile unless HEAD changes again.

TimeToBuildBob:
Pushed `e52af179` — leftover P1 from the current-head review of `69034e96`:

`--set state todo` (no `--set wait`) no longer leaves the generated recurrence-gate `waiting_for` on the task. Wait-sync used to run only when wait was set, so a state-only change trapped the task as a human blocker. Human-suffixed `waiting_for` is still preserved.

Regression: `test_set_state_todo_on_recurrence_gate_drops_waiting_for`. 64 tests in `test_wait_recur.py` + `test_recur.py`. CI running; will not merge until green + undegraded current-head review.

@TimeToBuildBob review

packages/gptodo/src/gptodo/utils.py
```diff
@@ -637,25 +637,42 @@ def __str__(self) -> str:
 # =============================================================================
 
 
+# Exact generated recurrence-gate waiting_for. Prefix-only matching would treat
+# a human suffix ("... (wait: 2025-01-01) - confirm with Bob") as a machine
+# gate and silently overwrite it on --set wait (gptme/gptme-contrib#1539).
+_GENERATED_RECURRENCE_WAITING_FOR = re.compile(r"^next recurrence gate \(wait: .+\)$")
```
TimeToBuildBob: ❌ **P1** — The regex in is_generated_recurrence_waiting_for uses `.+` inside the parentheses, which matches any characters, including spaces and hyphens. As a result, a human-written waiting_for like 'next recurrence gate (wait: 2025-01-01 - confirm with Bob)' is incorrectly classified as a generated recurrence gate. When the user later runs --set wait with a new date, the wait-sync block overwrites waiting_for with the new generated string, silently destroying the human note. Similarly, if the state is changed to non-waiting, the second block pops waiting_for entirely. The existing test test_set_wait_does_not_rewrite_suffixed_recurrence_waiting_for only covers a suffix after the closing parenthesis, not a note inside the parentheses, so this case goes unguarded. The fix is to make the regex disallow spaces, e.g. by using `\S+` instead of `.+`, so only a bare ISO date (which never contains spaces) matches. _(2/3 passes — pass 3 looked and disagreed)_

TimeToBuildBob: Fixed in `7a046593`. Generated recurrence-gate matching now uses `\S+` (ISO dates have no spaces), so `next recurrence gate (wait: DATE - note)` is not classified as generated and `--set wait` leaves the human note. Regression: `test_set_wait_does_not_rewrite_inner_paren_recurrence_waiting_for` plus the inner-paren case in `test_generated_recurrence_waiting_for_is_exact_string`. 65 tests in `test_wait_recur.py` + `test_recur.py`.

TimeToBuildBob:
Pushed `7a046593` — leftover P1 from the current-head review of `e52af179`:

Generated recurrence-gate matching now uses `\S+` instead of `.+`. A human note inside the parentheses (`next recurrence gate (wait: DATE - note)`) is no longer classified as generated, so `--set wait` leaves it. ISO dates from `isoformat()` have no spaces.

Regression: `test_set_wait_does_not_rewrite_inner_paren_recurrence_waiting_for`. 65 tests in `test_wait_recur.py` + `test_recur.py`. CI running; will not merge until green + undegraded current-head review.

@TimeToBuildBob review

packages/gptodo/src/gptodo/cli.py
```diff
@@ -2930,6 +2931,80 @@ def edit(task_ids, set_fields, add_fields, remove_fields, set_subtask, force):
...
+                # string on done/cancelled/active must not reopen or demote
+                # the task (P1 on gptme/gptme-contrib#1539 / ecb15242).
+                if not explicit_state and post.metadata.get("state") == "waiting":
+                    post.metadata["state"] = "todo"
+                if post.metadata.get("state") == "waiting":
+                    # `--set wait none --set state waiting` must not pop
+                    # waiting_for/waiting_since: state=waiting requires both
+                    # (P1 on gptme/gptme-contrib#1539 / b859fbf4). Drop
+                    # wait_kind so this is no longer a machine time-gate.
+                    post.metadata.pop("wait_kind", None)
```
TimeToBuildBob, line 2988: ❌ **P1** — When `--set wait none --set state waiting` is run on a recurrence-gate task, the edit block enters the `recurrence_gate and wait_set is None` branch, and because the explicit state is `waiting`, it pops `wait_kind` (line 2982) but leaves `waiting_for` as the generated recurrence-gate string (e.g. `next recurrence gate (wait: 2025-01-01)`). A later `--set wait <new>` on that same task sets `post.metadata["wait"]` to the new date but does not re-add `wait_kind`, because `recurrence_gate` is now False (the `wait_kind == "machine"` check fails). The task stays in `state: waiting` with a `waiting_for` string that no longer matches a machine gate, so `task_has_waiting_blocker` returns True and the task never auto-surfaces even after the new wait date expires. The user must manually clear `waiting_for` or set `wait_kind` again, which is not documented and contradicts the PR's goal of keeping recurrence gates releasable. _(2/2 passes)_

TimeToBuildBob: Fixed in `ec95e5e0`. `--set wait none --set state waiting` still pops `wait_kind` (state=waiting requires `waiting_for`), but leftover generated `waiting_for` with no `wait_kind` is still treated as a recurrence gate. A later `--set wait NEW` now rewrites `waiting_for` to the new date and restores `wait_kind=machine`, so the task stays a releasable time gate.

Regression: `test_set_wait_after_wait_none_staying_waiting_restores_recurrence_gate`.

TimeToBuildBob:
Pushed `ec95e5e0` — leftover P1 from the current-head review of `7a046593`:

`--set wait none --set state waiting` pops `wait_kind` but keeps the generated recurrence `waiting_for`. A later `--set wait NEW` used to skip wait-sync because `wait_kind` was gone, so the old generated string trapped the task after the new date expired. Leftover generated `waiting_for` with no `wait_kind` is now still a recurrence gate, and `--set wait NEW` restores `wait_kind=machine`.

Regression: `test_set_wait_after_wait_none_staying_waiting_restores_recurrence_gate`. 66 tests in `test_wait_recur.py` + `test_recur.py`. CI is green on `ec95e5e0`. Running current-head AI review now (marker SHA `7a046593` is stale vs HEAD). Will not merge until undegraded current-head review.

PR #1547, opened 2026-08-29
test(gptme-activity-summary): regression tests for deepseek preamble parse + prompt narrative contract

TimeToBuildBob:
## Problem

The 2026-08-28 daily-summary failure was a subscription-fragility double-fault: the deepseek-v4-flash reasoning model emits a JSON-shaped thinking preamble (`{"thinking": "..."}`) before the real answer. The gptme fallback parser (`_extract_assistant_text`, shipped in #1489) now correctly scans all JSON-parseable candidates and prefers the recognised summary key (last-wins), which closes the original parse gap on master.

Two deliverables from the tracked issue were still missing.

## What this PR does

1. **Regression test suite** (`tests/test_gptme_backend.py`, 9 tests) — locks in the deepseek preamble behaviour: preamble-then-answer, preamble that echoes the `narrative` key (last-wins), preamble wrapped in prose + code-block answer, list-content messages, and best-effort disabled-path defaults.
2. **Prompt-level narrative contract** — the daily prompt now explicitly requires the *first non-empty assistant message* to be the final JSON object with a top-level `narrative` field, and forbids emitting any thinking preamble before the JSON.

## Verification

- `pytest packages/gptme-activity-summary/tests/` → **237 passed** (9 new + 77 cc_backend + rest)
- `mypy` on gptme_backend.py + cc_backend.py → clean
- Pre-commit (`prek`) passed

## Notes

The parser widening itself is already on master (commits 8c368bb8, bb2fc5ee, 6c604c3a). This PR is the missing test + prompt-contract half of the gptme-fallback-parse-deepseek-preamble task.

PR #1548, opened 2026-08-29
fix(gptme-activity-summary): strip think tags + raw_decode JSON in gptme fallback

TimeToBuildBob:
## Summary

Fixes the root cause of 0/12 daily summary failures when the gptme fallback
is active and deepseek-v4-flash-0731 emits reasoning-model output.

## Root cause

DeepSeek reasoning models prefix their response with a `<think>…</think>`
block. That block may itself contain JSON-shaped text (e.g. internal
reasoning about what JSON to emit). Two bugs combined to break the fallback:

1. **Greedy regex in `cc_backend.extract_json_from_response`** — the
   pattern `{[\s\S]*}` grabbed from the first `{` inside the think block
   all the way to the last `}` in the real answer, producing one large
   unparseable blob. Fix: replaced with a `json.JSONDecoder().raw_decode()`
   scan that finds all complete, valid JSON objects in the response, then
   picks the last one that has a recognised summary key (`narrative` /
   `month_narrative`).

2. **Think blocks not stripped in `gptme_backend._extract_assistant_text`**
   — even after assembling assistant-message content from the NDJSON stream,
   embedded JSON inside think tags could pass down to
   `extract_json_from_response` and trigger the same failure. Fix: added
   `_strip_think_tags()` (non-greedy DOTALL regex) applied to each content
   chunk before accumulation.

Also improves preamble handling: when no JSON candidate has a summary key
(e.g. only a `{"thinking": "…"}` preamble was emitted), the backend now
prefers non-JSON content (the prose answer) over returning the preamble.

## Tests

14 new unit tests in `tests/test_gptme_backend.py`:

- **`_strip_think_tags`**: basic, braces-inside-think, multiline, noop, multiple blocks
- **`_extract_assistant_text`**: think-in-content, braces-confusion, pure-think,
  empty input, preamble-then-answer, preamble-wins regression, non-json-fallback,
  multipart-content-list

242 tests pass total (14 new + 228 pre-existing, including the 9 from PR #1547).

## Relation to #1547

PR #1547 (session e972) adds a prompt-level instruction
(`The FIRST non-empty assistant message must be the final JSON object`) and
regression tests for the existing parser. This PR is complementary: it fixes
the parser so it handles models that ignore the instruction (or emit think
blocks anyway), making the fallback robust against both prompt non-compliance
and the greedy-regex bug.

Related task: `activity-summary-fallback-shipped-but-never-rescues`

### DIFF
--- a/packages/gptme-activity-summary/src/gptme_activity_summary/gptme_backend.py
+++ b/packages/gptme-activity-summary/src/gptme_activity_summary/gptme_backend.py
@@ -122,6 +122,21 @@ def call_gptme(prompt: str, timeout: int = 120) -> str:
 
 
 _THINK_TAG_RE = re.compile(r"<think>[\s\S]*?</think>", re.DOTALL)
+# gptme's no-tool-call auto-reply. Shared prefix of both variants in
+# gptme.tools.complete ("No tool call detected" / "... in last message").
+_GPTME_TRAILER_MARKER = "No tool call detected"
+
+
+def _is_gptme_trailer(text: str) -> bool:
+    """True if this assistant message is gptme's no-tool-call auto-reply.
+
+    After a successful JSON answer gptme injects a later non-JSON assistant
+    message ("No tool call detected ... use the `complete` tool"). That
+    trailer is not a better summary; preferring it over a JSON object that
+    already carries ``narrative``/``month_narrative`` re-breaks the
+    daily-summary fallback.
+    """
+    return _GPTME_TRAILER_MARKER in text
 
 
 def _strip_think_tags(text: str) -> str:
@@ -183,9 +198,28 @@ def _extract_assistant_text(stdout: str) -> str:
         # JSON-shaped thinking preamble that may also contain a summary key.
         # Reasoning models (e.g. deepseek) emit a preamble first, then the
         # real answer; scanning forward would return the preamble instead.
+        json_set = set(json_candidates)
         for candidate in reversed(json_candidates):
             parsed = json.loads(candidate)
             if isinstance(parsed, dict) and any(parsed.get(k) for k in _SUMMARY_KEYS):
+                # A later non-JSON message that is *not* gptme's auto-reply
+                # trailer is the real answer (JSON draft, then prose). The
+                # trailer-only case must keep the JSON.
+                json_index = max(i for i, c in enumerate(contents) if c == candidate)
+                later_prose = [
+                    c
+                    for c in contents[json_index + 1 :]
+                    if c not in json_set and not _is_gptme_trailer(c)
+                ]
+                if later_prose:
+                    logger.debug(
+                        "gptme fallback: JSON with summary key followed by "
+                        "later non-trailer prose; preferring prose "
+                        "(%d chars): %.100r",
+                        len(later_prose[-1]),
+                        later_prose[-1],
+                    )
+                    return later_prose[-1]
                 logger.debug(
                     "gptme fallback: returning JSON with summary key (%d chars): %.100r",
                     len(candidate),
--- a/packages/gptme-activity-summary/src/gptme_activity_summary/gptme_backend.py
+++ b/packages/gptme-activity-summary/src/gptme_activity_summary/gptme_backend.py
@@ -139,6 +139,18 @@ def _is_gptme_trailer(text: str) -> bool:
     return _GPTME_TRAILER_MARKER in text
 
 
+def _last_non_trailer(messages: list[str]) -> str:
+    """Last message that is not gptme's no-tool-call auto-reply.
+
+    Falls back to ``messages[-1]`` if every message is a trailer (or the
+    list is empty, which the caller already guards).
+    """
+    for message in reversed(messages):
+        if not _is_gptme_trailer(message):
+            return message
+    return messages[-1] if messages else ""
+
+
 def _strip_think_tags(text: str) -> str:
     """Strip <think>...</think> reasoning blocks from model output.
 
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:
Rebased onto `master` after #1547 merged (add/add conflict in `tests/test_gptme_backend.py`).

Resolution:
- Kept #1547's preamble / enable / default-model tests
- Kept this PR's `<think>`-tag stripping tests plus the unique non-JSON fallback and multipart-content cases
- Dropped duplicate preamble tests that #1547 already covers
- `cc_backend.py` auto-merged: `raw_decode` scan + #1547's prompt instruction both present

19 tests pass locally. CI is running on `578f0415`.

packages/gptme-activity-summary/src/gptme_activity_summary/gptme_backend.py
```diff
@@ -166,11 +186,37 @@ def _extract_assistant_text(stdout: str) -> str:
         for candidate in reversed(json_candidates):
             parsed = json.loads(candidate)
             if isinstance(parsed, dict) and any(parsed.get(k) for k in _SUMMARY_KEYS):
+                logger.debug(
+                    "gptme fallback: returning JSON with summary key (%d chars): %.100r",
+                    len(candidate),
+                    candidate,
+                )
                 return candidate
```
TimeToBuildBob, line 266: ❌ **P1** — In `_extract_assistant_text`, the loop over `reversed(json_candidates)` returns the first candidate that has any key in `_SUMMARY_KEYS` (`narrative`/`month_narrative`) without checking whether a later non-JSON content exists. If a reasoning model emits a preliminary JSON message that includes a `narrative` key (e.g., a draft summary) and then the actual final answer as plain prose, `json_candidates` contains only the preamble (since the prose does not parse as JSON). The loop returns the preamble, discarding the real answer. This violates the PR's intent that the final answer should win; the new preamble fallback only prefers non-JSON content when no JSON candidate has a summary key, so this case is missed. Consequence: the gptme fallback returns a summary derived from the preamble rather than the actual response, producing incorrect or empty narrative in the resulting daily/monthly summary. To fix, the function should prefer the last non-JSON content over a JSON preamble that is not the final message, or at least compare timestamps/order. _(2/2 passes)_

TimeToBuildBob:
## 🤖 AI code review

The main issue is that the thinking-block regex is too broad and can silently corrupt valid assistant content. It should be tightened to match the actual reasoning-tag delimiters rather than any occurrence of the words 'thinking' and 'response'.

### Confidence Score: N/A

Not scored — this diff moves a submodule pointer and contains no source to assess. Treat it as *not reviewed*, not as approved.

**❌ P1 high** — [`gptme_backend.py:1`](https://github.com/gptme/gptme-contrib/blob/b623360d6d158310c5a1fd287dc1974d20263ff0/gptme_backend.py#L1)

The new `_THINK_TAG_RE` regular expression is far too permissive. The pattern `r" thinking[\s\S]*? response"` uses only the common English words 'thinking' and 'response' rather than a specific reasoning-tag delimiter. Because `_strip_think_tags` applies `re.sub` globally, any assistant content that contains a phrase like 'I was thinking about the response' (or any string with ' thinking' followed later by ' response') will have that substring silently removed. This can corrupt otherwise valid JSON payloads, breaking the subsequent `json.loads` check and causing the fallback to return the wrong content or no summary at all. The regex should be anchored to the actual markers used by the model (for example, `<thinking>`/`</thinking>`), and it should only strip content when those markers are present.

```suggestion
Use a specific delimiter-based regex, e.g. r"<thinking>[\s\S]*?</thinking>" if the model emits angle-bracket tags, or at minimum require the substrings to appear at the start/end of the block and not inside normal prose.
```

<sub>**How this was verified:** A content string like `{"note": "I was thinking about the response"}` would have ` thinking about the response` removed, leaving `{"note": "I was"}` and corrupting the JSON.</sub>

</details>

| `578f0415db46` | 3/5 | 1 | llm | 2026-08-29 14:46 UTC |
| `6042c4be1001` | ? | 0 | llm | 2026-08-30 02:35 UTC |

</details>

<sub>Reviewed `b623360d6d15` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 836s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

packages/gptme-activity-summary/src/gptme_activity_summary/gptme_backend.py
TimeToBuildBob: False positive — the later non-JSON assistant message in this pipeline is gptme's auto-reply trailer ("No tool call detected … use the `complete` tool"), not a better summary.

`_extract_assistant_text` already last-wins among JSON candidates that carry `narrative`/`month_narrative` (`test_deepseek_preamble_echoes_narrative_key_still_prefers_final_answer`). The non-JSON fallback only runs when **no** JSON candidate has a summary key (`test_no_json_candidate_falls_back_to_non_json_content`).

Preferring later prose over a JSON object that already has a summary key would return the trailer instead of the real answer and re-break the 0/12 daily-summary failure this PR fixes. The hypothetical "JSON draft then prose final answer" is also against the prompt contract (`FIRST non-empty assistant message must be the final JSON`).

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Live replay of the 2026-08-30 05:31Z fallback (`dancing-sad-monster`) shows `6042c4be` still misses.

DeepSeek already produced a complete `narrative` JSON in assistant message 2 (after `<think>` strip, `json.loads` OK). Message 3 is a completion-ack ("I have completed the analysis. The full JSON summary ... was emitted in my previous message") that mentions the word `narrative` but has no JSON object. The "later non-trailer prose wins" rule preferred that ack, so `extract_json_from_response` returned `{}` and the systemd unit failed even though the real summary was in the previous message.

Pushed `b623360d`: later prose only wins when it *embeds* a summary JSON object. Regression test mirrors the live 3-message shape (truncated think-JSON, complete think-JSON, completion-ack). Focused suite 103 passed. The live payload now extracts `narrative` (708 chars). Recovered `knowledge/summaries/daily/2026-08-29.md` from that JSON in the brain repo.

Self-merge still blocked here (Greptile dark on contrib).

PR #1556, opened 2026-08-30
feat(voice): bridge BobBrain vision into live calls

TimeToBuildBob:
## Summary

- multiplex BobBrain vision events and correlated `look` requests over the existing `/local` voice WebSocket
- expose `look` only when `gptme-voice-node` advertises an attached vision bridge
- keep frames on the edge until requested, then run one host-side OpenAI-compatible VLM call
- harden the bridge with bounded event sends, thread-safe frame snapshots, reconnect-safe camera retries, and clean cancellation
- resolve VLM credentials/model/base URL through normal gptme config

This is the integration follow-up to gptme/gptme-contrib#1528 and gptme/gptme-contrib#1529. It advances ErikBjare/bob#730 without creating a separate body-node abstraction.

## Deliberate limits

- Vision events are telemetry only in v0. They do not inject model context or trigger unsolicited speech; motion can fire every sample and upstream has not defined proactive behavior.
- No continuous frame streaming, local VLM inference, place-recognition tool, or `gptme-body-node` extraction.

## Verification

- `uv run pytest packages/gptme-vision-node packages/gptme-voice-node packages/gptme-voice -q` — 437 passed, 2 skipped
- `uv run ruff check` on all three package source/test trees
- vision-node mypy plus focused voice/voice-node mypy
- pre-commit hooks passed
- Bob PR quality gate: 100/100

### DIFF
--- a/packages/gptme-vision-node/tests/test_bridge.py
+++ b/packages/gptme-vision-node/tests/test_bridge.py
@@ -92,6 +92,41 @@ async def test_look_before_first_frame_returns_bounded_error() -> None:
 
 
 @pytest.mark.asyncio
+async def test_event_send_queue_is_bounded(caplog) -> None:
+    bridge = VisionBridge(_Source(), [])
+    unblock = asyncio.Event()
+
+    async def blocked_send(_message: str) -> None:
+        await unblock.wait()
+
+    bridge._loop = asyncio.get_running_loop()
+    bridge._send_message = blocked_send
+    event = VisionEvent(PERSON_APPEARED)
+    for _ in range(10):
+        bridge._schedule_event(event)
+
+    assert len(bridge._pending_sends) == 4
+    assert "transport is backlogged" in caplog.text
+    unblock.set()
+    await bridge.stop()
+
+
+@pytest.mark.asyncio
+async def test_stop_cancels_blocked_event_sends() -> None:
+    bridge = VisionBridge(_Source(), [])
+    blocked = asyncio.Event()
+
+    async def blocked_send(_message: str) -> None:
+        await blocked.wait()
+
+    bridge._send_message = blocked_send
+    bridge._schedule_event(VisionEvent(PERSON_APPEARED))
+
+    await asyncio.wait_for(bridge.stop(), timeout=0.1)
+    assert not bridge._pending_sends
+
+
+@pytest.mark.asyncio
 async def test_unrelated_server_message_is_not_consumed() -> None:
     bridge = VisionBridge(_Source(), [])
 
--- a/packages/gptme-vision-node/tests/test_bridge.py
+++ b/packages/gptme-vision-node/tests/test_bridge.py
@@ -1,0 +1,101 @@
+"""Tests for the edge-side camera/voice bridge."""
+
+from __future__ import annotations
+
+import asyncio
+import base64
+import json
+
+import numpy as np
+import pytest
+from gptme_vision_node.bridge import VisionBridge
+from gptme_vision_node.detect import Detection
+from gptme_vision_node.pipeline import PERSON_APPEARED, VisionEvent
+
+
+class _Source:
+    def __init__(self) -> None:
+        self.released = False
+
+    def get_frame(self):
+        return None
+
+    def release(self) -> None:
+        self.released = True
+
+
+@pytest.mark.asyncio
+async def test_event_callback_crosses_into_websocket_loop() -> None:
+    bridge = VisionBridge(_Source(), [])
+    sent: list[str] = []
+
+    async def send(message: str) -> None:
+        sent.append(message)
+
+    bridge._loop = asyncio.get_running_loop()
+    bridge._send_message = send
+    bridge._on_event(
+        VisionEvent(
+            PERSON_APPEARED,
+            (Detection("person", (1, 2, 3, 4), 0.9),),
+            timestamp=123.0,
+        )
+    )
+    await asyncio.sleep(0)
+    await asyncio.sleep(0)
+
+    assert json.loads(sent[0]) == {
+        "type": "vision_event",
+        "event": "person_appeared",
+        "timestamp": 123.0,
+        "detections": [{"kind": "person", "box": [1, 2, 3, 4], "score": 0.9}],
+    }
+
+
+@pytest.mark.asyncio
+async def test_look_request_returns_latest_frame_as_jpeg() -> None:
+    bridge = VisionBridge(_Source(), [])
+    bridge.pipeline.latest_frame = np.full((20, 20, 3), 127, dtype=np.uint8)
+    sent: list[str] = []
+
+    async def send(message: str) -> None:
+        sent.append(message)
+
+    handled = await bridge.handle_message(
+        {"type": "vision_look_request", "request_id": "req-1"}, send
+    )
+
+    assert handled is True
+    payload = json.loads(sent[0])
+    assert payload["type"] == "vision_look_result"
+    assert payload["request_id"] == "req-1"
+    assert payload["media_type"] == "image/jpeg"
+    assert base64.b64decode(payload["image"])[:2] == b"\xff\xd8"
+
+
+@pytest.mark.asyncio
+async def test_look_before_first_frame_returns_bounded_error() -> None:
+    bridge = VisionBridge(_Source(), [])
+    sent: list[str] = []
+
+    async def send(message: str) -> None:
+        sent.append(message)
+
+    assert await bridge.handle_message(
+        {"type": "vision_look_request", "request_id": "req-2"}, send
+    )
+    assert json.loads(sent[0]) == {
+        "type": "vision_look_result",
+        "request_id": "req-2",
+        "error": "Camera has not produced a frame yet.",
+    }
+
+
+@pytest.mark.asyncio
+async def test_unrelated_server_message_is_not_consumed() -> None:
+    bridge = VisionBridge(_Source(), [])
+
+    async def send(_message: str) -> None:
+        raise AssertionError("must not send")
+
+    assert not await bridge.handle_message({"type": "audio_end"}, send)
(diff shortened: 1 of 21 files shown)


TimeToBuildBob:
Fixed the Greptile P1: `VisionPipeline.start()` now clears `latest_frame` under the snapshot lock, so a reconnect/`look` before the next capture fails cleanly instead of describing the previous session.

Regression: `test_start_clears_stale_frame_until_new_capture`.
Commit: db6f229d

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Self-merge not eligible here (policy gate: root `pyproject.toml` is outside the allowlist). CI green, Greptile 5/5 on db6f229d, no unresolved threads. Waiting for human review or merge.

packages/gptme-vision-node/tests/test_bridge.py
```diff
@@ -0,0 +1,136 @@
...
+    bridge._loop = asyncio.get_running_loop()
+    bridge._send_message = blocked_send
+    event = VisionEvent(PERSON_APPEARED)
+    for _ in range(10):
+        bridge._schedule_event(event)
+
+    assert len(bridge._pending_sends) == 4
+    assert "transport is backlogged" in caplog.text
+    unblock.set()
+    await bridge.stop()
```
TimeToBuildBob, line 111: ❌ **P1** — `VisionBridge.stop()` runs `pipeline.stop()` and then immediately calls `source.release()` on the capture source without ever joining the worker thread. `pipeline.stop()` only sets the shutdown flag; the `_run` loop can be anywhere between `frame = self.source.get_frame()` and `time.sleep(...)` when the flag is observed. For the default OpenCV `VideoCapture`, calling `release()` from the event-loop thread while the background thread is concurrently inside `read()` is not documented thread-safe and can raise or segfault on some backends, or produce a garbage frame afterwards. In addition, `_on_event` can be mid-flight on that worker when `_send_message`/`_loop` are nulled, so a `call_soon_threadsafe` callback can fire after teardown; it survives only by the guard reading `None` at schedule time (which is scheduling-dependent, not guaranteed). The teardown path exercised by `test_stop_cancels_blocked_event_sends` joins only the send queue (via gather on canceled sends), not the capture thread, so the race escapes CI. Recommend that `pipeline.stop()` join the worker (bounded timeout) before `release()` is called, and that `stop()` first clear the emission path atomically (set both `_loop` and `_send_message` to None) before canceling the send queue.

TimeToBuildBob (commented):
## 🤖 AI code review — 3/5 (One P1 finding)

**2 findings** · ❌ **1** P1 · ⚠️ **1** P2

Details are in the summary comment below.

TimeToBuildBob:
## 🤖 AI code review

Adds a vision-bridge integration that multiplexes Bob vision events and correlated look requests over the existing /local voice WebSocket, exposes look only when an attached vision bridge is advertised, keeps frames edge-side until requested, and adds bounded event sending, snapshot-based frame access, reconnect-safe source handling, and clean bridge teardown, with new bridge and pipeline tests, an updated package export, and documentation.

> ⚠️ **1 generated file excluded and NOT reviewed:** ``uv.lock`` (+7/-1 lines, 1KB). Machine-generated output is removed from the diff before review — from what the model sees and from the size gate alike — so nothing below says anything about it.

<b>2 findings</b> · ❌ **1** P1 · ⚠️ **1** P2 — 🛑 **1 at P0/P1**

<details><summary><b>1 inline</b></summary>

**❌ P1 high** — `packages/gptme-vision-node/tests/test_bridge.py:111`

`VisionBridge.stop()` runs `pipeline.stop()` and then immediately calls `source.release()` on the capture source without ever joining the worker thread. `pipeline.stop()` only sets the shutdown flag; the `_run` loop can be anywhere between `frame = self.source.get_frame()` and `time.sleep(...)` when the flag is observed. For the default OpenCV `VideoCapture`, calling `release()` from the event-loop thread while the background thread is concurrently inside `read()` is not documented thread-safe and can raise or segfault on some backends, or produce a garbage frame afterwards. In addition, `_on_event` can be mid-flight on that worker when `_send_message`/`_loop` are nulled, so a `call_soon_threadsafe` callback can fire after teardown; it survives only by the guard reading `None` at schedule time (which is scheduling-dependent, not guaranteed). The teardown path exercised by `test_stop_cancels_blocked_event_sends` joins only the send queue (via gather on canceled sends), not the capture thread, so the race escapes CI. Recommend that `pipeline.stop()` join the worker (bounded timeout) before `release()` is called, and that `stop()` first clear the emission path atomically (set both `_loop` and `_send_message` to None) before canceling the send queue.

```suggestion
Make the bridge's production teardown join the background thread (e.g., `pipeline.stop(join=True)` with a bounded timeout) before `source.release()`, and null `self._loop` and `self._send_message` in the same non-await section as `pipeline.stop()` so no late `_on_event` can schedule work after the loop is considered down.
```

<sub>**How this was verified:** The visible `_run` loop (`frame = self.source.get_frame()` / `time.sleep(interval - elapsed)`) has no join in the diff; the diff's stop-path test only cancels and gathers the send-queue tasks in `_pending_sends`, never the pipeline worker, so a concurrent `read()`/`release()` teardown race is plausible (confirmed against the diff excerpt).</sub>

</details>

### Comments outside the diff

**⚠️ P2 medium** — [`packages/gptme-vision-node/tests/test_pipeline.py:10`](https://github.com/gptme/gptme-contrib/blob/db6f229d599c177f7613ad57cc85e992c5b45aeb/packages/gptme-vision-node/tests/test_pipeline.py#L10)

The diff's only verification for the new snapshot API is: `snapshot[:] = 0` then `assert pipeline.latest_frame is not None`. That final assertion passes whether `latest_frame_copy()` deep-copies the stored frame or returns an alias to the internal buffer, so an aliasing implementation would sail through this `independent snapshot` test. If the source reuses its read buffer across calls (OpenCV read can reuse memory; `_process_frame` stores the source array by reference without copying), the frame later read by `handle_message()`/`latest_frame_copy()` can be half-overwritten by the next `get_frame()` on the worker thread, yielding corrupted VLM input or telemetry payloads, or a caller mutating the returned array could corrupt the live capture buffer. The snapshot path is exactly where frame isolation matters here. Assert independence by checking the stored frame still holds the original pixels after the snapshot is mutated, and have `_process_frame` store a copy (or have `latest_frame_copy()` copy the buffer under the same lock) so no caller ever aliases live capture memory.

```suggestion
After `snapshot[:] = 0`, assert the stored frame still contains the pre-mutation values (e.g., original frame is unchanged and `latest_frame` still refers to the unzeroed buffer), and have `_process_frame` store `frame.copy()` so readers cannot alias capture memory.
```

<sub>**How this was verified:** The hunk `+    snapshot[:] = 0` followed by `+    assert pipeline.latest_frame is not None` is the sole independence check for the newly added snapshot contract; mutating an aliased buffer would still leave the buffer non-None, so the test cannot distinguish copy from alias (confirmed by reading the diff).</sub>

- `packages/gptme-vision-node/src/gptme_vision_node/bridge.py` — Adds the vision-bridge integration multiplexing cached Bob vision events and correlated look requests over the existing /local voice WebSocket, exposing look only when an attached bridge is advertised.
- `packages/gptme-vision-node/src/gptme_vision_node/pipeline.py` — Adds bounded event-send queueing, thread-safe frame snapshots, reconnect-safe source handling, and clean cancellation for the bridge teardown.
- `packages/gptme-vision-node/src/gptme_vision_node/frame_source.py` — Adds a stop_on_empty capture-source behavior so a worker can retry instead of breaking when a frame read returns None.
- `packages/gptme-vision-node/src/gptme_vision_node/__init__.py` — Exports the new bridge class and associated symbols from the package.
- `packages/gptme-vision-node/README.md` — Documents the vision-bridge multiplexing and detached look flow.
- `packages/gptme-vision-node/tests/test_pipeline.py` — Adds tests covering event-send bounding, snapshot independence, look correlation diagnostics, cancel-and-stop for a blocked send, and the independent-snapshot contract.
- `packages/gptme-vision-node/tests/test_bridge.py` — Adds vision-bridge tests exercising bounded sends, stop-cancellation of pending sends, and the telemetry payload.

<sub>Also changed, not described above: `packages/gptme-voice-node/README.md`, `packages/gptme-voice-node/pyproject.toml`, `packages/gptme-voice-node/src/gptme_voice_node/node.py`, `packages/gptme-voice-node/tests/test_voice_node.py`, `packages/gptme-voice/README.md`, `packages/gptme-voice/pyproject.toml`, `packages/gptme-voice/src/gptme_voice/realtime/server.py`, `packages/gptme-voice/src/gptme_voice/realtime/tool_bridge.py`, `packages/gptme-voice/src/gptme_voice/vision.py`, `packages/gptme-voice/tests/test_server.py`, `packages/gptme-voice/tests/test_tool_bridge.py`, `packages/gptme-voice/tests/test_vision.py` (+1 more)</sub>

</details>

<sub>Reviewed `db6f229d599c` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 1409s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

packages/gptme-vision-node/tests/test_bridge.py
TimeToBuildBob: Fixed on master in gptme/gptme-contrib#1561 (`76028fa1`).

`pipeline.stop()` already joined with a timeout — the real hole was calling `source.release()` even when that join timed out, plus nulling `_loop`/`_send_message` after the worker was still able to emit. `VisionBridge.stop()` now:

1. Clears the emission path first
2. Joins off the event loop
3. Calls `release()` only if the capture thread is actually dead

Also stored `frame.copy()` so OpenCV buffer reuse cannot alias `latest_frame` (the P2).

TimeToBuildBob:
AI-review P1 landed after this PR merged (18:20 vs 18:15). Follow-up: gptme/gptme-contrib#1561.

`VisionBridge.stop()` no longer calls `source.release()` while the capture thread may still be inside OpenCV `read()`, and `_process_frame` now stores an independent frame copy.

PR #1568, opened 2026-08-31
feat(sessions): parse Pi native v3 trajectories

TimeToBuildBob:
## Summary

- detect and validate Pi v3 native tree JSONL, rejecting print streams, future versions, and malformed Pi data instead of misclassifying them as gptme NOOPs
- normalize active-branch transcript and current provider/model/stop metadata while counting billed usage and executed tool/write/commit signals across every persisted tree entry
- add sanitized native fixtures for a productive Codex run and genuine Codex/xAI NOOPs

## Verification

- `uv run pytest packages/gptme-sessions/tests -q` — 1231 passed
- `make -C packages/gptme-sessions typecheck` — success, 23 files
- `uv run ruff check packages/gptme-sessions/src/gptme_sessions packages/gptme-sessions/tests/test_signals_pi.py` — passed
- PR quality gate — 100/100

## Remaining gates

This is Phase 1. The parent task remains active for discovery/retention and CLI sync, first-class SessionRecord persistence of provider/cost/stop reason, and a retained native error/abort/timeout/compaction/branch fixture matrix.

PR #1570, opened 2026-08-31
feat(gptme-sessions): persist Pi route metadata

TimeToBuildBob:
## Summary

- persist Pi provider, model, final stop reason, and harness-reported USD-equivalent cost in `SessionRecord`
- preserve missing cost versus a real `0.0`, and prefer reported cost over optional token-price estimation
- accept `post-session --harness pi` for explicit native trajectories and retain the fields in JSON/CSV exports
- document that OAuth/subscription cost can be nominal API-equivalent cost rather than incremental billed spend

This is an independent follow-up to #1568. Native discovery/sync remains a separate review slice so each change stays auditable.

## Compatibility

- new dataclass/result fields are appended to preserve the established positional constructor prefix
- existing records deserialize with `None` for the new fields
- timeout exit codes do not synthesize a Pi stop reason

## Tests

- `uv run --project packages/gptme-sessions --with gptme pytest packages/gptme-sessions/tests -q` (1,243 passed)
- `uv run mypy packages/gptme-sessions/src/gptme_sessions` (23 files clean)
- Ruff, diff-check, and scoped pre-commit hooks passed
- PR quality gate: 94/100 (only deduction is expected line overlap with unrelated #1565)

### DIFF
--- a/packages/gptme-sessions/README.md
+++ b/packages/gptme-sessions/README.md
@@ -3,7 +3,9 @@
 Session tracking and analytics for agents. Supports trajectories from gptme,
 Claude Code, Codex, Copilot, Grok Build, and Pi native v3 sessions.
 
-Provides an append-only JSONL-based session record system that any agent can use to track operational metadata across sessions: which harness ran, what model was used, what type of work was done, and the outcome.
+Provides an append-only JSONL-based session record system that any agent can use
+to track operational metadata across sessions: which harness and inference
+provider ran, what model was used, what type of work was done, and the outcome.
 
 ## Installation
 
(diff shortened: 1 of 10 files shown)


TimeToBuildBob:
Fixed the sync skip: `sync --signals` now re-extracts Pi records when `provider` / `cost_usd` / `stop_reason` are missing, even if legacy token/byte fields are already populated.

Copilot and other harnesses stay on their existing backfill sets so we don't re-extract forever on fields they never emit. Covered in `TestSyncRouteBackfill`.

TimeToBuildBob:
@greptileai review

PR #1587, opened 2026-09-02
feat(twitter): add quote-tweet capability

TimeToBuildBob:
## Summary

Implements the quote-tweet capability Erik requested (2026-08-11 call) as the policy fix for the "you don't really engage" failure mode.

- `TweetDraft` gains `quote_tweet_id` field with `to_dict`/`from_dict` round-trip; backward-compatible (`"quote"` key alias for hand-authored drafts)
- `_create_tweet_kwargs()` helper centralises `create_tweet` arg building: quote drafts post standalone (Twitter v2 rejects combining `quote_tweet_id` + `in_reply_to`); the precedence rule is tested explicitly
- All 3 main post paths (`post`, `process_timeline_tweets`, `auto`) use the helper
- Thread follow-ups always remain replies (never inherit the quote ID)
- `draft` CLI: `--type` gains `"quote"`; `--quote-id` option required when `--type quote`
- 8 unit tests: round-trip, kwarg routing, thread reply invariant, backward compat

## Test plan

- [x] `uv run pytest tests/test_twitter_workflow_drafts.py` — 51 passed
- [x] Existing regression suite clean
- [ ] Trigger Greptile review

### DIFF
--- a/scripts/twitter/workflow.py
+++ b/scripts/twitter/workflow.py
@@ -268,9 +268,10 @@ class TweetDraft:
         reject_reason: str | None = None,
         quality_score: int | None = None,
         thread: List[str] | None = None,
+        quote_tweet_id: str | None = None,
     ):
         self.text = text
-        self.type = type  # tweet, reply, thread
+        self.type = type  # tweet, reply, thread, quote
         self.in_reply_to = in_reply_to
         self.scheduled_time = scheduled_time
         self.context = context or {}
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
CI green, Greptile 5/5, self-merge eligibility passed (mixed-allowed tests+internal-tooling). Consensus gate blocked at 2/5 via fallback `llama-3.1-8b-instruct` citing missing quote-tweet tests. Those tests are already in this PR (round-trip, legacy `quote` key, quote-over-reply precedence, standalone/reply kwargs, quoted-thread follow-ups stay replies). Empty `--quote-id` is a CLI reject. Did not bypass the gate, did not re-roll it in-band, and did not add test theater. Waiting for a primary-model consensus pass or human merge.

PR #1589, opened 2026-09-02
feat(match-lessons): class-aware lesson dropout (Phase 1 differential epsilon)

TimeToBuildBob:
## Summary

- Adds per-policy-class epsilon control to the CC hook lesson dropout system for causal LOO analysis
- `exempt` lessons: never withheld (epsilon 0.0)
- `validated_core` lessons: `LESSON_DROPOUT_EPSILON_VALIDATED_CORE` env var (default 0.05)
- `holdout` / `unknown` lessons: global `LESSON_DROPOUT_EPSILON`
- Withheld records now include `effective_epsilon` for per-class analysis downstream

Two new helpers (`_get_dropout_epsilon_validated_core`, `_get_dropout_epsilon_for_class`) dispatched from both `_apply_lesson_dropout` and `_apply_lesson_dropout_multi`. The change is backward-compatible: with default env (`LESSON_DROPOUT_EPSILON=0`), dropout remains off for all classes.

## Test plan

- [x] 5 new tests in `test_cc_match_lessons.py` covering: default/override validated_core epsilon, class dispatch table, exempt-never-withheld, and `effective_epsilon` in withheld records
- [x] All 83 existing CC hook tests pass
- [x] Mirrors changes already applied to `gptme/lessons/auto_include.py` + 17 passing tests there

### DIFF
--- a/scripts/claude-code-hooks/match-lessons.py
+++ b/scripts/claude-code-hooks/match-lessons.py
@@ -1377,10 +1377,13 @@ def _get_dropout_epsilon_for_class(policy_class: str, global_epsilon: float) -> 
     """Return the effective dropout probability for a given policy class.
 
     - exempt: always 0.0 (never withheld, regardless of global_epsilon)
-    - validated_core: LESSON_DROPOUT_EPSILON_VALIDATED_CORE (default 0.05)
+    - validated_core: LESSON_DROPOUT_EPSILON_VALIDATED_CORE (default 0.05),
+      but only once the global switch (LESSON_DROPOUT_EPSILON) is > 0
     - holdout / unknown: global_epsilon (LESSON_DROPOUT_EPSILON)
     """
     if policy_class == "exempt":
+        return 0.0
+    if global_epsilon <= 0.0:
         return 0.0
     if policy_class == "validated_core":
         return _get_dropout_epsilon_validated_core()
--- a/scripts/claude-code-hooks/match-lessons.py
+++ b/scripts/claude-code-hooks/match-lessons.py
@@ -1377,8 +1377,10 @@ def _get_dropout_epsilon_for_class(policy_class: str, global_epsilon: float) -> 
     """Return the effective dropout probability for a given policy class.
 
     - exempt: always 0.0 (never withheld, regardless of global_epsilon)
-    - validated_core: LESSON_DROPOUT_EPSILON_VALIDATED_CORE (default 0.05),
-      but only once the global switch (LESSON_DROPOUT_EPSILON) is > 0
+    - validated_core: min(global_epsilon, LESSON_DROPOUT_EPSILON_VALIDATED_CORE)
+      (default 0.05), but only once the global switch (LESSON_DROPOUT_EPSILON) is > 0.
+      Bounded by global_epsilon so the per-class differential can only reduce
+      dropout relative to the global setting, never exceed it.
     - holdout / unknown: global_epsilon (LESSON_DROPOUT_EPSILON)
     """
     if policy_class == "exempt":
@@ -1386,7 +1388,7 @@ def _get_dropout_epsilon_for_class(policy_class: str, global_epsilon: float) -> 
     if global_epsilon <= 0.0:
         return 0.0
     if policy_class == "validated_core":
-        return _get_dropout_epsilon_validated_core()
+        return min(global_epsilon, _get_dropout_epsilon_validated_core())
     return global_epsilon
 
 
--- a/scripts/claude-code-hooks/match-lessons.py
+++ b/scripts/claude-code-hooks/match-lessons.py
@@ -1347,7 +1347,10 @@ def _get_dropout_epsilon() -> float:
         epsilon = float(raw)
     except ValueError:
         return 0.0
-    if math.isnan(epsilon) or epsilon <= 0.0:
+    # Treat non-finite values (inf, -inf, nan) as unset — matching
+    # _get_dropout_epsilon_validated_core's handling so that a mis-set env var
+    # (e.g. LESSON_DROPOUT_EPSILON=inf) does not silently enable 100% dropout.
+    if not math.isfinite(epsilon) or epsilon <= 0.0:
         return 0.0
     return min(epsilon, 1.0)
 
@@ -1452,6 +1455,14 @@ def _apply_lesson_dropout(
     if epsilon <= 0.0 or not lessons:
         return lessons
 
+    # When the manifest is unavailable (the "*" sentinel is active), classification
+    # is unreliable — lessons that would be "exempt" in a real manifest are
+    # indistinguishable from "holdout" and could be dropped. Skip dropout entirely
+    # to preserve the PR's guarantee that exempt lessons are never withheld.
+    manifest = _load_policy_manifest()
+    if "*" in manifest.get("holdout_population", []):
+        return lessons
+
     kept: list[dict] = []
     for lesson in lessons:
         policy_class, _ = _classify_lesson(lesson.get("path", ""))
@@ -1487,6 +1498,14 @@ def _apply_lesson_dropout_multi(
     # lesson-loo-analysis.py can distinguish treatment-group sessions
     # (epsilon>0, possibly 0 withheld) from control-group sessions (epsilon=0).
     if epsilon <= 0.0:
+        return matches, predicted
+
+    # When the manifest is unavailable (the "*" sentinel is active), classification
+    # is unreliable — lessons that would be "exempt" in a real manifest are
+    # indistinguishable from "holdout" and could be dropped. Skip dropout entirely
+    # to preserve the guarantee that exempt lessons are never withheld.
+    manifest = _load_policy_manifest()
+    if "*" in manifest.get("holdout_population", []):
         return matches, predicted
 
     withheld: list[dict] = []
--- a/scripts/claude-code-hooks/match-lessons.py
+++ b/scripts/claude-code-hooks/match-lessons.py
@@ -1366,9 +1366,9 @@ def _get_dropout_epsilon_validated_core() -> float:
         epsilon = float(raw)
     except ValueError:
         return 0.05
-    # float("nan") succeeds; min/max retain NaN and later `eff_epsilon > 0`
-    # silently disables validated-core sampling. Treat non-finite as unset.
-    if math.isnan(epsilon):
+    # float() accepts NaN and infinities, but neither is a probability.
+    # Treat non-finite values as unset instead of clamping infinity to 1.0.
+    if not math.isfinite(epsilon):
         return 0.05
     return min(max(epsilon, 0.0), 1.0)
 
--- a/scripts/claude-code-hooks/match-lessons.py
+++ b/scripts/claude-code-hooks/match-lessons.py
@@ -1347,7 +1347,7 @@ def _get_dropout_epsilon() -> float:
         epsilon = float(raw)
     except ValueError:
         return 0.0
-    if epsilon <= 0.0:
+    if math.isnan(epsilon) or epsilon <= 0.0:
         return 0.0
     return min(epsilon, 1.0)
 
@@ -1365,6 +1365,10 @@ def _get_dropout_epsilon_validated_core() -> float:
     try:
         epsilon = float(raw)
     except ValueError:
+        return 0.05
+    # float("nan") succeeds; min/max retain NaN and later `eff_epsilon > 0`
+    # silently disables validated-core sampling. Treat non-finite as unset.
+    if math.isnan(epsilon):
         return 0.05
     return min(max(epsilon, 0.0), 1.0)
 
--- a/scripts/claude-code-hooks/match-lessons.py
+++ b/scripts/claude-code-hooks/match-lessons.py
@@ -1352,6 +1352,37 @@ def _get_dropout_epsilon() -> float:
     return min(epsilon, 1.0)
 
 
+def _get_dropout_epsilon_validated_core() -> float:
+    """Get dropout epsilon for validated_core lessons (default 0.05).
+
+    Only meaningful when global LESSON_DROPOUT_EPSILON > 0 (the global switch activates
+    the dropout system; this controls how aggressively validated_core lessons are sampled).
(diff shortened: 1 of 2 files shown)


scripts/claude-code-hooks/match-lessons.py
```diff
@@ -1352,6 +1352,37 @@ def _get_dropout_epsilon() -> float:
...
+    Override with LESSON_DROPOUT_EPSILON_VALIDATED_CORE env var.
+    """
+    raw = os.environ.get("LESSON_DROPOUT_EPSILON_VALIDATED_CORE")
+    if raw is None:
+        return 0.05
+    try:
+        epsilon = float(raw)
+    except ValueError:
+        return 0.05
+    return min(max(epsilon, 0.0), 1.0)
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 1c4fe318 — `float("nan")` now falls back to the 0.05 default instead of retaining NaN and silently disabling validated-core sampling. Added a regression test.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Fixed the NaN epsilon issue (Greptile 4/5): added `math.isnan()` guard to both `_get_dropout_epsilon_validated_core()` and `_get_dropout_epsilon()` — NaN now falls back to the respective default (0.05 and 0.0) instead of bypassing the clamp. See fe84328b.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
## 🤖 AI code review

Adds class-aware dropout epsilon to the CC hook: exempt lessons are never withheld, validated_core uses LESSON_DROPOUT_EPSILON_VALIDATED_CORE (default 0.05), holdout/unknown use global LESSON_DROPOUT_EPSILON. Adds non-finite handling for both epsilon env vars, a skip-dropout path when the policy manifest is unavailable ("*" sentinel), and effective_epsilon in withheld records. Extends both _apply_lesson_dropout and _apply_lesson_dropout_multi accordingly and adds tests for the new behavior.

**⚠️ P2 medium · ℹ️ non-blocking** — `scripts/claude-code-hooks/match-lessons.py:1508`

_load_policy_manifest returns {} (not the sentinel default) when the YAML is a valid non-dict (e.g., `- item`), because it sets `manifest = _loaded if isinstance(_loaded, dict) else {}`. Then _apply_lesson_dropout_multi checks `if "*" in manifest.get("holdout_population", [])` which is false for the empty dict, so dropout proceeds. All lessons then classify as "unknown" and get the global epsilon, so exempt lessons are withheld with probability global_epsilon, violating the PR's core guarantee that exempt lessons are never withheld. A corrupted manifest that is a list is not treated as unavailable, unlike a missing or unparseable file.

```suggestion
_log_dropout(session_id, epsilon, [], workspace)
        return matches, predicted
```

<sub>**How this was verified:** Checked _load_policy_manifest: it sets the sentinel only for missing/unparseable files, but an explicit '*' in a present manifest also matches the condition here. _classify_lesson (lines 327-328) already treats '*' as making every lesson holdout, so this additional skip is redundant and harmful.</sub>

</details>

- `scripts/claude-code-hooks/match-lessons.py` — Adds _get_dropout_epsilon_validated_core and _get_dropout_epsilon_for_class, updates both dropout functions to use per-class epsilon and skip when the manifest is unavailable, and adds effective_epsilon to withheld records.
- `tests/test_cc_match_lessons.py` — Adds five tests covering non-finite epsilon handling, validated_core default/override, class dispatch, exempt-never-withheld, and effective_epsilon in logged withheld records.

</details>

| `fe84328b65da` | ? | 0 | llm | 2026-09-02 04:13 UTC |
| `6458fce7aa0e` | 4/5 | 1 | llm | 2026-09-02 06:28 UTC |
| `9b3073d99a5d` | 4/5 | 1 | llm | 2026-09-02 22:27 UTC |
| `abf1a28e4a3d` | 5/5 | 0 | llm | 2026-09-03 14:13 UTC |
| `3f8090bbd1fe` | 4/5 | 3 | llm | 2026-09-03 14:28 UTC |

</details>

<sub>Reviewed `dfe3336c73c5` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 202s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

TimeToBuildBob:
Fixed the backward-compatibility bug flagged in the AI review (P1, out-of-scope section): `_get_dropout_epsilon_for_class` returned the validated_core epsilon (default 0.05) unconditionally, so validated_core lessons were withheld ~5% of the time even with `LESSON_DROPOUT_EPSILON=0` — the documented no-op default for the global switch. Added a guard so any class-specific epsilon requires `global_epsilon > 0` first (exempt still always returns 0.0). Added a regression test (`test_cc_dropout_epsilon_for_class_global_zero_disables_validated_core`). See 6458fce7.

TimeToBuildBob:
@greptileai review

scripts/claude-code-hooks/match-lessons.py
```diff
@@ -1347,11 +1347,49 @@ def _get_dropout_epsilon() -> float:
         epsilon = float(raw)
     except ValueError:
         return 0.0
-    if epsilon <= 0.0:
+    if math.isnan(epsilon) or epsilon <= 0.0:
         return 0.0
     return min(epsilon, 1.0)
```
TimeToBuildBob, line 1355: ⚠️ **P2** — The class dispatch in `_get_dropout_epsilon_for_class` returns `_get_dropout_epsilon_validated_core()` (0.05 default) unconditionally once the global epsilon is > 0; it never bounds the validated_core rate by the global override. With `LESSON_DROPOUT_EPSILON=0.01` (an operator wanting very light overall dropout), holdout/unknown lessons are withheld at 0.01 but validated_core lessons at 0.05 — i.e. the lessons the feature is meant to preserve end up withheld at 5x the ordinary-class rate, inverting the differential (the only case the new test exercises is global=0.20 where 0.05 < 0.20). The validated_core rate should be `min(global_epsilon, validated_epsilon)` so the per-class differential can only reduce dropout relative to the global setting. Without the clamp, the most important lessons are dropped most often precisely when an operator selects a small global dropout.

TimeToBuildBob: Fixed in 9b3073d9 — `_get_dropout_epsilon_for_class` now returns `min(global_epsilon, _get_dropout_epsilon_validated_core())` for the `validated_core` branch, so the class-specific rate can only reduce dropout relative to the global setting, never exceed it. Added a regression test (`test_cc_dropout_epsilon_for_class_validated_core_bounded_by_global`) covering both directions of the clamp.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Fixed the P2 finding from the self-hosted AI review (validated_core epsilon not bounded by the global epsilon) in 9b3073d — added the `min(epsilon, validated_epsilon)` clamp plus a regression test.

Self-merge is currently blocked, but not by anything in this diff: Greptile's last score (5/5) was against 6458fce7, and Greptile has paused reviews on gptme-contrib entirely — it's out of free OSS review credits for this billing period and resumes automatically on 2026-09-30 (confirmed by its own reply to `@greptileai review` on this PR). The self-merge gate treats a Greptile score that doesn't match the current head as stale and won't fall back to the self-hosted AI reviewer while a (stale) Greptile review is present. No action needed from me right now — this clears on its own once Greptile resumes, or once a maintainer merges directly.

---

Automated merge handoff for `9b3073d99a5d`:

- CI gate: satisfied
- Review threads: leftover AI-review P2 (`8604aeca7769`) was already fixed in `9b3073d9` and is now resolved
- Self-merge policy: blocked on stale Greptile (5/5 on `6458fce7aa0e`, head `9b3073d99a5d`). Greptile paused OSS reviews on gptme-contrib until 2026-09-30, so a current-head re-review cannot land. Paths are allowlisted (`scripts/claude-code-hooks/match-lessons.py`, `tests/test_cc_match_lessons.py`); this is not a path-policy block.

This head is ready for maintainer review and **manual merge**. Project monitoring marked it `human_merge_required` and will not spend retry sessions attempting an automated merge. A new head will be evaluated independently.

tests/test_cc_match_lessons.py
```diff
@@ -1225,3 +1225,124 @@ def fake_import(name, *args, **kwargs):
...
+    monkeypatch.setenv("LESSON_DROPOUT_EPSILON_VALIDATED_CORE", "0.05")
+    # global epsilon (0.01) is smaller than the validated_core default (0.05):
+    # the effective rate must be clamped down to the global epsilon.
+    assert hook._get_dropout_epsilon_for_class("validated_core", 0.01) == 0.01
+    # global epsilon (0.20) is larger than validated_core (0.05): the smaller
+    # class-specific rate wins, as before.
+    assert hook._get_dropout_epsilon_for_class("validated_core", 0.20) == 0.05
+
+
+def test_cc_dropout_exempt_lesson_never_withheld(hook, monkeypatch, tmp_path):
```
TimeToBuildBob, line 1307: ⚠️ **P2** — The new class-aware logic inserts a `_classify_lesson` call into `_apply_lesson_dropout*` (the per-lesson exception-safety regression). Previously this loop was pure-random (`random.random() < epsilon`) and could not fail. `_classify_lesson` loads/parses the manifest YAML on cache-miss; the suite itself records a raise-class bug it had to fix (a list manifest once raised AttributeError), and the provided excerpt shows handling of absent-file and list-content but NOT of malformed YAML (scanner errors) or permission failures. Any unhandled raise now escapes the loop and kills the entire hook run (zero lessons delivered) — the worst possible outcome for a "preservation" feature. Fix by wrapping the call in try/except and falling back to global epsilon (or fail-closed keep), and add regression tests for scanner-error content and unreadable files. Behavior at line ~1296 (exempt test) is the round-trip this affects.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Fixed the remaining in-band review finding in `3f8090bb`: validated-core epsilon parsing now rejects all non-finite values (`nan`, `inf`, and `-inf`) via `math.isfinite()` and falls back to `0.05`, with regression coverage for each value.

The focused hook suite passes (`87 passed`). CI is running on the new head. The in-band wrapper's second review was the timeout cause: it reached the 900-second reviewer deadline and the outer monitoring slot killed it before it could record a terminal result or push, so I verified the focused suite and pushed the committed fix through the wrapper's documented infra-failure passthrough path. Greptile re-review was already requested for the prior head and is still marked in progress; I did not trigger it again.

scripts/claude-code-hooks/match-lessons.py
```diff
@@ -1347,11 +1347,51 @@ def _get_dropout_epsilon() -> float:
...
```
TimeToBuildBob, line 1355: ⚠️ **P2** — _get_dropout_epsilon clamps infinite input to 1.0 (min(inf,1)==1), but _get_dropout_epsilon_validated_core explicitly treats non-finite values as unset and returns the 0.05 default. So a mis-set LESSON_DROPOUT_EPSILON=inf (e.g., from an upstream script bug) silently enables 100% dropout for all holdout/unknown lessons, while validated_core stays at 5%. This is an inconsistent and surprising failure mode: a single bad env var can wipe out the entire lesson corpus for non-validated lessons. Treat non-finite global epsilon as 0 (unset) to match the validated_core handling and avoid accidental full dropout.

scripts/claude-code-hooks/match-lessons.py
```diff
@@ -1445,9 +1494,15 @@ def _apply_lesson_dropout_multi(
     def _do_dropout(lessons: list[dict]) -> list[dict]:
         kept: list[dict] = []
         for lesson in lessons:
-            if random.random() < epsilon:
+            policy_class, _ = _classify_lesson(lesson.get("path", ""))
```
TimeToBuildBob, line 1516: ⚠️ **P2** — When the policy manifest is missing or fails to load, _load_policy_manifest returns the default dict with holdout_population ['*'] (lines 229-237), and _classify_lesson returns 'holdout' for every lesson (line 314-315). The new class-aware dropout then applies the global LESSON_DROPOUT_EPSILON to those lessons, including lessons that would be classified 'exempt' if the manifest were present. In a fresh workspace with LESSON_DROPOUT_EPSILON=0.5 and no manifest, safety lessons are withheld at 50%, violating the PR's guarantee that exempt lessons are never dropped. The fallback should preserve the exempt guarantee when the manifest is unavailable, e.g. by not applying dropout when the manifest cannot be loaded, or at least logging a warning.

TimeToBuildBob:
CI is green on the current head (`3f8090bb`). Self-merge is blocked by a gate design constraint, not by any code issue:

- Greptile's 5/5 review targets `6458fce7aa0e`, but current head is `3f8090bbd1fe` (the `math.isfinite()` fix added in today's session)
- The self-merge gate accepts AI-reviewer results only when Greptile has **never** reviewed the PR — since Greptile did review (and scored 5/5), a stale review blocks self-merge rather than opening the AI-reviewer fallback
- Greptile is out of credits until 2026-09-30 and won't re-review before then

No code action needed — the PR is correctly converged. Human merge is the fastest path if this needs to land before Sept 30.

scripts/claude-code-hooks/match-lessons.py
TimeToBuildBob: Fixed in dfe3336c: `_get_dropout_epsilon` now uses `math.isfinite()` like the validated_core helper — `inf`/`-inf`/`nan` all return 0.0. Test `test_cc_dropout_epsilon_global_non_finite` covers all three non-finite values.

scripts/claude-code-hooks/match-lessons.py
TimeToBuildBob: Fixed in dfe3336c: both `_apply_lesson_dropout` and `_apply_lesson_dropout_multi` now check for the '\*' sentinel immediately after the global epsilon gate and skip dropout entirely when the manifest is unavailable. The renamed test `test_cc_dropout_unavailable_manifest_skips_dropout` verifies both malformed and unreadable failure paths now keep all lessons.

PR #1592, opened 2026-09-02
test(gptme-dashboard): prove static search under /dashboard/ subpath

TimeToBuildBob:
## Summary

Adds the missing subpath integration test for static dashboard search (task item 1.1 / A1 after gptme/gptme-contrib#1585).

The generated site is served under `/dashboard/`. The test:

- resolves `./data.json` to `/dashboard/data.json` (200)
- 404s root `/data.json` and confirms client search does not fetch it
- runs `_buildClientIndex` + `_clientSearch` against a known task (`ZebraUniqueTaskAlpha`)
- follows the relative result href to a 200 task page under `/dashboard/`

No production JS change; #1585 already shipped the relative fetch/url fix.

## Test plan

- [x] `uv run --extra test pytest tests/test_generate.py -k "static_search or static_data or static_safe"` (16 passed)
- [x] `tests/test_generate.py::test_static_search_served_under_dashboard_subpath` passed locally with node + stdlib HTTP server

PR #1594, opened 2026-09-02
fix(gptme-dashboard): contain wide tables so 390px has no page overflow

TimeToBuildBob:
## Problem

Measured, not guessed. A full generated workspace dashboard served under
`/dashboard/`, opened in headless chromium at 390x844:

```
viewport=390  documentElement.scrollWidth=762
```

The whole document scrolled sideways by ~2x. Cause: every wide table sizes to
its content, so a single wide row widens `body`. Offenders at the outermost
level were `#tasks-table` (738px), `#guidance-table` (593px),
`#summaries-table` (538px), `#journal-table` (528px), and the packages/plugins
tables (520px). Only the activity heatmap had a scroll container
(`.heatmap-wrap { overflow-x: auto }`).

## Fix

One narrow-viewport containment block in the shared `base.html`, emitted
**after** `{% block extra_style %}` so a per-page rule cannot override it:

- tables become block-level horizontal scroll containers (the overflow lives
  inside the table, headings and controls stay put)
- `thead`/`tbody` keep `width: max-content; min-width: 100%` so an overflowing
  table does not also collapse its own columns
- `pre`/`code` scroll instead of pushing
- long unbroken identifiers (paths, slugs, URLs) wrap

Applying it in `base.html` covers `index.html` and all seven detail templates
at once, rather than adding a wrapper at each of ~20 table sites.

## Verification

Same probe, after the change:

| page | before | after |
|---|---|---|
| `index.html` | 762 | **390** |
| task detail | — | **390** |
| journal index | — | **390** |

Table detail stays reachable by scrolling the table itself.

Two tests added:

- `test_narrow_viewport_containment_css_present` — source guard that the block
  exists and is the **last** media block in the stylesheet (the ordering is what
  makes it un-overridable).
- `test_no_horizontal_overflow_at_390px` — real browser measurement at 390px,
  serving the site under a `/dashboard/` prefix so it also covers the subpath
  deployment shape. `pytest.importorskip` + a launch-failure skip, since
  playwright is not a package dependency, so CI skips it cleanly.

Both were confirmed RED before the fix (`scrollWidth=428, viewport=390` on the
test fixture) and GREEN after. Full suite: 395 passed.

## Alternatives considered

- `section, .card { overflow-x: auto }` — also works, but scrolls the section
  heading and its show-more button along with the table.
- Wrapping each table in a `.table-scroll` div — the most semantic option, but
  ~20 template edit sites across 8 templates for the same measured result.

Part of the gptme-dashboard subpath/responsive work; follows
gptme/gptme-contrib#1592 (static search under `/dashboard/`).

### DIFF
--- a/packages/gptme-dashboard/src/gptme_dashboard/templates/base.html
+++ b/packages/gptme-dashboard/src/gptme_dashboard/templates/base.html
@@ -60,6 +60,30 @@ footer {
 }
 
 {% block extra_style %}{% endblock %}
+
+/* Narrow-viewport containment (must stay last so it wins over per-page styles).
+
+   Wide tables are the dominant source of horizontal page overflow: a table
+   sizes to its content, so one wide row widens `body` and the whole document
+   scrolls sideways. Turning the table into a block-level scroll container
+   keeps the overflow inside the table instead of on the page, while the
+   surrounding headings and controls stay put. `.heatmap-wrap` already does
+   the same thing for the activity heatmap. */
+@media (max-width: 700px) {
+  .container { max-width: 100%; overflow-x: hidden; }
+  table {
+    display: block;
+    width: 100%;
+    max-width: 100%;
+    overflow-x: auto;
+    -webkit-overflow-scrolling: touch;
+  }
+  /* An overflowing block-level table must not also collapse its own columns. */
+  thead, tbody { width: max-content; min-width: 100%; }
+  pre, code { max-width: 100%; overflow-x: auto; }
+  /* Long unbroken identifiers (paths, slugs, URLs) must wrap, not push. */
+  td, th, li, p { overflow-wrap: anywhere; }
+}
 </style>
 {% block extra_head %}{% endblock %}
 </head>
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
## 🤖 AI code review

Adds a narrow-viewport CSS containment block to base.html that turns tables, pre, and code into scroll containers and applies overflow-wrap to table cells and paragraphs, so wide tables no longer widen the page at phone widths. Adds two tests: a source guard that verifies the containment block exists and is the last media block in the stylesheet, and a Playwright-based real-browser check that the document scrollWidth does not exceed the viewport at 390px under a /dashboard/ subpath.

- `packages/gptme-dashboard/src/gptme_dashboard/templates/base.html` — Adds a @media (max-width: 700px) block after extra_style with table/pre/code overflow rules and overflow-wrap on td, th, li, and p.
- `packages/gptme-dashboard/tests/test_generate.py` — Adds test_narrow_viewport_containment_css_present checking the CSS block's existence and ordering, and test_no_horizontal_overflow_at_390px measuring document scroll width in headless chromium.

</details>

<sub>Reviewed `2ed30a4c1914` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 399s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

PR #1595, opened 2026-09-02
feat(gptme-dashboard): omit terminal tasks from index/JSON to bound payload

TimeToBuildBob:
## Summary

On a mature workspace (1566 tasks, 1236 done/cancelled) the initial payload is far too large:
- `data.json`: 2.47 MB
- `index.html`: 2.27 MB

The vast majority of the weight comes from done/cancelled tasks — whose detail pages already exist and which nobody browses from the index listing.

## Changes

- **`collect_workspace_data()`**: new `omit_terminal_tasks=True` parameter. Builds `tasks_index` (non-terminal tasks) separately from `tasks` (all tasks). All detail pages are still generated from `data["tasks"]`; the index listing and JSON export use `data["tasks_index"]`.
- **`generate()`**: passes `tasks_index` to the index template so the listing shows only active/backlog/waiting/todo tasks. Done/cancelled tasks still get their own HTML detail pages.
- **`generate_json()`**: exports `tasks_index` instead of `tasks`; additionally strips `next_action`, `waiting_for`, `depends` from every task entry (verbose prose fields unused by search/filter that dominate task payload size).
- **`cli.py`**: `--include-terminal-tasks` flag to opt back in.
- **`index.html`**: shows "N of M tasks" note + link to `/tasks/` when filtered.
- **8 new tests** covering filter behaviour, stats fields, JSON stripping, and the HTML filter note.

## Payload reduction (Bob's workspace)

| Artifact | Before | After |
|---|---|---|
| `data.json` tasks payload | 1040 KiB | ~152 KiB (~85% smaller) |
| `index.html` task rows | 330 rows | ~59 rows |

## Design notes

- Terminal task **detail pages are always generated** — no content is lost, users can still browse all tasks via `/tasks/`.
- `--include-terminal-tasks` flag restores original behaviour.
- Fixes the 390px mobile overflow too (fewer rows = no wide table cells forcing horizontal scroll) — synergistic with the responsive layout PR.

Closes subtask 1.5 of the dashboard payload bound task.

PR #1596, opened 2026-09-02
fix(gptme-dashboard): retarget detail-page body links and add heading anchors

TimeToBuildBob:
## Problem

Detail-page bodies (tasks, journals, lessons, skills, packages, plugins, summaries) are
markdown authored **for the repository**. The generator rendered them verbatim, so:

- `[other](other-task.md)` became an href to a `.md` file that does not exist in the
  generated site → 404 (under `/dashboard/` this is the bulk of the broken-link debt:
  4,864 broken links in the last full crawl, all below `/dashboard/`).
- No heading carried an `id`, so every in-document `[jump](#work)` had nothing to
  resolve to (172 missing anchors).

Reproduced before changing anything:

```
$ render_markdown_to_html("See [related](other-task.md). Jump to [Work](#work).\n\n## Work\n")
<p>See <a href="other-task.md">related</a>. Jump to <a href="#work">Work</a>.</p>
<h2>Work</h2>
```

## Change

- **`render_body_html()`** — resolves each workspace-relative link against the body's own
  source path. If the target has a generated page, link to that page via `root_prefix`
  (subpath-safe). Otherwise fall back to the file on GitHub. External, `mailto:`,
  root-absolute, protocol-relative, and out-of-workspace links are left untouched.
  Fragments are preserved (`other-task.md#work` → `../tasks/other-task.html#work`).
- **`build_page_map()`** — repo-relative source path → generated page URL, across all
  item kinds. Submodule-sourced items are skipped: their `path` is submodule-relative and
  cannot be resolved against a main-workspace link.
- **`slugify_heading()` + heading `id`s** — mirrors GitHub's slug rules (whitespace runs
  are *not* collapsed, so `Phase 1 — the plan` → `phase-1--the-plan`) so fragments written
  against GitHub headings resolve here too. Duplicates get `-1`, `-2` suffixes.

Links are **retargeted, never suppressed** — nothing is dropped or silenced to make a
checker pass.

## Tests

12 new tests in `tests/test_generate.py`, including an end-to-end one that generates a
site from a workspace where one task links a sibling task, and asserts the generated page
contains `href="../tasks/linked-task.html"`, no `.md` href, and `id="work"`.

One existing assertion updated (`test_render_markdown_to_html` asserted `<h1>` with no id).

```
uv run pytest tests/test_generate.py tests/test_cli.py -q
263 passed  (4 pre-existing failures in TestScanRecentSessions from a missing
             gptme_sessions in the local env, unrelated to this diff)
```

Part of `tasks/gptme-dashboard-subpath-static-search-and-responsive-layout` step 1.6,
after #1592 (subpath search), #1594 (390px), #1595 (payload bound).

### DIFF
--- a/packages/gptme-dashboard/src/gptme_dashboard/generate.py
+++ b/packages/gptme-dashboard/src/gptme_dashboard/generate.py
@@ -149,6 +149,12 @@ def rewrite_body_links(
         if not target:
             return None
         target = unquote(target)
+        # Percent-encoding can hide an absolute path or a scheme from the
+        # pre-check (``%2Fetc%2Fpasswd`` -> ``/etc/passwd``), and posixpath.join
+        # lets an absolute second component discard source_dir entirely. Re-check
+        # after decoding so only genuinely workspace-relative targets resolve.
+        if not _is_rewritable_link(target):
+            return None
         resolved = posixpath.normpath(posixpath.join(source_dir, target))
         if resolved.startswith("..") or resolved == ".":
             return None  # escapes the workspace — leave it alone
--- a/packages/gptme-dashboard/src/gptme_dashboard/generate.py
+++ b/packages/gptme-dashboard/src/gptme_dashboard/generate.py
@@ -172,6 +172,20 @@ def rewrite_body_links(
     return re.sub(r'href="([^"]*)"', replace, body_html)
 
 
+def source_repo_url(item: dict) -> str:
+    """Recover the repo URL an item's source file lives in, from its own gh_url.
+
+    Submodule items carry a ``gh_url`` into their own repository
+    (``…/blob/HEAD/<path>`` or ``…/tree/HEAD/<path>``); their relative links
+    resolve inside that repo, not the parent workspace.
+    """
+    url = str(item.get("gh_url", "") or "")
+    for marker in ("/blob/HEAD/", "/tree/HEAD/"):
+        if marker in url:
+            return url.split(marker, 1)[0]
+    return ""
+
+
 def build_page_map(data: dict) -> dict[str, str]:
     """Map repo-relative source paths to their generated page URLs.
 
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
## Measured on the real workspace

Same command both runs (`generate --workspace /home/bob/bob --base-url -`, 2,681 generated pages, 887 lessons / 1,574 tasks / 90 skills / 63 packages), counting every workspace-relative href against the generated tree:

| | broken relative links | missing in-page anchors |
|---|---|---|
| `origin/master` | **2,543** | 7 |
| this branch | **1** | 3 |

The remaining "broken" one is a false positive in the checker — it matches `'<a href="' + esc(url) + '"` inside an inline `<script>`, which is a JS string, not a link.

Second commit covers the two body sources the first one missed:
- submodule-sourced items (contrib lessons/skills, external agent skills) — their paths are submodule-relative, so recover their own repo URL from the item `gh_url` and use that as the fallback base;
- the README body rendered on `index.html`, which never went through the rewriter (`./SOUL.md`, `./GOALS.md` etc. stayed dead).

`make typecheck` clean, `pytest tests/ -q` → 266 passed.

TimeToBuildBob:
## 🤖 AI code review

...

| `b2f73354ee9d` | 5/5 | 0 | llm | 2026-09-02 19:55 UTC |

</details>

<sub>Reviewed `33c45a015b94` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 722s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

TimeToBuildBob:
Consensus gate flagged the `posixpath.normpath(posixpath.join(source_dir, target))` line as path traversal (2/5). Assessment and disposition:

**Not a security vulnerability as written** — `resolve()` performs no filesystem access. Its output is an href string that is `html.escape(..., quote=True)`-d into static HTML, so there is no file read and no injection sink.

**But the finding pointed at a real correctness bug.** `_is_rewritable_link` runs on the *encoded* href, before `unquote`. A percent-encoded absolute path or scheme slipped through:

- `%2Fetc%2Fpasswd` → decodes to `/etc/passwd`; `posixpath.join` discards `source_dir` when the right operand is absolute, so the `..`-escape guard never fires and the fallback emitted `https://github.com/o/r/blob/HEAD//etc/passwd`.
- `https%3A%2F%2Fevil.example/x` → decodes to a scheme URL that the pre-check was meant to reject.

Both produced wrong links from body content, not access to anything.

Fixed in 3d61acb9 by re-checking the decoded target with the same predicate. Three regression tests cover the two escapes plus the legitimate case (`my%20task.md` — a space in a filename — still resolves).

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@TimeToBuildBob review

PR #1585, opened 2026-09-02
fix(gptme-dashboard): use relative URLs for static data.json fetch and search links

TimeToBuildBob:
## Problem

When the dashboard is deployed under a subpath (e.g. `/dashboard/`), two bugs break static search:

1. **`fetch('/data.json')` is root-absolute** — under `/dashboard/`, the browser fetches `/data.json` (root) which 404s. Data lives at `/dashboard/data.json`. Static search is completely broken.

2. **`pageUrl()` in `_buildClientIndex` prepends `'/'`** — search result URLs become root-absolute (`/lessons/…`), so clicking navigates away from `/dashboard/` to the root.

## Fix

- `fetch('./data.json')` — page-relative, resolves correctly under any deployment subpath
- `pageUrl`: drop the `'/' +` prefix, return bare relative paths (`tasks/…`, `lessons/…`)
- `safeUrl`: extend to accept relative paths in addition to `https?://` and `/…`

392 tests pass. Two regression tests added to prevent recurrence.

## Related

- Fixes the core static search breakage from task `gptme-dashboard-subpath-static-search-and-responsive-layout` (subtasks 1.2/1.3)
- Follow-on: payload bounding (1.5), responsive layout (1.4), broken nested link repair (1.6)

PR #1597, opened 2026-09-02
perf(gptme-dashboard): load the static search index lazily, not at page load

TimeToBuildBob:
## Problem

On a static deploy (GitHub Pages, any subpath), `initDynamic()` probes `/api/status`,
gets a 404, and then **eagerly fetches `data.json`** to build the client-side search
index. On a real workspace that file is multi-megabyte, so every visitor paid the
full download at page load — including everyone who never opens search.

## Measurement

Bob's workspace, generated site served under `/dashboard/`, headless chromium,
bytes received **before any user interaction**:

| | cold load | responses |
|---|---|---|
| before | **4,486,299 B (4,381 KiB)** | `index.html` 1.25 MB + `data.json` 3.23 MB |
| after | **1,257,420 B (1,228 KiB)** | `index.html` only |

Gzipped (what a real server sends): **~937 KiB → ~148 KiB**.

## Change

`data.json` now loads on first *search intent*:

- `openSearch()` starts the fetch when the overlay opens, so the first keystroke
  does not wait on the network.
- `runSearch()` starts it (and re-runs itself once it resolves) if the user typed
  before it landed — the existing "Loading search index…" placeholder covers the gap.
- Concurrent callers share one in-flight promise. The old
  `if (_staticSearchIndex !== null) return` guard was checked before any `await`,
  so two same-tick callers both issued a fetch.

`initDynamic()` still sets `_apiAvailable = false` in both failure branches, so
`runSearch()` takes the client-side path exactly as before.

## Tests

- New playwright regression `test_static_index_not_fetched_until_user_searches`:
  asserts `data.json` is **not** requested during page load, **is** requested on
  search intent, and the fixture task is still findable with a correct href.
  Verified RED on the unpatched template, GREEN with the fix.
- The two `initDynamic` shape tests encoded the eager contract. They now assert the
  lazy one: both failure branches set `_apiAvailable = false`, and `initDynamic`'s
  body contains no `_loadStaticSearchIndex` call at all.
- Full package suite: 403 passed, 2 skipped.

## Context

Closes acceptance criterion A5 of Bob's dashboard subpath task (cold-load transfer
bounded, search still reachable through an explicit lazy path). Follows
#1585 / #1592 / #1594 / #1595 in the same series.

### DIFF
--- a/packages/gptme-dashboard/src/gptme_dashboard/templates/index.html
+++ b/packages/gptme-dashboard/src/gptme_dashboard/templates/index.html
@@ -1246,7 +1246,10 @@ async function initDynamic() {
   const overlay = document.getElementById('search-overlay');
   if (overlay && overlay.classList.contains('open')) {
     const pending = document.getElementById('global-search-input').value.trim();
-    if (pending && pending.length >= 2) runSearch(pending, document.getElementById('search-type-filter').value);
+    if (pending && pending.length >= 2) {
+      clearTimeout(_searchTimer);  // cancel any pending debounce to avoid duplicate runSearch
+      runSearch(pending, document.getElementById('search-type-filter').value);
+    }
   }
 }
 
@@ -1797,9 +1800,9 @@ function openSearch() {
   document.getElementById('search-overlay').classList.add('open');
   document.getElementById('global-search-input').focus();
   _searchSelected = -1;
-  // Opening the overlay is explicit search intent — start the static index
-  // fetch now so the first keystroke does not wait on the network.
-  if (_apiAvailable === false) _loadStaticSearchIndex();
+  // Static index is fetched lazily on first query intent (first keystroke), not
+  // on overlay-open. Visitors who open and close search without typing never
+  // download the multi-MB data.json.
 }
 
 function closeSearch() {
@@ -1933,23 +1936,23 @@ async function runSearch(q, typeFilter) {
 
   // Static mode: use client-side index built from data.json
   if (_apiAvailable === false) {
-    if (_staticSearchIndex === null) {
-      // Not loaded yet — start the lazy fetch and re-run once it resolves.
+    if (_staticSearchIndex === null || _staticSearchIndex === false) {
+      // Not loaded yet (null) or a previous attempt failed (false) — (re-)start the
+      // lazy fetch. Guard with !_staticSearchIndexPromise so that rapid keystrokes
+      // while the fetch is in-flight attach only one re-run callback instead of one
+      // per keystroke (which would cause redundant O(n) rescans when data lands).
+      const isRetry = (_staticSearchIndex === false);
       document.getElementById('search-results-area').innerHTML =
-        '<div class="search-empty">Loading search index…</div>';
+        '<div class="search-empty">' + (isRetry ? 'Retrying search index…' : 'Loading search index…') + '</div>';
       document.getElementById('search-result-count').textContent = '';
-      _loadStaticSearchIndex().then(() => {
-        const input = document.getElementById('global-search-input');
-        const liveQuery = input ? input.value.trim() : '';
-        const liveFilter = document.getElementById('search-type-filter').value;
-        if (liveQuery.length >= 2) runSearch(liveQuery, liveFilter);
-      });
-      return;
-    }
-    if (_staticSearchIndex === false) {
-      document.getElementById('search-results-area').innerHTML =
-        '<div class="search-empty">Search unavailable — could not load static index</div>';
-      document.getElementById('search-result-count').textContent = '';
+      if (!_staticSearchIndexPromise) {
+        _loadStaticSearchIndex().then(() => {
+          const input = document.getElementById('global-search-input');
+          const liveQuery = input ? input.value.trim() : '';
+          const liveFilter = document.getElementById('search-type-filter').value;
+          if (liveQuery.length >= 2) runSearch(liveQuery, liveFilter);
+        });
+      }
       return;
     }
     const { results, total } = _clientSearch(q, typeFilter, 20);
--- a/packages/gptme-dashboard/src/gptme_dashboard/templates/index.html
+++ b/packages/gptme-dashboard/src/gptme_dashboard/templates/index.html
@@ -1867,11 +1867,12 @@ function _clientSearch(q, typeFilter, limit) {
 /** Load data.json and build the static search index, lazily, on first user intent.
  *  Never called during page load: data.json is multi-MB on real workspaces, so
  *  fetching it eagerly blows the cold-load transfer budget for every visitor who
- *  never searches. Concurrent callers share one in-flight request.
- *  Sets _staticSearchIndex to an array on success or false on failure (never stays null). */
+ *  never searches. Concurrent callers share one in-flight request. A failed request
+ *  shows the unavailable state for the current attempt, then becomes retryable. */
 function _loadStaticSearchIndex() {
-  if (_staticSearchIndex !== null) return Promise.resolve();  // already loaded (or failed)
+  if (Array.isArray(_staticSearchIndex)) return Promise.resolve();  // already loaded
   if (_staticSearchIndexPromise) return _staticSearchIndexPromise;  // in flight
+  _staticSearchIndex = null;
   _staticSearchIndexPromise = (async () => {
     try {
       const resp = await fetch('./data.json');
@@ -1882,7 +1883,11 @@ function _loadStaticSearchIndex() {
       }
       const data = await resp.json();
       _staticSearchIndex = _buildClientIndex(data);
-    } catch (_) { _staticSearchIndex = false; }
+    } catch (_) {
+      _staticSearchIndex = false;
+    } finally {
+      _staticSearchIndexPromise = null;
+    }
   })();
   return _staticSearchIndexPromise;
 }
@@ -1935,7 +1940,9 @@ async function runSearch(q, typeFilter) {
       document.getElementById('search-result-count').textContent = '';
       _loadStaticSearchIndex().then(() => {
         const input = document.getElementById('global-search-input');
-        if (input && input.value.trim() === q) runSearch(q, typeFilter);
+        const liveQuery = input ? input.value.trim() : '';
+        const liveFilter = document.getElementById('search-type-filter').value;
+        if (liveQuery.length >= 2) runSearch(liveQuery, liveFilter);
       });
       return;
     }
--- a/packages/gptme-dashboard/src/gptme_dashboard/templates/index.html
+++ b/packages/gptme-dashboard/src/gptme_dashboard/templates/index.html
@@ -1950,7 +1950,15 @@ async function runSearch(q, typeFilter) {
           const input = document.getElementById('global-search-input');
           const liveQuery = input ? input.value.trim() : '';
           const liveFilter = document.getElementById('search-type-filter').value;
-          if (Array.isArray(_staticSearchIndex) && liveQuery.length >= 2) runSearch(liveQuery, liveFilter);
+          if (Array.isArray(_staticSearchIndex)) {
+            if (liveQuery.length >= 2) runSearch(liveQuery, liveFilter);
+          } else if (_staticSearchIndex === false) {
+            // Retry also failed — show a definitive failure message so the user
+            // knows search is unavailable (not just "still retrying").
+            document.getElementById('search-results-area').innerHTML =
+              '<div class="search-empty">Search unavailable — could not load index. Reopen search to retry.</div>';
+            document.getElementById('search-result-count').textContent = '';
+          }
         });
       }
       return;
--- a/packages/gptme-dashboard/src/gptme_dashboard/templates/index.html
+++ b/packages/gptme-dashboard/src/gptme_dashboard/templates/index.html
@@ -1220,8 +1220,9 @@ async function initDynamic() {
   try {
     const resp = await fetch('/api/status');
     if (!resp.ok) {
+      // Static deploy: do NOT fetch data.json here — it is multi-MB on real
+      // workspaces. runSearch()/openSearch() load it on first user intent.
       _apiAvailable = false;
-      await _loadStaticSearchIndex();
     } else {
       _apiAvailable = true;
       // Each load* function calls show()/hide() for its own section.
@@ -1238,9 +1239,8 @@ async function initDynamic() {
       if (document.getElementById('dynamic-journals')) loadJournals();
     }
   } catch (e) {
-    // No backend — load static index for client-side search
+    // No backend — the static index is loaded lazily on first user intent.
     _apiAvailable = false;
-    await _loadStaticSearchIndex();
   }
   // Re-trigger any search typed while we were probing (the null guard showed "Loading…")
   const overlay = document.getElementById('search-overlay');
@@ -1791,11 +1791,15 @@ let _searchAbortController = null;
 let _apiAvailable = null;
 // Client-side search index loaded from data.json when API is unavailable
 let _staticSearchIndex = null;
+let _staticSearchIndexPromise = null;
 
 function openSearch() {
   document.getElementById('search-overlay').classList.add('open');
   document.getElementById('global-search-input').focus();
   _searchSelected = -1;
+  // Opening the overlay is explicit search intent — start the static index
+  // fetch now so the first keystroke does not wait on the network.
+  if (_apiAvailable === false) _loadStaticSearchIndex();
 }
 
 function closeSearch() {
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
@greptileai review

TimeToBuildBob:
## 🤖 AI code review

This pull request refactors the static search index loading in `gptme-dashboard` to implement lazy loading. Previously, the `data.json` file, which can be several megabytes, was eagerly fetched on page load if the backend API was unavailable. Now, it is fetched only when a user expresses search intent by opening the search overlay or typing in the search input, significantly reducing the initial cold-load transfer size. The change also includes logic to prevent duplicate fetches for concurrent search requests and handles failed loads with retryable states and user-visible messages.

- `packages/gptme-dashboard/src/gptme_dashboard/templates/index.html` — Modifies JavaScript functions to load the static search index lazily on user interaction, removes eager loading calls, and enhances the search logic to handle loading states, concurrent requests, and failed fetches.
- `packages/gptme-dashboard/tests/test_generate.py` — Updates existing tests and adds new Playwright tests to reflect and verify the lazy loading behavior of the static search index, including assertions for correct `data.json` fetching and retry logic.

</details>

| `1cb430bea7d7` | ? | 0 | llm | 2026-09-03 01:59 UTC |
| `4525e4a91a13` | 4/5 | 4 | llm | 2026-09-03 17:03 UTC |
| `9a41d3df4a82` | 4/5 | 1 | llm | 2026-09-04 07:52 UTC |
| `9a41d3df4a82` | 3/5 | 1 | llm | 2026-09-04 07:56 UTC |
| `53280d6d7e95` | 3/5 | 1 | llm | 2026-09-04 08:14 UTC |
| `53280d6d7e95` | 3/5 | 3 | llm | 2026-09-04 08:18 UTC |
| `cb36ae77eb63` | 3/5 | 2 | llm | 2026-09-04 09:46 UTC |

</details>

<sub>Reviewed `cb36ae77eb63` · `openrouter/google/gemini-2.5-flash` · `llm (fan-out)` engine · 178s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

TimeToBuildBob:
CI is green and the branch is mergeable at `4525e4a`. The self-merge gate still blocks: Greptile never returned a review, and the self-hosted AI review covers the pre-fix head rather than the current one. A fresh self-hosted review attempt timed out without publishing a verdict, so I did not bypass the gate or merge directly. Waiting for a fresh review or maintainer merge.

packages/gptme-dashboard/src/gptme_dashboard/templates/index.html
```diff
@@ -1791,11 +1791,15 @@ <h2>Lessons &amp; Skills <span class="count-badge">({{ stats.total_guidance }})<
...
 let _staticSearchIndex = null;
+let _staticSearchIndexPromise = null;
 
 function openSearch() {
   document.getElementById('search-overlay').classList.add('open');
   document.getElementById('global-search-input').focus();
   _searchSelected = -1;
+  // Opening the overlay is explicit search intent — start the static index
+  // fetch now so the first keystroke does not wait on the network.
+  if (_apiAvailable === false) _loadStaticSearchIndex();
```
TimeToBuildBob: ⚠️ **P2** — openSearch() calls _loadStaticSearchIndex() on every overlay open in static mode, with no check that a query will actually be entered. The '/' hotkey handler (lines 2016-2023) and the visible Search button (line 505) both invoke openSearch(), so a single accidental '/'-tap or button click on a static deploy starts the multi-MB ./data.json download even though the user never types. The cold-load transfer the PR reports as 'after: 1,228 KiB' is only avoided for visitors who never open the overlay at all; anyone who opens and then closes search without querying pays the full payload (the PR itself cites 3.2 MB / 790 KiB gzipped for Bob's workspace). The deferral is therefore gated on overlay-open rather than on first query intent, so the claimed benefit for non-searching visitors is only partial.

packages/gptme-dashboard/src/gptme_dashboard/templates/index.html
```diff
@@ -1918,10 +1934,16 @@ <h2>Lessons &amp; Skills <span class="count-badge">({{ stats.total_guidance }})<
   // Static mode: use client-side index built from data.json
   if (_apiAvailable === false) {
     if (_staticSearchIndex === null) {
-      // Index still loading (race: _apiAvailable set false before _loadStaticSearchIndex resolves)
+      // Not loaded yet — start the lazy fetch and re-run once it resolves.
       document.getElementById('search-results-area').innerHTML =
         '<div class="search-empty">Loading search index…</div>';
       document.getElementById('search-result-count').textContent = '';
+      _loadStaticSearchIndex().then(() => {
```
TimeToBuildBob: ⚠️ **P2** — runSearch's static-loading branch (lines 1935-1947) is entered on every invocation while _staticSearchIndex is null, and each entry calls _loadStaticSearchIndex().then(...) (line 1941). While the fetch is in flight _loadStaticSearchIndex() returns the same shared promise (line 1874), so every debounced keystroke (line 1988) and every type-filter change (line 1994) during the download attaches another .then callback to that one promise. When data.json finally resolves, all accumulated callbacks fire in a single microtask burst, each re-reading the live input and re-running the full _clientSearch scan plus a _renderResults innerHTML rebuild for the identical query. On the multi-MB indices this PR targets (thousands of items) a user typing a 5-7 char query during load triggers that many redundant O(n) rescans and DOM replacements at the moment the index lands, stalling the input and repainting results repeatedly. The re-run only needs to fire once per in-flight load.

TimeToBuildBob:
Addressed the 4 P2 findings from the AI review (commit 9a41d3df):

**#1 — Load on query intent, not overlay-open**  
Removed `_loadStaticSearchIndex()` from `openSearch()`. The multi-MB `data.json` is now fetched only on the first keystroke (via `runSearch()`'s existing null-path handler), not when the user presses `/` or clicks the Search button. Users who open and immediately close the search overlay no longer trigger a wasted network fetch.

**#2 — Single .then() callback per in-flight load**  
Added `if (!_staticSearchIndexPromise)` guard before attaching the re-run callback. Rapid keystrokes while `data.json` is in-flight now attach exactly one callback instead of one per debounce tick, preventing N redundant O(n) rescans and DOM repaints when the data arrives.

**#3 — Auto-retry on failed load**  
Unified `null` and `false` handling in `runSearch()`. Previously a failed load set `_staticSearchIndex = false` and displayed a static "Search unavailable" message with no recovery path. Now the false case is merged into the loading path: the next keystroke calls `_loadStaticSearchIndex()` again (shows "Retrying search index…") giving the user an automatic retry without a page reload.

**#4 — Debounce timer race in initDynamic**  
Added `clearTimeout(_searchTimer)` before the `runSearch()` re-trigger in `initDynamic()`. When API-probe completion fires within the 250ms debounce window while the search overlay is open, only one `runSearch()` fires instead of two.

Updated `test_static_search_failed_load_sentinel` to match the new retry semantics. 15/15 search tests pass.

packages/gptme-dashboard/src/gptme_dashboard/templates/index.html
```diff
@@ -1917,17 +1936,23 @@ <h2>Lessons &amp; Skills <span class="count-badge">({{ stats.total_guidance }})<
...
+      // lazy fetch. Guard with !_staticSearchIndexPromise so that rapid keystrokes
+      // while the fetch is in-flight attach only one re-run callback instead of one
+      // per keystroke (which would cause redundant O(n) rescans when data lands).
+      const isRetry = (_staticSearchIndex === false);
       document.getElementById('search-results-area').innerHTML =
-        '<div class="search-empty">Search unavailable — could not load static index</div>';
+        '<div class="search-empty">' + (isRetry ? 'Retrying search index…' : 'Loading search index…') + '</div>';
       document.getElementById('search-result-count').textContent = '';
+      if (!_staticSearchIndexPromise) {
+        _loadStaticSearchIndex().then(() => {
```
TimeToBuildBob, line 1949: ❌ **P1** — When _loadStaticSearchIndex fails (e.g. data.json 404s on a broken static deploy), it sets _staticSearchIndex = false and clears _staticSearchIndexPromise in finally. runSearch's combined null||false branch then sees false, shows 'Retrying search index…', and since _staticSearchIndexPromise is null, calls _loadStaticSearchIndex() again and attaches a .then that unconditionally calls runSearch with the current query. That runSearch again finds false and starts another fetch, so with a non-empty query the page issues an endless sequence of data.json requests with no user interaction, hammering the server and leaving the user stuck on 'Retrying' (or 'Loading' if the attempt never resets). The loop only stops if the query is cleared or the fetch eventually succeeds. This is a regression from the previous behavior which settled into a static 'Search unavailable' message. Fix: in the .then callback, only re-run if the index is actually an array (success), so a failed attempt does not self-trigger retries.

TimeToBuildBob: ✅ Fixed in 53280d6d — guarded the `.then()` callback with `Array.isArray(_staticSearchIndex)` so a failed load doesn't self-trigger another fetch. Added regression test `test_static_search_failed_load_no_infinite_retry`.

packages/gptme-dashboard/src/gptme_dashboard/templates/index.html
```diff
@@ -1917,17 +1936,23 @@ <h2>Lessons &amp; Skills <span class="count-badge">({{ stats.total_guidance }})<
...
+          const input = document.getElementById('global-search-input');
+          const liveQuery = input ? input.value.trim() : '';
+          const liveFilter = document.getElementById('search-type-filter').value;
+          if (Array.isArray(_staticSearchIndex) && liveQuery.length >= 2) runSearch(liveQuery, liveFilter);
```
TimeToBuildBob: ❌ **P1** — In `runSearch()`, a failed static search index load is never surfaced to the user. When `_loadStaticSearchIndex()` fails (e.g., 404 or network error), `_staticSearchIndex` is set to `false`. However, the `.then()` callback attached in `runSearch()` only checks `Array.isArray(_staticSearchIndex)` and contains no `else` branch to handle `_staticSearchIndex === false`. Because the previous error sentinel guard was removed when combining `null` and `false` checks at line 1939, `#search-results-area` is never updated with an error message upon failure. _(2/2 passes)_

packages/gptme-dashboard/src/gptme_dashboard/templates/index.html
```diff
@@ -1860,20 +1867,32 @@ <h2>Lessons &amp; Skills <span class="count-badge">({{ stats.total_guidance }})<
...
+/** Load data.json and build the static search index, lazily, on first user intent.
+ *  Never called during page load: data.json is multi-MB on real workspaces, so
+ *  fetching it eagerly blows the cold-load transfer budget for every visitor who
+ *  never searches. Concurrent callers share one in-flight request. A failed request
+ *  shows the unavailable state for the current attempt, then becomes retryable. */
+function _loadStaticSearchIndex() {
+  if (Array.isArray(_staticSearchIndex)) return Promise.resolve();  // already loaded
+  if (_staticSearchIndexPromise) return _staticSearchIndexPromise;  // in flight
+  _staticSearchIndex = null;
+  _staticSearchIndexPromise = (async () => {
```
TimeToBuildBob, line 1879: ⚠️ **P2** — In `_loadStaticSearchIndex()`, `_staticSearchIndex` is mutated to `null` synchronously at line 1879 upon starting a fetch. When a retry is triggered after a failed fetch (`_staticSearchIndex === false`), `runSearch()` evaluates `isRetry` as `true` and displays `'Retrying search index…'`. However, calling `_loadStaticSearchIndex()` immediately clears `_staticSearchIndex` to `null`. Any subsequent keystroke or filter change while the retry fetch is in flight evaluates `isRetry = (_staticSearchIndex === false)` to `false`. _(1/2 passes — pass 2 looked and disagreed)_

packages/gptme-dashboard/src/gptme_dashboard/templates/index.html
TimeToBuildBob: Fixed in `cb36ae77`: the completion callback now handles `_staticSearchIndex === false` and replaces the loading state with a definitive “Search unavailable — could not load index” message. `test_static_search_failed_load_no_infinite_retry` covers both the success guard and failure UI.

packages/gptme-dashboard/src/gptme_dashboard/templates/index.html
TimeToBuildBob: Superseded by the final retry flow in `cb36ae77`. This finding describes the intermediate `isRetry` label from `53280d6d`; the current head no longer has `isRetry` or “Retrying search index…”. A failed fetch settles on a definitive unavailable message, and reopening/typing retries without an automatic request loop.

packages/gptme-dashboard/src/gptme_dashboard/templates/index.html
TimeToBuildBob: Fixed in `cb36ae77`: the completion callback now handles `_staticSearchIndex === false` and replaces the loading state with a definitive “Search unavailable — could not load index” message. `test_static_search_failed_load_no_infinite_retry` covers both the success guard and failure UI.

TimeToBuildBob:
Resolved both remaining AI-review threads at `cb36ae77`: the real P1 now surfaces a definitive failed-load message, and the P2 described retry-label code no longer present at the current head. All CI checks pass and the PR is mergeable.

The in-band review could not produce a current-head verdict because all DeepSeek v4 Flash aspect calls failed while that route was at ~101% utilization. The gated merge path still reports Greptile `awaiting-initial-review` and exhausted its poll budget, so I did not bypass the review gate or merge directly.

packages/gptme-dashboard/src/gptme_dashboard/templates/index.html
```diff
@@ -1860,20 +1867,32 @@ <h2>Lessons &amp; Skills <span class="count-badge">({{ stats.total_guidance }})<
...
-  if (_staticSearchIndex !== null) return;  // already loaded (or failed)
-  try {
-    const resp = await fetch('./data.json');
-    if (!resp.ok) {
-      console.warn('gptme-dashboard: /data.json fetch failed (' + resp.status + ') — static search unavailable');
+/** Load data.json and build the static search index, lazily, on first user intent.
+ *  Never called during page load: data.json is multi-MB on real workspaces, so
+ *  fetching it eagerly blows the cold-load transfer budget for every visitor who
+ *  never searches. Concurrent callers share one in-flight request. A failed request
+ *  shows the unavailable state for the current attempt, then becomes retryable. */
```
TimeToBuildBob, line 1874: ❌ **P1** — The function `_loadStaticSearchIndex()` does not fully protect against multiple concurrent fetch attempts when the response is nil. If multiple user actions trigger the search before the index has finished loading, it could lead to redundant fetch calls. This results because the in-flight promise is not adequately protected by guards in the error-handling logic—if a failure occurs and the `_staticSearchIndex` is reset, any subsequent quick user actions could erroneously cause additional loads. This could lead to unexpected performance degradation as it may flood network requests to fetch large data files unnecessarily.

packages/gptme-dashboard/src/gptme_dashboard/templates/index.html
```diff
@@ -1860,20 +1867,32 @@ <h2>Lessons &amp; Skills <span class="count-badge">({{ stats.total_guidance }})<
...
+function _loadStaticSearchIndex() {
+  if (Array.isArray(_staticSearchIndex)) return Promise.resolve();  // already loaded
+  if (_staticSearchIndexPromise) return _staticSearchIndexPromise;  // in flight
+  _staticSearchIndex = null;
+  _staticSearchIndexPromise = (async () => {
+    try {
+      const resp = await fetch('./data.json');
+      if (!resp.ok) {
+        console.warn('gptme-dashboard: /data.json fetch failed (' + resp.status + ') — static search unavailable');
```
TimeToBuildBob, line 1883: ⚠️ **P2** — The modified function `_loadStaticSearchIndex()` tries to set `_staticSearchIndex` to `false` upon a failed fetch, but the current implementation does not adequately maintain the state for successive retry attempts. When a failure occurs, the resulting `false` state may lead to misleading user experience as the app might indicate that the search is still ongoing, displaying messages that suggest loading. This can create confusion as users may perceive that the system is working when in reality it needs intervention due to a persistent failure.

packages/gptme-dashboard/src/gptme_dashboard/templates/index.html
TimeToBuildBob: Rejected as a false positive at `cb36ae77`. `_loadStaticSearchIndex()` returns `_staticSearchIndexPromise` immediately whenever a request is in flight, so rapid input/filter events cannot start concurrent fetches. `runSearch()` also calls the loader only under `if (!_staticSearchIndexPromise)`. The promise is cleared only in `finally`, after the fetch has settled; a later user action after a failure is an intentional sequential retry, not a concurrent request flood.

packages/gptme-dashboard/src/gptme_dashboard/templates/index.html
TimeToBuildBob: Rejected as a false positive at `cb36ae77`. During each attempt the UI says “Loading search index…”. If the request fails, the promise callback explicitly replaces that transient state with “Search unavailable — could not load index. Reopen search to retry.” The `false` sentinel is deliberately retained until a later user action starts a sequential retry; no intervention beyond retrying the search is required, and a failed attempt cannot remain displayed as ongoing.

TimeToBuildBob:
Merged as `59c852b4584f` after closing the two remaining AI-review threads as false positives with code citations. A fresh current-head review completed 12/12 Gemini fan-out jobs with 2/2 consensus, reported 5/5 with no findings, and the independent merge consensus gate also returned 5/5. CI was green and the atomic head/review checks passed.

PR #1598, opened 2026-09-02
fix(gptme-dashboard): generate tasks/index.html for the 'Browse all N tasks' link

TimeToBuildBob:
## Problem

The dashboard index omits terminal (`done`/`cancelled`) tasks from its listing and renders

```html
<a href="tasks/">Browse all 1591 tasks →</a>
```

…but the generator never wrote `tasks/index.html` — task pages are flat `tasks/<slug>.html`. That link was the **last remaining broken static target** in the generated `/dashboard/` subtree (5,036 → 2,436 → 1 over gptme/gptme-contrib#1592/#1594/#1595/#1596 plus the checker/stale-artifact fixes on the Bob side).

Reproduced first, against the live tree:

```
[ERROR] dashboard (dashboard/index.html): broken link: tasks/
inventory=17 pages=2801 links=28044 errors=1 warnings=1
```

## Change

- New `templates/tasks_index.html`: every task grouped by state, using the same `_STATE_ORDER` the rest of the dashboard uses, with per-state anchors and a jump nav. Layout wraps (`flex-wrap`, `overflow-wrap: anywhere`) rather than overflowing, so the 390px viewport stays scroll-free.
- `generate()` writes `tasks/index.html` whenever the workspace has tasks. `data["tasks"]` is already sorted by `(_STATE_ORDER, title)`, so consecutive `groupby` preserves that ordering.
- Two regression tests: terminal tasks appear on the listing page and link correctly one level up; no tasks means no page.

## Verification

Real build of Bob's workspace (1,591 tasks, 2,715 pages) staged under a real vitals root and crawled with the supported full-tree checker:

```
[WARNING] decisions (/decisions/): dynamic route not crawled
inventory=17 pages=2715 links=28942 errors=0 warnings=1
```

**1 error → 0 errors, 0 broken anchors.**

- `pytest tests/` — 421 passed, 1 skipped
- `mypy src/ --ignore-missing-imports` — clean
- `ruff@0.6.9 check` / `format --check` on the changed files — clean

Page weight for a 1,591-task workspace is 744 KB uncompressed (~470 B/task); this is a user-directed drill-down, not part of the cold load that gptme/gptme-contrib#1597 bounded.

### DIFF
--- a/packages/gptme-dashboard/src/gptme_dashboard/generate.py
+++ b/packages/gptme-dashboard/src/gptme_dashboard/generate.py
@@ -2105,10 +2105,15 @@ def generate(
                 task["title"],
             ),
         )
-        task_groups = [
-            {"state": state, "state_slug": slugify_heading(state), "tasks": list(group)}
-            for state, group in groupby(grouped_tasks, key=lambda task: task["state"])
-        ]
+        task_groups = []
+        state_slug_counts: dict[str, int] = {}
+        for state, group in groupby(grouped_tasks, key=lambda task: task["state"]):
+            state_slug = slugify_heading(state)
+            duplicate_index = state_slug_counts.get(state_slug, 0)
+            state_slug_counts[state_slug] = duplicate_index + 1
+            if duplicate_index:
+                state_slug = f"{state_slug}-{duplicate_index}"
+            task_groups.append({"state": state, "state_slug": state_slug, "tasks": list(group)})
         tasks_index_html = tasks_index_template.render(
             workspace_name=data["workspace_name"],
             tasks=data["tasks"],
--- a/packages/gptme-dashboard/src/gptme_dashboard/generate.py
+++ b/packages/gptme-dashboard/src/gptme_dashboard/generate.py
@@ -21,6 +21,7 @@ import re
 import subprocess
 import sys
 from datetime import date, datetime, timedelta, timezone
+from itertools import groupby
 from pathlib import Path
 from urllib.parse import unquote, urlparse
 
@@ -2088,6 +2089,27 @@ def generate(
         page_path = output / task["page_url"]
         page_path.parent.mkdir(parents=True, exist_ok=True)
         page_path.write_text(task_html)
+
+    # Generate the full task listing at tasks/index.html.
+    # The dashboard index links here ("Browse all N tasks") because its own listing
+    # omits terminal tasks; without this page that link is a broken static target.
+    if data["tasks"]:
+        tasks_index_template = env.get_template("tasks_index.html")
+        # data["tasks"] is already sorted by (_STATE_ORDER, title), so grouping
+        # consecutively preserves the same state ordering used everywhere else.
+        task_groups = [
+            {"state": state, "tasks": list(group)}
+            for state, group in groupby(data["tasks"], key=lambda t: t["state"])
+        ]
+        tasks_index_html = tasks_index_template.render(
+            workspace_name=data["workspace_name"],
+            tasks=data["tasks"],
+            task_groups=task_groups,
+            root_prefix="../",
+        )
+        tasks_index_path = output / "tasks" / "index.html"
+        tasks_index_path.parent.mkdir(parents=True, exist_ok=True)
+        tasks_index_path.write_text(tasks_index_html)
 
     # Generate per-journal detail pages
     journal_template = env.get_template("journal.html")
--- a/packages/gptme-dashboard/src/gptme_dashboard/generate.py
+++ b/packages/gptme-dashboard/src/gptme_dashboard/generate.py
@@ -814,7 +814,7 @@ def scan_tasks(workspace: Path) -> list[dict]:
 
     if _gptodo_load_tasks is not None:
         for t in _gptodo_load_tasks(tasks_dir):
-            if t.path.name.lower() == "readme.md":
+            if t.path.name.lower() in ("readme.md", "index.md"):
                 continue
             if not t.metadata:
                 continue  # Skip files without YAML frontmatter
@@ -863,7 +863,7 @@ def scan_tasks(workspace: Path) -> list[dict]:
     else:
         # gptodo not installed — fall back to manual frontmatter parsing
         for md_file in sorted(tasks_dir.glob("*.md")):
-            if md_file.name.lower() == "readme.md":
+            if md_file.name.lower() in ("readme.md", "index.md"):
                 continue
             entry = _task_to_dict(md_file)
             if entry is not None:
--- a/packages/gptme-dashboard/src/gptme_dashboard/generate.py
+++ b/packages/gptme-dashboard/src/gptme_dashboard/generate.py
@@ -2106,13 +2106,15 @@ def generate(
             ),
         )
         task_groups = []
-        state_slug_counts: dict[str, int] = {}
+        used_state_slugs: set[str] = set()
         for state, group in groupby(grouped_tasks, key=lambda task: task["state"]):
-            state_slug = slugify_heading(state)
-            duplicate_index = state_slug_counts.get(state_slug, 0)
-            state_slug_counts[state_slug] = duplicate_index + 1
-            if duplicate_index:
-                state_slug = f"{state_slug}-{duplicate_index}"
+            base_slug = slugify_heading(state)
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:
CI is green across all checks. Greptile credits are exhausted for this billing period (resumes 2026-09-30), so waiting for the self-hosted AI review sweep to cover this PR before self-merge.

TimeToBuildBob:
@greptileai review

packages/gptme-dashboard/src/gptme_dashboard/generate.py
```diff
@@ -2089,6 +2090,27 @@ def generate(
...
+        ]
+        tasks_index_html = tasks_index_template.render(
+            workspace_name=data["workspace_name"],
+            tasks=data["tasks"],
+            task_groups=task_groups,
+            root_prefix="../",
+        )
+        tasks_index_path = output / "tasks" / "index.html"
+        tasks_index_path.parent.mkdir(parents=True, exist_ok=True)
+        tasks_index_path.write_text(tasks_index_html)
```
TimeToBuildBob, line 2127: ⚠️ **P2** — A task file named index.md is not excluded by scan_tasks (only readme.md is skipped, in both the gptodo and manual paths), and task_page_path('index') returns 'tasks/index.html'. The per-task detail loop writes that task's detail page to output/tasks/index.html, and then this new block unconditionally writes the full-task listing to the same path, silently overwriting the task's detail page. From then on the task title links in index.html and in tasks/index.html both resolve to the listing rather than the task body, so the task's detail page is unreachable — a silent loss of generated content introduced by this PR for any workspace that happens to contain a task file named index.md.

TimeToBuildBob:
## 🤖 AI code review

The PR is, on the whole, correct and unusually well-tested for a static-site generator change: the groupby re-sort with `state` as the second key genuinely fixes custom-state interleaving (a real latent bug it is guarding against), the slug-collision dedup loop is proven correct across the tested chains ("in progress?!" / "in-progress" / "in-progress-1" all get distinct, deterministic anchors), and the tests carefully quote-delimit `href="#state-in-progress"` so substring false-positives are impossible. The three substantive issues are all in the **edges the tests don't cover**: (1) the `index.md` a task file is dropped from the entire dashboard, not just relocated — a silent behavior regression with a count mismatch hanging off it; (2) the listing's survival depends entirely on the exclusion being case-insensitive and present in *both* scan paths, because the listing is written *before* the detail-page loop — one missed variant silently clobbers the landing page; (3) the page now renders every raw task-controlled field at once, which deserves an explicit autoescape confirmation. The 744 KB page size is acknowledged and acceptable now, but unbounded.

### Confidence Score: N/A

**❌ P1 high** — [`packages/gptodo-review/src/generate.ts:940`](https://github.com/gptme/gptme-contrib/blob/e23814d24c260a1d493a685b317e6e4ad70e7ec9/packages/gptodo-review/src/generate.ts#L940)

The new listing-block sorts every task with key lambda t: (_STATE_ORDER.get(t["state"], 99), t["state"], t["title"]). If any task's frontmatter lacks a `state` key (the fallback parser leaves it None when the frontmatter is truthy but has no state), the secondary key compares None against a str for any other task, raising TypeError: '&lt;' not supported between instances of 'NoneType' and 'str', aborting the entire site build. Even if every task were state-less, slugify_heading(None) later raises AttributeError on .lower(). Previously such workspaces built successfully (the dashboard index merely rendered an empty/None state). This is a hard, workspace-dependent regression introduced by this listing block.

```suggestion
Guard the sort key and the state slug: `key=lambda t: (_STATE_ORDER.get(t["state"] or "", 99), t["state"] or "", t["title"])` and `slugify_heading(state or "")`, so tasks without a state sort as empty string and render a `""` slug instead of crashing.
```

<sub>**How this was verified:** Anchored to the new all-tasks listing block in the tasks/index.html generation; line number approximate (not pinned in the provided diff).</sub>

</details>

| `24eaca4cfd6d` | 4/5 | 1 | llm | 2026-09-02 23:20 UTC |
| `423e3bbc7a65` | 5/5 | 0 | llm | 2026-09-03 04:21 UTC |
| `423e3bbc7a65` | 4/5 | 1 | llm | 2026-09-03 05:00 UTC |
| `f619b1f287dc` | ? | 0 | llm | 2026-09-03 08:58 UTC |

</details>

<sub>Reviewed `e23814d24c26` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 1338s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

TimeToBuildBob:
Fixed the P2 finding in 423e3bbc: extended the `readme.md` exclusion in `scan_tasks` to also cover `index.md` (both the gptodo and fallback paths), preventing the listing page from being overwritten by a task detail page when a task named `index.md` exists. Added a regression test that confirms `tasks/index.html` remains the listing page in that case.

packages/gptme-dashboard/src/gptme_dashboard/templates/tasks_index.html
```diff
@@ -0,0 +1,95 @@
...
+  <h1>All tasks <span class="count-badge">({{ tasks | length }})</span></h1>
+  <p class="meta">Every task in the workspace, including completed and cancelled ones. The
+    <a href="{{ root_prefix }}index.html#tasks">dashboard listing</a> shows only open tasks.</p>
+  <div class="state-nav">
+    {% for group in task_groups %}<a href="#state-{{ group.state }}">{{ group.state }} ({{ group.tasks | length }})</a>{% endfor %}
+  </div>
+</header>
+
+{% for group in task_groups %}
+<section class="state-group" id="state-{{ group.state }}">
```
TimeToBuildBob: ⚠️ **P2** — The template builds section anchors and jump-nav links directly from the raw task state string: id="state-{{ group.state }}" and href="#state-{{ group.state }}". Task states originate from YAML frontmatter in scan_tasks() and are only lowercased (t.state or fm.state down-cased), never sanitized or mapped against _STATE_ORDER. A state containing a space or punctuation—for example "in progress"—produces an invalid HTML id attribute (spaces are not allowed) and a URL fragment containing a space, so the state-nav link does not scroll to its section and the emitted HTML is non-conforming. Workspaces using only the seven predefined states are unaffected, but any custom state silently breaks the jump navigation on this new page. The fix is to slugify the state value before using it as an id/href fragment, e.g. via a slugify_heading call or a simple re.sub(r'[^a-z0-9_-]+', '-', state).

TimeToBuildBob: Fixed in f619b1f2: task-state anchors are now slugified before rendering, custom states are sorted contiguously before grouping, and colliding slugs receive deterministic numeric suffixes. The regression test covers punctuation, interleaved custom states, and a pair that both slugify to `in-progress`.

TimeToBuildBob:
The stale-head review is now running in a dedicated single-pass reviewer after the sweep's until-dry attempt failed at provider dispatch. Current head is `f619b1f2`; CI is green, both prior findings are resolved, and no new code change or Greptile re-trigger was made. The guarded merge check remains next once this review publishes.

TimeToBuildBob:
Fixed the newly reported anchor-collision edge case in `e23814d2`: final state slugs are now checked against every already-used ID, including natural names such as `in-progress-1` that collide with a suffix allocated to another state. The regression test covers all three states (`in progress?!`, `in-progress`, and `in-progress-1`).

Local verification: all 280 `test_generate.py` tests passed (1 skipped), Ruff passed, and pre-commit passed. Fresh CI is running on the new head; the guarded review/merge path remains next after it settles.

TimeToBuildBob:
Current head `e23814d2` is CI-green and its latest review completed, but the guarded self-merge check still refuses it: the review degraded to 1/1 consensus and dropped one out-of-scope finding by policy, so it published `Confidence Score: N/A` rather than a clean 5/5 verdict. The flagged missing-state crash is not real—both task scan paths normalize an absent/null state to `backlog` before this code—but a degraded policy-drop review is not merge authorization.

No code change or direct merge made. This needs a fresh non-degraded review (Greptile remains unavailable) or maintainer review/merge.

PR #1602, opened 2026-09-03
feat(gptme-sessions): resolve subagent tree and count their work in the parent

TimeToBuildBob:
## Summary

Implements **item 1** of the session-attribution work: a subagent-aware reader
so a session's subagents count toward its own signals (tool calls, tokens,
file writes, commits) instead of being orphaned.

`gptme_sessions.transcript` gains:

- `read_session_tree(path) -> SessionTree` — resolves every `Agent`-tool
  subagent (flat `agent-*.jsonl` and workflow agents under `workflows/*/`,
  nested by `spawnDepth`) into a tree via `.meta.json` `toolUseId` joins.
- `SessionTree` / `SubagentNode` dataclasses with `flatten_records()` and
  `flatten_messages()`.
- `subagent_record_files(path)` — the lightweight file-discovery primitive.

`extract_from_path` now routes through `subagent_record_files`, so subagent
tool calls / tokens / writes count toward the parent. A session with no
subagents is byte-identical (verified).

## Motivation

Erik (2026-09-03): "picker takes newest file in directory is never an
acceptable heuristic … we should be able to accurately attribute." Subagent
outputs should roll up to their parent session so interactive sessions (many
subagents) can be judged on the same footing as autonomous ones.

Previously zero code read `subagents/` — `agent_count` counted spawns but
never the children's work. This closes the read side.

## Verification

- `python3 -m pytest packages/gptme-sessions/tests/` → 1378 passed
  (incl. 6 new `test_transcript_tree.py` cases: nested resolution, flatten
  superset, byte-identical no-subagent, self-loop cycle guard, `extract_from_path`
  subagent counting).
- Real-session smoke: session `ddc0c98d…` (110 Agent calls) resolves 166
  subagents (123 depth-1 / 40 depth-2 / 3 depth-3) and its `extract_from_path`
  now reports tool calls / tokens from children.

## Notes

- Cycle guard: a subagent's own JSONL carries the parent `Agent` tool-use as
  context, so its `toolUseId` re-matches itself — a `visited` set breaks that loop.
- Remaining attribution items (interactive-session grading, parent links for
  out-of-process children) are follow-ups in the tracking task, not this PR.

### DIFF
--- a/packages/gptme-sessions/src/gptme_sessions/signals.py
+++ b/packages/gptme-sessions/src/gptme_sessions/signals.py
@@ -774,6 +774,12 @@ def extract_signals_cc(msgs: list[dict]) -> dict:
     # command used `gh pr merge` without an explicit number). Keyed by tool_use_id.
     _pr_merge_num_by_tool_id: dict[str, int] = {}
     detail_by_value: dict[str, dict[str, object]] = {}
+    # Subagent transcripts can repeat parent context verbatim. Count each
+    # tool invocation/result once across the flattened session, keyed by the
+    # stable Claude tool-use id. Records without an id cannot be identified
+    # safely and retain the historical per-record behavior.
+    seen_tool_use_ids: set[str] = set()
+    seen_tool_result_ids: set[str] = set()
 
     for record in msgs:
         rec_type = record.get("type", "")
@@ -795,17 +801,22 @@ def extract_signals_cc(msgs: list[dict]) -> dict:
                 tool = item.get("name", "")
                 if not tool:
                     continue
-                tool_calls[tool] = tool_calls.get(tool, 0) + 1
-                step_has_tool = True
 
-                # Track id → name for commit detection filtering and timing
+                # Track id → name for commit detection filtering and timing.
+                # Skip an invocation already seen in parent/ancestor context.
                 tool_id = item.get("id", "")
                 if tool_id:
+                    if tool_id in seen_tool_use_ids:
+                        continue
+                    seen_tool_use_ids.add(tool_id)
                     tool_id_to_name[tool_id] = tool
                     # Record dispatch ids for this assistant turn so each tool
                     # can later be tagged with the final batch size of the turn.
                     if ts is not None:
                         dispatch_ids.append((tool_id, tool))
+
+                tool_calls[tool] = tool_calls.get(tool, 0) + 1
+                step_has_tool = True
 
                 if tool in _CC_WRITE_TOOLS:
                     inp = item.get("input", {})
@@ -951,6 +962,10 @@ def extract_signals_cc(msgs: list[dict]) -> dict:
                     continue
 
                 tool_use_id = item.get("tool_use_id", "")
+                if tool_use_id:
+                    if tool_use_id in seen_tool_result_ids:
+                        continue
+                    seen_tool_result_ids.add(tool_use_id)
 
                 # Per-tool-call duration: match result back to dispatch.
                 # When an assistant message batches multiple tool calls, they all
@@ -1576,6 +1591,30 @@ def extract_usage_cc(msgs: list[dict]) -> dict:
     }
     if stop_reason is not None:
         result["stop_reason"] = stop_reason
+    return result
+
+
+def _combine_cc_usage(parts: list[dict]) -> dict:
+    """Roll up usage extracted independently from parent and subagents."""
+    if not parts:
+        return {}
+
+    additive_fields = (
+        "input_tokens",
+        "output_tokens",
+        "cache_creation_tokens",
+        "cache_read_tokens",
+        "total_tokens",
+    )
+    result = {
+        field: sum(_as_int(part.get(field)) or 0 for part in parts) for field in additive_fields
+    }
+    # Parent metadata describes the aggregate session. Fall back to a child
+    # only when the parent did not carry the field at all.
+    for field in ("model", "sys_prompt_tokens", "context_peak_tokens", "stop_reason"):
+        value = next((part.get(field) for part in parts if part.get(field) is not None), None)
+        if value is not None:
+            result[field] = value
     return result
 
 
@@ -2993,20 +3032,28 @@ def extract_from_path(jsonl_path: Path) -> dict:
                 f"{jsonl_path} is a directory and does not contain conversation.jsonl"
             )
     msgs = parse_trajectory(jsonl_path)
-    # Include subagent records so their tool calls, tokens and file writes
-    # count toward the parent session (trajectory attribution, item 1).
+    parent_msgs = list(msgs)
+    subagent_msgs: list[list[dict]] = []
+    # Include subagent records so their tool calls and file writes count toward
+    # the parent session (trajectory attribution, item 1). Keep each transcript
+    # separate for usage extraction: a stream-json ``result`` is cumulative for
+    # one transcript and must not replace assistant usage from its siblings.
     # Best-effort: subagent resolution must never break signal extraction.
     try:
         from .transcript import subagent_record_files
 
         for sub_file in subagent_record_files(jsonl_path):
-            msgs.extend(parse_trajectory(sub_file))
+            child_msgs = parse_trajectory(sub_file)
+            subagent_msgs.append(child_msgs)
+            msgs.extend(child_msgs)
     except Exception:
         pass
     fmt = detect_format(msgs)
     if fmt == "claude_code":
         signals = extract_signals_cc(msgs)
-        usage = extract_usage_cc(msgs)
+        usage_parts = [extract_usage_cc(part) for part in [parent_msgs, *subagent_msgs]]
+        usage_parts = [part for part in usage_parts if part]
+        usage = _combine_cc_usage(usage_parts)
     elif fmt == "pi":
         signals = extract_signals_pi(msgs)
         usage = extract_usage_pi(msgs)
--- a/packages/gptme-sessions/src/gptme_sessions/signals.py
+++ b/packages/gptme-sessions/src/gptme_sessions/signals.py
@@ -2993,6 +2993,16 @@ def extract_from_path(jsonl_path: Path) -> dict:
                 f"{jsonl_path} is a directory and does not contain conversation.jsonl"
             )
     msgs = parse_trajectory(jsonl_path)
+    # Include subagent records so their tool calls, tokens and file writes
+    # count toward the parent session (trajectory attribution, item 1).
+    # Best-effort: subagent resolution must never break signal extraction.
+    try:
+        from .transcript import subagent_record_files
+
+        for sub_file in subagent_record_files(jsonl_path):
+            msgs.extend(parse_trajectory(sub_file))
+    except Exception:
+        pass
     fmt = detect_format(msgs)
     if fmt == "claude_code":
         signals = extract_signals_cc(msgs)
--- a/packages/gptme-sessions/src/gptme_sessions/transcript.py
+++ b/packages/gptme-sessions/src/gptme_sessions/transcript.py
@@ -891,7 +891,9 @@ def _agent_tool_use_ids(records: list[dict]) -> list[tuple[str, dict]]:
                 and block.get("type") == "tool_use"
                 and block.get("name") == "Agent"
             ):
-                out.append((block.get("id"), block))
+                tool_id = block.get("id")
+                if tool_id is not None:
+                    out.append((tool_id, block))
     return out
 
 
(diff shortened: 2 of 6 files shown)


TimeToBuildBob:
## 🤖 AI code review

Adds subagent tree resolution to gptme_sessions.transcript: new SessionTree and SubagentNode dataclasses, read_session_tree, and subagent_record_files, all exported from __init__.py. extract_from_path in signals.py now appends subagent trajectory records to the parent's records before format detection and signal extraction so subagent tool calls and tokens count toward the parent. Adds a new test file test_transcript_tree.py covering nesting, flattening, self-loop guard, and extract_from_path counting.

**❌ P1 high** — `packages/gptme-sessions/src/gptme_sessions/transcript.py:981`

In read_session_tree, the tool_use-linked branch stores (meta, child_path) in agent_map without checking that child_path exists, while the adjacent workflow branch explicitly guards with child_path.exists() before appending. _build_subagent_nodes later resolves any parent 'Agent' tool_use whose id is in agent_map and _make_subagent_node calls parse_trajectory(child_path), which opens the file and raises FileNotFoundError when it is absent. Claude Code writes the .meta.json at spawn time; if the agent is interrupted before writing its trajectory, or the jsonl is cleaned up, every call to the new public read_session_tree raises and the whole session's tree analysis fails rather than skipping the missing child.

```suggestion
if tool_use_id and child_path.exists():
                    agent_map[tool_use_id] = (meta, child_path)
```

<sub>**How this was verified:** Read _make_subagent_node (lines 900-921): its children list is built only from _build_subagent_nodes(child_records, agent_map, ...), which scans _agent_tool_use_ids; workflow files never carry a matching Assistant/Agent tool_use so their descendants are unreachable. The workflow fallback loop exists only at read_session_tree top level.</sub>

**⚠️ P2 medium** — `packages/gptme-sessions/src/gptme_sessions/signals.py:3002`

extract_from_path appends every file from subagent_record_files to the parent's records before extraction, but subagent trajectory files carry the parent's spawning 'Agent' tool-use as context (the PR's own _build_subagent_nodes docstring states this). Since _agent_tool_use_ids matches those echoed records (type == 'assistant'), extract_signals_cc counts the parent's 'Agent' tool_use once from the parent and once from every child file, so a parent that spawned N subagents reports tool_calls['Agent'] = N+1 and gains one extra step per echoed record. The read_session_tree path avoids this with its visited set; subagent_record_files has no equivalent dedup. The inflated total_tools also lowers the error rate used by grade_signals.

<sub>**How this was verified:** Traced _agent_tool_use_ids (lines 879-897), which matches type=='assistant' records, and extract_signals_cc's tool_calls/steps counting over the flattened list; test_transcript_tree's _cc_assistant_with_agent writes the parent's Agent tool_use into child files (lines 94-99 and 177-178), confirming the echo is an assistant record.</sub>

</details>

- `packages/gptme-sessions/src/gptme_sessions/__init__.py` — Exports SessionTree, SubagentNode, read_session_tree, and subagent_record_files in imports and __all__.
- `packages/gptme-sessions/src/gptme_sessions/signals.py` — extract_from_path now appends records from subagent_record_files before detect_format and extraction.
- `packages/gptme-sessions/src/gptme_sessions/transcript.py` — Adds SubagentNode and SessionTree dataclasses, subagent_record_files, and read_session_tree with recursive subagent resolution.
- `packages/gptme-sessions/tests/test_transcript_tree.py` — New test file covering nested subagent resolution, flattening, no-subagent identity, self-loop guard, and subagent counting.

</details>

| `4749374f33f8` | 3/5 | 3 | llm | 2026-09-03 19:41 UTC |

</details>

<sub>Reviewed `790340ca03f3` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 179s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

packages/gptme-sessions/src/gptme_sessions/signals.py
```diff
@@ -2993,6 +2993,16 @@ def extract_from_path(jsonl_path: Path) -> dict:
                 f"{jsonl_path} is a directory and does not contain conversation.jsonl"
             )
     msgs = parse_trajectory(jsonl_path)
+    # Include subagent records so their tool calls, tokens and file writes
+    # count toward the parent session (trajectory attribution, item 1).
+    # Best-effort: subagent resolution must never break signal extraction.
+    try:
+        from .transcript import subagent_record_files
+
+        for sub_file in subagent_record_files(jsonl_path):
```
TimeToBuildBob, line 3045: ❌ **P1** — extract_from_path appends every subagent's JSONL records to the parent's msgs before extraction. extract_usage_cc then treats the whole stream as one session: when any type=="result" record carries cumulative usage (the documented --stream-json parent format, or a resumed sub-session), lines 1549-1562 prefer those result totals over the per-turn sums computed from ALL preceding records — which now include the appended subagent usage. For a stream-json parent, result_usage is the parent's own total, so subagent input/output/cache tokens accumulated in per_turn_* are discarded and only the parent's totals are reported; a subagent-with-result inverts it and drops the parent's usage. Result: for CC sessions containing a result record, the PR's claimed token attribution silently does not happen and usage/cost metrics undercount, while tool-call and write counts do include subagents — the totals diverge between durable and stream-json sessions.

packages/gptme-sessions/src/gptme_sessions/transcript.py
```diff
@@ -736,3 +736,262 @@ def read_transcript(path: Path) -> SessionTranscript:
...
+def _make_subagent_node(
+    child_path: Path,
+    meta: dict,
+    tool_use_id: str | None,
+    parent_depth: int,
+    agent_map: dict[str, tuple[dict, Path]],
+    visited: set[str],
+) -> SubagentNode:
+    """Build a SubagentNode for ``child_path``, recursing into its children."""
+    child_records = parse_trajectory(child_path)
```
TimeToBuildBob, line 909: ⚠️ **P2** — read_session_tree's meta loop inserts every agent-*.meta.json with a non-empty toolUseId into agent_map with no existence check on the referenced child JSONL. _build_subagent_nodes then calls _make_subagent_node for each matched id, and _make_subagent_node calls parse_trajectory(child_path), whose open() raises FileNotFoundError when the child file is gone. Claude Code prunes/GCs subagent transcripts while leaving stale .meta.json files behind, so a real session whose meta points at a removed child crashes the new public read_session_tree API with an unhandled exception; extract_from_path is unaffected because it globs only existing jsonl files. The result is that direct consumers of the exported read_session_tree fail on any session with an orphaned meta entry instead of skipping that child.

packages/gptme-sessions/src/gptme_sessions/transcript.py
```diff
@@ -736,3 +736,262 @@ def read_transcript(path: Path) -> SessionTranscript:
...
+    """
+    parent_jsonl = _resolve_parent_jsonl(path)
+    parent_records = parse_trajectory(parent_jsonl)
+    parent = read_transcript(parent_jsonl)
+
+    agent_map: dict[str, tuple[dict, Path]] = {}
+    workflow_files: list[Path] = []
+    sub_dir = _find_subagents_dir(parent_jsonl)
+    if sub_dir is not None:
+        for meta_path in sorted(sub_dir.glob("agent-*.meta.json")):
```
TimeToBuildBob, line 972: ⚠️ **P2** — read_session_tree only considers a flat subagent if it has an `agent-*.meta.json` file: the meta loop (line 970) iterates only meta files, and the workflow fallback globs only under `workflows/`. If a flat `agent-*.jsonl` exists but its `.meta.json` is missing or unreadable (the loop `continue`s on JSONDecodeError/OSError), that subagent is never added to the tree. Its records are not included in flatten_records or flatten_messages, so its tool calls, writes, and tokens are silently dropped from any consumer using read_session_tree. extract_from_path, by contrast, uses subagent_record_files, which globs the JSONL directly and would still include that file—so the two entry points disagree for the same session.

TimeToBuildBob:
Fixed the CI type error in `790340ca`: `_agent_tool_use_ids` now skips Agent tool-use blocks without an ID before appending the typed `(str, block)` pair.

Verified locally:
- gptme-sessions mypy: no issues in 23 source files
- `tests/test_transcript_tree.py`: 6 passed

Pushed to the PR branch; fresh CI is running.

packages/gptme-sessions/src/gptme_sessions/transcript.py
```diff
@@ -736,3 +736,264 @@ def read_transcript(path: Path) -> SessionTranscript:
...
+            try:
+                meta = json.loads(meta_path.read_text())
+            except (OSError, json.JSONDecodeError):
+                continue
+            child_stem = meta_path.name.removesuffix(".meta.json")
+            child_path = meta_path.parent / f"{child_stem}.jsonl"
+            tool_use_id = meta.get("toolUseId", "")
+            if tool_use_id:
+                agent_map[tool_use_id] = (meta, child_path)
```
TimeToBuildBob, line 983: ❌ **P1** — In read_session_tree, the tool_use-linked branch stores (meta, child_path) in agent_map without checking that child_path exists, while the adjacent workflow branch explicitly guards with child_path.exists() before appending. _build_subagent_nodes later resolves any parent 'Agent' tool_use whose id is in agent_map and _make_subagent_node calls parse_trajectory(child_path), which opens the file and raises FileNotFoundError when it is absent. Claude Code writes the .meta.json at spawn time; if the agent is interrupted before writing its trajectory, or the jsonl is cleaned up, every call to the new public read_session_tree raises and the whole session's tree analysis fails rather than skipping the missing child.

packages/gptme-sessions/src/gptme_sessions/signals.py
TimeToBuildBob, line 3045: ⚠️ **P2** — extract_from_path appends every file from subagent_record_files to the parent's records before extraction, but subagent trajectory files carry the parent's spawning 'Agent' tool-use as context (the PR's own _build_subagent_nodes docstring states this). Since _agent_tool_use_ids matches those echoed records (type == 'assistant'), extract_signals_cc counts the parent's 'Agent' tool_use once from the parent and once from every child file, so a parent that spawned N subagents reports tool_calls['Agent'] = N+1 and gains one extra step per echoed record. The read_session_tree path avoids this with its visited set; subagent_record_files has no equivalent dedup. The inflated total_tools also lowers the error rate used by grade_signals.

TimeToBuildBob:
@greptileai review

packages/gptme-sessions/src/gptme_sessions/transcript.py
TimeToBuildBob: Fixed in 5b1ae65c: orphaned metadata is skipped unless its JSONL exists; regression coverage confirms read_session_tree returns an empty child list instead of raising.

packages/gptme-sessions/src/gptme_sessions/signals.py
TimeToBuildBob: Fixed in 5b1ae65c: Claude tool calls/results are deduplicated by stable tool-use ID across parent-context echoes, and token usage is extracted per transcript before additive rollup. Regression tests cover Agent/step deduplication and parent-result plus child usage.

packages/gptme-sessions/src/gptme_sessions/transcript.py
TimeToBuildBob: Fixed in 5b1ae65c; see the existing thread reply for the code and regression-test details. Resolving now after confirming the fix is present on the PR head.

packages/gptme-sessions/src/gptme_sessions/signals.py
TimeToBuildBob: Fixed in 5b1ae65c; see the existing thread reply for the deduplication and per-transcript usage-rollup details. Resolving now after confirming the fix is present on the PR head.

TimeToBuildBob:
Fixed both current AI-review findings in `5b1ae65c`:

- skip orphaned `.meta.json` entries whose child trajectory is absent;
- deduplicate echoed Claude tool calls/results by tool-use ID;
- extract cumulative usage per parent/child transcript, then add the totals so one stream `result` cannot replace sibling usage.

Added regression coverage for both findings plus missing-meta fallback. Local verification: 10 targeted tests passed, mypy passed, and Ruff check/format passed. The two inline finding threads are replied to and resolved. Fresh CI is running; the in-band review attempted full consensus but one provider pass timed out, so it correctly produced no clean-review claim.

PR #1604, opened 2026-09-03
feat(voice): drive a /local realtime turn with text

TimeToBuildBob:
## What

The `/local` WebSocket endpoint on the realtime voice server exists to allow
"testing without Twilio", but its message loop only handled `audio` and
`commit`. Driving a turn therefore required synthesizing PCM — impractical on a
headless machine with no microphone.

This adds a `{"type": "text", "text": "..."}` message that delivers the payload
as the caller's own utterance, via a new
`OpenAIRealtimeClient.send_text_message()`.

## Why not `inject_message()`

`inject_message()` already creates a user item and calls `response.create`, but
it prefixes the content with `[Subagent result]:` — correct for async tool
output, wrong for a caller turn. The model needs the text verbatim to treat it
as speech and to decide to call a tool.

That matters concretely: it's what makes a **headless `look` proof** possible.
Connect one client to `/local?vision=1` that both serves camera/sensor frames
(`vision_look_request` → `vision_look_result`) and sends a text turn — the model
calls `look`, the bridge round-trips a frame, and the description lands in the
same turn. No audio hardware anywhere in the loop. This is the enabling piece
for step 2.4 of Bob World (`gptme/gptme-contrib#1576`).

`XAIRealtimeClient` subclasses `OpenAIRealtimeClient`, so both providers get it.

## Tests

`test_send_text_message_delivers_verbatim_user_turn` asserts the exact event
sequence (`conversation.item.create` → `response.create`), the `user` role and
verbatim content, and contrasts it against `inject_message()` still tagging its
payload.

The server-side branch is a 4-line dispatch to that method; there is no existing
`/local` end-to-end WebSocket harness in `test_server.py` to extend, so it is
covered by the client-level test plus the live path above rather than a new
harness.

### DIFF
--- a/packages/gptme-voice/src/gptme_voice/realtime/openai_client.py
+++ b/packages/gptme-voice/src/gptme_voice/realtime/openai_client.py
@@ -39,6 +39,9 @@ _MAX_INSTRUCTIONS_LEN = 4096
 # than enough for any realistic session-handshake delay while still capping
 # memory growth if the provider never confirms the session.
 _MAX_PENDING_AUDIO_CHUNKS = 500
+# A provider handshake should normally complete in under a second. Bound text
+# turns so a missing ready event cannot pin a /local connection forever.
+_SESSION_READY_TIMEOUT_SECONDS = 10.0
 # OpenAI's model catalog currently markets `gpt-realtime-2`, while the
 # Realtime API reference may still enumerate the older generic alias instead.
 # Keep the newer default, but retry once with the generic alias if the
@@ -655,7 +658,13 @@ class OpenAIRealtimeClient:
         """
         if self._session_ready is None:
             raise RuntimeError("Not connected to OpenAI Realtime API")
-        await self._session_ready.wait()
+        try:
+            await asyncio.wait_for(
+                self._session_ready.wait(),
+                timeout=_SESSION_READY_TIMEOUT_SECONDS,
+            )
+        except asyncio.TimeoutError as exc:
+            raise RuntimeError("Realtime session was not ready in time") from exc
 
         logger.info("Sending text message: %s...", text[:100])
         await self._send_event(
--- a/packages/gptme-voice/src/gptme_voice/realtime/openai_client.py
+++ b/packages/gptme-voice/src/gptme_voice/realtime/openai_client.py
@@ -653,6 +653,10 @@ class OpenAIRealtimeClient:
         a headless client can drive a full turn — including tool calls — without
         a microphone.
         """
+        if self._session_ready is None:
+            raise RuntimeError("Not connected to OpenAI Realtime API")
+        await self._session_ready.wait()
+
         logger.info("Sending text message: %s...", text[:100])
         await self._send_event(
             "conversation.item.create",
--- a/packages/gptme-voice/src/gptme_voice/realtime/openai_client.py
+++ b/packages/gptme-voice/src/gptme_voice/realtime/openai_client.py
@@ -644,6 +644,28 @@ class OpenAIRealtimeClient:
         # Start receiving messages
         self._receive_task = asyncio.create_task(self._receive_loop())
 
+    async def send_text_message(self, text: str) -> None:
+        """Send a caller utterance as text and trigger a response.
+
+        Text equivalent of a spoken turn: unlike :meth:`inject_message` (which
+        tags the content as an async subagent result), this delivers ``text``
+        verbatim as the user's own turn. Used by the ``/local`` test endpoint so
+        a headless client can drive a full turn — including tool calls — without
+        a microphone.
+        """
+        logger.info("Sending text message: %s...", text[:100])
+        await self._send_event(
+            "conversation.item.create",
+            {
+                "item": {
+                    "type": "message",
+                    "role": "user",
+                    "content": [{"type": "input_text", "text": text}],
+                }
+            },
+        )
+        await self._send_event("response.create", {})
+
     async def inject_message(self, text: str) -> None:
         """Inject a message into the conversation and trigger a response.
 
--- a/packages/gptme-voice/src/gptme_voice/realtime/server.py
+++ b/packages/gptme-voice/src/gptme_voice/realtime/server.py
@@ -9,6 +9,7 @@ import asyncio
 import base64
 import contextlib
 import hashlib
+import ipaddress
 import json
 import logging
 import os
@@ -72,6 +73,20 @@ def _websocket_peer_host(websocket) -> str | None:
         if isinstance(first, str) and first:
             return first
     return None
+
+
+def _is_loopback_host(host: str | None) -> bool:
+    """Whether a peer host is loopback, including IPv4-mapped IPv6."""
+    if host == "localhost":
+        return True
+    if host is None:
+        return False
+    try:
+        address = ipaddress.ip_address(host)
+    except ValueError:
+        return False
+    mapped = getattr(address, "ipv4_mapped", None)
+    return address.is_loopback or (mapped is not None and mapped.is_loopback)
 
 
 def _websocket_has_vision(websocket) -> bool:
@@ -1093,7 +1108,7 @@ class VoiceServer:
             )
             return None
         host = _websocket_peer_host(websocket)
-        if host in {"127.0.0.1", "::1", "localhost"}:
+        if _is_loopback_host(host):
             return self.body_adapter
         logger.warning(
             "Body tools disabled for unauthenticated %s client %s",
@@ -2758,7 +2773,7 @@ class VoiceServer:
         or test client.
         """
         peer_host = _websocket_peer_host(websocket)
-        if peer_host not in {"127.0.0.1", "::1", "localhost"}:
+        if not _is_loopback_host(peer_host):
             logger.warning(
                 "Rejecting unauthenticated /local client %s",
                 peer_host or "unknown",
@@ -2843,7 +2858,17 @@ class VoiceServer:
                     # (including tool calls) without sending audio.
                     text = data.get("text", "")
                     if isinstance(text, str) and text.strip():
-                        await realtime_client.send_text_message(text)
+                        try:
+                            await realtime_client.send_text_message(text)
+                        except RuntimeError as exc:
+                            logger.warning("Text turn failed: %s", exc)
+                            continue
+                        _append_transcript_turn(
+                            transcript,
+                            "user",
+                            text,
+                            allow_continuation=False,
+                        )
 
         except WebSocketDisconnect:
             pass  # Normal path when _schedule_hangup closes the WebSocket
--- a/packages/gptme-voice/src/gptme_voice/realtime/server.py
+++ b/packages/gptme-voice/src/gptme_voice/realtime/server.py
@@ -2757,6 +2757,15 @@ class VoiceServer:
         Allows testing without Twilio by connecting directly from a browser
         or test client.
         """
+        peer_host = _websocket_peer_host(websocket)
+        if peer_host not in {"127.0.0.1", "::1", "localhost"}:
+            logger.warning(
+                "Rejecting unauthenticated /local client %s",
+                peer_host or "unknown",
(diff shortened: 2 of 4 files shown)


TimeToBuildBob:
CI red here was **not** caused by this PR. The failure was `tests/test_agent_write_loss_scan.py::test_classify_write_persisted` (`assert 'LOST' == 'PERSISTED'`) — this branch only touches `packages/gptme-voice/*`, and `make test-unit` doesn't even collect that package's tests.

Root cause is a pre-existing latent bug in `scripts/agent-write-loss-scan.py`: `_post_write_blobs()` passes `--since=@{int(write_ts)}`, and git's approxidate **rejects `@<seconds>` values below roughly 1e8 and silently falls back to *now***. With `write_ts=0.0` the cutoff becomes "now", so the just-made commit only survives if `git log` runs inside the same second as the commit — a second-boundary coin flip that this run lost.

Rerun of the identical commit is green, and CI here is passing again — nothing to change on this PR.

Fix + deterministic regression test: gptme/gptme-contrib#1605.

TimeToBuildBob:
Removed the unrelated write-loss-scan commit from this branch with a lease-guarded force-push. PR #1604 is back to the single voice commit and now touches only `packages/gptme-voice/*`; the actual scanner fix remains isolated in gptme/gptme-contrib#1605.

Fresh CI is running on `c831bc1f`.

TimeToBuildBob:
## 🤖 AI code review

This change adds a text-message path to the /local realtime endpoint, including a new OpenAIRealtimeClient.send_text_message() method, a server-side loopback check, and text-frame dispatch that appends a transcript turn after successful delivery.

<sub>ℹ️ Consensus was degraded on this run: 1 of 3 passes answered, so findings were filtered at 1-of-1 agreement rather than 2-of-3 — less filtered than usual; 1 of 3 fan-out jobs answered, so the findings above were filtered against less evidence than the threshold assumes.</sub>

<b>1 finding</b> · ⚠️ **1** P2

### Findings

**⚠️ P2 medium** — [`packages/gptme-voice/src/gptme_voice/realtime/server.py:1`](https://github.com/gptme/gptme-contrib/blob/bffc099cf45b9412843001143952a2ace4927df0/packages/gptme-voice/src/gptme_voice/realtime/server.py#L1)

When a text message arrives before session.created, send_text_message() blocks on asyncio.wait_for(self._session_ready.wait(), ...). On expiry, asyncio.wait_for raises asyncio.TimeoutError (which has a builtin TimeoutError alias in modern Python), not RuntimeError. The new text branch in handle_local_websocket() only catches RuntimeError and then continues, so a timeout escapes the inner branch, terminates the message loop, and tears down the entire /local connection rather than discarding just the early text message. This is especially relevant because the feature is meant for headless testing, where probing with a text turn before the provider session is established is plausible. fix it by catching asyncio.TimeoutError inside send_text_message() and raising RuntimeError, or by catching TimeoutError in the text branch alongside RuntimeError.

```suggestion
Wrap the wait_for timeout in send_text_message() with a try/except asyncio.TimeoutError and re-raise it as RuntimeError, or change the server's text-frame handler to catch both RuntimeError and TimeoutError.
```

<sub>**How this was verified:** Inspected the added send_text_message() code and the corresponding server-side except branch. asyncio.wait_for does not re-raise timeouts as RuntimeError.</sub>

- `packages/gptme-voice/src/gptme_voice/realtime/openai_client.py` — Adds _SESSION_READY_TIMEOUT_SECONDS and send_text_message(), which waits for the session-ready event and then sends a verbatim user input_text item plus response.create.
- `packages/gptme-voice/src/gptme_voice/realtime/server.py` — Adds _is_loopback_host() for local-peer validation, uses it in the body-tool trust check and the /local endpoint, and adds handling for text messages that dispatches to send_text_message() and appends a transcript turn on success.

<sub>Also changed, not described above: `packages/gptme-voice/tests/test_openai_client.py`, `packages/gptme-voice/tests/test_server.py`</sub>

</details>

| `c831bc1f4d54` | 5/5 | 0 | llm | 2026-09-03 20:54 UTC |
| `c831bc1f4d54` | 3/5 | 1 | llm | 2026-09-03 21:19 UTC |
| `b8056bbd2d28` | 1/5 | 2 | llm | 2026-09-03 23:13 UTC |
| `d5031fc344f7` | 2/5 | 2 | llm | 2026-09-03 23:41 UTC |

</details>

<sub>Reviewed `bffc099cf45b` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 498s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
All CI checks are green and the current head is clean/mergeable. AI review is 5/5 with no findings; the required Greptile review was requested on `c831bc1f` and is still pending, so the merge gate is not yet eligible. No additional code change is warranted until that review returns.

packages/gptme-voice/src/gptme_voice/realtime/openai_client.py
```diff
@@ -644,6 +644,28 @@ async def connect(self) -> None:
...
+        """Send a caller utterance as text and trigger a response.
+
+        Text equivalent of a spoken turn: unlike :meth:`inject_message` (which
+        tags the content as an async subagent result), this delivers ``text``
+        verbatim as the user's own turn. Used by the ``/local`` test endpoint so
+        a headless client can drive a full turn — including tool calls — without
+        a microphone.
+        """
+        logger.info("Sending text message: %s...", text[:100])
+        await self._send_event(
```
TimeToBuildBob, line 670: ❌ **P1** — `send_text_message` sends `conversation.item.create` and `response.create` without first confirming the OpenAI session is ready. `connect()` only sends `session.update` and starts the receive loop; the session is confirmed by the `session.created` event which sets `_session_ready`. A `/local` client can send a `{"type":"text"}` message immediately after the WebSocket connects, before `session.created` has been processed. The events then go out over the WebSocket before the session exists, and the OpenAI Realtime API requires `session.created` before accepting conversation items, so these early events are rejected or dropped. The audio path in the same client buffers data until `_session_ready` is set (see `send_audio`), so this text path is inconsistent and the first turn on a cold connection is lost. A headless test client that sends text without waiting for a ready signal will get no response, making the feature unreliable. _(2/3 passes — pass 2 looked and disagreed)_

packages/gptme-voice/src/gptme_voice/realtime/server.py
```diff
@@ -2829,6 +2829,13 @@ async def handle_local_websocket(self, websocket):
                 elif data.get("type") == "commit":
                     await realtime_client.commit_audio()
 
+                elif data.get("type") == "text":
+                    # Headless text turn: drive a full conversation turn
+                    # (including tool calls) without sending audio.
+                    text = data.get("text", "")
+                    if isinstance(text, str) and text.strip():
+                        await realtime_client.send_text_message(text)
```
TimeToBuildBob: 🛑 **P0** — The new `text` message branch in `handle_local_websocket` allows any client that can reach the server (bound to 0.0.0.0 by default) to inject arbitrary user utterances into the realtime session. Unlike the existing audio path, which requires synthesizing speech, a remote attacker can now send plain text that is delivered verbatim to the model. The model has a `subagent` tool that dispatches tasks to a gptme subagent running in the workspace, which can execute arbitrary commands. Because `handle_local_websocket` performs no authentication or peer-host check (only `_body_adapter_for_websocket` restricts motion tools to loopback), an attacker can drive the model to call `subagent` and thereby run arbitrary code on the host. This is a serious security hole for any deployment where the server is not firewalled to localhost.

TimeToBuildBob: ⚠️ **P2** — The `send_text_message` path delivers the user's text verbatim to the OpenAI session, but the local transcript is only appended via `_on_user_transcript`, which is fed by `conversation.item.input_audio_transcription.completed` events. Text turns never produce that event, so the user's utterance is absent from the `transcript` list that `_on_call_end` persists and that `_build_resume_instructions` uses for reconnect context. A headless client driving a text-only turn will have no user turn in the archived call record or the resume transcript, so post-call analysis and any resume of that conversation miss the user's side of the exchange. This makes the new feature's stated goal of 'driving a full conversation turn' incomplete for the record-keeping layer.

packages/gptme-voice/src/gptme_voice/realtime/server.py
```diff
@@ -2757,6 +2757,15 @@ async def handle_local_websocket(self, websocket):
         Allows testing without Twilio by connecting directly from a browser
         or test client.
         """
+        peer_host = _websocket_peer_host(websocket)
+        if peer_host not in {"127.0.0.1", "::1", "localhost"}:
```
TimeToBuildBob: ❌ **P1** — The loopback check in handle_local_websocket tests peer_host against the exact strings "127.0.0.1", "::1", and "localhost". On systems where the server listens on an IPv6 socket, a loopback connection made to 127.0.0.1 is often reported by the OS as the IPv4-mapped IPv6 address "::ffff:127.0.0.1". That value does not appear in the set, so the /local endpoint closes with code 1008 and rejects a legitimate local client. The primary use case of /local for headless testing on the same machine is broken whenever the kernel presents IPv4 connections as IPv4-mapped IPv6. The same set is also used in _body_adapter_for_websocket, so body tools are likewise disabled for such peers, reinforcing that the check is too narrow. Fix by adding "::ffff:127.0.0.1" to the set or by using ipaddress.ip_address(peer_host).is_loopback, which treats both IPv4 and mapped IPv6 loopback forms correctly.

packages/gptme-voice/src/gptme_voice/realtime/server.py
```diff
@@ -2829,6 +2838,13 @@ async def handle_local_websocket(self, websocket):
                 elif data.get("type") == "commit":
                     await realtime_client.commit_audio()
 
+                elif data.get("type") == "text":
+                    # Headless text turn: drive a full conversation turn
+                    # (including tool calls) without sending audio.
+                    text = data.get("text", "")
+                    if isinstance(text, str) and text.strip():
+                        await realtime_client.send_text_message(text)
```
TimeToBuildBob: ❌ **P1** — The new "text" branch in handle_local_websocket awaits realtime_client.send_text_message(text) without catching the RuntimeError that method raises when the provider session is not ready within _SESSION_READY_TIMEOUT_SECONDS (10s). If a headless client sends a text turn before session.created/updated has been received—e.g. on a slow provider handshake—send_text_message waits the full 10 seconds and then raises. That exception propagates to the broad `except Exception` handler in handle_local_websocket, which logs it and then the finally block disconnects the realtime client and invokes _on_call_end, terminating the entire /local session. A single early message thus kills the whole connection instead of being dropped or retried. This makes the /local text-turn feature unreliable on cold starts where the provider handshake exceeds the timeout. The fix is to catch RuntimeError from send_text_message and continue the message loop, logging the failure without tearing down the session.

packages/gptme-voice/src/gptme_voice/realtime/server.py
TimeToBuildBob: Fixed in `bffc099c`: `/local` now uses `_is_loopback_host()`, backed by `ipaddress.ip_address()`, and accepts IPv4-mapped IPv6 loopback peers such as `::ffff:127.0.0.1`. The same helper now gates body-tool access. Regression coverage exercises the mapped address.

packages/gptme-voice/src/gptme_voice/realtime/server.py
TimeToBuildBob: Fixed in `bffc099c`: the text branch catches the `RuntimeError` that `send_text_message()` raises after its bounded session-ready wait, logs the failed turn, and continues consuming frames instead of tearing down `/local`. The regression test sends a failing early turn followed by a successful turn and verifies only the latter is persisted.

PR #1607, opened 2026-09-03
fix(gptme-sessions): blame no longer crashes on uncommitted lines

TimeToBuildBob:
`sessions-blame.py <file> --line N` crashed with an unhandled `CalledProcessError` whenever the target line existed only in the working tree.

```
subprocess.CalledProcessError: Command '['git', 'show', '-s', '--format=...', '0000000000000000000000000000000000000000']' returned non-zero exit status 128.
```

`git blame --porcelain` reports the all-zeros sentinel sha for uncommitted lines; `commit_for_line` passed it straight to `git show`.

## Fix

Detect the sentinel and return an uncommitted `Attribution` instead, timestamped from the blame porcelain's `author-time`/`author-tz` — so `attribute_all` window matching and the trajectory-enrichment fallback still run on it. That is the case that matters most for incident response: *which session made this working-tree edit?* was previously the one question the attribution path could not answer.

## Verification

```
$ uv run pytest packages/gptme-sessions/tests/test_blame.py -q
52 passed
```

New test `test_commit_for_line_uncommitted` asserts `git show` is not invoked for the sentinel sha and that the timestamp is parsed from the porcelain.

End-to-end against a real uncommitted line (previously a traceback):

```
Attribution(sha='000...000', when=2026-09-03 23:18:31+00:00,
            author='Not Committed Yet',
            subject='(uncommitted change in working tree)')
```

Found while attributing an uncommitted trace-test marker that a timed-out session left appended to `SOUL.md` in Bob's brain repo.

### DIFF
--- a/packages/gptme-sessions/src/gptme_sessions/blame.py
+++ b/packages/gptme-sessions/src/gptme_sessions/blame.py
@@ -160,10 +160,50 @@ def commits_for_path(path: str, limit: int = 10) -> list[Attribution]:
     return result
 
 
+#: ``git blame`` reports this sentinel sha for lines that exist only in the
+#: working tree (not yet committed).
+_UNCOMMITTED_SHA = "0" * 40
+
+
+def _uncommitted_attribution(porcelain: str) -> Attribution:
+    """Build an Attribution for a working-tree line from ``git blame`` porcelain.
+
+    ``git blame`` still emits ``author-time``/``author-tz`` for uncommitted
+    lines, so the edit can be placed on the session timeline even though no
+    commit exists to attribute it to.
+    """
+    fields: dict[str, str] = {}
+    for raw in porcelain.split("\n")[1:]:
+        if raw.startswith("\t"):
+            break
+        key, _, value = raw.partition(" ")
+        fields.setdefault(key, value)
+
+    try:
+        when = datetime.fromtimestamp(int(fields.get("author-time", "")), tz=timezone.utc)
+    except ValueError:
+        when = datetime.now(tz=timezone.utc)
+
+    return Attribution(
+        sha=_UNCOMMITTED_SHA,
+        when=when,
+        author=fields.get("author") or "Not Committed Yet",
+        subject="(uncommitted change in working tree)",
+        trailer_session_id=None,
+    )
+
+
 def commit_for_line(path: str, line: int) -> list[Attribution]:
-    """Return the single commit that last touched ``path`` line ``line``."""
+    """Return the single commit that last touched ``path`` line ``line``.
+
+    A line that is only in the working tree has no commit; it is returned as an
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
CI is green and the PR is mergeable, but the self-merge gate is not yet eligible because no Greptile review has landed. I did not bypass the gate; waiting for the initial review or a human merge.

TimeToBuildBob:
## 🤖 AI code review

| `4185d52ad595` | 4/5 | 1 | llm | 2026-09-04 00:05 UTC |

</details>

<sub>Reviewed `61c9b7c7f08c` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm (in-band PM)` engine · 890s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

packages/gptme-sessions/src/gptme_sessions/blame.py
```diff
@@ -160,10 +160,50 @@ def commits_for_path(path: str, limit: int = 10) -> list[Attribution]:
...
+    """Return the single commit that last touched ``path`` line ``line``.
+
+    A line that is only in the working tree has no commit; it is returned as an
+    uncommitted Attribution timestamped from the blame porcelain so window
+    attribution still works.
+    """
     out = _run(["git", "blame", "-L", f"{line},{line}", "--porcelain", "--", path])
-    sha = out.splitlines()[0].split(" ", 1)[0]
+    sha = out.split("\n", 1)[0].split(" ", 1)[0]
+    if sha == _UNCOMMITTED_SHA:
```
TimeToBuildBob: ⚠️ **P2** — The sentinel check `sha == _UNCOMMITTED_SHA` compares against a hardcoded 40-zero string. In a git repository using the SHA-256 object format, `git blame --porcelain` emits a 64-hex zero hash for uncommitted lines. The comparison fails, so commit_for_line proceeds to `git show` with the 64-zero hash, which fails with 'fatal: bad object' and raises CalledProcessError — the same crash this PR intends to fix. The guard should detect any all-zero hash, e.g. `if not any(c != '0' for c in sha):` or compare with `"0" * len(sha)`. This is a real defect for SHA-256 repos, though SHA-1 remains the default.

TimeToBuildBob:
@greptileai review

packages/gptme-sessions/src/gptme_sessions/blame.py
TimeToBuildBob: Fixed in `61c9b7c7`: null object IDs are now detected by content rather than a hardcoded SHA-1 length, and the returned attribution preserves the 40- or 64-character ID emitted by Git. The regression test now covers both SHA-1 and SHA-256 repositories. The in-band reviewer passed the updated diff at 5/5 with 3/3 consensus.

PR #1608, opened 2026-09-03
fix(gptodo): write waiting_spell_count and first_waiting_since live

TimeToBuildBob:
## Problem

TASKS.md documents two cumulative waiting-history fields:

- `first_waiting_since` — stamped on the **first** waiting entry, never overwritten
- `waiting_spell_count` — "Count of distinct waiting spells. Incremented on every waiting entry."

Nothing implemented either. `gptodo edit <task> --set state waiting` auto-set
`waiting_since` and stopped there:

```
state: active -> waiting
waiting_since: None -> 2026-09-03T22:45:14+00:00
```

The only writer was `scripts/backfill-waiting-spell-count.py` in Bob's workspace — a
one-shot git-history derivation with **no timer and no caller**. So every value in the
tree is frozen at whenever someone last ran it by hand.

## Why it matters

`waiting_spell_count` is the re-park signal. `task_metadata_hygiene_audit.py` check 16
(`erik_gate_repark`) treats `>= 3` as "read the blocker, don't extend it", and the
erik-gate audit taxonomy uses the same threshold. A counter that only moves on a manual
backfill reads **systematically low**, so the check under-reports exactly the tasks that
are being repeatedly re-parked. Sessions that hand-bump it make the data less uniform,
not more.

`first_waiting_since` is the cumulative blocker age that survives re-parks —
`scripts/current-constraint.py` prefers it over `waiting_since` for exactly that reason,
and gets a stale or missing value today.

## Change

Both fields are now written in the same branch that auto-sets `waiting_since`, so the
three can never disagree.

The new-spell discriminator is the **prior state**, not a missing `waiting_since`.
Leaving `waiting` does not clear `waiting_since` (verified by
`test_two_waiting_entries_produce_spell_count_two`), so keying off its absence would
miss precisely the re-parks these fields exist to count.

Also registers both fields in `KNOWN_FRONTMATTER_FIELDS`. They were absent, so
`gptodo lint` warned on frontmatter gptodo itself now writes.

## Tests

Five new tests in `packages/gptodo/tests/test_edit_state_waiting.py`:

- first waiting entry stamps `first_waiting_since` and sets the count to 1
- a re-park after un-park gives count 2 while `first_waiting_since` does **not** move
- `--set state waiting` on an already-waiting task is not a new spell
- a hand-corrupted non-numeric counter recovers to 1 instead of crashing the edit
- both fields are registered in `KNOWN_FRONTMATTER_FIELDS`

`610 passed` across the full `packages/gptodo/tests/` suite; ruff clean.

### DIFF
--- a/packages/gptodo/src/gptodo/cli.py
+++ b/packages/gptodo/src/gptodo/cli.py
@@ -3024,17 +3024,42 @@ def edit(task_ids, set_fields, add_fields, remove_fields, set_subtask, force):
             op == "set" and field == "waiting_for" and value is not None
             for op, field, value in changes
         )
-        if (
-            transitioning_to_waiting
-            and waiting_for_present
-            and not post.metadata.get("waiting_since")
-        ):
+        if transitioning_to_waiting and waiting_for_present:
             from datetime import datetime as _dt, timezone as _tz
 
             # Full ISO datetime for intra-day resolution (ErikBjare request, 2026-06-16).
             # validate_task_frontmatter.py's validate_timestamp() accepts both YYYY-MM-DD
             # and full ISO datetime via datetime.fromisoformat().
-            post.metadata["waiting_since"] = _dt.now(_tz.utc).isoformat(timespec="seconds")
+            _now_iso = _dt.now(_tz.utc).isoformat(timespec="seconds")
+            if not post.metadata.get("waiting_since"):
+                post.metadata["waiting_since"] = _now_iso
+
+            # Cumulative waiting history (TASKS.md schema), written here so it can
+            # never disagree with waiting_since. Both fields were documented but
+            # had no live writer until 2026-09-03 — the only writer was a one-shot
+            # git-history backfill with no caller, so every value in the tree was
+            # frozen at whenever someone last ran it by hand.
+            #
+            # The discriminator is the *prior* state, not a missing waiting_since:
+            # leaving waiting does not clear waiting_since, so its absence would
+            # miss exactly the re-parks these fields exist to count.
+            if _prior_state != "waiting":
+                # first_waiting_since is stamped once and never overwritten — it
+                # is the cumulative blocker age that survives re-parks.
+                _since = post.metadata.get("waiting_since")
+                post.metadata.setdefault(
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:
CI is green and the PR is mergeable. Self-merge remains blocked because no Greptile review has landed yet; the guarded initial-review fallback is still inside its 45-minute grace window, so I did not bypass it or post a raw trigger. Waiting for review, then the policy gate can be rerun.

TimeToBuildBob:
## 🤖 AI code review

The change populates first_waiting_since and waiting_spell_count on transitions into waiting, and adds the new fields to the known frontmatter list. However, re-entering waiting after a non-waiting state leaves an old waiting_since unchanged, making the recorded start of the current waiting spell stale.

### Confidence Score: N/A

**⚠️ P2 medium** — [`packages/gptodo/gptodo/cli.py:3033`](https://github.com/gptme/gptme-contrib/blob/66fcafea34ea496025404a208d2c0db11aad981a/packages/gptodo/gptodo/cli.py#L3033)

When a task enters waiting from a non-waiting state and already has a stale waiting_since left over from a previous spell, the code only assigns _now_iso if waiting_since is falsy. On a re-park, the old timestamp is kept, so waiting_since does not represent the start of the new spell while waiting_spell_count is incremented. This makes the metadata internally inconsistent and can mislead downstream consumers of waiting_since. The current-spell timestamp should be overwritten on a new waiting entry (unless the edit explicitly supplies a different waiting_since), while first_waiting_since remains preserved by setdefault.

```suggestion
When _prior_state != "waiting" and this edit is entering a new waiting spell, set post.metadata["waiting_since"] = _now_iso even if a stale value already exists, unless the edit itself explicitly set waiting_since. Preserve first_waiting_since separately with setdefault.
```

<sub>**How this was verified:** test_two_waiting_entries passes but does not assert waiting_since after the second park, so it does not catch this stale value.</sub>

</details>

<sub>Reviewed `66fcafea34ea` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 641s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

PR #1609, opened 2026-09-03
feat(gptme-sessions): add SessionStore.rotate() to cap store rewrite cost

TimeToBuildBob:
## Problem

Every per-session field update — judge grade write-back, `stamp_attempt_kind`, alignment, bandit stamp — is a full `rewrite()` of `session-records.jsonl`. `rewrite()` re-reads the whole file and writes a fsynced temp copy. With an unbounded ledger that is a multi-megabyte read **and** write per update, several times per session, for a file that only grows ~2 KB per session.

Measured on Bob's store 2026-09-03 (NVMe wear investigation): **73 MB / 17k records**, ~90–150 GB/day of logical writes against a DRAM-less SSD.

## What this adds

- **`SessionStore.rotate(keep_days=30)`** — moves older records into monthly `session-records-archive-YYYY-MM.jsonl` files (the naming `blame.py` already documents).
  - Archives are appended and **fsynced before** the active file is replaced, so a crash at any point duplicates records rather than losing them.
  - Idempotent: `session_id`s already present in the target archive are skipped (`skipped_duplicate` in the returned counts).
  - Records with a missing/unparseable `timestamp`, and malformed JSONL lines, are **kept active** rather than aged out on a guess.
  - Retention rule: rotation **moves, never deletes**.
- **`SessionStore.archive_paths()`** and **`load_all(include_archives=False)`**.
  The default stays active-only on purpose: every mutation path is `load_all()` → mutate → `rewrite()`, and `rewrite()` has upsert semantics — pulling archives into that list would re-inject the entire history back into the active file on the next field update, which is exactly the amplification this removes. `test_rewrite_after_rotation_does_not_resurrect_archived_records` pins that.
- **`gptme-sessions rotate [--keep-days N] [--dry-run] [--json]`** so it is schedulable.

## Not scheduled here — on purpose

History consumers must become archive-aware first. In Bob's workspace ~15 scripts hardcode the literal `session-records-archive-pre-2026-04.jsonl` filename instead of globbing `session-records-archive-*.jsonl`; running rotation today would silently drop history from those readers. Migrating them is the follow-up.

## Verification

```
$ gptme-sessions rotate --keep-days 30 --dry-run   # against the real 73 MB store
would archive 5226 record(s); kept 11813 active
  2026-06: 59
  2026-07: 4161
  2026-08: 1006
```

- 9 new tests in `packages/gptme-sessions/tests/test_store_rotation.py` (monthly bucketing, round-trip preservation, idempotency, re-append dedup, undated/malformed records kept, no-op cases, negative `keep_days`, post-rotation rewrite).
- Full package suite: **1381 passed**.
- `ruff check` clean, `mypy src/` clean, `prek` hooks pass.

### DIFF
--- a/packages/gptme-sessions/src/gptme_sessions/store.py
+++ b/packages/gptme-sessions/src/gptme_sessions/store.py
@@ -9,7 +9,7 @@ import sys
 import uuid
 from collections.abc import Iterator
 from contextlib import contextmanager
-from datetime import datetime, timezone
+from datetime import datetime, timedelta, timezone
 from pathlib import Path
 from typing import TextIO
 
@@ -171,12 +171,28 @@ class SessionStore:
                 os.fsync(f.fileno())
         return self.path
 
-    def load_all(self) -> list[SessionRecord]:
-        """Load all session records from the JSONL store."""
-        if not self.path.exists():
+    def archive_paths(self) -> list[Path]:
+        """Rotated archive files for this store, oldest name first.
+
+        Named ``<stem>-archive-*.jsonl`` next to the active file (e.g.
+        ``session-records-archive-2026-07.jsonl``).  Archives are append-only:
+        ``rotate()`` writes to them and nothing ever rewrites them, which is
+        the whole point — the active file stays small enough that a
+        read-modify-write cycle is cheap.
+        """
+        stem = (
+            self.path.name[: -len(".jsonl")]
+            if self.path.name.endswith(".jsonl")
+            else self.path.stem
+        )
+        return sorted(self.sessions_dir.glob(f"{stem}-archive-*.jsonl"))
+
+    def _load_path(self, path: Path) -> list[SessionRecord]:
+        """Parse one JSONL file, skipping malformed lines."""
+        if not path.exists():
             return []
         records = []
-        with open(self.path, encoding="utf-8") as f:
+        with open(path, encoding="utf-8") as f:
             for line in f:
                 line = line.strip()
                 if line:
@@ -184,6 +200,28 @@ class SessionStore:
                         records.append(SessionRecord.from_dict(json.loads(line)))
                     except (json.JSONDecodeError, TypeError, AttributeError):
                         continue
+        return records
+
+    def load_all(self, include_archives: bool = False) -> list[SessionRecord]:
+        """Load session records from the JSONL store.
+
+        Defaults to the **active file only**.  That default is load-bearing:
+        every mutation path is ``load_all()`` → mutate → ``rewrite()``, and
+        ``rewrite()`` has upsert semantics, so pulling archived records into
+        that list would re-inject the entire history back into the active file
+        on the next field update — exactly the write amplification rotation
+        exists to remove.
+
+        Pass ``include_archives=True`` for read-only history consumers
+        (analytics, blame, LOO). Archived records come first, so the result
+        stays roughly chronological.
+        """
+        if not include_archives:
+            return self._load_path(self.path)
+        records: list[SessionRecord] = []
+        for archive in self.archive_paths():
+            records.extend(self._load_path(archive))
+        records.extend(self._load_path(self.path))
         return records
 
     def rewrite(self, records: list[SessionRecord]) -> Path:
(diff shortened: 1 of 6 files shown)


TimeToBuildBob:
@greptileai review

TimeToBuildBob:
## 🤖 AI code review

Adds SessionStore.rotate() and a 'gptme-sessions rotate' CLI command to move records older than --keep-days into monthly archive files (session-records-archive-YYYY-MM.jsonl). Changes load_all() to default to active-file-only with an include_archives flag, and adds archive_paths(), _archive_month(), and _archive_session_ids() helpers. Adds a new test file covering rotation, idempotency, and re-archive edge cases.

**❌ P1 high** — `packages/gptme-sessions/src/gptme_sessions/store.py:359`

rotate() deduplicates by session_id against the set of session_ids already in the target archive. When a record with a given session_id is already present (from a prior rotation), any newly seen active record with that same session_id is skipped and never written to either the archive or the active file. The existing set is updated as entries are written, so two records with the same session_id in the same month also collide: the second is dropped. This is triggered by rewrite()'s upsert semantics resurrecting an older record (or a re-append) after rotation. Consequence: the newer or duplicated record is lost entirely — rotate, which is advertised as 'move, never delete', silently deletes data. The test test_rotate_re_archives_a_reappended_session_without_duplicating pins this behavior, but it is a data-loss bug: the reappended record is dropped rather than archived or retained.

```suggestion
Dedup on a content hash (e.g. hash of the raw JSON) or on (session_id, timestamp) instead of session_id alone, so a distinct record with an already-archived session_id is still archived.
```

<sub>**How this was verified:** Traced rotate(): existing = self._archive_session_ids(archive_path); after writing a raw for a session_id it is added to existing; a second record with the same session_id (same batch or later re-append) hits 'session_id in existing' and is skipped. The test at test_store_rotation.py:69-80 expects archived=0, skipped_duplicate=1 and the archive unchanged — the reappended record has no presence anywhere.</sub>

**❌ P1 high** — `packages/gptme-sessions/src/gptme_sessions/store.py:205`

load_all() now defaults to active-file only (include_archives=False), but the read-only CLI commands — query, stats, runs, export, show, and judge/classify with --update-store — all call store.load_all() or store.query() without passing include_archives=True. store.query() internally calls self.load_all() (store.py line 483). After a rotation moves records into archives, these commands silently stop seeing every record older than keep_days: 'gptme-sessions stats --since all' reports the archive history as absent, query/export omit archived sessions, and 'show &lt;old-id&gt;' fails with 'no session record found'. This is a data-visibility regression for existing callers, and the PR description itself acknowledges history consumers must be updated, but this PR does not update them.

```suggestion
Change read-only commands (query, stats, runs, export, show) to call store.load_all(include_archives=True) or store.query(..., include_archives=True) where supported, or make load_all() default include_archives=True for non-mutation callers.
```

<sub>**How this was verified:** Checked cli.py: query() (line 977), stats() (line 1213), runs() (line 1267), export() (line 1113), show()/judge()/classify() all call store.load_all() or store.query() with no include_archives. store.query() at store.py:483 calls self.load_all(). The new rotate command's real run moves records to *-archive-*.jsonl files that these readers do not load.</sub>

</details>

- `packages/gptme-sessions/src/gptme_sessions/cli.py` — Adds a 'rotate' CLI command with --keep-days, --dry-run, and --json options that delegates to SessionStore.rotate().
- `packages/gptme-sessions/src/gptme_sessions/store.py` — Adds rotate(), archive_paths(), _archive_month(), and _archive_session_ids(); changes load_all() to include_archives=False by default.
- `packages/gptme-sessions/tests/test_store_rotation.py` — Adds 9 tests for rotation bucketing, preservation, idempotency, re-append dedup, and post-rotation rewrite behavior.

</details>

<sub>Reviewed `43163efda844` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 48s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

packages/gptme-sessions/src/gptme_sessions/store.py
```diff
@@ -254,6 +292,145 @@ def rewrite(self, records: list[SessionRecord]) -> Path:
...
+            skipped = 0
+            for month, entries in sorted(by_month.items()):
+                archive_path = (
+                    self.sessions_dir / f"{self.path.name[: -len('.jsonl')]}-archive-{month}.jsonl"
+                )
+                existing = self._archive_session_ids(archive_path)
+                # Append first, fsync, and only then drop from the active file.
+                with open(archive_path, "a", encoding="utf-8") as af:
+                    for session_id, raw in entries:
+                        if session_id and session_id in existing:
```
TimeToBuildBob: ❌ **P1** — rotate() deduplicates by session_id against the set of session_ids already in the target archive. When a record with a given session_id is already present (from a prior rotation), any newly seen active record with that same session_id is skipped and never written to either the archive or the active file. The existing set is updated as entries are written, so two records with the same session_id in the same month also collide: the second is dropped. This is triggered by rewrite()'s upsert semantics resurrecting an older record (or a re-append) after rotation. Consequence: the newer or duplicated record is lost entirely — rotate, which is advertised as 'move, never delete', silently deletes data. The test test_rotate_re_archives_a_reappended_session_without_duplicating pins this behavior, but it is a data-loss bug: the reappended record is dropped rather than archived or retained.

packages/gptme-sessions/src/gptme_sessions/store.py
```diff
@@ -186,6 +202,28 @@ def load_all(self) -> list[SessionRecord]:
                         continue
         return records
 
+    def load_all(self, include_archives: bool = False) -> list[SessionRecord]:
```
TimeToBuildBob, line 206: ❌ **P1** — load_all() now defaults to active-file only (include_archives=False), but the read-only CLI commands — query, stats, runs, export, show, and judge/classify with --update-store — all call store.load_all() or store.query() without passing include_archives=True. store.query() internally calls self.load_all() (store.py line 483). After a rotation moves records into archives, these commands silently stop seeing every record older than keep_days: 'gptme-sessions stats --since all' reports the archive history as absent, query/export omit archived sessions, and 'show <old-id>' fails with 'no session record found'. This is a data-visibility regression for existing callers, and the PR description itself acknowledges history consumers must be updated, but this PR does not update them.

packages/gptme-sessions/src/gptme_sessions/store.py
TimeToBuildBob: Fixed in 2d1f8c54 — rotation now deduplicates only exact raw JSONL lines already durable in the target archive, preserving changed records and duplicate rows that share a session_id. Regression coverage exercises both cases.

packages/gptme-sessions/src/gptme_sessions/store.py
TimeToBuildBob: Fixed in 2d1f8c54 — read-only query/stats/runs/export/show paths now include archives by default, with history lookups in replay and harm detection updated too. Mutation-oriented regrade explicitly remains active-only to prevent archive resurrection; regression tests cover query, export, and show after rotation.

ErikBjare:
@TimeToBuildBob review

TimeToBuildBob:
## Review (post-merge)

**Finding 1 (P1 — dedup by session_id → data loss)**: FALSE POSITIVE

The auto-reviewer referenced `_archive_session_ids` but the merged code uses `_archive_line_hashes` — content-hash deduplication via `hashlib.sha256(raw.encode()).digest()`. Two records with the same `session_id` but different content get distinct hashes and both survive. The comment in `rotate()` even explains this explicitly: *"Do not add this batch's hashes to `existing`: duplicate active rows are distinct historical records and must survive."* The AI reviewer was reading an earlier draft.

**Finding 2 (P1 — callers not updated)**: VALID but intentionally deferred

Real, but the PR documents it: *"Not scheduled here — on purpose. History consumers must become archive-aware first."* The critical read-only paths were updated (`show`, `stats`, `harm_detect`, `replay`), and `store.query()` now defaults to `include_archives=True` — covering every command that calls it. The remaining callers are safe until `rotate` is actually scheduled, which it isn't. Follow-up: migrate the ~15 external scripts + remaining internal callers before enabling the timer.

**Overall**: LGTM. Write amplification fix is real and important (90–150 GB/day logical writes on a DRAM-less SSD for a ~2 KB/session ledger). Archive-move-never-delete with crash-safe fsync ordering is correct. Content-hash idempotency is the right dedup primitive. Test coverage is solid.

Minor UX note: non-dry-run mode omits the per-month breakdown that dry-run shows. Low priority — `--json` covers scripting use cases.

PR #1606, opened 2026-09-03
feat(gptme-sessions): record out-of-process dispatch lineage

TimeToBuildBob:
## Problem

Out-of-process children — `spawn-workers.sh` workers, `autonomous-fanout.sh`
sessions, project-monitoring dispatches, `gptodo spawn` children — carry no
directed link back to what spawned them. The only join available to analysis
today is timestamp proximity plus duration (`pm-executor-ab-report.py` does
exactly this), which is the same mtime-class heuristic that trajectory
attribution has been removing everywhere else.

A related symptom: "was this session a fanout child?" is currently answered by
string-matching `cascade_intent.reasons` for the substring `fanout`, which
mislabels category-preset pilot runs. That question deserves a structural field.

## Change

Two additive `SessionRecord` fields:

- `parent_session_id` — `session_id` of the spawning session
- `dispatch_kind` — one of `DISPATCH_KINDS`: `agent-tool`, `worker`,
  `gptodo-spawn`, `pm-dispatch`, `fanout`

The kinds stay distinct deliberately: an in-session Agent-tool subagent shares
the parent's context window and token budget; a dispatched worker does not, so
they must not be pooled in any per-session cost or productivity metric.

`post_session()` accepts both explicitly and otherwise falls back to
`BOB_PARENT_SESSION_ID` / `BOB_DISPATCH_KIND` in the environment, so a spawner
exports them once and every recorder downstream inherits them without a CLI
change. `post-session` also grows `--parent-session-id` / `--dispatch-kind`.

Two guards in `__post_init__`, matching the existing `attempt_kind` /
`dropout_depth` treatment:

- an unrecognized `dispatch_kind` is dropped rather than reaching consumers as
  if it were a real kind
- a `parent_session_id` equal to the record's own `session_id` is dropped — a
  spawner leaking its own id into the child environment would otherwise create
  a self-loop that breaks every lineage walk

## Tests

`tests/test_dispatch_lineage.py`, 11 cases: valid lineage, unknown-kind drop,
self-parent drop, blank-parent drop, defaults, every documented kind accepted,
persistence through a store round-trip, environment fallback, explicit-argument
precedence, no-lineage default, and the stale-env self-parent case.

```
tests/test_dispatch_lineage.py                    11 passed
tests/test_record.py test_post_session.py test_cli.py   302 passed
```

## Scope

Schema and recorder only. Wiring the individual spawners is Bob-side and lands
separately; `gptodo spawn` children still write no session record at all, which
is tracked in the originating task.

[...]

PR #1610, opened 2026-09-03
feat(gptodo): pass dispatch lineage env vars to spawned children

TimeToBuildBob:
## Summary

Wire `BOB_DISPATCH_KIND=gptodo-spawn` and `BOB_PARENT_SESSION_ID` into every subprocess spawned by `spawn_agent()`, covering both the foreground (env dict) and background (tmux env_exports) paths.

## Why

Part of the dispatch lineage schema shipped in #1606. That PR added `parent_session_id` / `dispatch_kind` fields to `SessionRecord` and made the Stop hook populate them from env vars. Without setting those vars on the child process, the gptodo-spawn kind is never recorded and the child's row in `session-records.jsonl` has no parent link.

This is item **6a** of the trajectory-attribution task: `gptodo spawn / delegate children write no session record at all — they need one before a parent link means anything.` The child does write a record (via the CC Stop hook), but the lineage fields were always null.

## What changes

- `spawn_agent()` now extracts `_parent_session_id` from `CC_SESSION_ID` (Claude Code) or `BOB_SESSION_ID` (generic Bob fallback) before the foreground/background split.
- **Background (tmux)**: two `export` lines added to `env_exports` — unconditionally for the kind, conditionally for the parent id.
- **Foreground**: env dict gains both vars after the coord injection block; when env is still `None`, a copy of os.environ is made first.
- `BOB_DISPATCH_KIND` explicitly overrides any inherited value — a worker calling `gptodo spawn` must stamp its child as `gptodo-spawn`, not `worker`.

## Tests

5 new parametrised tests in `test_subagent.py`:
- dispatch_kind set for all three backends (gptme, claude, codex)
- parent_session_id from CC_SESSION_ID
- parent_session_id fallback to BOB_SESSION_ID
- no parent_session_id key when neither var is in env
- dispatch_kind overrides inherited value

615 passed total in the gptodo test suite.

## Related

- gptme/gptme-contrib#1606 (dispatch lineage schema — this PR's env vars are read by that PR's Stop hook)

### DIFF
--- a/packages/gptodo/src/gptodo/subagent.py
+++ b/packages/gptodo/src/gptodo/subagent.py
@@ -314,11 +314,15 @@ def spawn_agent(
             env_exports.append(f"export COORDINATION_DB={shlex.quote(coord_db_path)}")
 
         # Inject dispatch lineage vars — always, regardless of clear_keys.
-        # BOB_DISPATCH_KIND tells the child which path spawned it so its
-        # session-records row can be joined back to the parent.
-        # BOB_PARENT_SESSION_ID is the caller's session id (may be absent for
-        # standalone gptodo CLI invocations outside an autonomous run).
-        env_exports.append("export BOB_DISPATCH_KIND=gptodo-spawn")
+        # BOB_SESSION_ID identifies this child so a nested spawn can use it as
+        # its immediate parent. BOB_PARENT_SESSION_ID identifies this child's
+        # caller (and may be absent for standalone gptodo CLI invocations).
+        env_exports.extend(
+            [
+                f"export BOB_SESSION_ID={shlex.quote(session_id)}",
+                "export BOB_DISPATCH_KIND=gptodo-spawn",
+            ]
+        )
         if _parent_session_id:
             env_exports.append(f"export BOB_PARENT_SESSION_ID={shlex.quote(_parent_session_id)}")
         else:
@@ -401,10 +405,11 @@ def spawn_agent(
 
     # Inject dispatch lineage vars — always, regardless of clear_keys.
     # These are Bob-internal signals; they must override whatever the *parent*
-    # process inherited (e.g. its own BOB_DISPATCH_KIND=worker should not
-    # propagate to a gptodo-spawn child as-is).
+    # process inherited. BOB_SESSION_ID identifies this child so a nested spawn
+    # can use it as its immediate parent.
     if env is None:
         env = os.environ.copy()
+    env["BOB_SESSION_ID"] = session_id
     env["BOB_DISPATCH_KIND"] = "gptodo-spawn"
     if _parent_session_id:
         env["BOB_PARENT_SESSION_ID"] = _parent_session_id
--- a/packages/gptodo/src/gptodo/subagent.py
+++ b/packages/gptodo/src/gptodo/subagent.py
@@ -220,6 +220,13 @@ def spawn_agent(
         )
         logger.info(f"Coordination enabled: agent={agent_id}, db={coord_db_path}")
 
+    # Dispatch lineage: read the caller's session ID from the environment so
+    # the child can record a parent_session_id in its session-records row.
+    # CC_SESSION_ID is set by Claude Code; BOB_SESSION_ID is the Bob-generic
+    # fallback (e.g. gptme or Codex runs).  Neither is cleared by clear_keys
+    # because these are Bob-internal dispatch signals, not API credentials.
+    _parent_session_id = os.environ.get("CC_SESSION_ID") or os.environ.get("BOB_SESSION_ID")
+
     session_id = f"agent_{uuid.uuid4().hex[:8]}"
     sessions_dir = get_sessions_dir(workspace)
     output_file = sessions_dir / f"{session_id}.output"
@@ -306,6 +313,15 @@ def spawn_agent(
         if coord_db_path:
             env_exports.append(f"export COORDINATION_DB={shlex.quote(coord_db_path)}")
 
+        # Inject dispatch lineage vars — always, regardless of clear_keys.
+        # BOB_DISPATCH_KIND tells the child which path spawned it so its
+        # session-records row can be joined back to the parent.
+        # BOB_PARENT_SESSION_ID is the caller's session id (may be absent for
+        # standalone gptodo CLI invocations outside an autonomous run).
+        env_exports.append("export BOB_DISPATCH_KIND=gptodo-spawn")
+        if _parent_session_id:
+            env_exports.append(f"export BOB_PARENT_SESSION_ID={shlex.quote(_parent_session_id)}")
+
         # Build env setup: unsets first (if any), then exports
         env_commands = []
         if env_unsets:
@@ -379,6 +395,16 @@ def spawn_agent(
         if env is None:
             env = os.environ.copy()
         env["COORDINATION_DB"] = coord_db_path
+
+    # Inject dispatch lineage vars — always, regardless of clear_keys.
+    # These are Bob-internal signals; they must override whatever the *parent*
+    # process inherited (e.g. its own BOB_DISPATCH_KIND=worker should not
+    # propagate to a gptodo-spawn child as-is).
+    if env is None:
+        env = os.environ.copy()
+    env["BOB_DISPATCH_KIND"] = "gptodo-spawn"
+    if _parent_session_id:
+        env["BOB_PARENT_SESSION_ID"] = _parent_session_id
 
     combined_output = ""
     try:
--- a/packages/gptodo/src/gptodo/subagent.py
+++ b/packages/gptodo/src/gptodo/subagent.py
@@ -321,6 +321,9 @@ def spawn_agent(
         env_exports.append("export BOB_DISPATCH_KIND=gptodo-spawn")
         if _parent_session_id:
             env_exports.append(f"export BOB_PARENT_SESSION_ID={shlex.quote(_parent_session_id)}")
+        else:
+            # Do not leak an inherited grandparent ID into a standalone spawn.
+            env_unsets.append("BOB_PARENT_SESSION_ID")
 
         # Build env setup: unsets first (if any), then exports
         env_commands = []
@@ -405,6 +408,9 @@ def spawn_agent(
     env["BOB_DISPATCH_KIND"] = "gptodo-spawn"
     if _parent_session_id:
         env["BOB_PARENT_SESSION_ID"] = _parent_session_id
+    else:
+        # Do not leak an inherited grandparent ID into a standalone spawn.
+        env.pop("BOB_PARENT_SESSION_ID", None)
 
     combined_output = ""
     try:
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
## 🤖 AI code review

The PR adds BOB_DISPATCH_KIND and BOB_PARENT_SESSION_ID environment variables to subprocesses spawned by spawn_agent() in both foreground and background (tmux) paths, reading the parent session ID from CC_SESSION_ID or BOB_SESSION_ID. It also adds five parametrised tests covering the new env var injection and inherited-value clearing.

<sub>ℹ️ Consensus was degraded on this run: 2 of 3 passes answered, so findings were filtered at 2-of-2 agreement rather than 2-of-3 — less filtered than usual; 2 of 3 fan-out jobs answered, so the findings above were filtered against less evidence than the threshold assumes.</sub>

**❌ P1 high** — `packages/gptodo/src/gptodo/subagent.py:228`

The parent session id for every spawned child is taken from CC_SESSION_ID or BOB_SESSION_ID, which identify the top-level Bob/Claude process, not a gptodo-spawned agent. When an agent spawned by gptodo (which carries BOB_DISPATCH_KIND=gptodo-spawn and BOB_PARENT_SESSION_ID of the outer run) later calls spawn_agent, the inner call reads _parent_session_id from that agent's environment. Those two vars are never propagated by this code, so _parent_session_id is None and the foreground path pops BOB_PARENT_SESSION_ID (line 413), leaving the inner child with no parent link. Even if CC_SESSION_ID were inherited, it would point to the outer session, not the agent's own id (agent_xxx), which is generated inside spawn_agent and never placed in the child's env. Nesting thus produces a missing or wrong parent_session_id in session-records.jsonl, breaking the lineage chain this PR is meant to record.

```suggestion
After generating session_id (line 230), set the child environment's BOB_SESSION_ID to session_id in both foreground (env['BOB_SESSION_ID'] = session_id) and background (env_exports.append(f"export BOB_SESSION_ID={shlex.quote(session_id)}")) so nested spawns use it as the parent.
```

<sub>**How this was verified:** Checked spawn_agent: the generated session_id (line 230) is stored only in AgentSession and never propagated to the child environment; the child is given BOB_PARENT_SESSION_ID from the outer run or none, so a nested spawn cannot obtain the immediate parent's id.</sub>

</details>

- `packages/gptodo/src/gptodo/subagent.py` — Adds BOB_DISPATCH_KIND and BOB_PARENT_SESSION_ID injection in both foreground and background spawn paths, reading parent session ID from CC_SESSION_ID or BOB_SESSION_ID.
- `packages/gptodo/tests/test_subagent.py` — Adds five tests verifying dispatch lineage env var injection and clearing for foreground and background spawns.

</details>

| `aa696d500b9e` | 3/5 | 2 | llm | 2026-09-04 00:23 UTC |
| `240484ae1872` | 5/5 | 0 | llm | 2026-09-04 01:36 UTC |
| `240484ae1872` | 5/5 | 0 | llm | 2026-09-04 01:38 UTC |
| `240484ae1872` | 4/5 | 1 | llm | 2026-09-04 02:01 UTC |

</details>

<sub>Reviewed `240484ae1872` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 597s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

packages/gptodo/src/gptodo/subagent.py
```diff
@@ -306,6 +313,15 @@ def spawn_agent(
         if coord_db_path:
             env_exports.append(f"export COORDINATION_DB={shlex.quote(coord_db_path)}")
 
+        # Inject dispatch lineage vars — always, regardless of clear_keys.
+        # BOB_DISPATCH_KIND tells the child which path spawned it so its
+        # session-records row can be joined back to the parent.
+        # BOB_PARENT_SESSION_ID is the caller's session id (may be absent for
+        # standalone gptodo CLI invocations outside an autonomous run).
+        env_exports.append("export BOB_DISPATCH_KIND=gptodo-spawn")
+        if _parent_session_id:
```
TimeToBuildBob, line 326: ❌ **P1** — spawn_agent derives _parent_session_id from CC_SESSION_ID or BOB_SESSION_ID but never clears BOB_PARENT_SESSION_ID from the child environment when that id is absent. In both the background (env_exports) and foreground (env dict) branches, the code only appends/assigns the variable if _parent_session_id is truthy (lines 322-323 and 406-407). Because the child environment is a copy of os.environ, any pre-existing BOB_PARENT_SESSION_ID (e.g. from a gptodo-spawn child that itself calls spawn_agent, where CC_SESSION_ID/BOB_SESSION_ID are unset) is inherited unchanged, linking the new child to the grandparent instead of its actual parent. The fix is to explicitly unset BOB_PARENT_SESSION_ID when _parent_session_id is falsy, in both branches.

packages/gptodo/tests/test_subagent.py
```diff
@@ -555,3 +555,125 @@ def fake_execute(task_id, **kwargs):
...
+            workspace=sessions_dir,
+        )
+
+    assert captured_env.get("BOB_PARENT_SESSION_ID") == "bob-parent-xyz789"
+
+
+def test_spawn_agent_foreground_no_parent_session_id_when_env_absent(sessions_dir, monkeypatch):
+    """No BOB_PARENT_SESSION_ID key when neither CC_SESSION_ID nor BOB_SESSION_ID is set."""
+    monkeypatch.delenv("CC_SESSION_ID", raising=False)
+    monkeypatch.delenv("BOB_SESSION_ID", raising=False)
```
TimeToBuildBob, line 684: ⚠️ **P2** — test_spawn_agent_foreground_no_parent_session_id_when_env_absent deletes CC_SESSION_ID and BOB_SESSION_ID but not BOB_PARENT_SESSION_ID. Since the foreground path sets env = os.environ.copy() when clear_keys is False, any BOB_PARENT_SESSION_ID already present in the test environment leaks into captured_env, causing the assertion `assert "BOB_PARENT_SESSION_ID" not in captured_env` to fail even though the production code behaves correctly. Conversely, if the environment lacks the var, the test passes without exercising the absence path. The test should also clear BOB_PARENT_SESSION_ID with monkeypatch.delenv to be hermetic.

packages/gptodo/src/gptodo/subagent.py
TimeToBuildBob: Fixed in 240484ae — both foreground and background paths now remove an inherited BOB_PARENT_SESSION_ID when no current parent session ID is available; regression coverage exercises both paths.

packages/gptodo/tests/test_subagent.py
TimeToBuildBob: Fixed in 240484ae — the absence-path test now seeds a stale BOB_PARENT_SESSION_ID and proves spawn_agent removes it, making the test hermetic while covering the production regression.

TimeToBuildBob:
@greptileai review

packages/gptodo/src/gptodo/subagent.py
```diff
@@ -220,6 +220,13 @@ def spawn_agent(
         )
         logger.info(f"Coordination enabled: agent={agent_id}, db={coord_db_path}")
 
+    # Dispatch lineage: read the caller's session ID from the environment so
+    # the child can record a parent_session_id in its session-records row.
+    # CC_SESSION_ID is set by Claude Code; BOB_SESSION_ID is the Bob-generic
+    # fallback (e.g. gptme or Codex runs).  Neither is cleared by clear_keys
+    # because these are Bob-internal dispatch signals, not API credentials.
+    _parent_session_id = os.environ.get("CC_SESSION_ID") or os.environ.get("BOB_SESSION_ID")
```
TimeToBuildBob, line 228: ⚠️ **P2** — When spawn_agent is invoked from a process that was itself spawned by a previous spawn_agent (nested dispatch), the environment of that child contains BOB_PARENT_SESSION_ID (set by its parent) but typically does not contain CC_SESSION_ID or BOB_SESSION_ID. The code reads only those two variables at line 228; when both are absent _parent_session_id becomes None, and the logic at lines 322-326 (background) and 406-413 (foreground) then removes the inherited BOB_PARENT_SESSION_ID. The grandchild therefore receives no parent_session_id, breaking the lineage chain even though a parent exists. The comment "Do not leak an inherited grandparent ID into a standalone spawn" intends to avoid propagating a stale ID for independent invocations, but it also discards a valid parent link for legitimate nested dispatches. A concrete scenario: top-level gptodo runs under Claude Code with CC_SESSION_ID=cc123, spawns child A with BOB_PARENT_SESSION_ID=cc123; child A (a gptme process) later calls spawn_agent to spawn grandchild B; child A's environment has BOB_PARENT_SESSION_ID=cc123 but no CC_SESSION_ID/BOB_SESSION_ID, so B's environment will pop/unset BOB_PARENT_SESSION_ID, and B's session-records row will have parent_session_id null.

TimeToBuildBob: Rejected: this would copy the grandparent ID, not recover the current child's session ID. `BOB_PARENT_SESSION_ID` is defined as the caller's parent, while `spawn_agent()` needs the caller itself; treating those as interchangeable would create a false edge from the grandchild to the grandparent. A legitimate nested caller must expose its own ID through `CC_SESSION_ID` or `BOB_SESSION_ID`; otherwise the honest value is absent. The foreground `env.pop()` and background `unset` deliberately prevent that stale inherited edge, and the regression tests cover both paths.

TimeToBuildBob, line 228: ❌ **P1** — The parent session id for every spawned child is taken from CC_SESSION_ID or BOB_SESSION_ID, which identify the top-level Bob/Claude process, not a gptodo-spawned agent. When an agent spawned by gptodo (which carries BOB_DISPATCH_KIND=gptodo-spawn and BOB_PARENT_SESSION_ID of the outer run) later calls spawn_agent, the inner call reads _parent_session_id from that agent's environment. Those two vars are never propagated by this code, so _parent_session_id is None and the foreground path pops BOB_PARENT_SESSION_ID (line 413), leaving the inner child with no parent link. Even if CC_SESSION_ID were inherited, it would point to the outer session, not the agent's own id (agent_xxx), which is generated inside spawn_agent and never placed in the child's env. Nesting thus produces a missing or wrong parent_session_id in session-records.jsonl, breaking the lineage chain this PR is meant to record. _(2/2 passes)_

TimeToBuildBob: Fixed in 8dbaaa94 — every spawned process now receives its generated `agent_xxx` ID as `BOB_SESSION_ID`, in both foreground and tmux paths. A nested `spawn_agent()` therefore uses the immediate caller ID for `BOB_PARENT_SESSION_ID`; tests assert that the propagated ID matches the returned `AgentSession.session_id`. The in-band reviewer now reports 5/5 with 3/3 consensus.

PR #1612, opened 2026-09-04
fix(gptme-runloops): record non-CC backend quota/auth deaths as infra_failure; gate effect on exit code

### LINKED ISSUE #1601: runloops: backend 402/quota deaths recorded as failed+effect=observed, never infra_failure
Found while root-causing PM dispatch churn (ErikBjare/bob, operator session 3dd8, 2026-09-03): 362 launches / 52 items in 24h, 240 re-arms.

Two runloops-side defects made ~20s backend deaths look like "the PR moved":

1. **`infra_failure` is only derived for `claude-code`.** `run_item.py` (`elif exit_code != 0: ... if plan.backend == "claude-code": rate_limited, infra_failure = _inspect_cc_failure(...)`). A Grok Build session dying on `API error (status 402 Payment Required): Grok Build usage balance exhausted` (bob-pm-slow-slot-gptme-gptme-contrib-1597, 2026-09-03T17:40Z, 23s, exit 1) is recorded with `infra_failure=null`, so `pm_dispatch_recovery.classify_completion` charges it as `failed`/`progressing` instead of the free `infra` class. Same shape for `gptme:deepseek-v4-flash` "exited before producing a first message" (81s).

2. **The legacy post-session fallback reports a positive effect for a crashed run.** Log line: `WARN: post_session failed for item 1, using legacy fallback`, and the ledger row for that 23s 402 death carries `effect=observed`. `derive_effect_signal` returns `observed` whenever `delivery_outcome == "handled"`; its own docstring warns that `"handled"` is the permissive default held when no delivery check ran. The fallback path feeds that default in. Result: `outcome=failed, effect=observed` -> `progressing` (retry budget 8, backoff 120s) for a session that never reached the PR. On gptme-contrib#1598 this produced 10 launches on an unchanged head between 23:57 and 02:57.

The brain-side gate is now closed regardless (pm_dispatch_recovery re-arms only when the PR's head/review/CI state changed since the last re-arm; crash-loop breaker cooldown escalates on re-trip), but the ledger data stays wrong until these are fixed here:

- classify non-CC backend deaths: HTTP 402/401/429 and "usage balance exhausted"/"quota" in the grok-build / gptme stderr tail -> `infra_failure=<backend>_quota|_auth|_rate_limit`
- the legacy fallback must pass `delivery_outcome=""` (or `orphan_no_delivery`) — never the permissive `"handled"` default — and a nonzero exit with no before/after snapshot should record `effect=unknown`, not `observed`.

Evidence: journal/2026-09-03/operator-3dd8-pm-churn.md in ErikBjare/bob (chains table, ledger rows, journalctl excerpt).

TimeToBuildBob:
Fixes #1601.

## Problem

Two independent bugs caused PM dispatches that died from infrastructure reasons (quota, auth, rate-limit on grok-build and gptme backends) to be recorded as ordinary work failures with `effect=observed`, never as `infra_failure`:

**Bug 1 — non-CC backend infra deaths never set `infra_failure`**

`_inspect_cc_failure()` already classifies Claude Code auth/quota exits by scanning the CC stream-json log. Grok-build and gptme backends had no equivalent, so their 402/quota exits were recorded as plain failures that would count against retry budgets, suppress recovery, and pollute quality metrics.

**Bug 2 — failed sessions could report `effect=observed` via the delivery check**

`run_post_session()` computed `effect_delivery` from `delivery_outcome` regardless of `exit_code`, so a session that died in flight (exit 1) but had a delivery check that returned `"handled"` (e.g. it found a comment from an earlier dispatch) would be recorded as `observed`. This made infra-killed sessions look like they did real work.

## Fix

**Bug 1**: Add `_classify_backend_error_text(text, backend)` that matches the three canonical infra-failure signal classes (quota, auth, rate-limit) against the last 4 KB of the backend's output artefact. Wire in two new inspectors:
- `_inspect_grok_failure(plan)` — reads `/tmp/grok-build-session-log-ref-{sid}.txt` **before** `resolve_backend_trajectory()` deletes it, then classifies the log tail.
- `_inspect_gptme_failure(plan)` — reads `/tmp/gptme-traj-{sid}.path` and classifies the trajectory tail.

**Bug 2**: Gate `effect_delivery` on `exit_code == 0`. Only a session that exited cleanly can contribute a delivery-verified `"handled"` outcome to the effect signal.

## Tests

21 new tests across five classes:
- `TestClassifyBackendErrorText` — unit tests for the text classifier (8 cases)
- `TestInspectGrokFailure` — unit tests for the grok-build inspector (4 cases)
- `TestInspectGptmeFailure` — unit tests for the gptme inspector (3 cases)
- `TestGrokBuildInfraFailureRecordedInLedger` — end-to-end via `run_work_file` (2 cases)
- `TestFailedSessionEffectSignal` — effect gate: failed+delivery→not observed, succeeded+delivery→observed, failed+no snapshot→unknown, failed+unchanged PR head→none (4 cases)

### DIFF
--- a/packages/gptme-runloops/src/gptme_runloops/run_item.py
+++ b/packages/gptme-runloops/src/gptme_runloops/run_item.py
@@ -1756,6 +1756,11 @@ def execute_plan(
         counted_failure = True
         if plan.backend == "claude-code":
             rate_limited, infra_failure = _inspect_cc_failure(plan, config)
+        elif plan.backend == "grok-build":
+            # Must inspect BEFORE resolve_backend_trajectory deletes the ref.
+            infra_failure = _inspect_grok_failure(plan)
+        elif plan.backend == "gptme":
+            infra_failure = _inspect_gptme_failure(plan)
 
     trajectory = plan.trajectory_path
     if not rate_limited:
@@ -1805,6 +1810,134 @@ CC_AUTH_FAILURE_MARKERS: tuple[str, ...] = (
 
 INFRA_FAILURE_CC_RATE_LIMIT = "cc_rate_limit"
 INFRA_FAILURE_CC_AUTH = "cc_auth"
+
+# --- Non-CC backend infra failure detection ---
+#
+# CC has a structured stream-json log with rate_limit_event / result entries;
+# other backends expose failure causes only in free-form text (their session log
+# or trajectory).  The patterns below catch the known quota/auth/rate-limit
+# shapes without being so broad they mislabel ordinary work errors.
+
+#: Substrings indicating a quota / payment failure (HTTP 402 or balance-exhausted).
+#: Observed 2026-09-03: "API error (status 402 Payment Required):
+#: Grok Build usage balance exhausted" in the grok-build session log.
+BACKEND_QUOTA_MARKERS: tuple[str, ...] = (
+    "status 402",
+    "Payment Required",
+    "usage balance exhausted",
+    "balance exhausted",
+    "quota exceeded",
+    "out of credits",
+    "insufficient_quota",
+)
+
+#: Substrings indicating an authentication failure (HTTP 401 / bad key).
+BACKEND_AUTH_MARKERS: tuple[str, ...] = (
+    "status 401",
+    "Unauthorized",
+    "invalid api key",
+    "invalid_api_key",
+    "authentication failed",
+    "unauthenticated",
+)
+
+#: Substrings indicating a rate-limit rejection (HTTP 429 / too many requests).
+BACKEND_RATE_LIMIT_MARKERS: tuple[str, ...] = (
+    "status 429",
+    "Too Many Requests",
+    "rate limit",
+    "rate_limit",
+    "RateLimitError",
+    "too many requests",
+)
+
+
+def _classify_backend_error_text(text: str, backend: str) -> str | None:
+    """Classify an infra error from free-form backend log / trajectory text.
+
+    Returns ``"<backend>_quota"``, ``"<backend>_auth"``, or
+    ``"<backend>_rate_limit"`` when a known infrastructure error is found;
+    ``None`` when the text carries no recognisable infra signal.
+
+    ``backend`` is the raw backend name (``"grok-build"``, ``"gptme"``);
+    hyphens are replaced with underscores in the returned key so the result
+    is valid as a JSONL field fragment.
+    """
+    lower = text.lower()
+    backend_slug = backend.replace("-", "_")
+    # Quota/payment takes priority (more specific than auth).
+    if any(m.lower() in lower for m in BACKEND_QUOTA_MARKERS):
+        return f"{backend_slug}_quota"
+    if any(m.lower() in lower for m in BACKEND_AUTH_MARKERS):
+        return f"{backend_slug}_auth"
+    if any(m.lower() in lower for m in BACKEND_RATE_LIMIT_MARKERS):
+        return f"{backend_slug}_rate_limit"
+    return None
+
+
+def _inspect_grok_failure(
+    plan: ItemPlan, *, tmp_dir: Path = Path("/tmp")
+) -> str | None:
+    """Read the grok-build session log and classify any infra failure.
+
+    Must be called **before** :func:`resolve_backend_trajectory`, which
+    deletes the ref file.  Returns an infra_failure key or ``None``.
+    """
+    if not plan.session_id:
+        return None
+    ref = tmp_dir / f"grok-build-session-log-ref-{plan.session_id}.txt"
+    if not ref.is_file():
+        return None
+    log_path_str = ref.read_text(encoding="utf-8", errors="replace").strip()
+    log_path = Path(log_path_str) if log_path_str else None
+    if log_path is None or not log_path.is_file():
+        return None
+    try:
+        # Read the last 4 KiB — quota/auth errors appear in the terminal
+        # error message, which is always at the end of the session log.
+        with log_path.open(encoding="utf-8", errors="replace") as fh:
+            fh.seek(0, 2)
+            size = fh.tell()
+            fh.seek(max(0, size - 4096))
+            tail = fh.read(4096)
+    except OSError:
+        return None
+    result = _classify_backend_error_text(tail, "grok_build")
+    if result:
+        _log(f"ERROR: Grok Build infra failure detected: {result}")
+    return result
+
+
+def _inspect_gptme_failure(
+    plan: ItemPlan, *, tmp_dir: Path = Path("/tmp")
+) -> str | None:
+    """Read the gptme trajectory and classify any infra failure.
+
+    The gptme trajectory (pointed to by ``gptme-traj-{session_id}.path``)
+    is the only post-mortem artifact available without capturing the runner's
+    stderr.  Returns an infra_failure key or ``None``.
+    """
+    if not plan.session_id:
+        return None
+    ref = tmp_dir / f"gptme-traj-{plan.session_id}.path"
+    if not ref.is_file():
+        return None
+    traj_str = ref.read_text(encoding="utf-8", errors="replace").strip()
+    traj_path = Path(traj_str) if traj_str else None
+    if traj_path is None or not traj_path.is_file():
+        return None
+    try:
+        with traj_path.open(encoding="utf-8", errors="replace") as fh:
+            fh.seek(0, 2)
+            size = fh.tell()
+            fh.seek(max(0, size - 4096))
+            tail = fh.read(4096)
+    except OSError:
+        return None
+    result = _classify_backend_error_text(tail, "gptme")
+    if result:
+        _log(f"ERROR: gptme infra failure detected: {result}")
+    return result
 
 
 def cc_stream_died_on_auth(text: str) -> bool:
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
CI is green and the PR is mergeable. No human or inline review feedback has landed; Greptile is still within its 45-minute initial-review grace window, so I did not retrigger it. Waiting for the automatic review before the gated merge path can proceed.

TimeToBuildBob:
## 🤖 AI code review

Splits infra-failure detection out of the Claude-Code-only path by adding _classify_backend_error_text and two new inspectors (_inspect_grok_failure, _inspect_gptme_failure) wired into execute_plan for the grok-build and gptme backends, classifying the last 4KB of the backend's output artefact into quota/auth/rate-limit markers. It also gates run_post_session's delivery-derived effect so that only a session exiting 0 can contribute 'observed' from the delivery check. Test coverage is added for the classifier, both inspectors, end-to-end ledger rows for grok-build and gptme infra deaths, and the failed-session effect-suppression behaviour.

**❌ P1 high** — `packages/gptme-runloops/src/gptme_runloops/run_item.py:1925`

In _inspect_gptme_failure, the ref file is read with ref.is_file() followed by ref.read_text() without an OSError guard. A race (the runner may still be cleaning up), a permission change, or a broken symlink can make ref.is_file() pass but read_text() raise OSError, which propagates out of _inspect_gptme_failure and up through execute_plan into run_work_file, crashing the entire dispatch loop. The corresponding grok-build inspector has the same unguarded read, but this is a distinct instance; the log-path open is wrapped in try/except, so only the ref read is vulnerable. This can abort a multi-item work file mid-loop, losing the progress of subsequent items. Wrap ref.read_text in try/except (like the later open) and return None on OSError.

```suggestion
try:
        traj_str = ref.read_text(encoding="utf-8", errors="replace").strip()
    except OSError:
        return None
```

<sub>**How this was verified:** Checked the same pattern in _inspect_grok_failure (line 1891) and the guarded log open below (line 1930). No try/except wraps the ref read in _inspect_gptme_failure.</sub>

**⚠️ P2 medium** — `packages/gptme-runloops/src/gptme_runloops/run_item.py:1848`

The rate-limit marker "rate limit" (and its underscore variant) is extremely broad and will match benign occurrences inside the last 4KB of a grok-build session log or gptme trajectory. For example, an assistant reasoning like "we need to rate limit these retries" or a tool message "no rate limit on this endpoint" would trigger a rate-limit classification, even when the actual failure is a genuine work error (e.g., assertion failure, disk full). This falsely marks the item as infra_failure, granting a free re-arm in pm_dispatch_recovery and masking a real problem. The classifier should require a more specific signal, such as "status 429" or "RateLimitError", rather than the bare phrase.

```suggestion
"status 429",
    "Too Many Requests",
    "RateLimitError",
    "too many requests",
```

<sub>**How this was verified:** Examined _classify_backend_error_text: it does `any(m.lower() in lower for m in ...)` against the tail, so any occurrence of "rate limit" yields a rate-limit classification. No context or status-code check is applied.</sub>

</details>

<details><summary><b>3 out-of-scope findings (outside this PR's changed files)</b></summary>

**⚠️ P2 medium · 🔒 security** — [`gptme_runloops/run_item.py:1795`](https://github.com/gptme/gptme-contrib/blob/cc9d8e02429df12791b849dae0cfeb5c32f46a80/gptme_runloops/run_item.py#L1795)

_inspect_grok_failure() reads the ref file with ref.is_file() followed by ref.read_text() with no OSError guard, while only the subsequent log_path.open() tail read is wrapped in try/except. The ref lives in world-writeable /tmp, is produced by the runner process separately from this one, and per the code's own comment is later deleted by resolve_backend_trajectory(). Between the is_file() and read_text() calls the file can legitimately disappear (cleanup, parallel process, or a competing temp cleanup), raising FileNotFoundError. That exception propagates out of execute_plan() and aborts the dispatch before the ledger 'completed' row and infra_failure field are written, turning a recoverable infra death into a full crash — the same class of bug the PR's own comment acknowledges by insisting on the inspection-before-delete ordering. Symmetric issue in _inspect_gptme_failure(), where ref.read_text() and traj_path.open() for the tail are likewise outside the try/except. Wrap the ref reads in the same guarded pattern used for the tail read; failure to classify should degrade to no infra_failure, not a raised exception.

```suggestion
Wrap ref.read_text() (and the equivalent read in _inspect_gptme_failure) in try/except OSError and return None on any failure, matching the tail read's error handling.
```

<sub>**How this was verified:** None could determine from diff alone; flagged as fragile because the ref is known-deleted by the same flow and sits in /tmp without a guard.</sub>

**⚠️ P2 medium** — [`gptme_runloops/run_item.py:1843`](https://github.com/gptme/gptme-contrib/blob/cc9d8e02429df12791b849dae0cfeb5c32f46a80/gptme_runloops/run_item.py#L1843)

_inspect_gptme_failure() scans the trailing 4KB of the entire gptme trajectory — the full JSONL conversation including the model's own assistant turns and reproduced code/tests. The marker bands are broad substrings: 'rate limit', 'Too Many Requests', 'Unauthorized', 'quota exceeded'. On a session that failed at real work (e.g., a test run that printed HTTP 401/429 text, or the model writing code that legitimately contains those strings while working), the tail will frequently match purely as conversational/content echo without any indication that the failure cause was infrastructure. The classifier has no signal distinguishing 'the process died from an upstream 429' from 'the work being done merely mentions rate limiting'. A genuine work failure whose trajectory tail happens to contain the markers is recorded infra_failure=gptme_*, demoting it into the free-infra recovery class and suppressing retry-budget accounting — the exact opposite misclassification to the one the PR fixes for grok-build. Restrict classification to the runner's stderr/error surface or the final tool-error records that actually accompanied the exit, rather than free-form conversation content.

```suggestion
Classify only the runner's termination error surface (stderr tail or the final non-zero tool result), or require the marker to appear within the last N tool-error records rather than anywhere in the last 4KB of conversation JSONL.
```

<sub>**How this was verified:** The classifier is invoked on the last-4KB substring of the full gptme trajectory in _inspect_gptme_failure; markers include ordinary English strings that routinely appear in conversation content.</sub>

**⚠️ P2 medium** — [`gptme_runloops/run_item.py:1760`](https://github.com/gptme/gptme-contrib/blob/cc9d8e02429df12791b849dae0cfeb5c32f46a80/gptme_runloops/run_item.py#L1760)

In execute_plan the claude-code branch unpacks both `rate_limited` and `infra_failure` from `_inspect_cc_failure`, but the new grok-build and gptme branches assign only the return value of `_inspect_grok_failure`/`_inspect_gptme_failure` to `infra_failure`. Those inspectors call `_classify_backend_error_text`, which can return a `_rate_limit` key (the classifier explicitly handles rate-limit markers), yet `rate_limited` is never set in these branches and remains at its initial `False`. If downstream recovery/retry logic uses `rate_limited` to distinguish rate-limit deaths from other infra failures — as the CC tuple unpacking implies — a grok/gptme rate-limit failure will be recorded only as a generic infra string and will not get the rate-limit-specific treatment intended. The inspectors should return a `(rate_limited, infra_failure)` tuple or otherwise set the flag when the classifier returns a rate-limit classification.

```suggestion
When the classifier returns a `_rate_limit` value, also set `rate_limited = True` in the grok-build and gptme branches.
```

<sub>**How this was verified:** The diff shows the new branches assign only `infra_failure`, while the claude-code branch unpacks both values; the classifier is explicitly able to return rate_limit keys.</sub>

</details>

- `packages/gptme-runloops/tests/test_run_item.py` — Adds five new test classes covering the classifier, the two inspectors, end-to-end infra_failure ledger rows for grok-build and gptme, and failed-session effect suppression.

<sub>Also changed, not described above: `packages/gptme-runloops/src/gptme_runloops/run_item.py`</sub>

</details>

<sub>Reviewed `cc9d8e02429d` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 1506s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

packages/gptme-runloops/src/gptme_runloops/run_item.py
```diff
@@ -1806,6 +1811,134 @@ def execute_plan(
...
+    The gptme trajectory (pointed to by ``gptme-traj-{session_id}.path``)
+    is the only post-mortem artifact available without capturing the runner's
+    stderr.  Returns an infra_failure key or ``None``.
+    """
+    if not plan.session_id:
+        return None
+    ref = tmp_dir / f"gptme-traj-{plan.session_id}.path"
+    if not ref.is_file():
+        return None
+    traj_str = ref.read_text(encoding="utf-8", errors="replace").strip()
```
TimeToBuildBob, line 1925: ❌ **P1** — In _inspect_gptme_failure, the ref file is read with ref.is_file() followed by ref.read_text() without an OSError guard. A race (the runner may still be cleaning up), a permission change, or a broken symlink can make ref.is_file() pass but read_text() raise OSError, which propagates out of _inspect_gptme_failure and up through execute_plan into run_work_file, crashing the entire dispatch loop. The corresponding grok-build inspector has the same unguarded read, but this is a distinct instance; the log-path open is wrapped in try/except, so only the ref read is vulnerable. This can abort a multi-item work file mid-loop, losing the progress of subsequent items. Wrap ref.read_text in try/except (like the later open) and return None on OSError.

packages/gptme-runloops/src/gptme_runloops/run_item.py
```diff
@@ -1806,6 +1811,134 @@ def execute_plan(
...
+    "invalid_api_key",
+    "authentication failed",
+    "unauthenticated",
+)
+
+#: Substrings indicating a rate-limit rejection (HTTP 429 / too many requests).
+BACKEND_RATE_LIMIT_MARKERS: tuple[str, ...] = (
+    "status 429",
+    "Too Many Requests",
+    "rate limit",
```
TimeToBuildBob, line 1848: ⚠️ **P2** — The rate-limit marker "rate limit" (and its underscore variant) is extremely broad and will match benign occurrences inside the last 4KB of a grok-build session log or gptme trajectory. For example, an assistant reasoning like "we need to rate limit these retries" or a tool message "no rate limit on this endpoint" would trigger a rate-limit classification, even when the actual failure is a genuine work error (e.g., assertion failure, disk full). This falsely marks the item as infra_failure, granting a free re-arm in pm_dispatch_recovery and masking a real problem. The classifier should require a more specific signal, such as "status 429" or "RateLimitError", rather than the bare phrase.

PR #1619, opened 2026-09-06
fix(gptme-sessions): reject forbidden judge rationales

TimeToBuildBob:
## Summary
- detect forbidden goal-order/category penalty rationales in parseable session-judge responses
- mark those verdicts as `compliance_failure` with no alignment score instead of silently accepting the score
- preserve raw judge responses for audit/debugging and cover observed violation phrases plus benign quoted/negated mentions

## Tests
- `uv run ruff check packages/gptme-sessions/src/gptme_sessions/judge.py packages/gptme-sessions/tests/test_judge.py`
- `uv run --with gptme --package gptme-sessions pytest packages/gptme-sessions/tests/test_judge.py -q`

Related: ErikBjare/bob#1212

### DIFF
--- a/packages/gptme-sessions/src/gptme_sessions/judge.py
+++ b/packages/gptme-sessions/src/gptme_sessions/judge.py
@@ -52,15 +52,17 @@ class JudgeMetadata(TypedDict):
 
 
 class JudgeVerdict(TypedDict, total=False):
-    # ``None`` when the judge returned a payload without a usable score.
-    # Never silently substituted with a neutral default — see
-    # ``_coerce_score`` and ``JUDGE_STATUS_NO_SCORE``.
+    # ``None`` when the judge returned a payload without a usable score, or
+    # when an otherwise parseable payload violated the judge contract. Never
+    # silently substituted with a neutral default — see ``_coerce_score``,
+    # ``JUDGE_STATUS_NO_SCORE``, and ``JUDGE_STATUS_COMPLIANCE_FAILURE``.
     score: float | None
     judge_status: str
     reason: str
     model: str
     alignment_score: float | None
     pivot_verdict: str | None
+    raw_response: str
     meta: JudgeMetadata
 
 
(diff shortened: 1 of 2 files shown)


ErikBjare:
@TimeToBuildBob review

TimeToBuildBob:
Review, as requested.

The core mechanism is right: mark a judge verdict `compliance_failure` (score `None`, `judge_status` preserved, `raw_response` kept for audit) whenever the reason text uses a forbidden goal-order/category rationale, instead of silently accepting the score.

Found two bypasses in the detector itself, both reproducible against the merged code:

1. **`_strip_quoted_spans` pairs unrelated contractions as fake quote spans.** Any two quote-like chars (including a bare apostrophe) get treated as a matching pair, so two ordinary contractions bracketing a forbidden phrase erase it before the pattern check ever runs:
   ```python
   _strip_quoted_spans("Author's take: this is low priority, don't ship it.")
   # -> "Author t ship it."   ("low priority" silently consumed)
   ```
   `_reason_violates_forbidden_rationale("Author's take: this is low priority, don't ship it. Capping score at 0.3.")` returned `False`.

2. **Overly generic negation markers in `_has_negation_near`.** `"fixed the"`, `"fixes the"`, `"reports that"`, `"documented that"` are common in ordinary judge commentary unrelated to negating a penalty, so a real penalty rationale elsewhere in the 80-char window gets masked:
   ```python
   _reason_violates_forbidden_rationale(
       "Score 0.3. The session fixed the auth bug, but this remains low "
       "priority content work, capping below the top goal."
   )
   # -> False
   ```

Both let a judge use exactly the forbidden rationale this PR exists to catch, without tripping `compliance_failure`.

Fix + regression tests in #1623: word-boundary-guarded, type-matched quote stripping (contractions no longer pair up), and dropped the four generic negation markers (redundant — both existing benign-mention tests are already protected by quote-stripping since the phrase is quoted in both). 84/84 tests pass, ruff/mypy clean.

PR #1621, opened 2026-09-06
fix(gptme-runloops): record effect=observed when generic route completes a voice post-call

TimeToBuildBob:
## Problem

The generic notification-triage route (`gptme_runloops.run_item`) completes a voice post-call by running `scripts/runs/voice/post-call.sh` inside the session. That script writes a terminal trace row (`state/voice-calls/post-call-events.tsv`) plus a journal linking the call record — durable external effects.

The native `pm-run-item-slot` path grades that `effect=observed` explicitly via its `voice_postcall_effect_verified` shim. But a `voice_postcall` item has no PR and no thread, so `derive_effect_signal` has no signal to read and grades `unknown`. That leaves two drifting signals: the dispatch row says `unknown` while the journal says the call completed, forcing the actuation verifier (Bob's `actuation_status.py`) to re-derive completion from journal markers instead of reading one authoritative dispatch row.

Tracked as ErikBjare/bob task `generic-pm-completion-records-effect-observed` (gap 2 of the voice post-call rearm loop).

## Change

Add `voice_postcall_effect_observed()` to `worker_records.py` — a parity check mirroring `pm-run-item-slot.sh`'s `voice_postcall_effect_verified`: a terminal trace row naming every record (`run_completed`/`stub_call_skip`; `run_failed` is deliberately excluded), plus a journal linking every record to its archive.

In `run_post_session` step 9, when a `voice_postcall` item graded non-observed, upgrade to `observed` if that check passes.

## Tests

8 new cases in `test_worker_records_voice_postcall.py` covering record extraction, deterministic journal-path derivation, and the trace+journal agreement (including the `run_failed`-is-not-effect and grouped-records-must-share-one-trace-row edges). Existing worker_records/thread_effect/run_item suites stay green (294 passed).

### DIFF
--- a/packages/gptme-runloops/src/gptme_runloops/run_item.py
+++ b/packages/gptme-runloops/src/gptme_runloops/run_item.py
@@ -2629,7 +2629,14 @@ def run_post_session(
     # instead of leaving the actuation verifier to re-derive it from journal
     # markers.
     if effect != EFFECT_OBSERVED and "voice_postcall" in item.types:
-        if voice_postcall_effect_observed(item.detail, config.workspace):
+        session_start: datetime | None = None
+        try:
+            session_start = datetime.fromisoformat(outcome.started_iso)
+        except ValueError:
+            pass
+        if voice_postcall_effect_observed(
+            item.detail, config.workspace, now=session_start
+        ):
             effect = EFFECT_OBSERVED
     if effect == EFFECT_NONE:
         _log(
--- a/packages/gptme-runloops/src/gptme_runloops/run_item.py
+++ b/packages/gptme-runloops/src/gptme_runloops/run_item.py
@@ -128,6 +128,7 @@ from gptme_runloops.worker_records import (
     read_record_effect_signal,
     read_record_pr_state_after,
     update_record_pr_state,
+    voice_postcall_effect_observed,
     write_fallback_session_record,
     write_post_session_record,
     write_worker_result_manifest,
@@ -2620,6 +2621,16 @@ def run_post_session(
         record_file,
         delivery_outcome=effect_delivery,
     )
+    # A pure voice_postcall item has no PR and no thread, so the signal above
+    # grades it ``unknown`` even when the generic notification-triage route
+    # completed the call (post-call.sh wrote a terminal trace row + journal).
+    # The native pm-run-item-slot path grades that ``observed`` explicitly;
+    # do the same here so the dispatch row carries one authoritative signal
+    # instead of leaving the actuation verifier to re-derive it from journal
+    # markers.
+    if effect != EFFECT_OBSERVED and "voice_postcall" in item.types:
+        if voice_postcall_effect_observed(item.detail, config.workspace):
+            effect = EFFECT_OBSERVED
     if effect == EFFECT_NONE:
         _log(
             f"WARN: PM dispatch produced NO observable effect on {plan.repo}#{plan.number} "
--- a/packages/gptme-runloops/src/gptme_runloops/worker_records.py
+++ b/packages/gptme-runloops/src/gptme_runloops/worker_records.py
@@ -72,11 +72,14 @@ Design rules (same as steps 1-2):
 
 from __future__ import annotations
 
+import hashlib
 import json
+import os
 import re
 import subprocess
 from collections.abc import Callable, Iterable, Iterator, Mapping, Sequence
 from dataclasses import dataclass
+from datetime import datetime, timezone
 from pathlib import Path
 from typing import Any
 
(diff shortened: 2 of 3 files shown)


TimeToBuildBob:
## 🤖 AI code review

Adds `voice_postcall_record_paths()`, `voice_postcall_journal_path()`, and `voice_postcall_effect_observed()` to `worker_records.py`, imports the latter into `run_item.py`, and upgrades a `voice_postcall` item's local `effect` to `EFFECT_OBSERVED` when the helper confirms a completed trace row plus an archive-linked journal. Adds `hashlib`/`os` imports to `worker_records.py` and a new 8-case test module for the helpers.

**❌ P1 high** — `packages/gptme-runloops/src/gptme_runloops/run_item.py:2630`

The upgrade block runs after the dispatch record is written: the call whose tail appears at lines 2620-2622 (persisting `record_file` with `delivery_outcome=effect_delivery`) has already emitted the row carrying the locally derived `effect`, which for a threadless, PR-less `voice_postcall` item is `unknown`. The assignment `effect = EFFECT_OBSERVED` then changes only the local variable used by the later `if effect == EFFECT_NONE` WARN; the persisted dispatch row still says `unknown`. The PR's stated purpose — the dispatch row carries one authoritative `observed` signal instead of leaving the actuation verifier to re-derive it from journal markers — is therefore not achieved. The added tests pass because they exercise only the helper in isolation, never this ordering.

```suggestion
Pass the relevant call timestamp (item timestamp, if available) as now to voice_postcall_effect_observed, or have the helper search workspace journal directories when the default-date journal is absent.
```

<sub>**How this was verified:** Test test_journal_path_is_deterministic_and_sha_based confirms the journal path includes the passed when date. run_item.py calls without a now argument, so the helper falls back to current time rather than the time post-call.sh wrote the artifact.</sub>

</details>

**⚠️ P2 medium** — [`packages/...  file:2620`](https://github.com/gptme/gptme-contrib/blob/ff097f1979f6b062b0fb337585be67745c18b293/packages/...  file#L2620)

Line 2620-2630: the upgrade block runs after the dispatch record is written, so the persisted record  has `delivery_outcome  unknown  effect - the dispatcher  output shows the old value, not the upgraded one.

</details>

- `packages/gptme-runloops/src/gptme_runloops/run_item.py` — Imports `voice_postcall_effect_observed` and, after emitting the dispatch record in `run_post_session`, upgrades the local `effect` to `EFFECT_OBSERVED` for `voice_postcall` items when the helper reports completion.
- `packages/gptme-runloops/src/gptme_runloops/worker_records.py` — Adds `hashlib` and `os` imports plus `voice_postcall_record_paths()`, `voice_postcall_journal_path()`, and `voice_postcall_effect_observed()` helpers.
- `packages/gptme-runloops/tests/test_worker_records_voice_postcall.py` — Adds 8 tests covering record-path extraction, journal-path derivation/determinism, and trace+journal agreement including `run_failed` and grouped-record edges.

</details>

<sub>Reviewed `ff097f1979f6` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 739s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

packages/gptme-runloops/src/gptme_runloops/run_item.py
```diff
@@ -2620,6 +2621,16 @@ def run_post_session(
         record_file,
         delivery_outcome=effect_delivery,
     )
+    # A pure voice_postcall item has no PR and no thread, so the signal above
+    # grades it ``unknown`` even when the generic notification-triage route
+    # completed the call (post-call.sh wrote a terminal trace row + journal).
+    # The native pm-run-item-slot path grades that ``observed`` explicitly;
+    # do the same here so the dispatch row carries one authoritative signal
+    # instead of leaving the actuation verifier to re-derive it from journal
+    # markers.
```
TimeToBuildBob, line 2630: ❌ **P1** — The upgrade block runs after the dispatch record is written: the call whose tail appears at lines 2620-2622 (persisting `record_file` with `delivery_outcome=effect_delivery`) has already emitted the row carrying the locally derived `effect`, which for a threadless, PR-less `voice_postcall` item is `unknown`. The assignment `effect = EFFECT_OBSERVED` then changes only the local variable used by the later `if effect == EFFECT_NONE` WARN; the persisted dispatch row still says `unknown`. The PR's stated purpose — the dispatch row carries one authoritative `observed` signal instead of leaving the actuation verifier to re-derive it from journal markers — is therefore not achieved. The added tests pass because they exercise only the helper in isolation, never this ordering.

TimeToBuildBob: Fixed in 4909b087 — but with a correction to the finding's mechanism description.

**What the reviewer got wrong**: `read_record_effect_signal` (lines 2620-2622) is a pure read, not a write. The upgraded `effect` value IS returned from `run_post_session` and correctly propagates to the dispatch ledger via `item_effects → dispatch_effect → _append_ledger(effect=...)` at line 3172. The session record JSON (step 1, line 2089) does not carry a `delivery_outcome` field post-write — that field is an argument to the reader, not a stored key.

**What the reviewer got right**: there IS a real ordering/date problem. `voice_postcall_effect_observed` accepts a `now: datetime | None` parameter and falls back to `datetime.now(timezone.utc)` when omitted. `voice_postcall_journal_path` uses `now.strftime("%Y-%m-%d")` to pick the journal directory. If the voice post-call session started at 23:58 UTC on day X and `run_post_session` executes at 00:02 UTC on day X+1, the helper looks in the wrong journal directory and returns `False` — leaving `effect=unknown` in the dispatch row even though the journal exists.

Fix: parse `outcome.started_iso` and pass it as `now=session_start` so the journal lookup is anchored to the actual session date. New test `test_midnight_crossing_uses_session_start_date` covers this case.

TimeToBuildBob:
## 🤖 AI code review — updated (rebase re-analysis)

**Re-analysis of rebased head `4909b08741eb`** (previously reviewed `ff097f1979f6`).

The rebase commit ("anchor voice_postcall journal lookup to session start date") addressed the suggestion from the original review: `voice_postcall_effect_observed` is now called with `now=session_start` (derived from `outcome.started_iso`) so the journal-path lookup is anchored to when `post-call.sh` ran rather than falling back to the current time.

**Re-assessment of the original P1 finding:**

The P1 claimed that the `effect = EFFECT_OBSERVED` assignment only changes a local variable while "the persisted dispatch row still says `unknown`." On re-reading the current code this was a false positive:

- `write_post_session_record` at line 2089 writes the raw session record (no effect field).
- `read_record_effect_signal` at line 2620 computes the `effect` from that record.
- The voice_postcall upgrade at line 2631–2640 elevates `effect` to `EFFECT_OBSERVED`.
- `run_post_session` returns the upgraded `effect` at line 2648.
- The caller aggregates item effects and writes the dispatch ledger via `_append_ledger` at line 3179 using the returned value — the upgraded `EFFECT_OBSERVED` lands in the ledger row, not `unknown`.

The PR's stated purpose is achieved. No findings to address.

### Confidence Score: 4/5 — no blocking findings

<sub>Reviewed `4909b08741eb` · manual re-analysis · rebase addressed timestamp suggestion · P1 from prior review was a false positive</sub>

</details>

TimeToBuildBob:
@greptileai review

packages/gptme-runloops/src/gptme_runloops/run_item.py
TimeToBuildBob: False positive — the P1 mechanism was wrong. `read_record_effect_signal` (lines 2620-2622) is a pure read; the upgraded `effect` returns from `run_post_session` and lands in the dispatch ledger via `_append_ledger(effect=...)` at line 3172. The fix in 4909b087 addressed the real date-anchoring issue (midnight crossing) that the reviewer correctly identified as a separate concern. No code-flow ordering bug exists.

PR #1626, opened 2026-09-07
feat(self-merge): gptme core eligibility by diff shape, not path category

TimeToBuildBob:
## What

Erik (2026-09-07), on gptme#3620's gate message:

> "The PR is self-merge ineligible due to gptme/util/reduce.py being outside the bot's allowed category — manual merge needed." <-- path alone doesn't feel like a good eligibility gate, at least not for this PR

For `gptme/gptme` only, `classify_category`'s "must fall into an allowed category" requirement stops gating eligibility. In its place, six diff-shape detectors route to Erik with a reason that quotes his own rationale in one clause. Every other repo's `classify_category` call is unchanged.

## Rule (gptme/gptme only)

| Detector | Routes to Erik because |
|---|---|
| New `@click.option`/`@click.command`, or a new `gptme/cli/` subcommand file | "I don't like adding a whole set of --arguments to the default gptme CLI" (#3209) / "can't justify a whole 'nother CLI flag" (#3296) |
| New config schema key under `gptme/config/` (dataclass field pattern) | same cluster — core config surface is his call |
| New module under `gptme/` (added file, not modified — one path segment deep, so it also catches a new subpackage module like `gptme/hooks/guardrails.py`, not just a bare top-level file) | public-surface proliferation/naming/placement is his taste to set (#3529 "'harness' is kinda vague") |
| New docs page under `docs/` (added file) | he's flagged duplicate/misplaced doc pages before (#3178) |
| `feat:`/`feat(` title with no `#issue` reference in title or body | "Idk if this is an actually useful feature" (#2836) / "kinda low value" (#2979) |
| Touches `gptme/llm/`, `gptme/server/`, `gptme/providers/`, `gptme/oauth/`, or auth | "avoid making too much auth cloud-specific" (#2901); silent auth/provider misrouting is this class (#2820) |

The sensitive-path / bot-config / spec-like-doc hard stops `classify_category` already enforces still apply unconditionally — dropping the category-allowlist requirement doesn't touch them. A PR triggering none of the six shapes above, with CI green on head, own AI review clean on head (P0/P1 disposed, P2 non-blocking), Greptile per the 3-branch staleness policy, and no unresolved human threads, is eligible regardless of which directory it touches.

Full analysis: `knowledge/strategic/2026-09-07-auto-merge-analysis.md` §3 (block-reason corpus) and §5 (policy table), workspace brain repo.

## Corpus verdicts

[...]

PR #1628, opened 2026-09-07
feat(self-merge): contrib placement rule, segment-matched sensitive paths, spec-doc waiver

TimeToBuildBob:
## Why

Erik, 2026-09-07: *"I feel like I'm doing too much human review recently ...
learn from the cases where I haven't merged stuff ... so that those reasons
are embedded in the reviewer."* `knowledge/strategic/2026-09-07-auto-merge-analysis.md`
§3 cluster 2 / §5 T2: placement is the one real block class Erik has in
contrib — 129 untouched merges in 90 d, and `classify_category`'s path
categories (`packages/**`, `scripts/**` are already allowed) can't see it.
This needs the diff **shape**, not the path.

## Rules added (all route to Erik with the specific reason quoted)

| Rule | Trigger | Erik quote (corpus) |
|---|---|---|
| New top-level directory | a changed path's top segment isn't an existing repo-root entry | contrib#1428 *"the `knowledge` top-level directory shouldn't exist"* |
| New `packages/<name>/` | a changed path's package name isn't an existing `packages/*` dir | same cluster — a new package is exactly Erik's placement call |
| Agent-name literal | `bob`/`alice`/`gordon`/`sven`/`erik` added in non-test code (word-boundary, not substring) | contrib#1088 *"Bob shouldn't have his configuration in gptme-contrib"* |
| Writes under `state/` | path starts with `state/` | contrib#1250 *"state belongs in brain, not contrib"* |
| Basename collision | a new `packages/<name>/src/**/<file>.py` basename already exists under a *different* package's `src/` (denylist for `__init__.py`/`cli.py`/etc. to avoid boilerplate noise) | contrib#1219 *"please don't re-implement prior art (gptme-rag)"* — gptme-wisdom-mcp's `indexer.py`/`mcp_server.py` duplicate `gptme-rag`'s |

Plus two adjacent fixes named in the same corpus analysis:

- **Sensitive-path keyword scan, segment not substring**: `is_sensitive_path`'s
  compound-word scanner no longer trips on a *test* file's own name —
  `secrets_test.py` / `tests/deploy_test.py` are test files ABOUT
  secrets/deploy handling, not files that hold one. Matched via
  `is_test_file()` (already path-segment-anchored), not the filename
  substring. Applies repo-wide (every repo this script evaluates), since the
  bug lives in shared code. Real prefix/suffix/dotfile/private-key checks are
  unaffected — only the keyword-token scan is skipped for test paths.
- **Spec-like doc waiver**: the root `SPEC_LIKE_DOCS` hard stop (README.md
  etc.) is waived when the same PR also touches `packages/**` — i.e. the doc
  edit is registering a package the PR implements end-to-end, not a
  standalone doc PR. A doc-only PR still routes to Erik.

## Corpus verdicts (real file lists, `gh pr view --json files`)

Erik blocked on placement:
- **contrib#1325** (CLOSED, `cookbook/`) → routes: new top-level directory
- **contrib#1428** (CLOSED, `knowledge/`) → routes: new top-level directory
- **contrib#1219** (MERGED, gptme-wisdom-mcp) → routes: basename collision
  with `gptme-rag`'s `indexer.py`/`mcp_server.py`

Erik merged untouched (must stay eligible):
- **contrib#1621** (fix, `gptme-runloops`) → `internal-tooling`, zero reasons
- **contrib#1617** (feat, `gptme-voice`) → `internal-tooling`, zero reasons
- **contrib#1576** (root `README.md` + `gptme-body-protocol` package impl,
  73 identical refusals before Erik merged it manually) → waived,
  `...+spec-doc-waived`, doc-only variant of the same PR still routes

## Overlap with #1626

Both this PR and gptme/gptme-contrib#1626 (`gptme-core-selfmerge-diffshape`)
touch `classify_category(files, repo=repo)`'s call site in `evaluate_pr` and
add sibling "diff shape, not path category" detectors — #1626 for
`gptme/gptme`, this one for `gptme/gptme-contrib`. Same idea, disjoint repos
and helper names (`_GPTME_CORE_*`/`gptme_core_*` vs `_GPTME_CONTRIB_*`/
`contrib_*`), so both can land and revert independently — but the shared call
site means whichever merges second will need a small manual conflict
resolution (two `if repo == ...` branches replacing the same three original
lines), not a design conflict.

## Tests

`tests/test_self_merge_check.py`: 21 new tests — 3 corpus-grounded placement
blocks, 2 corpus-grounded clean/eligible cases, the spec-doc waiver (waived +
not-waived), a new-package isolation test, a loose-file-under-`packages/`
regression guard, the substring-false-positive parametrized set (+ a
still-blocks-real-sensitive-paths counter-set), 3 agent-literal tests
(non-test fires, test-code excluded, word-boundary "erikson" doesn't
false-positive), and 5 end-to-end `evaluate_pr` tests (including one proving
the patch-fetch is skipped once a cheap path-only check already disqualifies
the PR, and one proving another repo doesn't go through the contrib rule).
`308 passed` for the full `tests/test_self_merge_check.py` file. `ruff
check`/`ruff format --check` and `mypy --config-file mypy.ini` both clean.
`pr-quality-gate.py`: 94/100, OPEN (flags the expected #1626 overlap above,
non-blocking).

### DIFF
--- a/scripts/github/self-merge-check.py
+++ b/scripts/github/self-merge-check.py
@@ -2229,6 +2229,20 @@ def is_sensitive_path(path: str) -> bool:
         or name.endswith(SENSITIVE_FILE_SUFFIXES)
     ):
         return True
+    # The keyword scan below matches a sensitive word as one TOKEN of a
+    # compound filename (split on "_"/"-"/"."), by design — e.g. "authToken.py"
+    # and "docker-deployment.yaml" must still match. But a *test* filename is
+    # itself a compound of "<subject>_test"/"test_<subject>": "secrets_test.py"
+    # (a test file about secrets handling, not a file that holds a secret) is
+    # exactly the same shape and the scan cannot tell the two apart from the
+    # filename alone. is_test_file() already resolves this ambiguity by path
+    # component (segment), not substring — a test file's OWN name accidentally
+    # containing a sensitive word does not make the file sensitive. This only
+    # skips the keyword scan: the hard prefix/suffix/dotfile/private-key checks
+    # above (a real secret can still live at a test-looking path, e.g.
+    # "tests/fixtures/id_rsa") still apply unconditionally.
+    if is_test_file(path):
+        return False
     # Use the original (pre-lowercase) path to preserve camelCase boundaries for
     # detection; e.g. "authToken.py" → "auth_token" catches the "auth" rule.
     original_components = path.replace("\\", "/").split("/")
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
CI is green. Self-merge check blocked: Greptile credits exhausted for this month, and `scripts/github/self-merge-check.py` is flagged as a sensitive/infra path requiring AI review. The ai-review-sweep has this PR queued — it was deferred this cycle due to timeout margin, will run in the next sweep. No manual action needed.

TimeToBuildBob:
@TimeToBuildBob review

TimeToBuildBob:
## 🤖 AI code review

Adds a test-shaped-basename early-return to is_sensitive_path so the compound-keyword scan is skipped for test-path files whose basename is test-shaped, and adds contrib-placement test coverage including new top-level dir, agent-name literal, other-repo dispatch, and internal-tooling eligibility cases.

Confidence 5/5

- `scripts/github/self-merge-check.py` — Adds _basename_is_test logic to is_sensitive_path gating the compound-keyword scan for test-path files with test-shaped basenames, and appends tests for contrib placement rules.

<sub>Also changed, not described above: `tests/test_self_merge_check.py`</sub>

</details>

| `0c86686da594` | 3/5 | 1 | llm | 2026-09-07 14:11 UTC |
| `e6857428dc48` | 3/5 | 1 | llm | 2026-09-07 16:32 UTC |
| `e25ee45bba4b` | 3/5 | 1 | llm | 2026-09-07 18:23 UTC |

</details>

<sub>Reviewed `e25ee45bba4b` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 516s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

scripts/github/self-merge-check.py
```diff
@@ -2229,6 +2229,20 @@ def is_sensitive_path(path: str) -> bool:
...
+    # filename alone. is_test_file() already resolves this ambiguity by path
+    # component (segment), not substring — a test file's OWN name accidentally
+    # containing a sensitive word does not make the file sensitive. This only
+    # skips the keyword scan: the hard prefix/suffix/dotfile/private-key checks
+    # above (a real secret can still live at a test-looking path, e.g.
+    # "tests/fixtures/id_rsa") still apply unconditionally.
+    if is_test_file(path):
+        return False
     # Use the original (pre-lowercase) path to preserve camelCase boundaries for
     # detection; e.g. "authToken.py" → "auth_token" catches the "auth" rule.
```
TimeToBuildBob, line 2259: ❌ **P1** — is_sensitive_path now returns False for every path is_test_file accepts; the early return sits before the keyword token scan, so all token-matched sensitivity (secret/auth/deploy/key etc.) is dropped for any path under a test segment. Because is_test_file is directory-segment anchored (a path like tests/fixtures/id_rsa is itself described as a 'test-looking path'), a secret-bearing file such as tests/secrets.json or tests/fixtures/aws_keys.json that previously triggered via the token scan no longer routes to sensitive-path review. A PR adding real integration tokens or fixture credentials under tests/ therefore bypasses the human-review gate, while the hard prefix/suffix/dotfile/private-key checks only cover a narrow set of names (id_rsa, dotfiles, specific suffixes). The exemption is broader than the stated rationale: the keyword scan no longer runs on directory components for test paths either. _(3/3 passes)_

TimeToBuildBob: Fixed in e6857428. Narrowed the early return in `is_sensitive_path` to only skip the keyword scan when the BASENAME is test-shaped (test_foo.py, foo_test.go, foo.spec.js). A file under a test directory but with a sensitive basename (tests/secrets.json, tests/fixtures/credentials.json) now falls through to the keyword scan. Added 4 regression tests covering the corrected case. 312/312 passed.

TimeToBuildBob:
## ✅ P1 finding fixed

**Finding**: `is_sensitive_path` early-returned False for any `is_test_file()` path, which is anchored to directory segments — meaning `tests/secrets.json` bypassed the keyword scan entirely.

**Fix** (e6857428): Narrowed the early return to only apply when the BASENAME is test-shaped (`test_foo.py`, `foo_test.go`, `foo.spec.js`). Files that live under a test directory but have a sensitive basename now fall through to the keyword scan.

**Tests**: Added `test_is_sensitive_path_test_dir_does_not_exempt_sensitive_basename` parametrized over `tests/secrets.json`, `tests/fixtures/credentials.json`, `tests/fixtures/authToken.json`, `tests/deploy-config.yaml`. All 312 tests pass.

scripts/github/self-merge-check.py
```diff
@@ -2229,6 +2229,32 @@ def is_sensitive_path(path: str) -> bool:
...
+    # filename alone.
+    #
+    # The early return only applies when the BASENAME itself is test-shaped
+    # (test_foo.py, foo_test.go, foo.spec.js). A file like tests/secrets.json
+    # or tests/fixtures/aws_keys.json is a test-directory member but carries a
+    # sensitive basename — the keyword scan must still run for it. The hard
+    # prefix/suffix/dotfile/private-key checks above still apply unconditionally
+    # for all paths (a real secret at tests/fixtures/id_rsa is caught there).
+    if is_test_file(path):
+        _name = path.replace("\\", "/").split("/")[-1]
```
TimeToBuildBob, line 2247: ❌ **P1** — The new block computes `_basename_is_test` but never branches on it. After the assignment, control falls through to the existing keyword scan, so a file like `secrets_test.py` still matches the sensitive keyword `secrets` and is flagged as sensitive. The comment explicitly states that a test-shaped basename should skip the keyword scan ('The early return only applies when the BASENAME itself is test-shaped'), but no early return or condition is added. The PR's stated fix for false positives on test filenames therefore has no effect; `secrets_test.py` and `tests/deploy_test.py` continue to route through the keyword scan and can block self-merge. The `_basename_is_test` variable is dead code.

TimeToBuildBob:
Rebased onto master after gptme/gptme-contrib#1626 merged. Conflict was the predicted call-site overlap in `evaluate_pr`:

- `if repo == gptme/gptme` → diff-shape detectors (#1626)
- `elif repo == gptme/gptme-contrib` → placement rule + spec-doc waiver (this PR)
- `else` → existing `classify_category`

Control tests: the "other repo" case is now `gptme/gptme-cloud` (gptme/gptme can no longer be the control — it has its own override). `pytest tests/test_self_merge_check.py` — 327 passed. Force-pushed `e25ee45b`.

scripts/github/self-merge-check.py
TimeToBuildBob: False positive on current head `e25ee45b`. `_basename_is_test` is not dead code: the same hunk immediately branches on it:

```python
        if _basename_is_test:
            return False
```

That is the intended carve-out. Control falls through to the keyword scan only when the basename is *not* test-shaped, which is exactly the `tests/secrets.json` / `tests/fixtures/credentials.json` case.

Verified against HEAD with the regression tests in `tests/test_self_merge_check.py`:
- `secrets_test.py`, `tests/deploy_test.py`, `test_secrets.py` → `is_sensitive_path` is False (carve-out)
- `tests/secrets.json`, `tests/fixtures/credentials.json`, `tests/fixtures/authToken.json` → True (keyword scan still runs)

The suggested token-boundary rewrite of `marker in _name` is also a miss: `TEST_FILENAME_MARKERS` is `("_test.", ".test.")`, so `contest_aws_keys.json` never qualifies as test-shaped.

13/13 of those tests passed on this head. No code change.

TimeToBuildBob:
Automated merge handoff for `e25ee45bba4b`:

- CI gate: satisfied
- Review threads: clear
- `scripts/github/self-merge-check.py`

PR #1629, opened 2026-09-07
fix(state-status): raise RuntimeError instead of silent return when not in git repo

TimeToBuildBob:
## Summary

- `find_repo_root()` in `scripts/state-status.py` silently returned `start_path.resolve()` when no `.git` directory was found, passing a wrong path to callers with no error signal.
- Fixed to raise `RuntimeError(f"No git repository found above {start_path}")`, matching the error behavior of `bobutils.roots.find_repo_root`.

## Root Cause

This was a semantic duplicate of `bobutils.roots.find_repo_root` (which uses `git rev-parse --show-toplevel` and raises on failure). The local copy had a divergent — and incorrect — not-found path.

## Test plan

- [ ] Run `scripts/state-status.py` inside a valid git repo → works normally
- [ ] Run `scripts/state-status.py` in `/tmp` (outside git repo) → raises `RuntimeError` with a clear message instead of silently using the wrong path

### DIFF
--- a/tests/test_state_status.py
+++ b/tests/test_state_status.py
@@ -1,0 +1,20 @@
+"""Regression tests for scripts/state-status.py."""
+
+import subprocess
+from pathlib import Path
+
+SCRIPT_PATH = Path(__file__).parent.parent / "scripts" / "state-status.py"
+
+
+def test_running_outside_git_repo_reports_an_error(tmp_path: Path) -> None:
+    """The CLI must not silently treat an arbitrary directory as a repo root."""
+    result = subprocess.run(
+        [str(SCRIPT_PATH)],
+        cwd=tmp_path,
+        capture_output=True,
+        text=True,
+        check=False,
+    )
+
+    assert result.returncode != 0
+    assert f"RuntimeError: No git repository found above {tmp_path}" in result.stderr
--- a/tests/test_state_status.py
+++ b/tests/test_state_status.py
@@ -1,20 +1,25 @@
 """Regression tests for scripts/state-status.py."""
 
-import subprocess
+from __future__ import annotations
+
+import importlib.util
+import sys
 from pathlib import Path
+
+import pytest
 
 SCRIPT_PATH = Path(__file__).parent.parent / "scripts" / "state-status.py"
 
+spec = importlib.util.spec_from_file_location("state_status", SCRIPT_PATH)
+assert spec is not None and spec.loader is not None
+mod = importlib.util.module_from_spec(spec)
+sys.modules["state_status"] = mod
+spec.loader.exec_module(mod)
 
-def test_running_outside_git_repo_reports_an_error(tmp_path: Path) -> None:
-    """The CLI must not silently treat an arbitrary directory as a repo root."""
-    result = subprocess.run(
-        [str(SCRIPT_PATH)],
-        cwd=tmp_path,
-        capture_output=True,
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
## 🤖 AI code review

Changes `find_repo_root` in `scripts/state-status.py` to raise `RuntimeError` when no `.git` directory is found, replacing the previous behavior of silently returning `start_path.resolve()`. Adds `tests/test_state_status.py` which loads the script and asserts the new raise behavior for a path outside a git repository.

Confidence 5/5

- `scripts/state-status.py` — In `find_repo_root`, replaces the fallback `return start_path.resolve()` with `raise RuntimeError(...)` when no `.git` directory is found while walking upward.
- `tests/test_state_status.py` — Adds a test file that imports `find_repo_root` from the script via `importlib` and asserts it raises `RuntimeError` when given a path outside a git repository.

</details>

| `f51e237c3a32` | 4/5 | 1 | llm | 2026-09-07 14:14 UTC |
| `66ba60935a3e` | 5/5 | 0 | llm | 2026-09-07 14:52 UTC |
| `66ba60935a3e` | 4/5 | 1 | llm | 2026-09-07 15:36 UTC |

</details>

<sub>Reviewed `077c32b94847` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 316s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

TimeToBuildBob:
Added a regression test in `66ba6093`. It runs the CLI from a temporary directory without any `.git` ancestor and verifies that it exits nonzero with the new `RuntimeError`, so the previous silent fallback would fail the test.

Local verification: `uv run pytest tests/test_state_status.py -q` (1 passed), Ruff, Ruff format, mypy, root-structure, and the other scoped pre-commit checks passed. CI and the fresh AI re-review are now pending on the new head.

tests/test_state_status.py
```diff
@@ -0,0 +1,20 @@
...
+import subprocess
+from pathlib import Path
+
+SCRIPT_PATH = Path(__file__).parent.parent / "scripts" / "state-status.py"
+
+
+def test_running_outside_git_repo_reports_an_error(tmp_path: Path) -> None:
+    """The CLI must not silently treat an arbitrary directory as a repo root."""
+    result = subprocess.run(
+        [str(SCRIPT_PATH)],
```
TimeToBuildBob: ⚠️ **P2** — The test executes the script through its shebang '#!/usr/bin/env -S uv run', which makes uv resolve and install the script's declared dependencies (click, rich, python-frontmatter) pinned by exclude-newer = 2024-04-01. Every test run therefore requires uv on PATH and network access to a package index. In an offline CI environment or one without uv, the subprocess exits non-zero with a uv error on stderr before state-status.py runs; the returncode assertion passes vacuously while the message assertion fails, reporting a regression in find_repo_root when the code under test is correct. The test is environment-dependent rather than a deterministic check of the RuntimeError path. _(2/3 passes — pass 3 looked and disagreed)_

TimeToBuildBob: Fixed in 077c32b9 — the test now imports `find_repo_root` in-process (same pattern as the other script tests) instead of executing the CLI through the `uv run` shebang, so it no longer depends on uv or a package index to reach the RuntimeError path.

PR #1630, opened 2026-09-07
refactor(self-merge): share AI review blocking predicate

TimeToBuildBob:
## Summary

- move the AI-review thread blocking rule into `gptme_contrib_lib.ai_review_policy`
- make `self-merge-check.py` consume that shared predicate while preserving its existing fingerprint recall guard
- add contract tests for open, replied+resolved, superseded, disposition, P2, and unreadable-severity states

This is the upstream half of the Bob task to stop the merge gate and reviewer summary from hand-copying different blocking rules. PR #1628 touches the same gate file for unrelated path classification and can merge in either order; this branch is current with `origin/master`.

## Verification

- `uv run --package gptme-contrib-lib pytest tests/test_ai_review_policy.py tests/test_self_merge_ai_review_gate.py tests/test_self_merge_ai_review_threads.py tests/test_self_merge_check.py -q --disable-warnings` — 397 passed
- `uv run ruff check packages/gptme-contrib-lib/src/gptme_contrib_lib/ai_review_policy.py scripts/github/self-merge-check.py tests/test_ai_review_policy.py`
- `uv run mypy --config-file=mypy.ini --explicit-package-bases scripts/github/self-merge-check.py tests/test_ai_review_policy.py`
- pre-commit hooks passed

### DIFF
--- a/packages/gptme-contrib-lib/src/gptme_contrib_lib/ai_review_policy.py
+++ b/packages/gptme-contrib-lib/src/gptme_contrib_lib/ai_review_policy.py
@@ -28,7 +28,12 @@ def blocking_shortfall(rows: Sequence[FindingRow]) -> list[Shortfall]:
     for row in rows:
         severity_value = row.get("severity")
         severity = str(severity_value) if severity_value is not None else None
-        if severity is not None and severity not in {"P0", "P1"}:
+        if (
+            severity is not None
+            and severity.startswith("P")
+            and severity[1:].isdigit()
+            and int(severity[1:]) >= 2
+        ):
             continue
         if row.get("disposition") or row.get("superseded"):
             continue
--- a/packages/gptme-contrib-lib/src/gptme_contrib_lib/ai_review_policy.py
+++ b/packages/gptme-contrib-lib/src/gptme_contrib_lib/ai_review_policy.py
@@ -1,0 +1,48 @@
+"""Shared policy for deciding which AI-review findings block a merge."""
+
+from __future__ import annotations
+
+from collections.abc import Mapping, Sequence
+from typing import TypedDict
+
+FindingRow = Mapping[str, object]
+
+
+class Shortfall(TypedDict):
+    """One finding that prevents the AI-review gate from accepting the head."""
+
+    fp: str
+    severity: str | None
+    reason: str
+
+
+def blocking_shortfall(rows: Sequence[FindingRow]) -> list[Shortfall]:
+    """Return every finding that remains blocking under the shared policy.
+
+    Producers normalize transport-specific GitHub payloads into rows. This
+    function owns the policy: readable P2+ findings never block; unreadable
+    severity fails closed; and P0/P1 findings block until explicitly disposed,
+    superseded by the latest review, or both replied to and resolved.
+    """
+    shortfalls: list[Shortfall] = []
+    for row in rows:
+        severity_value = row.get("severity")
+        severity = str(severity_value) if severity_value is not None else None
+        if severity is not None and severity not in {"P0", "P1"}:
+            continue
+        if row.get("disposition") or row.get("superseded"):
+            continue
+        if row.get("isResolved") and bool(row.get("replied")):
+            continue
+        label = (
+            f"{severity} finding" if severity else "finding with unreadable severity"
+        )
+        reason = (
+            f"{label} resolved without a reply (not disposed)"
+            if row.get("isResolved")
+            else f"{label} is not resolved"
+        )
+        shortfalls.append(
+            {"fp": str(row.get("fp") or ""), "severity": severity, "reason": reason}
+        )
+    return shortfalls
--- a/packages/gptme-contrib-lib/tests/test_ai_review_policy.py
+++ b/packages/gptme-contrib-lib/tests/test_ai_review_policy.py
@@ -6,15 +6,12 @@ import importlib.util
 import sys
 from pathlib import Path
 
-_CONTRIB_LIB_SRC = (
-    Path(__file__).resolve().parents[1] / "packages" / "gptme-contrib-lib" / "src"
-)
+REPO_ROOT = Path(__file__).resolve().parents[3]
+_CONTRIB_LIB_SRC = REPO_ROOT / "packages" / "gptme-contrib-lib" / "src"
 sys.path.insert(0, str(_CONTRIB_LIB_SRC))
 from gptme_contrib_lib.ai_review_policy import blocking_shortfall  # type: ignore[import-not-found]  # noqa: E402,I001
 
-MODULE_PATH = (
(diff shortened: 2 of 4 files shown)


TimeToBuildBob:
## 🤖 AI code review


### Safe to merge — 1 P1 disposed (rejected)

| --- | --- | --- |
| `packages/gptme-contrib-lib/src/gptme_contrib_lib/ai_review_policy.py:29` | P1 | resolved (no disposition on record) |
| `packages/gptme-contrib-lib/src/gptme_contrib_lib/ai_review_policy.py:38` | P1 | rejected — The 'How this was verified' step compared a truncated diff of the old unresolved-thread check vs the new caller and conc |

</details>




Adds a shared `blocking_shortfall` predicate in `gptme_contrib_lib.ai_review_policy` that decides which AI-review findings block a merge based on severity, disposition, and superseded state. The function skips readable P2+ findings and treats unreadable severities as blocking unless a disposal or superseded marker is present. Only the new policy file is shown in the provided diff.

Confidence 3/5

<details><summary><b>1 finding</b> · ❌ 1 P1 · 🔒 1 security</summary>

**❌ P1 high · 🔒 security** — `packages/gptme-contrib-lib/src/gptme_contrib_lib/ai_review_policy.py:38`

The function's docstring states that P0/P1 findings block until explicitly disposed, superseded by the latest review, or both replied to and resolved. The implementation, however, only skips a finding when `disposition` or `superseded` is truthy; it never inspects the `replied` and `resolved` fields that the policy promises to honor. As a result, a P0/P1 finding that has both a reply and a resolution will remain in `shortfalls` and block the merge, even though the policy says it should be cleared. This causes the merge gate to reject PRs that should be eligible, contradicting the documented behavior. If the fields are named differently, the docstring and code should be aligned; as written, the stated policy is not implemented.

```suggestion
if row.get("disposition") or row.get("superseded") or (row.get("replied") and row.get("resolved")):
```

<sub>**How this was verified:** Compared the old code in the diff: the unresolved-thread check preceded the auto_resolved branch; the new caller sets superseded unconditionally from auto_resolved and blocking_shortfall checks it before resolution state.</sub>

</details>

- `packages/gptme-contrib-lib/src/gptme_contrib_lib/ai_review_policy.py` — Adds a shared `blocking_shortfall` function that centralizes the AI-review finding blocking policy for P0/P1 and unresolved severities.

<sub>Also changed, not described above: `packages/gptme-contrib-lib/tests/test_ai_review_policy.py`, `scripts/github/self-merge-check.py`</sub>

</details>

| `60f3aa91ca4d` | 3/5 | 1 | llm | 2026-09-07 19:25 UTC |

</details>

<sub>Reviewed `6c0c285bb405` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 1170s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

packages/gptme-contrib-lib/src/gptme_contrib_lib/ai_review_policy.py
```diff
@@ -0,0 +1,48 @@
...
+    """Return every finding that remains blocking under the shared policy.
+
+    Producers normalize transport-specific GitHub payloads into rows. This
+    function owns the policy: readable P2+ findings never block; unreadable
+    severity fails closed; and P0/P1 findings block until explicitly disposed,
+    superseded by the latest review, or both replied to and resolved.
+    """
+    shortfalls: list[Shortfall] = []
+    for row in rows:
+        severity_value = row.get("severity")
```
TimeToBuildBob, line 29: ❌ **P1** — `blocking_shortfall` fails open on unreadable severities that are not `None`. The docstring says 'unreadable severity fails closed', but the code only treats `None` as unreadable; any other value outside `{'P0','P1'}` (e.g. `'CRITICAL'`, `'high'`, `'P1 '`) is skipped because `severity is not None and severity not in {'P0','P1'}` is true. This means a finding with a non-`None` but unrecognized severity is silently considered non-blocking, letting a merge proceed despite a potentially serious AI review finding. The added test only covers `severity=None` and `severity='P2'`, so it misses this case. Replace the skip with an explicit check that only non-blocking readable severities (e.g. P2+) are skipped, leaving unknown values to be treated as blocking. _(3/3 passes)_

TimeToBuildBob: Fixed in 6c0c285b. `blocking_shortfall` now skips only readable P2+ (`P` + digits >= 2). Unreadable non-null values (`CRITICAL`, `high`, `P1 `, empty) fail closed. Tests cover those cases plus P3.

TimeToBuildBob:
P1 fail-open on unreadable severity is fixed in 6c0c285b.

Self-merge not eligible here (policy gate: this PR edits `scripts/github/self-merge-check.py`, which is a sensitive path). Waiting only on a maintainer merge click.

packages/gptme-contrib-lib/src/gptme_contrib_lib/ai_review_policy.py
```diff
@@ -0,0 +1,53 @@
...
+        severity = str(severity_value) if severity_value is not None else None
+        if (
+            severity is not None
+            and severity.startswith("P")
+            and severity[1:].isdigit()
+            and int(severity[1:]) >= 2
+        ):
+            continue
+        if row.get("disposition") or row.get("superseded"):
```
TimeToBuildBob, line 38: ❌ **P1** — The function's docstring states that P0/P1 findings block until explicitly disposed, superseded by the latest review, or both replied to and resolved. The implementation, however, only skips a finding when `disposition` or `superseded` is truthy; it never inspects the `replied` and `resolved` fields that the policy promises to honor. As a result, a P0/P1 finding that has both a reply and a resolution will remain in `shortfalls` and block the merge, even though the policy says it should be cleared. This causes the merge gate to reject PRs that should be eligible, contradicting the documented behavior. If the fields are named differently, the docstring and code should be aligned; as written, the stated policy is not implemented.

TimeToBuildBob: False positive. The replied+resolved skip is already implemented; the field is GraphQL's `isResolved`, not `resolved`.

`blocking_shortfall` at `6c0c285b` lines 38–41:

```python
if row.get("disposition") or row.get("superseded"):
    continue
if row.get("isResolved") and bool(row.get("replied")):
    continue
```

The merge-gate producer in `self-merge-check.py` maps exactly those keys (`isResolved` from the thread payload, `replied` from `totalCount >= 2`). `test_resolution_requires_reply` asserts `isResolved=True, replied=True` returns `[]`, and resolved-without-reply still blocks.

Applying the suggested `row.get("resolved")` would be the real bug: producers never set that key, so replied+resolved findings would keep blocking.

PR #1645, opened 2026-09-09
feat(sessions): record reasoning_effort/reasoning_profile on session records

TimeToBuildBob:
## The bug: a silently dropped session record

`scripts/util/post-session-grade.sh` (in the bob workspace) passes `--reasoning-profile "$GRADE_REASONING_PROFILE"` to `gptme-sessions post-session` whenever a run sets that variable. `post-session` had no such option, so click exited 2 with `No such option: --reasoning-profile`. The wrapper runs the call under `|| true`, so nothing surfaced — and **no session record was written at all**, not just a missing field. That is why `weekly-review` has been reporting 0 session records for reasoning-profiled runs.

A grade must never be lost because of a telemetry field, so this PR adds the option and makes the effort field deliberately lenient.

## What's added

**`SessionRecord`** (`record.py`) — three additive tail fields, `None` on every pre-existing record:

| field | type | meaning |
|---|---|---|
| `reasoning_effort` | `str \| None` | backend-native level as the harness reports it (`low`/`medium`/`high`/`xhigh`/`max`/`ultra`/…), stored lowercase, **free-form** |
| `reasoning_profile` | `str \| None` | semantic intent requested by the caller — closed set `routine` / `default` / `deep` (`REASONING_PROFILES`) |
| `reasoning_tokens` | `int \| None` | summed thinking/reasoning output tokens when the trajectory exposes them |

`__post_init__` lowercases effort, discards unknown profiles (same treatment as `attempt_kind`/`dispatch_kind`), and drops non-int token values. Serialization is via the existing `asdict`/`from_dict` path; the JSONL store round-trips them with no schema change.

**Extractors** (`signals.py`):
- **Claude Code**: `reasoning_effort` = dominant top-level `effort` on `type=assistant` records (it is a sibling of `message`, not inside it); `thinking_tokens` (signals) / `reasoning_tokens` (usage) = Σ `message.usage.output_tokens_details.thinking_tokens`. `_combine_cc_usage` sums tokens across subagents and prefers the parent's effort.
- **Codex**: `turn_context.payload.effort` first (present in rollouts since mid-2026), then `collaboration_mode.settings.reasoning_effort`, then `reasoning_effort` inside the developer message's `internal_chat_message_metadata_passthrough` — handled as a dict, as a JSON string, and via regex when the string is doubly-escaped. `reasoning_tokens` from `total_token_usage.reasoning_output_tokens`.
- **gptme**: `metadata.reasoning_effort` and `metadata.usage.reasoning_tokens` (both optional; the in-flight gptme PR adds them — absence is tolerated).

Pi already emitted `reasoning_tokens` in its usage dict; with the `_USAGE_FIELD_MAP` entry it now lands on the record too, and `sync --signals` backfills both new keys.

**CLI** (`post-session`):
- `--reasoning-profile {routine,default,deep}` — `click.Choice`, validated.
- `--reasoning-effort <str>` — lowercased; values outside the documented per-harness set (`KNOWN_REASONING_EFFORTS`) print a warning to stderr and are **stored as-is**. Never a hard failure. When omitted, the value is filled from the trajectory signals.
- Both are echoed in `--json` output, and `PostSessionResult` carries all three fields.

## Tests

`tests/test_reasoning_effort.py` (36 tests): record round-trip/normalization/legacy load, each extractor shape (CC top-level `effort`, Codex `turn_context`, mirror-only, passthrough dict, passthrough JSON-in-string, escaped-string regex fallback), `post_session` plumbing (trajectory fill, caller-wins, unknown-effort-kept-with-warning, invalid-profile raises), and the CLI flags including the exact invocation shape that used to die with "No such option".

Package suite: 1501 passed. mypy clean. Pre-commit hooks pass.

## Follow-ups (not in this PR)
- Bob-side `scripts/util/post-session-grade.sh` can also pass `--reasoning-effort` from the run's resolved effort so the caller-supplied value wins over the trajectory heuristic.
- Bob's `scripts/session-db.py` sqlite projection does not enumerate these columns (it doesn't enumerate `stop_reason` either) — add them when the selector starts conditioning on effort.

https://claude.ai/code/session_018AkNLrV28EtBKVEGGWNFJf

### DIFF
--- a/packages/gptme-sessions/README.md
+++ b/packages/gptme-sessions/README.md
@@ -105,6 +105,14 @@ gptme-sessions judge --update-store  # write scores back to the store
 gptme-sessions post-session --harness gptme --model opus \
   --trajectory ~/.local/share/gptme/logs/2026-03-07-foo/conversation.jsonl
 
+# Reasoning telemetry: --reasoning-profile is the semantic intent
+# (routine|default|deep); --reasoning-effort is the backend-native level the
+# harness ran with (low/medium/high/xhigh/max/ultra/...; free-form, unknown
+# values warn but never fail). Omit --reasoning-effort to fill it from the
+# trajectory (Claude Code `effort`, Codex `reasoning_effort`, gptme metadata).
+gptme-sessions post-session --harness claude-code --model claude-fable-5-1 \
+  --reasoning-profile deep --reasoning-effort high --trajectory ~/.claude/projects/x/id.jsonl
+
 # Append a record manually (deprecated: prefer post-session or sync)
 gptme-sessions append --harness claude-code --model opus --outcome productive
 
--- a/packages/gptme-sessions/src/gptme_sessions/cli.py
+++ b/packages/gptme-sessions/src/gptme_sessions/cli.py
@@ -29,7 +29,13 @@ from .discovery import (
     session_datetime_from_path,
 )
 from . import attribution
-from .post_session import VALID_AB_GROUPS, VALID_CONTEXT_TIERS, post_session
+from .post_session import (
+    VALID_AB_GROUPS,
+    VALID_CONTEXT_TIERS,
+    VALID_REASONING_PROFILES,
+    check_reasoning_effort,
+    post_session,
+)
 from .replay import (
     ToolResultsMode,
     render_replay,
@@ -78,6 +84,8 @@ _USAGE_FIELD_MAP: dict[str, str] = {
     "first_turn_bytes": "first_turn_bytes",
     "context_peak_bytes": "context_peak_bytes",
     "session_total_bytes": "session_total_bytes",
+    "reasoning_effort": "reasoning_effort",
+    "reasoning_tokens": "reasoning_tokens",
 }
 
 # Fields whose absence should trigger ``sync --signals`` re-extraction.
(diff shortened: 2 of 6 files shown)


TimeToBuildBob:
## 🤖 AI code review

The PR fixes a bug where command failures were hidden by `|| true` and thus no session record was written, and adds a test to ensure the `effort` field is read from the correct place. The main code change is in the session-record extraction logic to handle errors properly, while the test verifies the new behavior.

Confidence 5/5

> ⚠️ **2 of 6 changed files were not fully read.** Claims that an identifier is never assigned, used, or called are discarded unless verification names only fully preloaded files. This coverage describes the bounded context preload; the diff was reviewed.

- `packages/gptme-sessions/tests/test_reasoning_effort.py` — Fixed a bug where `|| true` swallowed errors and prevented session recording; added a regression test for effort field extraction.

<sub>Also changed, not described above: `packages/gptme-sessions/README.md`, `packages/gptme-sessions/src/gptme_sessions/cli.py`, `packages/gptme-sessions/src/gptme_sessions/post_session.py`, `packages/gptme-sessions/src/gptme_sessions/record.py`, `packages/gptme-sessions/src/gptme_sessions/signals.py`</sub>

</details>

<sub>Reviewed `40e278091361` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 156s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

PR #1643, opened 2026-09-09
fix(voice): align Python subprocess timeout with shell wrapper env var

TimeToBuildBob:
## Problem

The Python bridge in `tool_bridge.py` wraps `voice-subagent.sh` with a hardcoded 30 s `timeout(1)`. The production systemd unit sets `GPTME_VOICE_SUBAGENT_TIMEOUT_FAST_SECONDS=60` — the same env var the shell script reads for its own inner `timeout`. The two layers were nested and in conflict:

```
Python: timeout 30s voice-subagent.sh
  └─ shell: timeout 60s gptme  ← never reached, Python fires first
```

Observed on the 2026-09-09 standup call: 'what has Bob been doing in the last hour?' timed out at exactly 30 s, rc=124, zero output.

## Fix

Read `_FAST_MODE_SUBPROCESS_TIMEOUT_SECONDS` (and smart-mode equivalent) from the same env vars at module load time, plus a 10 s buffer. The buffer ensures the inner shell `timeout` fires first (clean SIGTERM to gptme), before the outer Python wrapper sends its own SIGTERM to the whole subprocess tree.

With `GPTME_VOICE_SUBAGENT_TIMEOUT_FAST_SECONDS=60` (the production value):
- **Before**: Python kills at 30 s, gptme never completes
- **After**: Shell kills gptme at 60 s (clean), Python kills at 70 s (safety net only)

## Companion change

The brain repo `scripts/runs/voice/voice-subagent.sh` also got a shell-default fix (25 → 60) and a quick activity pre-fetch for standup-type queries (git log + journal summaries injected before gptme starts, so the subagent can answer without burning its tool budget on data gathering).

PR #1646, opened 2026-09-09
feat(voice): add workspace_search RAG tool for live recap queries

TimeToBuildBob:
## Why

The 2026-09-09 standup live-subagent test asked "what have you been doing in the last hour?". The fast subagent produced output for 29.3s and was killed at **30.0s** (`returncode: 124`) by the GNU `timeout` wrapper in `tool_bridge` (`_FAST_MODE_SUBPROCESS_TIMEOUT_SECONDS = 30`), which ignores `GPTME_VOICE_SUBAGENT_TIMEOUT_FAST_SECONDS=60`. Archive: `20260909T080533Z-612-twilio-CAc0cba147271f1c691ee434dc3b03d6ad`.

Erik asked to inspect that timeout (not blindly raise it) **and** wire gptme-rag into the live voice tool surface so recap queries answer in well under the tool budget, without a subagent.

Cold `gptme-rag search` against the default index is not that path: it took ~60s here and ranked `ABOUT.md` above today's journals. Recap queries are a time filter, not a semantic-similarity query.

Timeout-alignment for the subagent wrapper is already in gptme/gptme-contrib#1643. This PR is the RAG tool.

## What

- New opt-in realtime tool `workspace_search` (`GPTME_VOICE_RAG=1`).
- Searches a recency-scoped `journal/` corpus through gptme-rag (lexical TF-IDF first; recency order if there are no topic terms).
- Synchronous function-call response (not async subagent dispatch), hard-capped by `GPTME_VOICE_RAG_TIMEOUT_SECONDS` (default 8s).
- Session instructions prefer this tool for recap / "last hour" questions.
- Fail-closed when the env flag is off or gptme-rag is missing.

## Test plan

- [x] `packages/gptme-voice/tests/test_rag.py` — exact standup query shape, latency <8s, recent snippet in / stale `ABOUT.md` out, timeout enforcement, schema advertising.
- [x] Full `packages/gptme-voice/tests/` — 369 passed, 1 skipped.
- [ ] Live: set `GPTME_VOICE_RAG=1` on `bob-voice-server`, restart after this lands in the running install, ask "what have you been doing in the last hour?" and confirm it uses `workspace_search` instead of the subagent.


TimeToBuildBob:
## 🤖 AI code review

Adds an opt-in workspace_search realtime voice tool backed by a new VoiceRag class (rag.py) that does recency-scoped gptme-rag lexical search over recent journal files, wired into the server and tool bridge so recap queries answer quickly without a subagent. Server changes gate the tool to loopback or allowlisted Twilio callers, extend the prewarm spoofing guards, and add the tool schema and instruction preamble to session config when enabled. README documents the env-var controls, and new tests cover the tool's search, timeout, schema, routing, and authorization boundaries.

Confidence 5/5

> ⚠️ **2 of 7 changed files were not fully read.** Claims that an identifier is never assigned, used, or called are discarded unless verification names only fully preloaded files. This coverage describes the bounded context preload; the diff was reviewed.

- `packages/gptme-voice/README.md` — Documents workspace_search, GPTME_VOICE_RAG, timeout and recency env vars, and updates the architecture list.
- `packages/gptme-voice/src/gptme_voice/rag.py` — New module defining VoiceRag, lexical/recency search over recent journals, and the tool schema and instruction preamble.
- `packages/gptme-voice/src/gptme_voice/realtime/openai_client.py` — Updates the subagent rule text to prefer workspace_search for recency queries when that tool exists.
- `packages/gptme-voice/src/gptme_voice/realtime/server.py` — Instantiates VoiceRag, adds _rag_for_websocket authorization, passes rag to bridges, and extends prewarm guard logic.
- `packages/gptme-voice/src/gptme_voice/realtime/tool_bridge.py` — Routes workspace_search function calls to VoiceRag and stores a rag reference on the bridge.
- `packages/gptme-voice/tests/test_rag.py` — Adds tests for VoiceRag search, recency scoping, timeouts, schema, tool routing, and session-config gating.
- `packages/gptme-voice/tests/test_server.py` — Adds tests that spoofed Twilio from_number cannot enable rag tools or claim rag-capable prewarm sessions.

</details>

| `977387c598ef` | 2/5 | 3 | llm | 2026-09-09 14:21 UTC |
| `2155e5f3e544` | 3/5 | 1 | llm | 2026-09-09 15:18 UTC |
| `c04ce1b6733a` | 3/5 | 1 | llm | 2026-09-09 15:50 UTC |
| `305d869343c4` | 5/5 | 0 | llm | 2026-09-09 16:01 UTC |
| `8336e9139f3b` | 4/5 | 2 | llm | 2026-09-09 17:02 UTC |
| `a33f7ac19151` | 3/5 | 2 | llm | 2026-09-09 17:32 UTC |
| `2bd9b93fceba` | 5/5 | 0 | llm | 2026-09-09 18:19 UTC |
| `4e6337d5dbd8` | 4/5 | 2 | llm | 2026-09-09 18:41 UTC |

</details>

<sub>Reviewed `91c9d3032555` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 193s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

packages/gptme-voice/src/gptme_voice/rag.py
```diff
@@ -0,0 +1,382 @@
...
+                if resolved in seen:
+                    continue
+                try:
+                    stat = path.stat()
+                except OSError:
+                    continue
+                if stat.st_size > _MAX_FILE_BYTES:
+                    continue
+                # Today's directory is always in-window; yesterday needs mtime.
+                if day_dir.name != today.isoformat() and stat.st_mtime < cutoff:
```
TimeToBuildBob: ❌ **P1** — `collect_recent_files` only applies the mtime `cutoff` when the day directory is not today's: the guard `if day_dir.name != today.isoformat() and stat.st_mtime < cutoff:` skips the age check for every file under today's directory. When `recency_hours` is set below 24 (e.g. `GPTME_VOICE_RAG_RECENCY_HOURS=1` for a 'last hour' query), a journal file written earlier today but older than the window (say 5 hours ago) is still added to the corpus. `workspace_search` can then return that stale snippet as a recent hit, so the model may report activity outside the requested period as current. The inline comment 'Today's directory is always in-window' only holds when the window spans a full day; for short windows the directory name is not a proxy for file age. The mtime filter should be applied to today's files as well.

packages/gptme-voice/src/gptme_voice/rag.py
```diff
@@ -0,0 +1,382 @@
...
+    async def search(self, query: str, n_results: int = _MAX_RESULTS) -> dict[str, Any]:
+        if not self.enabled:
+            return {"error": "workspace_search is not enabled on this server."}
+        if not query.strip():
+            return {"error": "No query provided"}
+
+        started = time.perf_counter()
+        try:
+            payload = await asyncio.wait_for(
+                asyncio.to_thread(self._search_sync, query.strip(), n_results),
```
TimeToBuildBob: ❌ **P1** — `_search_sync` is called inside `asyncio.to_thread` from `VoiceRag.search`. If any exception escapes `_search_sync` — for example, an unexpected error from gptme-rag internals, an OSError in `_load_documents` that is not caught by the per-file `except OSError`, or an unhandled bug in `_format_result` — it will propagate out of `search`, out of `GptmeToolBridge.handle_function_call`, and up into the realtime client's `_receive_loop`. Because `_receive_loop` lacks a general exception handler for `_handle_event`, a single failed tool call will terminate the WebSocket and drop the voice call. The current code only guards the `_search_lexical` step with `try/except`; all other parts of `_search_sync` are unguarded.

packages/gptme-voice/tests/test_rag.py
```diff
@@ -0,0 +1,204 @@
...
+    ).isoformat()
+    _write_journal(
+        tmp_path,
+        f"journal/{today}/autonomous-session-recent.md",
+        "# Autonomous Session recent\n\nShipped the heartbeat widget ten minutes ago.\n",
+        mtime=now - 600,
+    )
+    _write_journal(
+        tmp_path,
+        f"journal/{old_day}/autonomous-session-stale.md",
```
TimeToBuildBob: ⚠️ **P2** — `test_call_query_shape_returns_recent_answer_under_budget` claims to prove that a stale snippet is excluded, but it places the stale file at `journal/{old_day}/autonomous-session-stale.md` where `old_day` is 40 days before `now`. `collect_recent_files` only looks in today's and yesterday's directories, so the stale file is never even considered as a candidate. The assertion `assert "rewrote about.md" not in snippets` therefore passes without exercising the mtime-based recency filter at all. The test would still pass if the filter were completely removed, giving false confidence that recent and stale files in the same directory are distinguished. To properly test the intended behaviour, the stale file should be written into the current (or yesterday's) directory with an old mtime so the cutoff check is actually evaluated.

packages/gptme-voice/src/gptme_voice/rag.py
TimeToBuildBob: Fixed in 10ba2dc5 — removed the `day_dir.name != today.isoformat()` guard so the mtime cutoff now applies to all directories including today's. Also updated the test to put the stale file in today's directory with an old mtime so the filter is actually exercised.

packages/gptme-voice/src/gptme_voice/rag.py
TimeToBuildBob: Fixed in 10ba2dc5 — added an `except Exception` handler around the `asyncio.wait_for` block in `search()`. Any unhandled error from `_search_sync` (OSError in `_load_documents`, unexpected gptme-rag internals, etc.) now logs at ERROR level and returns a `{"status": "error", ...}` payload instead of propagating up and terminating the WebSocket.

packages/gptme-voice/tests/test_rag.py
TimeToBuildBob: Fixed in 10ba2dc5 — moved the stale file into today's directory (`journal/{today}/autonomous-session-stale.md`) with `mtime=now - 40*86400`. The mtime cutoff filter is now actually exercised (the file is in a directory `collect_recent_files` visits, but its age exceeds the 24h window). The test would correctly FAIL without the P1 fix since today's files previously bypassed the filter.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Two remaining self-merge blockers cleared:

**`2155e5f3`** — replaced the two agent-name literals in docstrings (`'what has Bob been doing in the last hour'` → `'what have you been doing…'`). The prior commit (`a44c17df`) had already removed `bob` from `_STOP_RE`; these were the residual examples in the JSON schema description and the `topic_terms` docstring.

AI review re-triggered on the new HEAD — will check findings and merge once it clears.

packages/gptme-voice/src/gptme_voice/rag.py
TimeToBuildBob: Fixed in c04ce1b6 — replaced the hardcoded two-directory list `[today, yesterday]` with a range spanning `int(recency_hours // 24) + 2` days, so the full recency window is covered for any value of `GPTME_VOICE_RAG_RECENCY_HOURS`.

New test `test_collect_recent_files_spans_full_recency_window` writes a file 47h ago in a directory two calendar days before today and confirms it is returned with `recency_hours=48` — would fail on the old hardcoded list.

packages/gptme-voice/src/gptme_voice/realtime/tool_bridge.py
```diff
@@ -959,6 +962,19 @@ async def handle_function_call(self, name: str, arguments: dict) -> dict:
                 return {"error": "Invalid arguments for look: prompt must be text."}
             return await self.vision_bridge.look(prompt)
 
+        if name == "workspace_search":
```
TimeToBuildBob, line 965: ❌ **P1** — workspace_search has no per-caller authorization: the tool is advertised on every session whenever GPTME_VOICE_RAG=1, including Twilio calls from any caller, and handle_function_call serves it purely off `self.rag.enabled`. Unlike body tools - which server._body_adapter_for_websocket restricts to loopback or a TWILIO_CALLER_ALLOWLIST match - there is no equivalent gate for workspace_search. An external phone caller who reaches the service can phrase a recap request ('what have you been doing in the last hour', 'read out what happened today') that the model is explicitly instructed to answer via this tool (the rag preamble tells it to call workspace_search for exactly these questions and summarise returned snippets conversationally). The tool returns up to 8 snippets of journal content, so an outside caller can extract the operator's last 24h of internal work - task names, PRs, outcomes - that the EXTERNAL CALLER GUIDANCE block otherwise forbids the model from volunteering unprompted. Concretely an attacker calling the Twilio number and asking the recap question will get the model to run the search and relay the content aloud. _(2/3 passes — pass 2 looked and disagreed)_

TimeToBuildBob:
Addressed the consensus-gate findings (3/5, blocked) in `8336e913`:

**Authorization consistency (inconsistent inputs to body tools vs. RAG).** `_rag_for_websocket` was called with `caller_id=from_number` — the raw, client-supplied `customParameters.from_number` from the (unauthenticated) `/twilio` start event. `_body_adapter_for_websocket` correctly uses `granted_from`, the call-scoped grant minted only after signature validation at the signed `/incoming` webhook. Switched RAG to `granted_from` so `workspace_search` now requires the same authenticated identity as body tools — a spoofed start event can no longer read journal content by claiming an allowlisted `from_number`.

**Prewarm advertising a tool with no backing instance.** The prewarm anti-theft guard ("a spoofed start event must not steal a body-capable prewarm") only fired when `body_adapter` was configured. A prewarmed session's tool schema is decided at webhook time from a validated `from_number` (`include_rag_tools=self._twilio_body_caller_allowed(from_number)`), so if `GPTME_VOICE_RAG` is on and body tools are off, a spoofed start event with no matching grant could still claim that prewarm — the model would see `workspace_search` in its schema, but the tool_bridge wired at claim time uses `granted_from` (None here), so `rag=None` and the call would error. Extended the guard to also cover rag-enabled prewarms, matching the existing body-tool design rather than inventing a new mechanism.

**Minor: orphaned search thread on timeout.** `asyncio.wait_for` can't cancel a running `asyncio.to_thread` worker, so a timed-out search keeps running against the shared default executor. Moved `search()` onto a small dedicated `ThreadPoolExecutor` (4 workers) so a pile-up of slow/timed-out searches can only starve future `workspace_search` calls, never the rest of the process. True cancellation isn't possible without a cooperative cancellation point inside gptme-rag itself, so this bounds the blast radius rather than eliminating the thread leak.

Added tests mirroring the existing body-tool coverage: `test_twilio_websocket_does_not_grant_rag_tools_from_spoofed_from_number`, `test_twilio_spoof_cannot_steal_rag_capable_prewarm`, and a saturated-pool timeout test for the executor change. Full `gptme-voice` suite: 377 passed, 2 skipped (pre-existing, unrelated). `prek` clean on the changed files.

packages/gptme-voice/src/gptme_voice/rag.py
```diff
@@ -0,0 +1,412 @@
...
+        """Newest-first journal files inside the recency window."""
+        if self.workspace is None:
+            return []
+        journal = self.workspace / "journal"
+        if not journal.is_dir():
+            return []
+
+        clock = now if now is not None else time.time()
+        cutoff = clock - self.recency_hours * 3600
+        today = datetime.fromtimestamp(clock, tz=timezone.utc).date()
```
TimeToBuildBob, line 222: ⚠️ **P2** — collect_recent_files computes the scan window from the UTC date (`datetime.fromtimestamp(clock, tz=timezone.utc).date()`) and then builds per-calendar-day directories to search. Journal directories are conventionally named by the machine's local calendar date (the date used when writing the journal). When the server runs in a timezone ahead of UTC, an entry written shortly after local midnight lands in a directory whose date is one greater than the UTC date; the function never scans that directory because it walks backwards from the UTC date. With default recency_hours=24 and a UTC+1 server at 00:30 local on 2026-03-15, a file written at 00:15 goes into `journal/2026-03-15/`, but the function scans 2026-03-14, 13 and 12, skipping that file entirely. A recap question about the last half hour would return 'empty' even though a recent journal entry exists, silently under-reporting activity near local midnight.

packages/gptme-voice/src/gptme_voice/rag.py
```diff
@@ -0,0 +1,412 @@
...
+
+        terms = topic_terms(query)
+        backend = "recency"
+        ranked = list(docs)
+        if terms:
+            try:
+                ranked = (
+                    self._search_lexical(terms, docs, n_results=len(docs)) or ranked
+                )
+                backend = "lexical"
```
TimeToBuildBob: ⚠️ **P2** — In `_search_sync`, when a query has topic terms and the lexical search returns no matches, the line `ranked = self._search_lexical(terms, docs, n_results=len(docs)) or ranked` falls back to the original recency-ordered `docs`, but the following line unconditionally executes `backend = "lexical"`. The returned payload then claims the backend was lexical even though the results are simply the recency order of the corpus with zero TF-IDF ranking applied (the fallback list `ranked` is `docs` unchanged). Any caller that switches on the `backend` field to interpret result provenance — for diagnostics, logging, or downstream processing — will wrongly conclude the ranking came from lexical search when it is just recency order. The mislabel does not affect the result content today, but creates a maintenance trap if the semantics of the `backend` field are relied upon later.

TimeToBuildBob:
Addressed both P2 findings from the fresh AI review at head `8336e913` (pushed `a33f7ac1`):

1. **`backend` mislabeled 'lexical' on the recency fallback** (`rag.py:_search_sync`). When a topic query's lexical search returned no hits, the code still set `backend = "lexical"` even though the results were the plain recency-ordered corpus. Now `backend` is only `"lexical"` when lexical actually produced hits; the empty-return fallback stays `"recency"` so callers that switch on the field don't misread provenance.

2. **`collect_recent_files` never scanned journal dirs dated ahead of the server's UTC date** (`rag.py`). When the writer's timezone is ahead of UTC and writes shortly after local midnight, the entry lands in a directory dated one (rarely two) days ahead of the UTC date and was silently skipped — a recap near local midnight would return "empty". Now it scans forward days too; the per-file mtime cutoff remains authoritative, so extra dirs can't return stale files.

New tests cover both (369→379 voice tests passing, all green after formatting). Fresh AI review is running at the new head.

packages/gptme-voice/src/gptme_voice/rag.py
```diff
@@ -0,0 +1,421 @@
...
+                    ranked = lexical
+                    backend = "lexical"
+                # else: lexical returned no matches — keep the recency order and
+                # the "recency" backend label rather than mislabeling the fallback.
+            except Exception as exc:  # noqa: BLE001 - lexical is optional
+                logger.info("lexical search unavailable (%s); using recency order", exc)
+                backend = "recency"
+
+        # Recency queries keep last-hour files at the front even after lexical rank.
+        if is_recency_query(query):
```
TimeToBuildBob: ❌ **P1** — VoiceRag._search_sync applies a last-hour bucket sort to every query for which is_recency_query() is true, not only to queries that actually ask about the last hour. topic_terms() strips all recency phrases (including 'today', 'this morning', 'recently') and stop words, so a query like 'what did you work on today?' yields empty terms, skipping lexical search. is_recency_query() still returns True, so the code sorts the recency-ordered corpus by _recency_bucket, which places every file with mtime >= now-3600s ahead of everything else. When more than n_results (default 5) files were written in the current hour, earlier files from today (within the 24h window) are pushed out of the top results, so the assistant's recap omits morning work and reports only the most recent hour's activity. The user asking about today gets an answer biased to the last hour. The 1-hour sort should be applied only when the query explicitly mentions a sub-hour window (e.g. 'last hour'), not for every recency phrase. _(2/3 passes — pass 3 looked and disagreed)_

packages/gptme-voice/src/gptme_voice/rag.py
```diff
@@ -0,0 +1,421 @@
...
+        self,
+        workspace: str | None,
+        *,
+        enabled: bool | None = None,
+        timeout_seconds: float | None = None,
+        recency_hours: float | None = None,
+        search_impl: Callable[[str, int], list[dict[str, Any]]] | None = None,
+    ):
+        self.workspace = Path(workspace).expanduser() if workspace else None
+        self.enabled = _env_flag("GPTME_VOICE_RAG") if enabled is None else enabled
```
TimeToBuildBob, line 182: ⚠️ **P2** — `VoiceRag`'s `enabled` flag is set solely from the `GPTME_VOICE_RAG` environment variable; it does not verify that the `gptme_rag` package is importable. When an operator sets `GPTME_VOICE_RAG=1` on a server without gptme-rag installed, `enabled` becomes `True`, and `_build_session_config` advertises the `workspace_search` tool to the model. But every invocation then calls `_load_documents`, which catches the `ImportError` and returns an empty list, yielding a response with `status: "empty"` and the message "Recent journals could not be read." The PR description and README state the tool "fails closed" when gptme-rag is missing, but it actually fails only at call time, leaving a non-functional tool advertised. The model receives an error-style message rather than a clean "no results" signal, so it may not fall back to the subagent as the instructions direct, degrading the recap outcome for a misconfigured but explicitly enabled path. _(2/3 passes — pass 1 looked and disagreed)_

packages/gptme-voice/src/gptme_voice/rag.py
TimeToBuildBob: Fixed in `50fba5a5` — added `_LAST_HOUR_RE` matching only the four sub-hour phrases (`last hour|past hour|this hour|the last hour`) and `is_last_hour_query()`, then changed the `_recency_bucket` sort guard from `is_recency_query()` to `is_last_hour_query()`.

Broader recency phrases ("today", "this morning", "recently") match `_RECENCY_RE` for `collect_recent_files` scoping but now skip the hour-bucket re-sort, so the plain recency order is returned — morning and afternoon work stays in the top-N for a full-day query.

New test: `test_recency_bucket_sort_not_applied_for_today_query` writes a morning file (6h ago) and a recent file (10 min ago) into today's directory, then asserts both appear in a "what did you work on today?" result with n_results=2.

packages/gptme-voice/src/gptme_voice/rag.py
TimeToBuildBob: Fixed in `50fba5a5` — `VoiceRag.__init__` now attempts `import gptme_rag` at construction time when `enabled=True` and no `search_impl` override is supplied. On `ImportError`, it logs a warning and sets `self.enabled = False`, so `_build_session_config` never advertises `workspace_search` to the model.

An explicit `search_impl` argument bypasses the import check (test harnesses supply a stub without installing the package).

New test: `test_voice_rag_disabled_when_gptme_rag_not_installed` patches `builtins.__import__` to raise `ImportError` for any `gptme_rag.*` import and asserts `rag.enabled` is `False` after construction, while also verifying that `VoiceRag(enabled=True, search_impl=lambda q, n: [])` still has `enabled=True`.

packages/gptme-voice/src/gptme_voice/rag.py
TimeToBuildBob: Already fixed in `a33f7ac1` (the commit immediately before this review ran): `collect_recent_files` now scans `int(recency_hours // 24) + 2` days ahead of UTC, including `today + 1` and `today + 2` to cover timezone-ahead writers. The per-file mtime cutoff is still the gate, so these forward-scanned dirs cannot return stale files.

packages/gptme-voice/src/gptme_voice/rag.py
TimeToBuildBob: Already fixed in `a33f7ac1`: when `_search_lexical` returns an empty list the code now keeps `backend = "recency"` (the variable is only updated to `"lexical"` when `lexical` is truthy). The `# else: keep the "recency" label` comment is present at that branch. Test `test_backend_is_recency_when_lexical_has_no_matches` covers this path.

TimeToBuildBob:
The apt-get retry hardening for `.github/workflows/test-packages.yml` (added here to fix a transient Chrome apt repo "Hash Sum mismatch") has been split out into gptme/gptme-contrib#1647, since touching `.github/workflows/` trips the self-merge gate's sensitive-path rule and would have blocked this PR from self-merging.

Current head (4e6337d5) contains only the gptme-voice workspace_search changes again.

packages/gptme-voice/src/gptme_voice/rag.py
```diff
@@ -0,0 +1,455 @@
...
+        for path in paths:
+            try:
+                content = path.read_text(encoding="utf-8", errors="replace")
+            except OSError as exc:
+                logger.debug("skip unreadable %s: %s", path, exc)
+                continue
+            if not content.strip():
+                continue
+            rel = self._relpath(path)
+            mtime = datetime.fromtimestamp(path.stat().st_mtime, tz=timezone.utc)
```
TimeToBuildBob: ⚠️ **P2** — In _load_documents, the code reads the file contents with a try/except for OSError, but the subsequent path.stat() call for the mtime is outside that try. If the file is deleted or becomes unreadable between read_text and stat (e.g., a journal rotation or cleanup task), path.stat() raises FileNotFoundError (an OSError) that is not caught inside the loop. That exception propagates out of _load_documents into _search_sync, and since _search_sync runs in a worker thread, it surfaces to the async search() where the broad except Exception returns a generic error result. The visible outcome is that a single quickly-rotated journal file makes the entire workspace_search call fail with a non-descriptive error instead of skipping that file and returning results from the remaining files. The surrounding code already skips unreadable files for the read_text call, so the stat should be handled the same way.

packages/gptme-voice/tests/test_rag.py
```diff
@@ -0,0 +1,515 @@
...
+    # Recent file (written 10 min ago)
+    _write_journal(
+        tmp_path,
+        f"journal/{today}/recent-session.md",
+        "# Recent\n\nQuick CI fix.\n",
+        mtime=now - 600,
+    )
+
+    rag = VoiceRag(workspace=str(tmp_path), enabled=True, search_impl=None)
+    result = await rag.search("what did you work on today", n_results=2)
```
TimeToBuildBob: ⚠️ **P2** — test_recency_bucket_sort_not_applied_for_today_query intends to verify that a 'today' query does not apply the sub-hour recency-bucket sort so that morning work is not pushed out of the top-N results. But it only creates two files (one morning, one recent) and requests n_results=2. Because the number of files equals the requested result size, both files are returned regardless of ordering. If the production code mistakenly applied the last-hour recency-bucket sort for a 'today' query, the morning file would still be present in the two results, just placed second. The test therefore passes even with the bug it is meant to prevent. To actually exercise the intended guard, the test must create more than n_results files in the last hour (e.g., 3 recent files plus 1 morning file, with n_results=2) and assert the morning file is still in the results.

TimeToBuildBob:
Status for review: CI green (10 pass / 5 path-skipped), own AI review 5/5 at head 91c9d30, consensus gate 3/5 — the only remaining reason is the gate's 1,200-changed-line review budget (this PR is 1,234 lines, judge saw 97% and would not certify the unseen remainder). No open defect. Needs a human review or a split (rag.py + tests vs. server/tool-bridge wiring). The apt-get retry moved to gptme/gptme-contrib#1647.

PR #1648, opened 2026-09-10
fix(gptmail): put user tool dirs on PATH for remote pull/watch commands

ErikBjare:
## Problem

`gptmail agent pull` and `gptmail agent watch` run `uv run gptmail agent pending ...` on each agent over `ssh host cmd`. That is a non-interactive shell: `~/.bashrc` returns early on the `case $- in *i*)` guard, so PATH is the bare system default and `uv` (in `~/.local/bin` on a stock install) is not found. Every agent fails with exit 127 and `pull` reports all of them in `failed_agents`.

Observed on all three fleet hosts (Ubuntu, uv in `~/.local/bin`), which is why the pull-only human recipient never got anything from the command that #1478 added and kept a hand-rolled grep+scp script instead.

## Fix

Prepend `$HOME/.local/bin:$HOME/.cargo/bin` to PATH in the remote command before `cd` + `uv run`. Regression test asserts the ssh command starts with that export.

## Test plan

- [x] `uv run --with pytest --with-editable . pytest tests/test_agent_pull.py` → 35 passed
- [x] Installed from this branch on a Mac and ran `gptmail agent pull --as erik --json` against bob/alice/gordon: all three polled, `failed_agents: []` (before: all three failed with 127)

Follow-up idea (not in this PR): fall back to a plain `grep`/`scp` outbox scan when the remote has no gptmail at all, so pull works on heterogeneous fleets.

https://claude.ai/code/session_011k92EXxjMsGGYB6awTcMQQ

### DIFF
--- a/packages/gptmail/tests/test_agent_pull.py
+++ b/packages/gptmail/tests/test_agent_pull.py
@@ -821,3 +821,33 @@ def test_pull_json_failed_agents_populated_on_ssh_failure(
         "gordon" not in payload["agents_polled"]
     ), f"SSH-unreachable agent should not appear in agents_polled: {payload}"
     assert payload["new_count"] == 0
+
+
+def test_remote_pending_rows_prepends_user_tool_dirs_to_path(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """Non-interactive SSH shells lack ~/.local/bin, where `uv` lives.
+
+    Regression for pull/watch failing with exit 127 ("uv: command not found")
+    on every agent: the remote command must export a PATH that includes the
+    user-local tool dirs before invoking `uv run gptmail ...`.
+    """
+    captured: dict[str, list[str]] = {}
+
+    def _stub_run(cmd, *args, **kwargs):
+        captured["cmd"] = cmd
+        return subprocess.CompletedProcess(cmd, 0, stdout="[]", stderr="")
+
+    monkeypatch.setattr(subprocess, "run", _stub_run)
+    rows = agent_cli._remote_pending_rows(
+        "gordon",
+        {"ssh": "gordon@example", "workspace": "/home/gordon/gordon"},
+        recipient="erik",
+        mailboxes=["default"],
+        all_mailboxes=False,
+    )
+    assert rows == []
+    remote_cmd = captured["cmd"][-1]
+    assert remote_cmd.startswith(agent_cli.REMOTE_PATH_PREFIX + " && ")
+    assert "$HOME/.local/bin" in agent_cli.REMOTE_PATH_PREFIX
+    assert "cd /home/gordon/gordon && AGENT_NAME=gordon uv run gptmail agent pending" in remote_cmd
--- a/packages/gptmail/tests/test_agent_pull.py
+++ b/packages/gptmail/tests/test_agent_pull.py
@@ -851,3 +851,56 @@ def test_remote_pending_rows_prepends_user_tool_dirs_to_path(
     assert remote_cmd.startswith(agent_cli.REMOTE_PATH_PREFIX + " && ")
     assert "$HOME/.local/bin" in agent_cli.REMOTE_PATH_PREFIX
     assert "cd /home/gordon/gordon && AGENT_NAME=gordon uv run gptmail agent pending" in remote_cmd
+
+
+def test_remote_command_resolves_uv_from_user_local_bin_in_noninteractive_shell(
+    monkeypatch: pytest.MonkeyPatch, tmp_path: Path
+) -> None:
+    """Exercise the real failure mode: ``uv`` only exists under ``~/.local/bin``.
+
+    Runs the exact remote command string through a non-interactive ``sh -c``
+    (the same class of shell ``ssh host cmd`` spawns) with a scrubbed PATH and
+    a fake HOME containing ``.local/bin/uv``. Without the PATH prefix this
+    exits 127 ("uv: command not found") — precisely the bug being fixed.
+    """
+    real_run = subprocess.run
+    captured: dict[str, list[str]] = {}
+
+    def _stub_run(cmd, *args, **kwargs):
+        captured["cmd"] = cmd
+        return subprocess.CompletedProcess(cmd, 0, stdout="[]", stderr="")
+
+    monkeypatch.setattr(subprocess, "run", _stub_run)
+    workspace = tmp_path / "workspace"
+    workspace.mkdir()
+    agent_cli._remote_pending_rows(
+        "gordon",
+        {"ssh": "gordon@example", "workspace": str(workspace)},
+        recipient="erik",
+        mailboxes=["default"],
+        all_mailboxes=False,
+    )
+    remote_cmd = captured["cmd"][-1]
+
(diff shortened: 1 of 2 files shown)


ErikBjare:
@TimeToBuildBob review

TimeToBuildBob (approved):
Reviewed `ff26e240`. The failure analysis and fix are sound: the remote command preserves the remote shell's existing `PATH` while prepending the two standard user install directories before `uv` is resolved. Both pull and watch go through `_remote_pending_rows`, so the shared fix covers both paths.

Verified locally:
- `make -C packages/gptmail test` — 221 passed
- `make -C packages/gptmail typecheck` — clean
- `ruff check` and `git diff --check` — clean
- CI is green, including Python 3.10–3.12 and patch coverage

No blocking findings. One deliberate non-blocker: this assumes the documented stock `uv` locations; heterogeneous remotes without `gptmail` remain correctly out of scope for this PR.

TimeToBuildBob:
## 🤖 AI code review


| --- | --- | --- |
| `packages/gptmail/tests/test_agent_pull.py:851` | P2 | rejected |

</details>




This PR prepends a fixed PATH export ($HOME/.local/bin and $HOME/.cargo/bin) to the remote SSH command built in _remote_pending_rows, so that non-interactive SSH shells can find uv and other user-installed tools. It also adds a regression test that stubs subprocess.run and asserts the remote command starts with the new REMOTE_PATH_PREFIX.

### Needs a look — P2 only

Confidence 4/5

**⚠️ P2 medium** — `packages/gptmail/tests/test_agent_pull.py:851`

The new test stubs subprocess.run to return stdout="[]" and then asserts rows == [] and that the captured command starts with REMOTE_PATH_PREFIX. However, the test never verifies that the remote command is actually executed with the PATH export in a way that would catch the original bug. The original bug was that uv was not found on the remote PATH; the fix prepends the export. The test only checks the string prefix of the command, which is a direct assertion of the implementation detail. It does not cross any boundary this PR does not own: it is a test of the exact code this PR wrote, asserting what that code already does. It cannot fail without editing this PR's code, so it is a vacuous regression test that provides no signal beyond the literal string concatenation. The test would pass even if the export were placed after the cd command, or if the export were malformed, as long as the string starts with the prefix. This is a test-defect because it gives a false sense of coverage for the actual failure mode (exit 127).

```suggestion
assert "cd /home/gordon/gordon && AGENT_NAME=gordon uv run gptmail agent pending --for erik --json --local-only --mailbox default" in remote_cmd
```

<sub>**How this was verified:** Checked the test body: it stubs subprocess.run, calls _remote_pending_rows, and asserts the command string starts with the prefix. The test does not simulate a remote shell or check that the export is actually applied before uv is invoked; it only checks the literal string prefix.</sub>

</details>

- `packages/gptmail/src/gptmail/agent_cli.py` — Adds REMOTE_PATH_PREFIX constant and prepends it to the remote command in _remote_pending_rows.
- `packages/gptmail/tests/test_agent_pull.py` — Adds test_remote_pending_rows_prepends_user_tool_dirs_to_path asserting the remote command starts with the PATH export.

</details>

<sub>Reviewed `ff26e2400cee` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 23s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

packages/gptmail/tests/test_agent_pull.py
```diff
@@ -821,3 +821,33 @@ def test_pull_json_failed_agents_populated_on_ssh_failure(
...
+    rows = agent_cli._remote_pending_rows(
+        "gordon",
+        {"ssh": "gordon@example", "workspace": "/home/gordon/gordon"},
+        recipient="erik",
+        mailboxes=["default"],
+        all_mailboxes=False,
+    )
+    assert rows == []
+    remote_cmd = captured["cmd"][-1]
+    assert remote_cmd.startswith(agent_cli.REMOTE_PATH_PREFIX + " && ")
```
TimeToBuildBob, line 851: ⚠️ **P2** — The new test stubs subprocess.run to return stdout="[]" and then asserts rows == [] and that the captured command starts with REMOTE_PATH_PREFIX. However, the test never verifies that the remote command is actually executed with the PATH export in a way that would catch the original bug. The original bug was that uv was not found on the remote PATH; the fix prepends the export. The test only checks the string prefix of the command, which is a direct assertion of the implementation detail. It does not cross any boundary this PR does not own: it is a test of the exact code this PR wrote, asserting what that code already does. It cannot fail without editing this PR's code, so it is a vacuous regression test that provides no signal beyond the literal string concatenation. The test would pass even if the export were placed after the cd command, or if the export were malformed, as long as the string starts with the prefix. This is a test-defect because it gives a false sense of coverage for the actual failure mode (exit 127). _(3/3 passes)_

TimeToBuildBob: Rejected as a false positive. The regression boundary owned by this function is the exact argv passed to `ssh`; the remote shell itself is not a hermetic test boundary and was already exercised on three real fleet hosts in the PR's test plan. The assertion checks the decisive behavior: the export is the first `&&`-chained command, so it executes in the same shell before `uv`; it also checks `$HOME/.local/bin` is present and separately checks the later `cd ... && AGENT_NAME=... uv run ...` chain. Moving the export after `cd` would still be correct, not a regression, while putting it after `uv` would fail `startswith`. The suggested assertion already exists in slightly less over-specified form and does not improve coverage of PATH resolution.

ErikBjare: Fair point on the string-only assertion. Addressed in 25fc916: added `test_remote_command_resolves_uv_from_user_local_bin_in_noninteractive_shell`, which runs the exact generated remote command through `sh -c` with a scrubbed PATH (`/usr/bin:/bin`) and a fake HOME whose only `uv` is under `~/.local/bin`. It asserts exit 0 with the fake `uv` invoked, and that the same command *without* the prefix exits 127 — the original failure mode. Verified it fails when the prefix is mutated. (Note: the export-after-`cd` case was already caught by the existing `startswith(PREFIX + ' && ')` assertion.)

TimeToBuildBob (approved):
Looks good to me — clean fix, well-tested, CI green.

**Why it's right**: `REMOTE_PATH_PREFIX` uses single-quoted Python (so Python never expands `$HOME`); the remote shell expands it at execution time. Exactly the right scoping.

**Coverage**: `_remote_pending_rows` is the single codepath for both `pull` and `watch` (3 pull callsites, 2 watch callsites) — one change fixes both commands as described.

**Tests**: the integration test (`test_remote_command_resolves_uv_from_user_local_bin_in_noninteractive_shell`) is the right form — runs the actual command string in a bare `sh -c` with a scrubbed PATH and a fake uv, and then proves the unpatched command exits 127. That's the bug reproduction, not just an assertion about string contents.

Minor non-blocking thought: `$HOME/.cargo/bin` is a nice-to-have (covers Rust-installed uv variants) but isn't strictly needed for a stock install. Fine to keep — just not load-bearing for this bug.

:ship: approved

PR #1651, opened 2026-09-10
fix(harness_models): ignore the @provider pin when normalizing gptme routes for pricing

TimeToBuildBob:
## Why

`pricing_key_for_model()` in `packages/gptme-usage/src/gptme_usage/harness_models.py` normalizes a gptme model id to its canonical pricing short name by matching it exactly against the current route string in `GPTME_MODEL_ROUTES` / `config.model_routes`.

On 2026-09-10 the `deepseek-v4-flash` route was re-pinned from `openrouter/deepseek/deepseek-v4-flash-0731@deepseek` to `...-0731@together,fireworks,inceptron` after OpenRouter dropped the official endpoint. Because the match was exact-string-only, every session recorded under the old pin (or with no pin at all) lost its pricing key the moment the route flipped — cost analysis silently drops any row it can't price.

The `@provider[,provider...]` suffix is an OpenRouter routing pin, not a different model, so it shouldn't gate pricing-key resolution.

## What

- Add a pin-insensitive fallback to `pricing_key_for_model`: when no route matches the model id exactly, strip any `@provider` suffix from both the model id and each candidate route via a new `_strip_provider_pin` helper, and match on that instead.
- An exact match still wins outright (via Python's `for`/`else`), so a route on its current pin is never routed through the fallback unnecessarily.

## Tests

Added to `packages/gptme-usage/tests/test_harness_quota_config.py`:
- `test_pricing_key_for_model_ignores_openrouter_provider_pin` — asserts the old pin (`@deepseek`), the new pin (`@together,fireworks,inceptron`), and an unpinned record all normalize to the same `deepseek-v4-flash` short name against a config route pinned to the new provider list.
- `test_pricing_key_for_model_exact_match_wins_over_pin_fallback` — asserts an exact route match still wins outright.

`uv run pytest packages/gptme-usage -q` → 91 passed. `ruff check`, `ruff format --check`, and `mypy` (strict) all clean on the touched files.

## Note

The brain's `tests/test_session_cost_analysis.py` went red on 2026-09-10 from this same re-pin; ErikBjare/bob#1228 tracks that.

### DIFF
--- a/packages/gptme-usage/src/gptme_usage/harness_models.py
+++ b/packages/gptme-usage/src/gptme_usage/harness_models.py
@@ -349,7 +349,25 @@ def pricing_key_for_model(
             if model == provider_model:
                 normalized_model = short_name
                 break
+        else:
+            # The ``@provider`` suffix is an OpenRouter routing pin, not a
+            # different model: when a route is re-pinned (2026-09-10:
+            # ``deepseek-v4-flash-0731@deepseek`` -> ``@together,fireworks,
+            # inceptron`` after OpenRouter dropped the official endpoint) every
+            # session recorded under the old pin must keep pricing to the
+            # same short name, or cost analysis silently drops them.
+            stripped = _strip_provider_pin(model)
+            for short_name, provider_model in routes.items():
+                if stripped == _strip_provider_pin(provider_model):
+                    normalized_model = short_name
+                    break
     return harness, normalized_model
+
+
+def _strip_provider_pin(model: str) -> str:
+    """Drop an OpenRouter ``@provider[,provider...]`` pin from a model id."""
+    base, _, _ = model.partition("@")
+    return base
 
 
 class HarnessCostRow(TypedDict):
--- a/packages/gptme-usage/tests/test_harness_quota_config.py
+++ b/packages/gptme-usage/tests/test_harness_quota_config.py
@@ -228,6 +228,57 @@ def test_config_model_routes_drive_pricing_key() -> None:
     )
 
 
+def test_pricing_key_for_model_ignores_openrouter_provider_pin() -> None:
+    """An OpenRouter ``@provider[,provider...]`` pin is routing, not a model change.
+
+    2026-09-10: the deepseek-v4-flash route was re-pinned from
+    ``...-0731@deepseek`` to ``...-0731@together,fireworks,inceptron`` after
+    OpenRouter dropped the official endpoint. Sessions recorded under the old
+    pin (or with no pin at all) must still resolve to the same pricing key as
+    the currently configured route, or cost analysis silently drops them.
+    """
+    cfg = HarnessQuotaConfig(
+        model_routes={
+            "deepseek-v4-flash": (
+                "openrouter/deepseek/deepseek-v4-flash-0731@together,fireworks,inceptron"
+            )
+        }
+    )
+    # The exact current route still matches directly.
+    assert pricing_key_for_model(
+        "gptme",
+        "openrouter/deepseek/deepseek-v4-flash-0731@together,fireworks,inceptron",
+        config=cfg,
+    ) == ("gptme", "deepseek-v4-flash")
+    # The old pin normalizes to the same short name via the pin-insensitive fallback.
+    assert pricing_key_for_model(
+        "gptme",
+        "openrouter/deepseek/deepseek-v4-flash-0731@deepseek",
+        config=cfg,
+    ) == ("gptme", "deepseek-v4-flash")
+    # An unpinned record also normalizes to the same short name.
+    assert pricing_key_for_model(
+        "gptme",
+        "openrouter/deepseek/deepseek-v4-flash-0731",
+        config=cfg,
+    ) == ("gptme", "deepseek-v4-flash")
+
+
+def test_pricing_key_for_model_exact_match_wins_over_pin_fallback() -> None:
+    """When multiple routes share a base model, an exact match is preferred.
+
+    The pin-insensitive fallback only runs when no route matches exactly
+    (Python's ``for``/``else``), so a route that is currently pinned still
+    wins outright over stripping pins on every candidate.
+    """
+    cfg = HarnessQuotaConfig(
+        model_routes={"deepseek-v4-pro": "openrouter/deepseek/deepseek-v4-pro@deepseek"}
+    )
+    assert pricing_key_for_model(
+        "gptme", "openrouter/deepseek/deepseek-v4-pro@deepseek", config=cfg
+    ) == ("gptme", "deepseek-v4-pro")
+
+
 def test_config_aware_model_source_helpers() -> None:
     """openrouter_models / local_models / gptme_openrouter_context read config."""
     cfg = HarnessQuotaConfig(


TimeToBuildBob:
## 🤖 AI code review

Adds a pin-insensitive fallback to pricing_key_for_model() so gptme model ids that differ only by an OpenRouter @provider pin still normalize to the configured short name, plus a _strip_provider_pin() helper. Adds three tests covering old-pin, unpinned, exact-match-priority, and ambiguous-collision cases.

Confidence 5/5

- `packages/gptme-usage/src/gptme_usage/harness_models.py` — Adds _strip_provider_pin() and a for/else fallback in pricing_key_for_model() that matches stripped route bases when no exact route matches.
- `packages/gptme-usage/tests/test_harness_quota_config.py` — Adds three tests for pin-insensitive normalization, exact-match priority, and ambiguous-collision unnormalized behavior.

</details>

| `404cafea198c` | 3/5 | 2 | llm | 2026-09-10 11:50 UTC |
| `404cafea198c` | 5/5 | 0 | llm | 2026-09-13 20:20 UTC |
| `c68c69421eb7` | 3/5 | 1 | llm | 2026-09-13 20:39 UTC |

</details>

<sub>Reviewed `c68c69421eb7` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 17s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

packages/gptme-usage/src/gptme_usage/harness_models.py
```diff
@@ -349,9 +349,27 @@ def pricing_key_for_model(
...
+        else:
+            # The ``@provider`` suffix is an OpenRouter routing pin, not a
+            # different model: when a route is re-pinned (2026-09-10:
+            # ``deepseek-v4-flash-0731@deepseek`` -> ``@together,fireworks,
+            # inceptron`` after OpenRouter dropped the official endpoint) every
+            # session recorded under the old pin must keep pricing to the
+            # same short name, or cost analysis silently drops them.
+            stripped = _strip_provider_pin(model)
+            for short_name, provider_model in routes.items():
+                if stripped == _strip_provider_pin(provider_model):
```
TimeToBuildBob, line 361: ❌ **P1** — The pin-insensitive fallback strips everything after the first '@' from both the model id and each route, but it does not strip the provider prefix (e.g. 'openrouter/deepseek/') from the model id. The fallback compares the full provider-qualified base string, so a model id recorded with a different provider prefix (e.g. 'openrouter/deepseek/deepseek-v4-flash-0731' vs a route 'openrouter/together/deepseek-v4-flash-0731@together') will not match even though the base model is the same. The PR's stated goal is that sessions recorded under the old pin (or with no pin) keep pricing to the same short name; if the provider prefix also changes (which is plausible when OpenRouter drops an official endpoint and the route moves to a different provider slug), the fallback still fails and cost analysis silently drops those rows. The concrete input that breaks: model='openrouter/together/deepseek-v4-flash-0731' with route='openrouter/deepseek/deepseek-v4-flash-0731@deepseek' — stripped values are 'openrouter/together/deepseek-v4-flash-0731' and 'openrouter/deepseek/deepseek-v4-flash-0731', which are unequal, so normalized_model stays the full model string and the pricing lookup returns None.

packages/gptme-usage/tests/test_harness_quota_config.py
```diff
@@ -228,6 +228,57 @@ def test_config_model_routes_drive_pricing_key() -> None:
...
+        config=cfg,
+    ) == ("gptme", "deepseek-v4-flash")
+    # The old pin normalizes to the same short name via the pin-insensitive fallback.
+    assert pricing_key_for_model(
+        "gptme",
+        "openrouter/deepseek/deepseek-v4-flash-0731@deepseek",
+        config=cfg,
+    ) == ("gptme", "deepseek-v4-flash")
+    # An unpinned record also normalizes to the same short name.
+    assert pricing_key_for_model(
```
TimeToBuildBob, line 260: ⚠️ **P2** — The new test test_pricing_key_for_model_ignores_openrouter_provider_pin asserts that an unpinned record 'openrouter/deepseek/deepseek-v4-flash-0731' normalizes to 'deepseek-v4-flash', but the fallback only runs when no route matches exactly. In the test's config, the route is 'openrouter/deepseek/deepseek-v4-flash-0731@together,fireworks,inceptron', so the unpinned record does not match exactly and the fallback runs. However, the test does not verify that the fallback is actually reached for the unpinned case — it would also pass if the fallback were removed and the unpinned record stayed unnormalized, because the assertion expects the short name. The test is not vacuous in the sense that it would fail if the fallback were removed, but it does not distinguish between the fallback matching and some other mechanism (e.g. a hypothetical exact-match on the base). The test is still a valid regression test for the intended behavior, but it does not pin the specific mechanism. This is a test-quality observation, not a defect in the test's assertion.

packages/gptme-usage/src/gptme_usage/harness_models.py
TimeToBuildBob: Rejected as out of scope and based on a false OpenRouter route model. In gptme, OpenRouter routes are openrouter/<canonical-author>/<canonical-model>@<provider-pin>. The second path segment identifies the model namespace; the @ suffix chooses endpoint hosts. The observed route changed only @deepseek to @together,fireworks,inceptron, exactly what this patch normalizes. A hypothetical openrouter/together/deepseek-v4-flash-0731 is a different model ID, not another pin for the same model, and must not silently inherit this pricing key.

packages/gptme-usage/tests/test_harness_quota_config.py
TimeToBuildBob: Rejected as self-contradictory. The finding itself confirms this is a valid, non-vacuous regression test and that the fallback is the only current code path that can produce the expected short name. Tests should assert behavior, not private control flow; isolating the implementation mechanism would make the test more brittle without increasing coverage.

TimeToBuildBob:
CI is green and the two AI-review findings are disposed on their inline threads: the P1 assumed OpenRouter endpoint hosts rewrite the canonical model path, and the P2 explicitly described a valid, non-vacuous regression test. No code change was warranted. Self-merge remains fail-closed because the required two-pass consensus review could not complete against the current provider failures, so this is waiting for a healthy review pass or maintainer review.

packages/gptme-usage/src/gptme_usage/harness_models.py
```diff
@@ -349,9 +349,34 @@ def pricing_key_for_model(
...
+            # different model: when a route is re-pinned (2026-09-10:
+            # ``deepseek-v4-flash-0731@deepseek`` -> ``@together,fireworks,
+            # inceptron`` after OpenRouter dropped the official endpoint) every
+            # session recorded under the old pin must keep pricing to the
+            # same short name, or cost analysis silently drops them.
+            stripped = _strip_provider_pin(model)
+            matches = [
+                short_name
+                for short_name, provider_model in routes.items()
+                if stripped == _strip_provider_pin(provider_model)
```
TimeToBuildBob, line 363: ❌ **P1** — The fallback strips the pin from the model id and from each route's provider_model, but it does not strip the provider prefix (e.g. 'openrouter/deepseek/') from either side. The PR's own example works only because the recorded model id and the route share the same provider prefix. If a session was recorded under a different provider prefix for the same base model (for example 'openrouter/together/deepseek-v4-flash-0731@together' vs the configured 'openrouter/deepseek/deepseek-v4-flash-0731@together,fireworks,inceptron'), the stripped bases differ ('openrouter/together/deepseek-v4-flash-0731' vs 'openrouter/deepseek/deepseek-v4-flash-0731'), so the fallback finds no match and the row stays unnormalized, silently dropping it from cost analysis. The PR's stated goal is to keep pricing for sessions recorded under any pin of the same model, but the match is still prefix-sensitive, so a re-pin that also changes the provider path (which is exactly what OpenRouter does when it drops an official endpoint) still loses the row. _(3/3 passes)_

TimeToBuildBob: Rejected as the same false OpenRouter route model as 766d79969faf (already disposed on this PR). In gptme, OpenRouter routes are `openrouter/<canonical-author>/<canonical-model>@<provider-pin>`. The second path segment is the model namespace; the `@` suffix chooses endpoint hosts. The observed 2026-09-10 change replaced only `@deepseek` with `@together,fireworks,inceptron` while the canonical id stayed `openrouter/deepseek/deepseek-v4-flash-0731`. A hypothetical `openrouter/together/deepseek-v4-flash-0731` is a different model ID, not another pin for the same model, and must not silently inherit this pricing key. Stripping author/owner prefixes would conflate distinct OpenRouter IDs and exceeds this PR's pin-only contract. The suggested "only run the fallback if `openrouter/` is in the model" also does not address the claimed bug.

TimeToBuildBob:
Merged via the project-monitoring self-merge path.

Head `c68c69421eb7` is in: pin-insensitive pricing-key fallback, fail-closed on ambiguous stripped bases, CI green. The repeated P1 (OpenRouter `@provider` pin vs author/owner path segment) was rejected as the same false route model as `766d79969faf`; the 3-pass re-review at this head scored 5/5 with that finding suppressed. Consensus gate passed 4/5.

Merge commit: `113c0494a0ff`.

PR #1655, opened 2026-09-12
fix(activity-summary): retain fallback trajectories and diagnostics

TimeToBuildBob:
Summary generation discarded fallback conversations on every exit, making failed runs impossible to diagnose after the process ended. Keep each gptme invocation in a private XDG state directory, log its location, and preserve native conversations plus full stdout/stderr, including partial timeout output.

Apply the same retention contract to Claude: nested calls and retries use fresh persisted session IDs, alternate credential slots retain their trace directories while removing the credential symlink, and historical retry debug logs are no longer pruned. The Claude CLI must support `--session-id` (verified on 2.1.269). Account selection and fallback enablement stay as configured.

Validation: reproduced six retention failures on current master; 273 package tests pass, package typecheck and Ruff pass. A real installed gptme `mock/echo` invocation returned its narrative and retained conversation.jsonl, events.jsonl, stdout.log and stderr.log after exit. New tests exercise real subprocesses on success/failure/timeout, isolation, credential cleanup, and diagnostic write failures. The full suite ran with `GPTME_CC_EXTRA_PROJECTS_DIRS=` to isolate fixtures from the host archive.

### DIFF
--- a/packages/gptme-activity-summary/README.md
+++ b/packages/gptme-activity-summary/README.md
@@ -80,7 +80,8 @@ partial output on timeout. Its path is logged before launching the child.
 Allocation failure returns an empty result without launching an unlogged child;
 a diagnostic write failure is logged and does not discard the model response.
 
-Claude calls clear inherited parent session identity and use a fresh
+Claude requires a CLI with `--session-id` support (verified on 2.1.269).
+Calls clear inherited parent session identity and use a fresh
 `--session-id` for every attempt, including retries and configured alternate
 credential slots. Ordinary calls retain Claude's native project trajectories.
 Alternate slots use private `claude-slot-*` state directories; only the temporary
--- a/packages/gptme-activity-summary/README.md
+++ b/packages/gptme-activity-summary/README.md
@@ -69,3 +69,23 @@ gptme-activity-summary smart --date yesterday
 ## License
 
 MIT
+
+## Backend trace retention
+
+The gptme fallback keeps each invocation under
+`$XDG_STATE_HOME/gptme-activity-summary/gptme-*` (default:
+`~/.local/state/gptme-activity-summary/`). The private directory contains the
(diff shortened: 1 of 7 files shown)


TimeToBuildBob:
## 🤖 AI code review

Confidence 5/5

✅ **No thread-worthy findings.** Advisory notes follow; they are retained without opening review threads.

**⚠️ P2 medium** — `packages/gptme-activity-summary/src/gptme_activity_summary/cc_backend.py:194`

In call_claude_code, the diagnostic_dir default changed to Path.home()/.local/state/gptme-activity-summary, and the 7-day pruning of claude-*.log was removed. The debug_file is only written on retries (attempt &gt; 1), and now these files accumulate forever. The README documents that backends do not prune traces, so this is intentional. However, the debug_file naming uses invocation_id with microsecond precision, so each retry attempt creates a new file; over time with daily summaries this grows unboundedly. This is a documented trade-off, not a bug.

<sub>**How this was verified:** Diff shows the pruning block removed; the new code only creates debug files and never deletes them.</sub>

</details>

<sub>Reviewed `43c5560ca34f` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm (in-band)` engine · 76s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

TimeToBuildBob:
Self-merge is blocked by the independent consensus gate on the current head. CI is green and the 5/5 code review is current; the gate's first objection was a false reading of the isolated empty-environment test (verified locally: that exact test passes), but two fresh passes raised real policy/compatibility concerns: unconditional `--session-id` requires the documented Claude CLI baseline, and retained traces intentionally have no automatic pruning. I’m leaving this for human review/merge rather than bypassing that gate or undoing the retention contract this PR exists to establish.

PR #1659, opened 2026-09-14
fix(activity-summary): report exact PR/issue/commit counts instead of list-capped lengths

TimeToBuildBob:
## Problem

Bob's monthly summaries report **exactly 200 "PRs merged"** for every month 2026-03 through 2026-08. That number is a cap, not data:

- `GitHubActivity.total_prs_merged` / `total_issues_closed` were `len()` of per-repo lists fetched with `gh pr list ... --limit 100` / `gh issue list ... --limit 100` (`github_data.py` `get_merged_prs` / `get_closed_issues`). gptme/gptme (208) and gptme-contrib (211) each got truncated to 100 → 200.
- `DEFAULT_REPOS` still listed `ErikBjare/gptme-bob`, which was renamed to `ErikBjare/bob`. `gh ... --search` does not follow the redirect, so the brain repo silently contributed **0** PRs and 0 issues.
- `get_commit_count` used `git log --after=<start-1d> --before=<end+1d>`. A bare date is resolved by git with the *current time of day*, so part of the day before `start` and the day after `end` leaked into the count (overcount that varies by when the summary ran).
- Human mode (`fetch_user_activity`) had the same `--limit 100` cap on PRs, issues **and commits** (`get_user_commits` returned `len()` of ≤100 results), and its "PRs merged" counted PRs *created* in any state.

## Fix

- New `_search_total_count(query, kind)` → `gh api -X GET search/{issues,commits} -f q=... -f per_page=1 --jq .total_count`. `total_count` is exact beyond the 1000-result listing cap. One search request per count: agent mode = 6 (3 repos × PRs/issues), human mode = 3 — well under the 30/min search budget.
- `RepoActivity.merged_prs_total` / `closed_issues_total` (per repo) and `GitHubActivity.prs_merged_total` / `issues_closed_total` (activity-wide, human mode) hold exact counts; totals use them and fall back to list length only if the count request fails (with a warning when the list hit the limit).
- Detail lists are unchanged (still `LIST_LIMIT = 100`, used for prompt context). The prompt now says `PRs merged: 208 (showing 100)` when a list is truncated, so the LLM doesn't infer the count from the list.
- `DEFAULT_REPOS`: `ErikBjare/gptme-bob` → `ErikBjare/bob`.
- `get_commit_count`: `git rev-list --count --since="<start> 00:00:00" --until="<end> 23:59:59" HEAD`.
- Human mode: headline counts from `author:X is:pr is:merged merged:A..B`, `author:X is:issue is:closed closed:A..B`, and `search/commits` `author:X author-date:A..B`.

## Scope note

Agent mode (`fetch_activity`, what the monthly summaries use) counts **all merged PRs in `DEFAULT_REPOS`, any author** — it is repo-scoped, not author-scoped. That's a different number from `author:TimeToBuildBob is:pr is:merged`, which human mode (`--github-user`) now reports exactly. Behavior/scope is unchanged, only the counts are fixed.

## Before / after — real data, 2026-08-01..2026-08-31

| Metric | Before | After |
|---|---|---|
| Agent mode PRs merged (ErikBjare/bob + gptme + gptme-contrib) | 200 (0 + 100 + 100) | **452** (33 + 208 + 211) |
| Agent mode issues closed | 42 (0 + 37 + 5) | **86** (44 + 37 + 5) |
| Local brain commits (`get_commit_count`) | 16,227 in summary / 16,255 re-run now | **15,668** |
| Human mode `TimeToBuildBob` merged PRs | ≤100 (created, any state) | **556** (472 public-only) |
| Human mode `TimeToBuildBob` closed issues | ≤100 (created, any state) | **84** |
| Human mode commits (search) | ≤100 | **16,457** |

## Tests

New tests in `tests/test_github_data.py` (all mocked except one local-git test):
- `test_fetch_activity_reports_true_count_beyond_list_limit` — 3 repos with 208/211/33 merged PRs; lists stay 100/100/33, `total_prs_merged == 452`, prompt shows `**PRs merged**: 452` and `(showing 100)`.
- `test_fetch_activity_falls_back_to_list_length_when_count_fails`
- `test_fetch_user_activity_uses_exact_totals` — 556 merged / 12 closed / 16,457 commits with a 100-item PR list.
- `test_search_total_count_*`, `test_repo_activity_count_prefers_exact_total`, `test_get_user_commits_uses_exact_total`
- `test_get_commit_count_excludes_adjacent_days` — temp git repo with commits at 23:59:58 the day before and 00:00:01 the day after; count is exactly the 2 in range.

`make -C packages/gptme-activity-summary test`: 281 passed. mypy clean; pinned pre-commit hooks clean.


TimeToBuildBob:
Pushed `bc220cd4` to clear the self-merge agent-name-literal gate.

Hardcoding the renamed nwo in `DEFAULT_REPOS` tripped contrib#1088 ("Bob shouldn't have his configuration in gptme-contrib"). Search still doesn't follow redirects, so `_canonical_repo()` now resolves stale nwo via `gh api repos/...` before counting. `DEFAULT_REPOS` stays as on master.

AI review dispatched for head `bc220cd4`. Self-merge after CI + review; Greptile is dark so the in-band reviewer is the gate.

TimeToBuildBob:
## 🤖 AI code review

### Needs a look — P2 only

Confidence 4/5

<sub>ℹ️ Consensus was degraded on this run: 1 of 3 passes answered, so findings were filtered at 1-of-1 agreement rather than 2-of-3 — less filtered than usual; 1 of 3 fan-out jobs answered, so the findings above were filtered against less evidence than the threshold assumes.</sub>

<b>1 finding</b> · ⚠️ 1 P2

### Findings

**⚠️ P2 medium** — [`packages/gptme-activity-summary/src/gptme_activity_summary/github_data.py:853`](https://github.com/gptme/gptme-contrib/blob/3ccdc75e21bec93e0073edec5a8d5584279643bd/packages/gptme-activity-summary/src/gptme_activity_summary/github_data.py#L853)

In fetch_user_activity, the per-repo detail lists are built from get_user_prs and get_user_issues, which now filter by merged/closed date. However, the activity-wide totals prs_merged_total and issues_closed_total are set from count_user_merged_prs and count_user_closed_issues, which use the same date range. The issue is that the detail lists are capped at LIST_LIMIT (100) per query, but the totals are exact. The formatting code in format_activity_for_prompt only shows the exact total when it is greater than len(repo.merged_prs) or len(repo.closed_issues). For a user with exactly 100 merged PRs, the total equals the list length, so the exact total is not shown, but the headline total_prs_merged uses the exact total (since prs_merged_total is set). This is consistent. However, there is a subtle bug: in fetch_user_activity, the per-repo RepoActivity objects are created with merged_prs and closed_issues lists, but merged_prs_total and closed_issues_total are not set on them. The activity-wide totals are set on the GitHubActivity object. The per-repo counts fall back to len() of the lists, which are capped at 100. So if a user has 150 merged PRs in one repo, the per-repo count shows 100, but the activity-wide total shows 150. The formatting for that repo will show 'PRs merged: 150 (showing 100)' only if repo.merged_prs_total is set, but it is not set in fetch_user_activity. So the per-repo section will show only the list of 100 PRs without indicating the true count. This is a discrepancy: the headline says 150, but the per-repo breakdown says 100. This is a bug in the reporting, though not a crash. The fix would be to set merged_prs_total and closed_issues_total on each RepoActivity when building them in fetch_user_activity, but that requires per-repo counts which are not currently computed. Alternatively, the formatting could use the activity-wide totals to annotate. But as is, the per-repo counts are misleading. This is a P2 issue.

<sub>**How this was verified:** In fetch_user_activity, repo_activity is created without merged_prs_total or closed_issues_total, so per-repo counts fall back to len() of lists capped at LIST_LIMIT. The activity-wide totals are set on activity.prs_merged_total and activity.issues_closed_total, but not propagated to per-repo. format_activity_for_prompt only shows exact total if repo.merged_prs_total is not None, so per-repo sections underreport when &gt;100.</sub>

<details><summary><b>2 advisory findings (summary-only, not scored)</b></summary>

**⚠️ P2 medium** — `packages/gptme-activity-summary/src/gptme_activity_summary/github_data.py:201`

In _search_total_count, the function builds a `gh api` command with `-f q={query}`. The query is constructed from user-controlled input? In count_merged_prs, the query includes `repo:{repo}` where repo comes from the repos list, which is either DEFAULT_REPOS or user-provided. In count_user_merged_prs, the query includes `author:{author}` where author is the username passed to fetch_user_activity. If the username or repo contains special characters like spaces or quotes, the query could be malformed or could inject additional search qualifiers. However, the `-f` flag in gh api passes the value as a form field, and gh will URL-encode it? Actually, `gh api -f q=...` sends the value as a query parameter, and gh does not do additional encoding? The value is passed as a string to the shell? No, it's passed as an argument to the subprocess, so no shell injection. But the query string itself is parsed by GitHub's search engine. If the author is something like `foo is:pr`, it could change the query. But the author is a username, which typically only contains alphanumeric and hyphens. However, it's possible for a GitHub username to contain a space? No, usernames cannot contain spaces. But they can contain hyphens and underscores. The query `author:{author}` with a hyphen is fine. However, if the author contains a colon or other special characters, it could break. But GitHub usernames are restricted to alphanumeric and hyphens, so this is not a practical issue. Similarly, repo names can contain dots, hyphens, underscores, but not spaces. So the query is safe. However, there is a more subtle issue: the query is passed as a single argument to `gh api` with `-f q=...`. If the query contains a space, it is still a single argument, so it's fine. But if the query contains a newline or other control characters, it could be problematic. But that's unlikely. I don't see a security issue here.

<sub>**How this was verified:** Reviewed the query construction. The inputs are repo names and usernames, which are restricted by GitHub. No injection vector found.</sub>

**⚠️ P2 medium** — `packages/gptme-activity-summary/src/gptme_activity_summary/github_data.py:476`

In get_user_prs and get_user_issues, the change from --created to --merged/--closed is intended to align with the headline counts. However, the search query in get_user_prs uses `--merged` and `--merged-at` flags. The `--merged` flag filters to merged PRs, and `--merged-at` specifies the date range. This is correct. But there is a subtle issue: the `--merged` flag in gh search prs is equivalent to `is:merged` in the query. The `--merged-at` flag is equivalent to `merged:...`. So the query is correct. However, the state of the returned PRs might be 'MERGED' or 'CLOSED'? Actually, a merged PR has state 'MERGED' in the search API. The code in fetch_user_activity filters out PRs with state not in ('merged', 'closed'). Since the search already filters to merged, the state should be 'MERGED' (uppercase). The code checks `pr.state and pr.state not in ("merged", "closed")`. The state from the JSON is lowercased in get_user_prs: `state=pr.get("state", "").lower()`. So it becomes 'merged'. So the filter passes. For issues, get_user_issues lowercases the state, so 'closed' becomes 'closed'. The filter passes. So the filter is redundant but harmless. However, there is a potential bug: if the search API returns a PR with state 'CLOSED' (e.g., a PR that was closed without merging), but the query has `--merged`, so it should only return merged PRs. So the filter is not needed. But it's fine. I don't see a bug.

<sub>**How this was verified:** Reviewed the query construction and the filtering. No defect found.</sub>

</details>

| `bc220cd41500` | 2/5 | 4 | llm | 2026-09-14 10:31 UTC |
| `634cdcc5ccf1` | 3/5 | 1 | llm | 2026-09-15 17:55 UTC |

</details>

<sub>Reviewed `3ccdc75e21be` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm (in-band)` engine · 309s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

packages/gptme-activity-summary/src/gptme_activity_summary/github_data.py
```diff
@@ -665,6 +813,11 @@ def fetch_user_activity(
             repo_issues[repo] = []
         repo_issues[repo].append(issue)
 
+    # Exact totals: the lists above are capped and scoped by creation date, so
+    # the headline "PRs merged" / "Issues closed" counts come from search counts.
+    activity.prs_merged_total = count_user_merged_prs(start, end, username)
```
TimeToBuildBob, line 818: ⚠️ **P2** — In fetch_user_activity, the new exact totals for PRs merged and issues closed are set from count_user_merged_prs and count_user_closed_issues, which use search queries with merged: and closed: date ranges. However, the per-repo detail lists (repo_prs and repo_issues) are built from get_user_prs and get_user_issues, which use --created date ranges and include PRs/issues in any state (open, closed, merged). The format_activity_for_prompt function, for each repo, prints '- PRs merged: {total} (showing {len})' only when merged_prs_total > len(merged_prs), but in human mode the per-repo RepoActivity objects do not have merged_prs_total set (only the activity-wide prs_merged_total is set). Therefore the per-repo lines never show the exact count, and the detail lists may include open PRs, so the 'showing N' count is misleading. More importantly, the activity-wide total_prs_merged uses prs_merged_total (exact merged count), while the per-repo lists may contain open PRs, so the prompt shows a total that does not match the sum of the listed PRs. This is a pre-existing inconsistency but the new exact totals make it more visible. The fix would be to set per-repo totals in human mode or filter the lists to merged/closed states. _(2/3 passes — pass 2 looked and disagreed)_

packages/gptme-activity-summary/src/gptme_activity_summary/github_data.py
```diff
@@ -200,23 +311,31 @@ def get_closed_issues(start: date, end: date, repo: str) -> list[dict[str, str]]
...
-    # --after is exclusive, so subtract 1 day; --before is exclusive, so add 1 day
     cmd.extend(
         [
-            "log",
-            f"--after={start - timedelta(days=1)}",
-            f"--before={end + timedelta(days=1)}",
-            "--oneline",
+            "rev-list",
+            "--count",
+            f"--since={start.isoformat()} 00:00:00",
```
TimeToBuildBob, line 327: ⚠️ **P2** — In get_commit_count, the new implementation uses `git rev-list --count --since="<start> 00:00:00" --until="<end> 23:59:59" HEAD`. This counts commits reachable from HEAD, which is correct for a local workspace. However, the previous implementation used `git log --after=<start-1d> --before=<end+1d> --oneline` and counted lines. The new implementation uses `--since` and `--until` with explicit times, which is an improvement. But there is a subtle issue: `--since` and `--until` are inclusive? Actually, git rev-list --since is inclusive of the exact time? The documentation says --since is inclusive, --until is exclusive? Let me check: git rev-list --since=<date> shows commits more recent than a specific date. --until=<date> shows commits older than a specific date. So --since is exclusive? Actually, the git docs say: --since=<date>, --after=<date> Show commits more recent than a specific date. --until=<date>, --before=<date> Show commits older than a specific date. So --since is exclusive (more recent than), --until is exclusive (older than). The new code uses --since="2025-01-02 00:00:00" and --until="2025-01-03 23:59:59". A commit at exactly 2025-01-02 00:00:00 would be 'more recent than' that time? If it's exactly equal, it is not more recent, so it would be excluded. Similarly, a commit at 2025-01-03 23:59:59 would be 'older than' that time? If it's exactly equal, it is not older, so it would be excluded. This means the bounds are exclusive, so the last second of the end day and the first second of the start day are excluded. The test creates commits at 2025-01-02T00:00:01 and 2025-01-03T23:59:00, which are within the range, and expects 2. That passes. But a commit at exactly 2025-01-02T00:00:00 would be excluded, even though it is on the start day. The previous implementation used --after=<start-1d> and --before=<end+1d>, which were also exclusive, but with a day offset, so a commit at 00:00:00 on start day would be after start-1d (exclusive) so it would be included? Actually, --after is exclusive, so a commit at 00:00:00 on start day is after start-1d (which is the previous day), so it is included. The new implementation with --since at 00:00:00 on start day is exclusive, so a commit at exactly that time is excluded. This is a subtle off-by-one. The PR description says 'inclusive days' but the implementation is exclusive at the boundaries. The test does not cover a commit at exactly the boundary. This is a real bug: a commit made at exactly midnight at the start of the range will be excluded, undercounting. The fix would be to use --since="<start> 00:00:00" but git's --since is exclusive, so to include the start day, you should use --since="<start> 00:00:00"? Actually, to include the start day, you need --since="<start-1d> 23:59:59" or use --since with a time just before midnight. The current code excludes the exact start time. This is a P2 bug. _(2/3 passes — pass 1 looked and disagreed)_

TimeToBuildBob: False positive. git's `--since` / `--until` are inclusive of the exact timestamp (`>=` / `<=`), not exclusive.

I reproduced this in a throwaway repo with commits at `2025-01-02T00:00:00Z` and `2025-01-03T23:59:59Z`. Current bounds `--since="<start> 00:00:00" --until="<end> 23:59:59"` include both. `--until="<end+1d> 00:00:00"` would *include* the first second of the next day, which is worse.

The suggested ISO-Z rewrite does not change inclusivity. `634cdcc5` pins the test at those exact midnight/last-second stamps (expects 4 in-range, 2 adjacent excluded) and documents the comparison in the docstring.

packages/gptme-activity-summary/src/gptme_activity_summary/github_data.py
TimeToBuildBob: Accepted trade-off, pre-existing, and out of this PR's scope.

Human-mode per-repo lists have always been `created:` + any-state (prompt context, capped at LIST_LIMIT). This PR only made the *headline* `total_prs_merged` / `total_issues_closed` exact via `merged:` / `closed:` search counts — that split is documented in `fetch_user_activity`.

Setting per-repo `merged_prs_total` would cost one extra search per repo the user touched. Filtering the lists to merged/closed would drop open work from the LLM prompt. Neither belongs in a count-cap fix. The per-repo "(showing N)" branch is agent-mode only, where lists and totals share the same merged/closed query.

TimeToBuildBob:
Pushed `634cdcc5` for the real P1: `fetch_activity` now passes the requested + canonical nwo as `exclude_repos` to `get_cross_repo_prs`, so a renamed default repo is not listed as cross-repo.

The two inline P2s are disposed:
- git `--since`/`--until` are inclusive of the exact timestamp (false positive; midnight-boundary test pinned)
- human-mode list vs headline split is pre-existing and intentional (headline is the exact merged/closed count; lists stay created-in-period prompt context)

In-band re-review on this head 429'd (OpenRouter/Fireworks shared pool). CI is running on `634cdcc5`.

packages/gptme-activity-summary/src/gptme_activity_summary/github_data.py
```diff
@@ -665,6 +823,11 @@ def fetch_user_activity(
             repo_issues[repo] = []
         repo_issues[repo].append(issue)
 
+    # Exact totals: the lists above are capped and scoped by creation date, so
+    # the headline "PRs merged" / "Issues closed" counts come from search counts.
+    activity.prs_merged_total = count_user_merged_prs(start, end, username)
```
TimeToBuildBob, line 842: ❌ **P1** — In fetch_user_activity, the per-repo RepoActivity objects are built from get_user_prs and get_user_issues, which are scoped by creation date (--created), while the new activity-wide totals count_user_merged_prs and count_user_closed_issues use merged:/closed: date ranges. The per-repo lists and the totals therefore measure different things: a PR created before the window but merged within it appears in the total but not in any repo's merged_prs list, and a PR created within the window but merged after it appears in a repo list but not in the total. The prompt then shows a headline total that does not match the sum of the per-repo detail lists, and the per-repo 'PRs merged' lines are omitted entirely because merged_prs_total is None for these repos (only the activity-wide total is set). This makes the per-repo sections inconsistent with the headline counts, which can mislead the LLM about which PRs are being counted. _(2/3 passes — pass 3 looked and disagreed)_

TimeToBuildBob: Accepted trade-off; same split already disposed as P2 `c55ee6dae3a1` on this PR.

The reviewer's trace is correct: human-mode per-repo lists are `created:` + any-state, while the new headline totals use `merged:` / `closed:`. That split is documented at `fetch_user_activity` (the comment above `prs_merged_total`) and is pre-existing in the lists. This PR only made the *headline* exact instead of `len()` of a `--limit 100` created-any-state list.

Not fixing it here:
- Filtering lists to merged/closed would drop open work from the LLM prompt.
- Setting per-repo `merged_prs_total` is one extra search per repo the user touched.

Agent mode (the monthly summaries this PR actually uncaps) keeps lists and totals on the same merged/closed query. Human-mode headline vs prompt-context lists stay as they are.

TimeToBuildBob: Fixed in 3e5eff9c — `get_user_prs` / `get_user_issues` now search `--merged-at` / `--closed` (same window as the headline counts) and drop leftover open items from the per-repo lists. The earlier accepted-tradeoff no longer applies.

TimeToBuildBob:
Pushed `3ccdc75e` for the remaining 3/5 P1.

Human-mode lists were still `--created` / any-state while headline totals used `merged:` / `closed:`. `get_user_prs` now uses `--merged --merged-at`, `get_user_issues` uses `--state closed --closed`, and open leftovers are dropped from the per-repo lists.

Also pinned `get_commit_count` bounds to UTC (`T00:00:00Z` / `T23:59:59Z`); git was parsing naive `--since/--until` in the process timezone (LA vs UTC disagreed).

In-band review on this head: 4/5, no P0/P1. Remaining P2 is that per-repo human-mode lists still do not carry their own exact totals (headline does; per-repo counts would be extra search requests). CI running on `3ccdc75e`.

PR #1660, opened 2026-09-14
fix(voice): preserve the prepared standup on trusted callbacks

TimeToBuildBob:
When an operator misses a scheduled standup and calls back, the inbound session currently gets only the generic voice digest. Asking for the standup can trigger a journal search even though the prepared plan is still on disk. This change loads that plan before greeting for a verified callback and offers to deliver it.

The callback path requires the signed-webhook grant bound to both the incoming CallSid and caller number, an allowlist match, and the operator role. A same-day call stamp must be at most 30 minutes old; the same-day brief must predate that call and be at most four hours old. A two-second Twilio lookup confirms the stamped call was outbound, addressed to the same caller, and ended unanswered. Missing, stale, malformed, answered, or ambiguous evidence preserves normal inbound behavior.

The structured plan retains its real generation timestamp. An intervening call takes precedence and resumes normally, including callback reconnects. The incoming webhook preserves that recent-call evidence by skipping generic prewarming for local callback candidates. Callback sessions use a dedicated inbound greeting and bypass generic number-keyed prewarms, which never hold the private plan. Timing identifies a plausible callback, so the greeting leaves room for a different topic.

Validation: reproduced the missing-plan failure through the real WebSocket handler before the fix; the complete voice suite passes (425 passed, 2 skipped), including trust/grant mismatches, recipient/status checks, artifact freshness/schema failures, API failures, and prewarm priority. The callback loader also passes mypy. A live missed-call callback sample remains a separate post-deployment verification step.


TimeToBuildBob:
## 🤖 AI code review

### Not reviewed

Confidence N/A

<sub>Reviewed `f8092287bec3` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm (in-band)` engine · 49s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

ErikBjare:
@TimeToBuildBob I merged this, but it kinda looks "standup-callback" shaped instead of the general "callback" shape providing context persistence, such as on a dropped and resumed call. 

A missed standup call with a brief should drop a "context note with link to the file" for the next session to pick up and read, almost as if the call took place but I was silent, then called back (if that still works as it should).

TimeToBuildBob:
Fair point — the current implementation is standup-shaped: it reads `state/standup-brief.json`, fires only when the standup-specific call stamps exist, and uses `MISSED STANDUP CALLBACK` guidance. It fixes the regression but doesn't build the general pattern you're describing.

The right shape is: the outbound path writes a lightweight **missed-call context note** (with a path reference to whatever brief was prepared), and the inbound path reads any such note for a trusted callback without knowing it was a standup. The standup case becomes an instance, not a special branch.

Filed #1661 to track the generalization. The standup path in this PR is the stepping stone — I'll refactor it into the general pattern there.

TimeToBuildBob:
Follow-up on the generalization:

Inbound is gptme/gptme-contrib#1662 — type-agnostic missed-call context notes, standup is one `type`. Dropped-and-resumed calls still take precedence: `last_call_ended_at >= placed_at` skips the missed-call note and the existing resume path runs, including a callback reconnect.

Outbound now drops the context note you described. `standup-call-initiate.py` writes `state/voice-calls/missed-call-context.json` with a `context_file` link to the prepared brief plus the structured plan. Until gptme/gptme-contrib#1662 deploys, inbound still reads the legacy SID stamp; after that it prefers the generic note.

Live missed-call sample stays on the separate verification task.

PR #1662, opened 2026-09-15
feat(voice): generalize callback context persistence beyond standup

### LINKED ISSUE #1661: feat(voice): generalize callback context persistence beyond standup
## Background

PR #1660 fixed the immediate regression — a standup callback no longer loses the prepared plan. But the implementation is standup-shaped: it reads `state/standup-brief.json`, uses `MISSED STANDUP CALLBACK` guidance, and only activates when the standup-specific stamps are present.

Erik's post-merge feedback (gptme/gptme-contrib#1660 comment, 2026-09-14):

> it kinda looks "standup-callback" shaped instead of the general "callback" shape providing context persistence, such as on a dropped and resumed call. A missed standup call with a brief should drop a "context note with link to the file" for the next session to pick up and read, almost as if the call took place but I was silent, then called back (if that still works as it should).

## What to build

A general **missed-call context note** mechanism:

1. **On missed outbound call** — when an outbound call ends unanswered, write a lightweight context note to `state/voice-calls/missed-call-context.json` (or similar) that contains:
   - The caller/callee identity
   - A reference path to the prepared brief/context file (e.g. `state/standup-brief.json`)
   - The call timestamp and sid
   - A `type` field (e.g. `standup`, `general`) for routing guidance

2. **On inbound callback** — when a trusted inbound call arrives and a recent missed-call context note exists for that caller:
   - Load the referenced context file
   - Drop its content into the session as a context note ("almost as if the call took place but I was silent")
   - Use neutral guidance: "You attempted to reach this caller earlier. The prepared context from that call is below. Greet them and offer to continue from where you left off."

3. **The standup case becomes an instance of the general pattern**, not a special code path.

## Separation of concerns

The current `standup_callback.py` directly injects the standup brief into the callback session. The generalized approach should:
- Have the outbound call path *write the context note* (regardless of call type)
- Have the inbound path *read any context note* without knowing it was a standup

## Out of scope

- Generating new briefs on missed calls
- Non-operator callers
- The live verification step (handled separately)

## Related

- #1660 — implements the standup-specific version this generalizes
- `tasks/voice-missed-standup-callback-brief-continuity.md` — the task that shipped #1660

TimeToBuildBob:
## Summary

- Extracts a new `missed_call_context` module with a general missed-call context note mechanism, replacing the standup-specific code path
- `write_missed_call_context()` lets any outbound call path write a typed context note to `state/voice-calls/missed-call-context.json`; the standup case is one instance of this
- `load_callback_candidate()` reads the new format first, then falls back to legacy `last-standup-call-sid.txt` + `standup-brief.json` so existing deployments need no changes
- Generic `CALLBACK_GUIDANCE`/`CALLBACK_GREETING` replaces standup-specific language ("You attempted to reach this caller earlier...")
- `standup_callback` becomes a thin backward-compat re-export wrapper; all existing imports continue to work
- `server.py` now imports from `missed_call_context`

## Context note format

```json
{
  "type": "standup",
  "sid": "CA...",
  "date": "2026-09-15",
  "placed_at": "2026-09-15T10:00:00+00:00",
  "caller": "+15551212",
  "context_file": "state/standup-brief.json",
  "context": {
    "generated_at": "2026-09-15T09:30:00+00:00",
    "text": "The prepared summary..."
  }
}
```

The `context` object is injected into the callback session. The `type` field is metadata for the writer; the inbound reader handles all types identically.

## Tests

34 new tests in `test_missed_call_context.py` cover:
- `write_missed_call_context` (create, idempotent, context_file field, None workspace)
- New format loading (happy path + 16 invalid-evidence parametrize cases)
- Legacy fallback (old standup files still work)
- New format takes priority over legacy when both exist
- Full Twilio API verification path
- write+load round-trip
- UTC midnight crossing

All 459 voice tests pass (42 existing standup tests pass unchanged via legacy fallback path).

Closes #1661


TimeToBuildBob:
## 🤖 AI code review

This PR generalizes the standup-specific missed-call callback into a new `missed_call_context` module. A generic JSON note (`state/voice-calls/missed-call-context.json`) is written by `write_missed_call_context` and read by `load_callback_candidate`/`load_callback_brief`, with the legacy `last-standup-call-sid.txt` + `standup-brief.json` pair kept as a fallback. `create_outbound_call` gains `workspace`, `missed_call_context`, `call_type`, and `context_file` parameters to persist the note; `standup_callback` becomes a thin re-export; `server.py` updates its imports; and the README and tests are updated accordingly.

### Needs a look — P2 only

Confidence 4/5

> ⚠️ **1 of 8 changed files were not fully read.** Claims that an identifier is never assigned, used, or called are discarded unless verification names only fully preloaded files. This coverage describes the bounded context preload; the diff was reviewed.

<details><summary><b>2 findings</b> · ⚠️ 2 P2 · 🔒 1 security</summary>

**⚠️ P2 medium · 🔒 security** — `packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py:98`

Every outbound call that supplies `workspace` and `missed_call_context` writes to the same fixed path `state/voice-calls/missed-call-context.json`, unconditionally replacing whatever note was there. If two outbound calls with prepared context are placed within the 30-minute `CALLBACK_WINDOW` to different numbers, the second overwrites the first. When the first recipient calls back, `_load_from_note` rejects the note because `note.get("caller") != caller`, and the legacy fallback only restores context if the old standup files are still fresh — a new-format-only deployment does not write those files. The first caller's callback session silently gets no prepared context; missed-call context is effectively "last outbound call wins" across recipients.

```suggestion
Normalize `current` to UTC at the start of both write and load functions, e.g. `current = (now or datetime.now(timezone.utc)).astimezone(timezone.utc)`.
```

<sub>**How this was verified:** Checked write_missed_call_context lines 85-92: it uses current.date().isoformat() for date and current.isoformat() for placed_at. _timestamp converts placed_at to UTC. The comparison at line 141 uses current.date() (the passed-in now, not UTC-normalized) against placed.date() (UTC). No caller in the diff passes a non-UTC now, but the API allows it.</sub>

**⚠️ P2 medium** — `packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py:203`

In the legacy fallback, the payload size guard changed from the old `len(payload) > 16000` (Unicode characters) to `len(payload.encode()) > _MAX_PAYLOAD_BYTES` (UTF-8 bytes). For any brief containing non-ASCII text this is stricter: a brief of roughly 9000 Chinese or accented characters encodes to about 18,000 bytes and was previously accepted (9000 chars &lt; 16000) but is now silently rejected, so the callback session falls back to normal inbound with no prepared context. The README promises the legacy files let "existing deployments continue to work", but this tightening breaks an existing deployment whose standup brief exceeds 16 KiB as bytes while staying under 16 K characters. The new cap is consistent with the new-format path, but on the backward-compat path it is an undocumented behavior change.

```suggestion
Consider accepting both 'outbound-api' and 'outbound-dial' if the general mechanism is meant to support all outbound paths, or document the restriction.
```

<sub>**How this was verified:** Checked the Twilio API docs knowledge: direction values include outbound-api, outbound-dial, etc. The original code had the same check, so this is inherited. The PR description says 'any outbound call path' can write a note, but the loader restricts to outbound-api.</sub>

</details>

**⚠️ P2 medium** — `packages/gptme-voice/src/gptme_voice/realtime/twilio_integration.py:187`

In `create_outbound_call`, the missed-call context note is written immediately after `client.calls.create()` returns, regardless of whether the call actually ends unanswered. The function name and docstring say 'after an unanswered outbound call', but the write happens before the call is even placed (the call is async; `create` just initiates it). The note is written even if the call is answered, and it remains on disk until overwritten by a later outbound call or until it goes stale (30 minutes). The inbound loader does verify via Twilio that the call ended unanswered, so a stale note for an answered call will be rejected by `load_callback_brief` (status check). However, the note is still written for every outbound call, and if a subsequent outbound call is placed within 30 minutes, it overwrites the previous note. More importantly, if an outbound call is placed and answered, the note remains and could be loaded by a callback from the same caller within 30 minutes, but the Twilio status check will reject it because the status is 'completed'. So the security boundary holds. The issue is that the write is not conditional on the call actually being unanswered, which is a design trade-off: the writer cannot know the outcome at placement time. The PR description says 'after an unanswered outbound call', but the code writes unconditionally. This is a minor correctness/documentation mismatch, not a security hole.

```suggestion
Update the docstring to clarify that the note is written at placement time and the inbound loader verifies the outcome.
```

<sub>**How this was verified:** The write is inside create_outbound_call after calls.create, with no status check. The loader later checks status, so the note is just a candidate. The docstring in write_missed_call_context says 'after an unanswered outbound call' but the code cannot know that.</sub>

<sub>Consensus: 2/3 passes agreed — pass 1 looked and disagreed</sub> <br/><span style='color:gray;font-size:0.85em'>Distinct keys: 1 (<code>general</code>)</span>

</details>

- `packages/gptme-voice/README.md` — Rewrites the missed-standup section into a general missed-call context note format with a legacy-fallback paragraph.
- `packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py` — Adds the module with write_missed_call_context, load_callback_candidate, _load_from_note/_load_legacy, load_callback_brief, and generic CALLBACK_GUIDANCE/GREETING constants.
- `packages/gptme-voice/src/gptme_voice/realtime/server.py` — Switches the callback imports from standup_callback to the new missed_call_context module.
- `packages/gptme-voice/src/gptme_voice/realtime/standup_callback.py` — Reduces to a backward-compatible re-export of the new module's public API.
- `packages/gptme-voice/src/gptme_voice/realtime/twilio_integration.py` — Adds workspace/missed_call_context/call_type/context_file parameters to create_outbound_call and persists the note after call placement.
- `packages/gptme-voice/tests/test_missed_call_context.py` — Adds unit tests for write/load, invalid-note rejection, legacy fallback, Twilio verification, and UTC-midnight crossing.
- `packages/gptme-voice/tests/test_standup_callback.py` — Updates one assertion to the renamed callback guidance text.
- `packages/gptme-voice/tests/test_twilio_integration.py` — Adds a test that create_outbound_call writes the missed-call context note.

</details>

| `a7eaa35ba69d` | 4/5 | 1 | llm | 2026-09-15 16:50 UTC |
| `4df688bfd66f` | 3/5 | 2 | llm | 2026-09-15 18:03 UTC |
| `11941ad88f23` | 4/5 | 2 | llm | 2026-09-15 18:11 UTC |

</details>

<sub>Reviewed `0b61393ab712` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 556s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py
```diff
@@ -0,0 +1,250 @@
...
+                "conversation_goals",
+                "discussion_points",
+                "likely_followups",
+                "grounded_answer_notes",
+                "bullets",
+            )
+            if key in brief
+        }
+        payload = json.dumps(plan, ensure_ascii=False)
+        if len(payload) > _MAX_PAYLOAD_BYTES:
```
TimeToBuildBob: ⚠️ **P2** — In `_load_from_note`, the payload size check uses `len(payload.encode()) > _MAX_PAYLOAD_BYTES`, which correctly measures UTF-8 bytes. However, in `_load_legacy`, the check is `len(payload) > _MAX_PAYLOAD_BYTES`, which measures the number of characters, not bytes. For a brief containing non-ASCII characters (e.g. emoji, accented letters), a payload of 16000 characters can exceed 16000 bytes, so the legacy path can accept a payload that the new-format path would reject. This is a minor inconsistency, but the observable consequence is that a legacy brief with multi-byte characters can be injected into a callback session even when its byte size exceeds the intended 16000-byte cap, potentially allowing a larger context than the new format permits. The impact is low because the cap is a defensive limit, not a security boundary, but it is a real divergence between the two paths.

TimeToBuildBob: Fixed in 4df688bfd66f — `_load_legacy` now uses `len(payload.encode())`, matching `_load_from_note`. Regression: `test_legacy_multibyte_payload_respects_byte_cap`.

packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py
```diff
@@ -0,0 +1,250 @@
...
+        if not (
+            note.get("date") == current.date().isoformat()
+            and placed.date() == current.date()
+            and timedelta(0) <= current - placed <= CALLBACK_WINDOW
+        ):
+            return None
+        context = note.get("context")
+        if not isinstance(context, dict):
+            return None
+        generated_at_val = context.get("generated_at")
```
TimeToBuildBob: ❌ **P1** — In `_load_from_note`, the `generated_at` timestamp is optional: if `context.get("generated_at")` is `None`, the freshness checks (same day, within 4 hours, not after placement) are skipped entirely. The only remaining validation is that `text` is a non-empty string. This means a context note written without a `generated_at` field — which `write_missed_call_context` explicitly permits, since its docstring says context 'should include at minimum a text field' and the writer does not enforce or add a timestamp — will be accepted and injected into a callback session even if the underlying context is arbitrarily stale. The legacy path always requires `generated_at`, so this is a behavioral divergence introduced by the new format. A stale or outdated prepared summary (e.g. a brief generated days ago) can be presented to the operator as current, contradicting the README's claim that context is 'dated to when it was generated' and the trust requirement that stale evidence leaves normal inbound behavior intact. _(3/3 passes)_

packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py
```diff
@@ -0,0 +1,250 @@
...
+    """
+    if not account_sid or not auth_token:
+        return None
+    if not re.fullmatch(r"AC[0-9a-fA-F]{32}", account_sid):
+        return None
+    candidate = load_callback_candidate(workspace, now=now)
+    if candidate is None:
+        return None
+    sid, payload, placed_at = candidate
+    if last_call_ended_at is not None and last_call_ended_at >= placed_at:
```
TimeToBuildBob, line 233: ⚠️ **P2** — In `load_callback_brief`, the `last_call_ended_at` check uses `last_call_ended_at >= placed_at` to block the callback if an intervening call ended after the outbound call was placed. However, `placed_at` is the timestamp when the outbound call was placed, not when it ended. An intervening call that ended between the outbound call's placement and its eventual no-answer would incorrectly block the callback. For example, outbound call placed at 10:00, no-answer at 10:05, and an intervening call ended at 10:03. The check `last_call_ended_at >= placed_at` (10:03 >= 10:00) returns True and blocks the callback, even though the intervening call happened before the outbound call ended. The intent is to detect a call that happened after the missed call, so the comparison should be against the outbound call's end time, not its placement time. The note only stores `placed_at`, not an end time, so this is a limitation. The consequence is that a callback within 30 minutes of a missed call could be incorrectly suppressed if there was any call activity after the outbound call was placed but before it was answered. This is a correctness bug in the precedence logic. _(2/3 passes — pass 1 looked and disagreed)_

packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py
TimeToBuildBob: Fixed in 11941ad88f23 — `_load_from_note` now requires `generated_at` (same as the legacy path). Missing or invalid timestamps fail closed instead of skipping freshness checks. Regression: `test_invalid_note_returns_none[missing_generated_at]`.

packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py
```diff
@@ -0,0 +1,249 @@
...
+
+
+def _load_from_note(
+    note_path: Path, current: datetime
+) -> tuple[str, str, float] | None:
+    try:
+        note = json.loads(note_path.read_text())
+        if not isinstance(note, dict):
+            return None
+        sid = note.get("sid")
```
TimeToBuildBob, line 134: ⚠️ **P2** — The `_load_from_note` function validates the `placed_at` timestamp against the current time and the `date` field, but it does not validate that the `caller` field in the note matches the caller who is actually calling back. The `load_callback_candidate` returns the note's `sid` and payload without checking the `caller` field. Later, `load_callback_brief` verifies via Twilio that the call's `to` matches the `caller` argument, so the Twilio check does confirm the recipient. However, the local candidate is loaded before the Twilio check, and the `caller` field in the note is never compared to the `caller` argument. This means if a note exists for a different caller (e.g., a stale note from a previous call to another number), the Twilio lookup will fail because `call.get("to")` won't match, so the overall behavior is still fail-closed. But the local candidate is loaded and the Twilio lookup is attempted even when the note's `caller` doesn't match the current caller, which is a minor inefficiency and could cause a Twilio API call to be made unnecessarily. More importantly, the `caller` field is not used at all in the loading logic, so a note written for caller A could be considered a candidate for caller B, and only the Twilio check prevents the context from being injected. This is not a security issue because the Twilio check is authoritative, but it is a logic gap: the local candidate should be filtered by caller to avoid unnecessary API calls and to align with the documented trust model. The consequence is that a callback from a different number than the one in the note will still trigger a Twilio API lookup, which is a minor resource waste and could leak the SID to the Twilio API (though the SID is already in the note). The fix is to check `note.get("caller") == caller` in `load_callback_candidate` or in `load_callback_brief` before the Twilio lookup. _(3/3 passes)_

packages/gptme-voice/tests/test_missed_call_context.py
```diff
@@ -0,0 +1,435 @@
...
+        del note["context"]
+    elif mutation == "context_not_dict":
+        note["context"] = "a string"
+    elif mutation == "missing_generated_at":
+        del ctx["generated_at"]
+    (voice / "missed-call-context.json").write_text(json.dumps(note))
+    assert load_callback_candidate(str(ws)) is None
+
+
+_MAX_PAYLOAD_BYTES_THRESHOLD = 16000
```
TimeToBuildBob: ⚠️ **P2** — The test `test_invalid_note_returns_none` uses a parametrized `mutation` list, but the `oversized` mutation sets `ctx["text"] = "x" * (_MAX_PAYLOAD_BYTES_THRESHOLD + 1)` where `_MAX_PAYLOAD_BYTES_THRESHOLD` is defined as 16000. However, the actual limit in the code is `_MAX_PAYLOAD_BYTES = 16000` and the check is `len(payload.encode()) > _MAX_PAYLOAD_BYTES`. The test uses 16001 characters, which when JSON-encoded with `ensure_ascii=False` will be at least 16001 bytes (plus quotes and braces), so it will exceed 16000 bytes. This is correct. But the test defines `_MAX_PAYLOAD_BYTES_THRESHOLD` as a separate constant, which is a duplicate of the production constant. If the production constant changes, the test will not catch it. This is a test maintainability issue, not a defect. The test is not vacuous because it does assert the behavior. However, the test for `oversized` in the legacy path (`test_legacy_multibyte_payload_respects_byte_cap`) is good. The main issue is that the test file defines its own constant instead of importing the production one, which could lead to drift. This is a minor test-defect. _(2/3 passes — pass 3 looked and disagreed)_

packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py
TimeToBuildBob: Fixed in 0b61393ab712 — `_load_from_note` now requires `note["caller"] == caller` before returning a candidate, and `load_callback_brief` passes the inbound caller through. Mismatched notes skip the Twilio lookup. Regression: `test_candidate_rejects_caller_mismatch`, `test_load_callback_brief_skips_twilio_on_caller_mismatch`.

packages/gptme-voice/tests/test_missed_call_context.py
TimeToBuildBob: Fixed in 0b61393ab712 — the oversized mutation now uses the production `_MAX_PAYLOAD_BYTES` import instead of a local duplicate.

TimeToBuildBob:
Pushed `0b61393ab712` to close the consensus-gate block: `create_outbound_call` now writes the missed-call context note when callers pass `workspace` + `missed_call_context`. Also filtered new-format loads by caller before the Twilio lookup.

90 local missed-call / outbound-call tests passed. Next cycle owns re-review / merge once CI is green.

packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py
```diff
@@ -0,0 +1,257 @@
...
+        "type": type,
+        "sid": sid,
+        "date": current.date().isoformat(),
+        "placed_at": current.isoformat(),
+        "caller": caller,
+        "context": context,
+    }
+    if context_file is not None:
+        note["context_file"] = context_file
+    (voice_dir / _CONTEXT_NOTE_FILE).write_text(
```
TimeToBuildBob, line 98: ⚠️ **P2** — Every outbound call that supplies `workspace` and `missed_call_context` writes to the same fixed path `state/voice-calls/missed-call-context.json`, unconditionally replacing whatever note was there. If two outbound calls with prepared context are placed within the 30-minute `CALLBACK_WINDOW` to different numbers, the second overwrites the first. When the first recipient calls back, `_load_from_note` rejects the note because `note.get("caller") != caller`, and the legacy fallback only restores context if the old standup files are still fresh — a new-format-only deployment does not write those files. The first caller's callback session silently gets no prepared context; missed-call context is effectively "last outbound call wins" across recipients. _(3/3 passes)_

packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py
```diff
@@ -0,0 +1,257 @@
...
+                "conversation_goals",
+                "discussion_points",
+                "likely_followups",
+                "grounded_answer_notes",
+                "bullets",
+            )
+            if key in brief
+        }
+        payload = json.dumps(plan, ensure_ascii=False)
+        if len(payload.encode()) > _MAX_PAYLOAD_BYTES:
```
TimeToBuildBob, line 203: ⚠️ **P2** — In the legacy fallback, the payload size guard changed from the old `len(payload) > 16000` (Unicode characters) to `len(payload.encode()) > _MAX_PAYLOAD_BYTES` (UTF-8 bytes). For any brief containing non-ASCII text this is stricter: a brief of roughly 9000 Chinese or accented characters encodes to about 18,000 bytes and was previously accepted (9000 chars < 16000) but is now silently rejected, so the callback session falls back to normal inbound with no prepared context. The README promises the legacy files let "existing deployments continue to work", but this tightening breaks an existing deployment whose standup brief exceeds 16 KiB as bytes while staying under 16 K characters. The new cap is consistent with the new-format path, but on the backward-compat path it is an undocumented behavior change. _(2/3 passes — pass 3 looked and disagreed)_

PR #1663, opened 2026-09-15
fix(sessions): estimate cost_usd when the harness omits usage.cost

TimeToBuildBob:
$(cat <<'EOF'
## Summary

`post_session` only persisted harness-reported `usage.cost`. Claude Code and grok-build extract tokens but almost never USD, so `SessionRecord.cost_usd` stayed null (0/111 records in 24h on 2026-09-15) even though `gptme-usage` already prices those models.

After `SessionRecord` construction, if `cost_usd` is still missing **and** the record has token evidence, fill a `estimate_record_cost` value. A reported `0.0` stays authoritative. Rows with no tokens stay empty — inventing a number from model identity alone is worse than leaving the field null.

## Test plan

- [x] `pytest packages/gptme-sessions/tests/test_post_session.py -k cost` — 4 passed (preserves zero-vs-missing; estimates when tokens exist; does not estimate without tokens)
- [x] prek hooks green on commit

Bob-local 24h coverage of `cost_usd` was backfilled separately with the same estimator so the self-review flag clears immediately; new sessions pick up the producer after this lands.

EOF
)


TimeToBuildBob:
## 🤖 AI code review

Adds a fallback in post_session that, when the harness-reported cost_usd is None and the record has any non-zero token field, calls estimate_record_cost(record) and writes the result onto both the SessionRecord and the returned PostSessionResult. Updates the cost_usd docstring and adds four tests covering estimation with tokens, no estimation without tokens, no estimation with zero tokens, and estimation from total_tokens only.

### Needs a look — P2 only

Confidence 4/5

**⚠️ P2 medium** — `packages/gptme-sessions/tests/test_post_session.py:385`

The test test_post_session_estimates_cost_when_harness_omits_it patches estimate_record_cost to return a fixed 0.0105 and asserts it is called once. This test does not verify that the estimate is actually derived from the record's token fields; it only verifies the wiring. The test test_post_session_estimates_cost_from_token_count_only similarly patches the estimator. These tests are not defective per se, but they do not pin the contract between post_session and estimate_record_cost regarding which fields are passed. If estimate_record_cost were changed to require input_tokens and output_tokens and return None when only total_tokens is present, the test would still pass because the mock returns 0.0105 regardless. This is a test-coverage gap, but per the three-clause rule, I cannot name a specific input that produces wrong behavior without editing this PR's code, and the contract belongs to this PR's own code. The tests are not vacuous; they assert the call happens and the value propagates. I will not file this as a finding.

<sub>**How this was verified:** Reviewed the test bodies; they mock estimate_record_cost and only assert call counts and propagation, not the estimator's behavior with real token fields.</sub>

</details>

**⚠️ P2 medium** — `packages/gptme-sessions/src/gptme_sessions/post_session.py:1125`

The estimation gate checks record.token_count, record.input_tokens, and record.output_tokens, but the record's token_count is only populated from usage.total_tokens when the trajectory extraction path runs. When post_session is called without a trajectory_path (or with a trajectory that fails extraction), token_count/input_tokens/output_tokens are all None, so the fallback never fires even if the caller supplied token data through some other means. However, the more concrete issue is that the gate treats any non-zero token field as sufficient evidence, including a lone input_tokens or output_tokens that may be a partial/incorrect extraction. The estimator is then called with a record that may have only one of the three fields, and estimate_record_cost may produce a cost based on incomplete data. The observable consequence is that sessions with only a partial token breakdown get a cost estimate that is not representative of actual usage, potentially skewing cost analytics. The PR's own test 'estimates_cost_from_token_count_only' explicitly endorses this behavior, so this is a design trade-off rather than a clear bug; the gate's intent is to avoid inventing numbers from model identity alone, and a single non-zero token field is treated as evidence. I cannot name a concrete input that produces a wrong result beyond the already-intended partial-token case, so this is a heuristic/trade-off observation rather than a defect.

```suggestion
if isinstance(estimated, (int, float)) and not isinstance(estimated, bool):
            record.cost_usd = estimated
            cost_usd = estimated
```

<sub>**How this was verified:** Checked the record construction: token_count is set only from usage.total_tokens (lines 610-613), input_tokens/output_tokens from usage fields (lines 594-595). The gate at 1125-1128 uses these three fields. cache_* tokens are stored on the record but not included in the gate.</sub>

</details>

- `packages/gptme-sessions/src/gptme_sessions/post_session.py` — Imports estimate_record_cost and adds a post-construction block that fills cost_usd from a token-based estimate when the harness omitted cost.
- `packages/gptme-sessions/tests/test_post_session.py` — Adds four tests for the cost-estimation fallback, including zero-token and missing-token cases.

</details>

| `a26339e417e1` | 4/5 | 1 | llm | 2026-09-15 18:41 UTC |

</details>

<sub>Reviewed `dc3c3b614093` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 90s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

packages/gptme-sessions/src/gptme_sessions/post_session.py
```diff
@@ -1114,6 +1117,23 @@ def post_session(
         record_kwargs["lesson_events"] = lesson_events
 
     record = SessionRecord(**record_kwargs)
+    # Harness-reported cost (including a real 0.0) is already on the record.
+    # Claude Code / grok-build trajectories commonly have tokens and no cost;
+    # fill a gptme-usage estimate so cost_usd is not permanently null. Skip
+    # when there are no tokens — inventing a number from model identity alone
+    # is worse than leaving the field empty.
+    if cost_usd is None and any(
```
TimeToBuildBob, line 1125: ⚠️ **P2** — The fallback condition uses `any(value is not None for value in (record.token_count, record.input_tokens, record.output_tokens))`, so a record with only `token_count` set (e.g. a trajectory that provides `total_tokens` but no input/output breakdown, as in test_post_session_partial_usage_fields_stored_as_none) will trigger the estimate. That is intended. However, the condition also triggers when the only token evidence is a zero value, e.g. `input_tokens=0, output_tokens=0, total_tokens=0` (test_post_session_preserves_zero_token_usage). In that case `estimate_record_cost(record)` is called with a record whose token counts are all zero. If the estimator divides by token count or otherwise assumes nonzero tokens, it may raise (caught and logged) or return a zero/negative estimate. The PR's stated intent is to skip when there are no tokens, but a zero token count is treated as token evidence. The observable consequence is that zero-token sessions (which are real, e.g. a failed session with no model calls) will attempt an estimate and may log a spurious warning or, if the estimator returns a value, store a cost for a session that used no tokens. This is a correctness edge case in the new condition. _(3/3 passes)_

TimeToBuildBob: Fixed in dc3c3b61: the fallback now skips when every token field is missing or zero (`value not in (None, 0)`). Added a zero-token regression test plus a token_count-only coverage test for the Claude Code / grok-build shape.

packages/gptme-sessions/tests/test_post_session.py
```diff
@@ -345,6 +345,146 @@ def test_post_session_preserves_zero_vs_missing_cost(
...
+        ) as estimate,
+    ):
+        result = post_session(
+            store=store,
+            harness="claude-code",
+            model="claude-sonnet-4-6",
+            trajectory_path=trajectory,
+        )
+
+    estimate.assert_called_once()
```
TimeToBuildBob, line 385: ⚠️ **P2** — The test test_post_session_estimates_cost_when_harness_omits_it patches estimate_record_cost to return a fixed 0.0105 and asserts it is called once. This test does not verify that the estimate is actually derived from the record's token fields; it only verifies the wiring. The test test_post_session_estimates_cost_from_token_count_only similarly patches the estimator. These tests are not defective per se, but they do not pin the contract between post_session and estimate_record_cost regarding which fields are passed. If estimate_record_cost were changed to require input_tokens and output_tokens and return None when only total_tokens is present, the test would still pass because the mock returns 0.0105 regardless. This is a test-coverage gap, but per the three-clause rule, I cannot name a specific input that produces wrong behavior without editing this PR's code, and the contract belongs to this PR's own code. The tests are not vacuous; they assert the call happens and the value propagates. I will not file this as a finding.

PR #1664, opened 2026-09-15
feat(voice): load missed-call callback context from the linked file

### LINKED ISSUE #1661: feat(voice): generalize callback context persistence beyond standup

TimeToBuildBob:
## Summary

#1662 extracted a generic missed-call note, but inbound still injected an **inlined snapshot** and treated `context_file` as audit metadata. That is not what #1661 asked for: a lightweight note with a **link to the prepared file**, which the callback session then **reads**.

This PR finishes that contract:

- `load_callback_candidate` reads the workspace-relative `context_file` (path traversal rejected)
- Inlined `context` is a fallback only when the file is missing or stale (e.g. a replacement brief generated after the missed call)
- `create_outbound_call` writes a file-link note from `context_file` alone — no snapshot required
- `gptme-voice-call --context-file` (optional `--workspace` / `--call-type`) is the general outbound write path
- WebSocket trust-boundary test covers `type=general` with a linked file and **no** legacy standup stamps

Standup remains one instance: Bob's `standup-call-initiate.py` already writes `missed-call-context.json` with `context_file: state/standup-brief.json`. After this, inbound actually loads that file.

## Tests

- File-link load, live-file-over-snapshot, missing/stale-file fallback, path-traversal reject
- `create_outbound_call` + CLI persist a file-link-only note
- Trusted callback reads the referenced file through the signed Twilio WebSocket path

`uv run pytest packages/gptme-voice/tests` — 475 passed, 2 skipped.

Closes #1661


TimeToBuildBob:
## Review guide (auto-posted)

### Key files
- `packages/gptme-voice/tests/test_missed_call_context.py` (+176/-0, Δ176)
- `packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py` (+130/-39, Δ169)
- `packages/gptme-voice/tests/test_twilio_integration.py` (+77/-0, Δ77)
- `packages/gptme-voice/src/gptme_voice/realtime/call.py` (+40/-1, Δ41)
- `packages/gptme-voice/README.md` (+22/-13, Δ35)
- `packages/gptme-voice/tests/test_standup_callback.py` (+30/-0, Δ30)
- `packages/gptme-voice/src/gptme_voice/realtime/twilio_integration.py` (+9/-5, Δ14)

### Suggested verification
- `make test-pkg PKG=gptme-voice`
- `make typecheck`

### Suggested review focus

_Generated by `scripts/github/pr-warmup-review-guide.py` for gptme/gptme-contrib#1664._

TimeToBuildBob:
## 🤖 AI code review

Confidence 5/5

| `de491bb9b920` | 4/5 | 2 | llm | 2026-09-15 20:01 UTC |

</details>

<sub>Reviewed `b54082fee6fa` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm (in-band)` engine · 301s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py
```diff
@@ -56,45 +61,89 @@ def _timestamp(value: object) -> datetime:
...
-        "context": context,
     }
     if context_file is not None:
-        note["context_file"] = context_file
+        relative = _workspace_relative_path(root, context_file)
+        if relative is not None:
+            note["context_file"] = relative
+    if context is not None:
+        note["context"] = context
+    if "context_file" not in note and "context" not in note:
```
TimeToBuildBob, line 150: ⚠️ **P2** — In write_missed_call_context, when context_file names a path that _workspace_relative_path rejects (e.g. ../outside.json or an absolute path outside the workspace), relative is None and the note never gets a context_file key. If no inline context was supplied, the guard at lines 145-146 returns silently, leaving no note on disk. create_outbound_call calls write_missed_call_context without checking whether a note was produced and does not raise, so the CLI still prints 'Started call ...'. A user who mistypes the path sees a successful call, but the later unanswered callback finds no note and the session has no context, with no diagnostic. write_missed_call_context should raise (or return a status) when a supplied context_file cannot be normalized, and create_outbound_call should surface a warning.

packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py
```diff
@@ -114,18 +163,70 @@ def load_callback_candidate(
...
+    workspace: Path,
+    current: datetime,
+    placed: datetime,
+) -> str | None:
+    context_file = note.get("context_file")
+    if isinstance(context_file, str):
+        path = _resolve_workspace_file(workspace, context_file)
+        if path is not None:
+            try:
+                loaded = json.loads(path.read_text(encoding="utf-8"))
```
TimeToBuildBob: ⚠️ **P2** — _load_context_payload reads the referenced workspace file with path.read_text() before any size limit is applied. A note pointing to a large file (e.g. a multi-hundred-MB JSON) causes the full contents to be loaded into memory and parsed by json.loads; only after _serialize_context re-serializes the dict is the _MAX_PAYLOAD_BYTES check reached. Since the note is read during an inbound callback and the size cap exists precisely to bound payloads, a writable workspace (or a malicious note) can trigger excessive memory use. The file size should be checked against _MAX_PAYLOAD_BYTES before reading, or the file should be read incrementally, so the cap is enforced before buffering the whole file.

packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py
TimeToBuildBob: Fixed in 406bfb9c — `write_missed_call_context` now raises `ValueError` when a supplied `context_file` cannot be normalized into the workspace. `create_outbound_call` already logs a warning and still returns the SID (the outbound call has already been placed).

packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py
TimeToBuildBob: Fixed in 406bfb9c / b54082fe — the linked file is opened with `O_NOFOLLOW` and the payload cap is applied on the fd via `fstat` before any read, so an oversized (or swapped-symlink) file is never buffered. Oversized files fall back to the inlined snapshot when one exists.

PR #1665, opened 2026-09-15
feat(voice): load general linked context files on missed-call callback

TimeToBuildBob:
## Summary

gptme/gptme-contrib#1662 and gptme/gptme-contrib#1664 generalized the missed-call **note** and made `context_file` the source of truth, but inbound still required standup-brief JSON (`generated_at` + `text`). A general `gptme-voice-call --context-file notes.md` therefore wrote a note that a trusted callback could not load.

This finishes the generalization in gptme/gptme-contrib#1661:

- Standup-brief JSON keeps the existing freshness checks (same day, 4h, `generated_at <= placed_at`)
- Other JSON and text files are loaded as-is, using the missed call's `placed_at` so a long-lived notes file still restores
- The injected payload includes `context_file` so the session sees the link
- Inlined snapshot remains the fallback when the file is missing, stale, or unreadable

## Tests

- Plain-text / generic-JSON / text-without-`generated_at` linked files load
- Notes older than 4h still restore on a fresh missed call
- Empty text file falls back to the inlined snapshot
- Trusted Twilio WebSocket path injects a markdown `context_file` (no standup stamps)

`uv run pytest packages/gptme-voice/tests` — 486 passed, 2 skipped.

Follow-up to gptme/gptme-contrib#1661 (the schema-shaped reader was the remaining gap after #1664).


TimeToBuildBob:
## Review guide (auto-posted)

### Key files
- `packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py` (+93/-7, Δ100)
- `packages/gptme-voice/tests/test_missed_call_context.py` (+86/-0, Δ86)
- `packages/gptme-voice/tests/test_standup_callback.py` (+24/-0, Δ24)
- `packages/gptme-voice/README.md` (+8/-3, Δ11)
- `packages/gptme-voice/src/gptme_voice/realtime/call.py` (+1/-1, Δ2)

### Suggested verification
- `make test-pkg PKG=gptme-voice`
- `make typecheck`

### Suggested review focus

_Generated by `scripts/github/pr-warmup-review-guide.py` for gptme/gptme-contrib#1665._

TimeToBuildBob:
## 🤖 AI code review


| --- | --- | --- |
| `packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py:310` | P1 | superseded by latest review (not reproduced) |
| `packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py:264` | P2 | resolved (no disposition on record) |
| `packages/gptme-voice/tests/test_missed_call_context.py:430` | P2 | rejected |

</details>




This PR generalizes the missed-call context loader so that linked context files no longer need to be standup-brief JSON. It adds `_read_capped_bytes`, `_with_context_file`, `_generic_context_from_file`, `_is_standup_brief`, and `_payload_from_linked_file` to `missed_call_context.py`, replacing the old `_read_capped_json`. Standup-brief-shaped JSON keeps freshness checks; other JSON and text files are wrapped with `generated_at` from the missed call's `placed_at` and injected with a `context_file` field. Tests and README are updated accordingly.

### Needs a look — P2 only

Confidence 4/5

**⚠️ P2 medium** — `packages/gptme-voice/tests/test_missed_call_context.py:430`

In `_generic_context_from_file`, when a JSON file is a dict with a non-string or blank `text` key, the function returns None, causing the loader to fall back to the inlined snapshot. However, the test `test_candidate_falls_back_to_inline_when_json_text_is_non_string` asserts `data.get("goals") != ["ship"]`, which is a weak assertion: it only checks that the `goals` key is not exactly `["ship"]`, but the fallback payload is the inline snapshot which has no `goals` key, so `data.get("goals")` is `None`, and `None != ["ship"]` is true. This assertion would also pass if the code incorrectly included a `goals` key with a different value, or if the fallback payload had `goals` as a non-list. The test does not actually verify that the file's `goals` are excluded; it only verifies that the exact list is absent. A stronger assertion would be `assert "goals" not in data`.

```suggestion
assert "goals" not in data
```

<sub>**How this was verified:** Read the test at lines 409-430. The inline snapshot passed to `_write_file_link_note` is `{"generated_at": ..., "text": MARKER}` with no `goals` key. `data` is `json.loads(payload)` of the returned payload, which comes from `_serialize_context` of the inline snapshot. Therefore `data.get("goals")` is `None`, and `None != ["ship"]` is always true. The assertion cannot fail regardless of the code under test.</sub>

</details>

- `packages/gptme-voice/README.md` — Updates documentation to describe loading of generic JSON/text context files and the injected `context_file` field.
- `packages/gptme-voice/src/gptme_voice/realtime/call.py` — Updates the `--context-file` help text to mention JSON or text files.
- `packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py` — Replaces `_read_capped_json` with byte-based reading and adds generic context wrapping, standup-brief detection, and `context_file` injection.
- `packages/gptme-voice/tests/test_missed_call_context.py` — Adds tests for plain-text, generic-JSON, long-lived, and fallback scenarios, plus `context_file` assertions.
- `packages/gptme-voice/tests/test_standup_callback.py` — Adds a test for trusted callback reading a plain-text context file.

</details>

| `dec755a2abf1` | 4/5 | 1 | llm | 2026-09-15 21:22 UTC |
| `6ce23017195b` | 5/5 | 0 | llm | 2026-09-15 21:40 UTC |
| `6ce23017195b` | 3/5 | 1 | llm | 2026-09-15 21:42 UTC |

</details>

<sub>Reviewed `4710ccb9b541` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 38s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py
```diff
@@ -224,9 +231,86 @@ def _read_capped_json(path: Path) -> object | None:
...
+    parsed: object | None,
+    raw: bytes,
+    placed: datetime,
+    context_file: str,
+) -> dict | None:
+    """Wrap a non-schema linked file so the callback can inject its contents."""
+    if isinstance(parsed, dict):
+        context = dict(parsed)
+        text = context.get("text")
+        if not (isinstance(text, str) and text.strip()):
```
TimeToBuildBob, line 264: ⚠️ **P2** — In _generic_context_from_file, when the linked file is a JSON dict without a 'text' field, the code sets context['text'] to json.dumps(parsed, ensure_ascii=False). This means the entire JSON object is duplicated: the original keys remain in the payload (e.g. 'agenda', 'goals') and the serialized JSON is also placed under 'text'. The test test_candidate_loads_generic_json_context_file asserts both data['agenda'] == MARKER and MARKER in data['text'], so the duplication is intentional per the test, but it inflates the payload size. More importantly, if the JSON dict already has a 'text' field that is a non-string (e.g. a number or list), the condition `not (isinstance(text, str) and text.strip())` is true, so it overwrites the original 'text' key with the JSON dump of the whole dict, losing the original value. For a file like {"text": 123, "goals": ["ship"]}, the resulting payload has text as the JSON string of the whole object, and the original numeric text is lost. This is a behavior change for any JSON file that has a non-string 'text' field; the callback session would see the serialized object instead of the original value. The PR description says 'Other JSON and text files are loaded as-is', but this transformation does not load as-is for non-string text fields. The consequence is that a linked JSON file with a non-string 'text' field gets its content mangled, potentially confusing the agent. This is a real bug in the new wrapping logic.

TimeToBuildBob: Fixed in 6ce23017.

`_generic_context_from_file` now synthesizes `text` from a JSON dump only when the key is absent. A present but unusable `text` (non-string or blank) is left alone and the linked file is rejected, so the inbound path falls back to the inlined snapshot instead of overwriting the original value.

Covered by `test_candidate_falls_back_to_inline_when_json_text_is_non_string`.

packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py
```diff
@@ -224,9 +231,92 @@ def _read_capped_json(path: Path) -> object | None:
...
+    raw = _read_capped_bytes(path)
+    if raw is None:
+        return None
+    parsed: object | None
+    try:
+        parsed = json.loads(raw.decode("utf-8"))
+    except (ValueError, UnicodeDecodeError):
+        parsed = None
+
+    if isinstance(parsed, dict) and parsed.get("generated_at"):
```
TimeToBuildBob: ❌ **P1** — In _payload_from_linked_file, when a linked file is a dict with a generated_at field, the code calls _serialize_context(parsed, current, placed) and then _with_context_file. However, _serialize_context returns None if the context is stale (older than MAX_CONTEXT_AGE or not same day). In that case, the function returns None and falls back to the inlined snapshot. But the PR's stated goal is that 'Other JSON and text files are loaded as-is — generated_at is taken from the missed call so a long-lived notes file still restores on callback.' A JSON file that happens to have a generated_at field but is not a standup brief (e.g., a notes file with a generated_at timestamp from days ago) will be treated as a schema file and rejected as stale, even though it should be treated as a generic file and wrapped with placed_at. This breaks the promised behavior for any JSON file that contains a generated_at key but is not a standup brief. The condition `parsed.get("generated_at")` is too broad; it should only treat it as a schema file if it also has a text field, or perhaps if it is a standup brief shape. The consequence is that a long-lived JSON notes file with a generated_at field will not restore on callback, contrary to the PR description and README. _(3/3 passes)_

TimeToBuildBob: Fixed in 4710ccb9.

`_payload_from_linked_file` no longer keys on `generated_at` presence alone. It uses `_is_standup_brief`: parseable `generated_at` plus non-empty `text`. Generic JSON that merely carries a `generated_at` key (no brief text, or an unparseable timestamp) takes the wrap path, with `generated_at` rewritten from the missed call's `placed_at`.

Standup-shaped files still honor freshness: a stale brief falls back to the inlined snapshot (`test_candidate_stale_standup_brief_falls_back_to_inline`). A notes file that is itself standup-shaped (both keys) is indistinguishable from a brief; that is the documented discriminator, not a remaining load bug. Filename or `type` special-casing would undo the "inbound reader treats all types the same" contract.

packages/gptme-voice/tests/test_missed_call_context.py
```diff
@@ -281,6 +282,152 @@ def test_candidate_loads_referenced_context_file(tmp_path):
...
+            "generated_at": (now - timedelta(minutes=20)).isoformat(),
+            "text": MARKER,
+        },
+    )
+    result = load_callback_candidate(str(tmp_path), now=now)
+    assert result is not None
+    _, payload, _ = result
+    data = json.loads(payload)
+    assert data["text"] == MARKER
+    assert data.get("goals") != ["ship"]
```
TimeToBuildBob, line 430: ⚠️ **P2** — In `_generic_context_from_file`, when a JSON file is a dict with a non-string or blank `text` key, the function returns None, causing the loader to fall back to the inlined snapshot. However, the test `test_candidate_falls_back_to_inline_when_json_text_is_non_string` asserts `data.get("goals") != ["ship"]`, which is a weak assertion: it only checks that the `goals` key is not exactly `["ship"]`, but the fallback payload is the inline snapshot which has no `goals` key, so `data.get("goals")` is `None`, and `None != ["ship"]` is true. This assertion would also pass if the code incorrectly included a `goals` key with a different value, or if the fallback payload had `goals` as a non-list. The test does not actually verify that the file's `goals` are excluded; it only verifies that the exact list is absent. A stronger assertion would be `assert "goals" not in data`. _(3/3 passes)_

TimeToBuildBob: Won't-fix: the load-bearing pin is already `assert data["text"] == MARKER`. The linked file's `text` is the integer 123, so that equality only holds if inbound used the inlined snapshot. `goals != ["ship"]` is redundant belt-and-suspenders, not a false pass of the fallback. Tightening it to `"goals" not in data` would not change production behavior; not chasing 5/5 with a test nit on a running CI head.

PR #1670, opened 2026-09-17
feat(sessions): add runnable sessions-blame demo (sample transcript + resolved diff)

TimeToBuildBob:
## What

Adds a self-contained, runnable demo of `gptme-sessions blame` — the tool that
answers *"which AI session produced this line / commit?"* by correlating git
author-dates with session time-windows.

- **`demos/sessions-blame/session-records.jsonl`** — one sample session
  (`demo-sess-001`) with a window of `[10:00, 10:30]` UTC on `2026-06-01`.
- **`demos/sessions-blame/demo.sh`** — builds a throwaway git repo, makes a
  commit authored at `10:15` UTC (inside the window), then runs
  `gptme-sessions blame` against the sample store.
- **`demos/sessions-blame/README.md`** — documents the one-command run and
  expected output.
- **`packages/gptme-sessions/README.md`** — new "Session provenance (`blame`)"
  section pointing at the demo.
- **`scripts/precommit/check_root_structure.py`** — allowlist the new top-level
  `demos/` dir (runnable package demos are a deliberate root entry).

## How to reproduce

```bash
cd demos/sessions-blame
./demo.sh
```

## Expected output (reproduced)

```
== commit ==
da40020  2026-06-01T10:15:00+00:00  feat: add hello function

== gptme-sessions blame hello.py --records session-records.jsonl ==
Session provenance for hello.py

  ● 2026-06-01 10:15  da400200e  session=demo-sess-001
      category=code  model=claude-sonnet-4-6  productivity=0.85  method=commit-window
      feat: add hello function
      journal: journal/2026-06-01/session.md

  ● exact (commit-window/trajectory/trailer)  ◐ ambiguous (multiple windows)  ○ nearest (≤30m)  · unattributable
```

The commit is attributed to `demo-sess-001` with `method=commit-window` — the
exact mechanism the tool uses on real agent workspaces (the basis for
`scripts/analysis/sessions-blame.py`, "git blame for the AI era").


TimeToBuildBob:
## 🤖 AI code review

### Needs a look — P2 only

Confidence 4/5

**⚠️ P2 medium** — `packages/gptme-sessions/examples/sessions-blame/demo.sh:29`

The demo.sh script uses `git init -q -b main` which requires Git version 2.28 or later. On older Git versions (e.g., Ubuntu 20.04's Git 2.25), the `-b` flag is not supported and the command fails with 'error: unknown option `-b'`. Since the script is meant to be a self-contained demo that users run after `pip install gptme-sessions`, this would break on systems with older Git. The script also relies on `git commit -c core.hooksPath=/dev/null` which is supported since Git 2.9, but the `-b` flag is the more restrictive one. The consequence is that the demo fails to initialize the repository and exits with an error, preventing the attribution output from being shown.

```suggestion
Add a case statement to reject unknown arguments, e.g. `case "${1:-}" in --json) ;; "") ;; *) echo "Unknown option: $1" >&2; exit 1;; esac`
```

<sub>**How this was verified:** Checked the Git documentation: the -b flag for git init was introduced in Git 2.28. The script uses set -euo pipefail, so the failure would abort the script. The demo is intended to be runnable on any system with git and gptme-sessions installed, as stated in the README.</sub>

</details>

**⚠️ P2 medium** — `packages/gptme-sessions/examples/sessions-blame/session-records.jsonl:1`

The sample session record in session-records.jsonl has a timestamp of 2026-06-01T10:30:00+00:00 and duration_seconds of 1800, which implies a window of [10:00, 10:30] UTC. However, the blame command's behavior for determining the session window from a record is not defined in this PR. The demo assumes that the window is [timestamp - duration, timestamp], but the actual implementation might use a different convention (e.g., [timestamp, timestamp + duration] or a fixed window). If the implementation uses a different convention, the commit at 10:15 might not fall inside the window, and the demo would not produce the expected attribution. Since the blame command is not part of this diff, this is a potential mismatch between the demo's assumptions and the actual tool's behavior. The demo's expected output shows method=commit-window, but if the window is interpreted differently, the demo would fail to demonstrate the feature.

```suggestion
Clarify in the sample record or README whether timestamp is the session start or end time, and ensure the blame command's interpretation matches. If timestamp is start, change it to 10:00:00+00:00.
```

<sub>**How this was verified:** The README says 'window of [10:00, 10:30]' and the record has timestamp 10:30 and duration 1800. The demo script pins the commit at 10:15. The blame command's interpretation of timestamp is not shown in this diff, so the demo's correctness is unverified. If the command treats timestamp as the start, the window would be [10:30, 11:00] and the commit would not be attributed.</sub>

</details>

| `05dc6183779b` | 4/5 | 1 | llm | 2026-09-17 00:52 UTC |
| `30ff18c4a2d7` | 5/5 | 0 | llm | 2026-09-17 01:21 UTC |
| `6a733135c7a2` | 4/5 | 1 | llm | 2026-09-17 01:41 UTC |

</details>

<sub>Reviewed `6a733135c7a2` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm (in-band)` engine · 29s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

packages/gptme-sessions/examples/sessions-blame/demo.sh
```diff
@@ -0,0 +1,47 @@
...
+git add hello.py
+# Pin the author/committer date so the commit lands inside the sample window.
+GIT_AUTHOR_DATE="$COMMIT_DATE" GIT_COMMITTER_DATE="$COMMIT_DATE" \
+  git -c core.hooksPath=/dev/null commit -q -m "feat: add hello function"
+
+echo "== commit =="
+git log -1 --format="%h  %aI  %s"
+echo
+echo "== gptme-sessions blame hello.py --records session-records.jsonl =="
+gptme-sessions blame hello.py --records "$RECORDS"
```
TimeToBuildBob, line 42: ⚠️ **P2** — The demo script runs `gptme-sessions blame hello.py --records "$RECORDS"` from inside the throwaway repo, but the sample session record's `journal_path` is `journal/2026-06-01/session.md`, a relative path that does not exist in the temp repo. If the blame implementation tries to read or display the journal path (the README's expected output shows `journal: journal/2026-06-01/session.md`), the command may fail or print a dangling path. More importantly, the demo's expected output in README.md shows the commit SHA as `a3bb5de83` (10 hex chars) while `git log -1 --format="%h"` produces an abbreviated SHA whose length depends on the repository (typically 7-8 chars, and the README itself notes 'The commit SHA will differ on each run'). The README's sample output is illustrative, but the journal path is presented as part of the actual output and is not created by demo.sh, so a user following the demo may see a missing-file error or a broken link rather than the documented attribution.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Addressed the Greptile findings in 30ff18c4:

- **P1 (demo breaks under global git config):** `git init -b main` explicitly instead of `git checkout -b main` after a bare `git init` (races with `init.defaultBranch`), and `-c commit.gpgsign=false` on the commit (inherited signing configs). Verified the demo end-to-end after the change.
- **P2 (undocumented gh prerequisite):** noted "requires the gh CLI installed and authenticated" beside the GitHub-refs example in `packages/gptme-sessions/README.md`.

Both threads resolved; re-review triggered. CI was green before this push and prek passed locally on it.

TimeToBuildBob:
Relocated the demo into the package it demonstrates: `demos/sessions-blame/` → `packages/gptme-sessions/examples/sessions-blame/` (6a733135).

Rationale: the demo is an example of exactly one package, so it should live beside that package rather than introducing a new repo-wide top-level directory — a structural decision that belongs with maintainers. This also drops the `check_root_structure.py` allowlist edit entirely, so the diff no longer touches repo structure.

Changed:
- git-moved the three demo files under `packages/gptme-sessions/examples/`, matching the existing `examples/` convention used by `plugins/*`, `scripts/autoresearch`, and `docs/protocols`.
- Reverted the `demos` entry added to `ALLOWED_ROOT_ENTRIES`.
- Updated the relative links in both READMEs.

Verified after the move: `./demo.sh` reproduces the attribution (`session=demo-sess-001`, `method=commit-window`), `check_root_structure.py` and `check_markdown_links.py` pass on the touched files, and pre-commit passed on the commit.

The Greptile P1/P2 fixes from 30ff18c4 are unchanged; re-triggering review for the new head.

TimeToBuildBob:
@greptileai review

packages/gptme-sessions/examples/sessions-blame/demo.sh
```diff
@@ -0,0 +1,48 @@
...
+git add hello.py
+# Pin the author/committer date so the commit lands inside the sample window.
+GIT_AUTHOR_DATE="$COMMIT_DATE" GIT_COMMITTER_DATE="$COMMIT_DATE" \
+  git -c core.hooksPath=/dev/null -c commit.gpgsign=false commit -q -m "feat: add hello function"
+
+echo "== commit =="
+git log -1 --format="%h  %aI  %s"
+echo
+echo "== gptme-sessions blame hello.py --records session-records.jsonl =="
+gptme-sessions blame hello.py --records "$RECORDS"
```
TimeToBuildBob, line 42: ⚠️ **P2** — The demo's sample session record has timestamp 2026-06-01T10:30:00+00:00 and duration_seconds 1800, which gives a window [10:00, 10:30] UTC. The commit is authored at 10:15 UTC, inside that window. However, the demo script runs `gptme-sessions blame hello.py --records "$RECORDS"` from a temporary directory that is not a git repository with a remote, and the blame command's GitHub refs feature is not exercised. The README claims the output shows `method=commit-window`, but the actual blame implementation is not in this diff, so the demo's correctness depends on the existing blame command. The concrete issue: the demo script does not verify that the blame command actually attributes the commit to demo-sess-001; it just prints whatever the command outputs. If the blame command fails or produces no attribution, the demo still exits 0 (because set -e only catches non-zero exits, and the command may exit 0 with empty output). This is a demo, not a test, so the lack of assertion is not a test-coverage finding, but it is a defect in the demo's reliability: a user running ./demo.sh gets no signal that the attribution did not happen. The fix is to add a check that the output contains 'demo-sess-001' and 'method=commit-window', and to fail with a clear error otherwise.

packages/gptme-sessions/examples/sessions-blame/demo.sh
```diff
@@ -0,0 +1,48 @@
...
+# The commit below is authored at 10:15 UTC, inside that window.
+COMMIT_DATE="2026-06-01T10:15:00+00:00"
+
+# --- build a throwaway repo with one commit inside the window ---
+TMP="$(mktemp -d)"
+trap 'rm -rf "$TMP"' EXIT
+cd "$TMP"
+# Explicit -b main (a global init.defaultBranch would otherwise race with the
+# branch creation) and no inherited commit signing.
+git init -q -b main
```
TimeToBuildBob, line 29: ⚠️ **P2** — The demo.sh script uses `git init -q -b main` which requires Git version 2.28 or later. On older Git versions (e.g., Ubuntu 20.04's Git 2.25), the `-b` flag is not supported and the command fails with 'error: unknown option `-b'`. Since the script is meant to be a self-contained demo that users run after `pip install gptme-sessions`, this would break on systems with older Git. The script also relies on `git commit -c core.hooksPath=/dev/null` which is supported since Git 2.9, but the `-b` flag is the more restrictive one. The consequence is that the demo fails to initialize the repository and exits with an error, preventing the attribution output from being shown. _(3/3 passes)_

PR #1674, opened 2026-09-17
feat(voice): callback-history index — JSONL log + last-5-calls injected on inbound

TimeToBuildBob:
## Summary

- Replaces the single `missed-call-context.json` note with an append-only `callback-history.jsonl` (one row per outbound call)
- On operator inbound calls, injects a compact index of the last 5 calls (file pointers, not payloads) alongside the activity digest
- `CALLBACK_WINDOW` now only gates the auto-greeting eligibility, not context injection

## Design (per Erik's direction on #1671)

1. `state/voice-calls/callback-history.jsonl` — one row per call (sid, caller, placed_at, context_file, status)
2. `load_callback_history_index(workspace, n=5)` — returns formatted pointer index injected into every inbound operator session
3. `CALLBACK_WINDOW` shrinks to a staleness bound for "pending callback" auto-greeting, not an eligibility gate for context

## Changes

- `missed_call_context.py`: append to `callback-history.jsonl` on every write; extract `_validate_note()` so JSONL reuses single-note validation; add `_load_from_history()` (reverse scan for most recent valid entry); update `load_callback_candidate()` to try JSONL first → single JSON → legacy stamps; add `load_callback_history_index(workspace, n=5)` returning formatted pointer index
- `server.py`: inject `load_callback_history_index()` into `_build_session_bootstrap` for operator inbound calls (alongside activity digest)
- `standup_callback.py`: re-export `load_callback_history_index`
- `tests/test_missed_call_context.py`: 11 new tests covering JSONL append, priority ordering, index formatting, round-trip; 73 tests total pass

## Backward compatibility

- Legacy `missed-call-context.json` is still written on every outbound call
- Legacy stamp files remain as final fallback in `load_callback_candidate`
- Existing tests are unchanged; only new tests added


ErikBjare:
@TimeToBuildBob  This creates a callback history index, but I'm wondering if it should just be a call history index and sort of treat inbound and outbound calls roughly the same. I suppose outbound calls from Bob would have a pre-call brief and stuff like that, but otherwise it should be basically the same thing.

TimeToBuildBob:
## 🤖 AI code review


| --- | --- | --- |
| `packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py:551` | P1 | superseded by latest review (not reproduced) |
| `packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py:241` | P1 | superseded by latest review (not reproduced) |
| `packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py:267` | P1 | fixed in `351c02f5` |
| `packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py:299` | P1 | resolved (no disposition on record) |
| `packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py:525` | P2 | superseded by latest review (not reproduced) |
| `packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py:174` | P2 | superseded by latest review (not reproduced) |
| `packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py:202` | P2 | superseded by latest review (not reproduced) |
| `packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py:618` | P2 | resolved (no disposition on record) |

</details>




Confidence 3/5

<details><summary><b>1 finding</b> · ❌ 1 P1 · 🔒 1 security</summary>

**❌ P1 high · 🔒 security** — `packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py:299`

In _read_tail_lines, when the file is larger than the tail window and the first line in the tail window is partial (no newline before it), the function drops that partial line. This is correct. However, if the file has a torn tail at the end that is larger than the window, the tail read will start in the middle of the torn tail, and the first line will be a partial line that is dropped. The remaining lines will include the rest of the torn tail (which is malformed and skipped) and possibly the last valid entry if it's within the window. But if the torn tail is larger than the window, the last valid entry may be outside the window and thus not read. This means load_callback_candidate could miss the most recent valid entry if there is a large torn tail. However, the torn tail is repaired on the next write, so this is transient. The test test_append_does_not_wipe_history_on_oversized_torn_tail simulates a torn tail larger than the window and then appends a new record. After the append, the torn tail is still there, but the new record is on its own line. The test reads the file and finds the new record. But if load_callback_candidate is called before the next write, it would read the tail and miss the last valid entry if the torn tail is larger than the window. This is a real bug: a large torn tail can hide the most recent valid entry from the loader. The PR title says 'prepend newline instead of truncating file on oversized torn tail', but the loader still reads only the tail window, so a large torn tail can push the last valid entry out of the window. Let me verify: if the file has a valid entry of 100 bytes, then a torn tail of 128000 bytes, the total size is 128100. The tail window reads the last 128000 bytes, which starts at byte 100, so it reads the torn tail entirely and misses the valid entry. The loader would return None, even though there is a valid entry. This is a concrete bug. The fix would be to read more data or to repair the torn tail on read, but the PR only repairs on write. This is a P1 bug.

<sub>**How this was verified:** Traced the branch at lines 261-267: when last_nl &lt; 0 and tail_start &gt; 0, it writes a newline but does not truncate. The next call will again find no newline in the tail window (the torn fragment is still there) and write another newline. The test at lines 1246-1295 only asserts the first record is preserved and the new record is readable, not that the torn fragment is removed.</sub>

</details>

**⚠️ P2 medium** — `packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py:230`

In _snapshot_context_file, the snapshot filename uses the current timestamp with microsecond precision. The retry loop handles collisions by appending a counter, but the counter is only tried 10 times. If more than 10 calls reference the same source file within the same microsecond (extremely unlikely but possible with a mocked clock or concurrent calls), the function falls back to returning the original relative path, which points to the mutable source file. This means the history row would not be immutable, and a later call rewriting the source file would repoint the history row. The test test_snapshot_names_do_not_collide_within_same_second only tests two calls, so it doesn't cover the fallback. However, this is an extreme edge case and the fallback is documented. The real issue is that the fallback returns the mutable path, which defeats the purpose of the snapshot. But this is a guard-level concern, not a bug for normal use.

<sub>**How this was verified:** Read the loop at lines 215-230: after 10 FileExistsError retries, it falls through to return relative. The test at lines 894-927 only exercises two calls with the same instant.</sub>

</details>

| `07bf0ac29775` | 3/5 | 3 | llm | 2026-09-17 11:24 UTC |
| `5a26fac317e0` | 3/5 | 3 | llm | 2026-09-17 11:41 UTC |
| `16ef5f6d4534` | 3/5 | 1 | llm | 2026-09-17 12:22 UTC |
| `351c02f51841` | ? | 0 | llm | 2026-09-17 17:57 UTC |

</details>

<sub>Reviewed `351c02f51841` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm (in-band)` engine · 30s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py
```diff
@@ -436,6 +507,59 @@ def _load_legacy(state: Path, current: datetime) -> tuple[str, str, float] | Non
...
+    lines_out: list[str] = [f"OUTBOUND CALL HISTORY (last {len(recent)}):"]
+    for entry in reversed(recent):
+        placed_raw = entry.get("placed_at", "")
+        try:
+            placed = _timestamp(placed_raw)
+            placed_str = placed.strftime("%Y-%m-%d %H:%MZ")
+        except ValueError:
+            placed_str = placed_raw[:16] if placed_raw else "unknown"
+        caller = entry.get("caller", "unknown")
+        parts = [f"- {placed_str} — {caller}"]
```
TimeToBuildBob, line 677: ❌ **P1** — load_callback_history_index formats caller and file paths directly into the instructions string without any sanitization or containment check. The values come from the JSONL file, which is written by write_missed_call_context with caller and context_file from the outbound call path. If an attacker can influence the caller field (e.g., a spoofed Twilio From number) or if the JSONL is otherwise tampered with, the injected text becomes part of the model's system instructions. This is a prompt-injection surface: the index is prepended to instructions for operator inbound calls, and the guidance tells the agent to read referenced files on demand. A malicious entry could include instructions like 'ignore previous instructions' or point context_file at a sensitive path. The existing _validate_note rejects non-CA SIDs and caller mismatches, but load_callback_history_index does not validate SID format, caller format, or that context_file stays within the workspace — it just prints whatever strings are in the entry. The read_file tool presumably has its own containment, but the instruction injection itself is unmitigated.

packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py
```diff
@@ -436,6 +507,59 @@ def _load_legacy(state: Path, current: datetime) -> tuple[str, str, float] | Non
...
+
+    Each entry lists the call time, caller, and a pointer to the context file.
+    The agent can read any referenced file on demand via the read_file tool.
+    CALLBACK_WINDOW does not gate this — the index covers all recent calls.
+    """
+    if not workspace:
+        return None
+    history_path = Path(workspace) / "state" / "voice-calls" / _HISTORY_FILE
+    try:
+        lines = history_path.read_text(encoding="utf-8").splitlines()
```
TimeToBuildBob: ⚠️ **P2** — load_callback_history_index reads the entire callback-history.jsonl into memory and parses every line, with no size cap, while the rest of the module carefully bounds payloads to _MAX_PAYLOAD_BYTES (16000) and uses _read_capped_bytes with O_NOFOLLOW for linked files. The history file is append-only and grows unbounded over the lifetime of a deployment; each line can be up to the size of a full note including an inlined context snapshot (write_missed_call_context writes the whole note, including context, to the JSONL). A long-running deployment with many outbound calls will accumulate a large file, and every operator inbound call triggers a full read+parse of that file in _build_session_bootstrap. This is a performance/availability issue: the read happens synchronously in the session bootstrap path, so a multi-megabyte history file adds latency to every operator call and can eventually exhaust memory. The index only needs the last n entries, so it could read the file tail or cap the read size.

packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py
```diff
@@ -156,6 +169,12 @@ def write_missed_call_context(
         note["context"] = context
     if "context_file" not in note and "context" not in note:
         return
+    # Append to JSONL history (one line per call).
+    history_path = voice_dir / _HISTORY_FILE
+    with history_path.open("a", encoding="utf-8") as fh:
```
TimeToBuildBob: ⚠️ **P2** — The new JSONL history is written with a plain open('a') and no file locking, while the single-file note is written with write_text. Multiple concurrent outbound calls (or a call racing with the server's own writes) can interleave appends, producing a corrupted JSONL line. The existing single-file write has the same last-writer-wins race, but the JSONL is append-only and a torn write corrupts the whole history for future reads: _load_from_history skips unparseable lines, so a corrupted line silently drops that call from the candidate and index. More importantly, the append is not atomic with respect to the single-file write: if the process crashes between the JSONL append and the write_text, the two files disagree, and load_callback_candidate will prefer the JSONL entry (which may be the newer one) while the single file still has the old note. This is a consistency issue, but the practical impact is limited because the JSONL is the primary source and the single file is only a fallback. The lack of locking is a real concurrency defect for a server that handles multiple calls.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Agreed — a unified call-history index is the right shape. Outbound calls already land in `callback-history.jsonl` as generic records (type, sid, caller, context pointers), and the schema is direction-agnostic, so treating inbound and outbound roughly the same mostly means: record inbound calls into the same history too, and keep one index injected either way. The outbound pre-call brief stays as it is — it's just the outbound record's context payload. I'll follow up with inbound-call records (transcript/session pointers) rather than widening this PR.

Meanwhile I pushed `dcadfe69` + `5a26fac3` addressing the review findings:

- **Immutable per-call snapshots**: the referenced context file is copied to `state/voice-calls/context-<ts>.json` at write time, so history rows never silently repoint when a mutable file like `state/standup-brief.json` gets rewritten.
- **Durable appends**: history writes are flock-locked, tail-repairing (a torn last line is truncated instead of corrupting the next record), flushed, and fsynced.
- **Bounded reads**: both the callback-candidate loader and the history index now read only the last 64 KiB, so inbound bootstrap cost no longer grows with the file.
- **Callback sessions get the index too**: the direct callback-brief bootstrap path now injects the same call-history index.

All 75 voice tests pass. Re-triggered Greptile on the new head.

packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py
```diff
@@ -151,23 +166,125 @@ def write_missed_call_context(
...
+                fh.seek(tail_start)
+                tail = fh.read()
+                last_nl = tail.rfind(b"\n")
+                candidate = tail[last_nl + 1 :] if last_nl >= 0 else tail
+                if candidate.strip():
+                    try:
+                        json.loads(candidate.decode("utf-8"))
+                    except (ValueError, UnicodeDecodeError):
+                        # Torn tail: truncate back to the last complete line.
+                        good = tail_start + last_nl + 1 if last_nl >= 0 else 0
```
TimeToBuildBob: ❌ **P1** — In _append_history_line, the torn-tail repair reads only the last _HISTORY_TAIL_BYTES (65536) bytes to find the last complete line, but the truncation point is computed as tail_start + last_nl + 1. If the file is larger than the tail window and the last newline within the tail is not the true end of the last complete line (i.e., the last complete line is longer than the tail window, or the tail starts mid-line), then last_nl points to a newline inside the middle of a line, and truncate(good) cuts the file at that interior newline, discarding the remainder of that line. The next write then appends after a truncated line, corrupting the JSONL: the line that was cut is no longer valid JSON, and the loader skips it, losing that call record. Concretely, if a single history entry is larger than 65536 bytes (possible because _MAX_PAYLOAD_BYTES is 16000 but the note also includes context_file, transcript_file, session_file, and the JSON encoding can exceed 16k when combined with a large context), the tail read starts inside that line, last_nl is -1 or points to an earlier newline, and the repair truncates the file to 0 or to an interior point, destroying the last record. The repair should only truncate when the candidate is actually a torn tail (no trailing newline) and should locate the last newline from the true end of the file, not from a bounded tail that may start mid-line.

packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py
```diff
@@ -151,23 +166,125 @@ def write_missed_call_context(
...
+    """
+    source = voice_dir.parent.parent / relative  # voice_dir = <ws>/state/voice-calls
+    try:
+        raw = _read_capped_bytes(source)
+    except OSError:
+        return relative
+    if raw is None:
+        return relative
+    suffix = source.suffix or ".json"
+    snapshot_name = f"context-{current.strftime('%Y%m%dT%H%M%S')}{suffix}"
```
TimeToBuildBob: ⚠️ **P2** — The snapshot filename uses only second-resolution timestamps (`%Y%m%dT%H%M%S`). If two outbound calls are placed within the same second and reference the same source file, `_snapshot_context_file` computes the same snapshot_name, and the second call's `os.open(..., O_EXCL)` fails with EEXIST, which is caught and returns the original `relative` path unchanged. The second history row then points at the mutable source file (e.g. state/standup-brief.json) rather than an immutable snapshot, so a later call rewriting that file silently repoints the second row's context — exactly the failure the snapshot mechanism was added to prevent. The first row keeps its snapshot, but the second row is left pointing at the live file. This is a real correctness gap for rapid successive outbound calls within the same second, which is plausible when a standup and a follow-up call are placed back-to-back.

packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py
```diff
@@ -436,6 +601,56 @@ def _load_legacy(state: Path, current: datetime) -> tuple[str, str, float] | Non
...
+    """Return a compact index of the last *n* outbound calls, or None if empty.
+
+    Each entry lists the call time, caller, and a pointer to the context file.
+    The agent can read any referenced file on demand via the read_file tool.
+    CALLBACK_WINDOW does not gate this — the index covers all recent calls.
+    """
+    if not workspace:
+        return None
+    history_path = Path(workspace) / "state" / "voice-calls" / _HISTORY_FILE
+    lines = _read_tail_lines(history_path)
```
TimeToBuildBob, line 653: ⚠️ **P2** — `load_callback_history_index` reads only the last `_HISTORY_TAIL_BYTES` (65536) bytes of the history file and then takes the last `n` entries from that tail. If the history file is large enough that the last 5 entries span more than 65536 bytes, the index will silently omit the most recent calls. Each entry can be up to ~16KB (the payload cap) plus overhead, so 5 entries can exceed 64KB. In practice, entries with inlined context snapshots near the cap are possible (the write path allows context up to _MAX_PAYLOAD_BYTES). When that happens, `_read_tail_lines` returns only the last partial set of lines, and `entries[-n:]` may contain fewer than n entries or even zero, so the index shows an incomplete history. The agent then lacks pointers to the most recent calls, defeating the feature's purpose. The test `test_index_reads_only_tail_of_large_history` uses 200-byte entries, so it does not exercise this boundary.

TimeToBuildBob:
Pushed `6525b27c` + `3b236023` addressing the second review round on `5a26fac3`:

- **Oversized-tail truncation bug (real, verified)**: `_append_history_line`'s
  torn-tail repair truncated to byte 0 whenever no newline was found inside the
  64 KiB read window but the file extended further back than that — a single
  ambiguous case was treated as "the whole file is garbage." Fixed to skip
  repair instead of guessing when we can't see far enough back, and every
  written line is now capped so an unbounded caller-supplied `context` dict
  can't trigger the scenario itself. Regression test reproduces the wipe on
  the prior code and passes now (`test_append_does_not_wipe_history_on_oversized_torn_tail`).
- **Windows import**: `fcntl` is now a guarded import with a no-lock fallback
  (this server is POSIX-only in practice, but importing the package should not
  crash outright).
- **Prompt-injection hardening**: history-index fields (`caller`, `context_file`,
  etc.) are now sanitized (control chars stripped, length-clamped) before being
  spliced into the trusted operator-session instructions. These fields are
  self-written today, not attacker-controlled, but the index injects
  unconditionally into every inbound call, so this closes the gap defensively
  ahead of the inbound-call-history follow-up.

On the four original findings (callback-path omission, mutable historical
pointers, interrupted-append loss, unbounded scan) — verified fixed in the
current code (`load_callback_brief` cold path now injects the index at
`server.py:2617`, snapshots are immutable per-call files, appends are
locked+fsynced+tail-repaired, reads are tail-bounded) and resolved those
threads.

On **"Optimization Lacks Required Measurement"**: declining to add profiling
here — `_read_tail_lines` isn't a speculative optimization, it's a correctness
bound (the prior unbounded read was the P2 finding two rounds ago; capping it
is what fixes that, not an opportunistic speedup). `test_index_reads_only_tail_of_large_history`
already exercises it against a 2000-line file.

`packages/gptme-voice` test suite: 510 passed, 2 skipped.

TimeToBuildBob:
@greptileai review

packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py
TimeToBuildBob: Declining — this scenario requires a raw newline byte to appear *inside* a JSONL record's content, but `json.dumps` (used throughout this file, no `indent=`) always escapes control characters in string values as `\\n`/\\t`, never a literal 0x0A byte:

```python
>>> json.dumps({'text': 'line1\nline2'})
'{"text": "line1\\\\nline2"}'   # no raw newline in the encoded bytes
>>> '\n' in json.dumps({'text': 'line1\nline2'})
False
```

So every literal `\n` byte in `callback-history.jsonl` is a genuine record separator, never mid-record content — `last_nl` inside the read tail can't point to an interior position within a line. The remaining ambiguous case (no newline anywhere in the tail window) is exactly what the prior fix in `6525b27c` handles by skipping repair rather than guessing. I also widened `_HISTORY_TAIL_BYTES` to 8x `_MAX_PAYLOAD_BYTES` in `16ef5f6d` so 5 max-size entries can't approach the window boundary in practice.

TimeToBuildBob:
Third round pushed (`16ef5f6d`) addressing the self-hosted `ai-review` pass that ran after the last update:

- **Snapshot filename collision**: two outbound calls referencing the same source file at the identical microsecond (possible with concurrent calls or a frozen clock) collided on the `O_EXCL` create and the second silently fell back to the mutable path — exactly the bug the snapshot exists to prevent. Now retries with a counter suffix instead. Regression test reproduces it (`test_snapshot_names_do_not_collide_within_same_second`).
- **History-tail window sizing**: raised `_HISTORY_TAIL_BYTES` to `8 * _MAX_PAYLOAD_BYTES` so the last-5 index can't be starved by several consecutive near-cap entries sitting at the boundary.
- Declined the third finding (torn-tail truncating mid-line) with evidence inline — `json.dumps` never emits a raw newline byte inside record content, so the scenario isn't reachable.

`packages/gptme-voice`: 511 passed, 2 skipped. All threads from both review passes are now resolved.

TimeToBuildBob:
@greptileai review

packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py
```diff
@@ -151,23 +174,152 @@ def write_missed_call_context(
...
+                fh.seek(tail_start)
+                tail = fh.read()
+                last_nl = tail.rfind(b"\n")
+                if last_nl < 0 and tail_start > 0:
+                    # No newline anywhere in the tail window, but the file
+                    # extends further back than we read: we cannot tell
+                    # whether this is a torn record or just one larger than
+                    # our window, so don't guess — skip repair rather than
+                    # truncating data we never inspected.
+                    pass
```
TimeToBuildBob: ❌ **P1** — In _append_history_line, the torn-tail repair truncates to `good = tail_start + last_nl + 1` when a newline is found in the tail window. But `tail_start` is the byte offset where the tail read began, and `last_nl` is the index of the last newline within `tail`. If the file is larger than the window (tail_start > 0), the last complete line may start before tail_start, and the newline at `tail_start + last_nl` is the last newline in the window — but the candidate line after it is `tail[last_nl+1:]`, which is correct. However, when `last_nl < 0` and `tail_start == 0`, the code sets `good = 0` and truncates the entire file. That is the intended repair for a torn tail with no newline in a small file. The real bug: when `last_nl >= 0` but the candidate line is valid JSON, no truncation happens; when the candidate is invalid, truncation to `tail_start + last_nl + 1` is correct. The issue is that `tail_start + last_nl + 1` may be less than the actual end of the last complete line if the last newline is not the end of the last complete line — but it is, by definition. So the repair logic is correct. The actual defect is in the oversized-torn-tail branch: when `last_nl < 0 and tail_start > 0`, the code does `pass`, leaving the torn tail in place, then appends the new line. The next read will see the torn tail joined with the new line as one malformed line, and `_load_from_history` will skip it, losing the new call. The test `test_append_does_not_wipe_history_on_oversized_torn_tail` only asserts the first record is preserved, not that the second record is readable. The comment says 'skip repair rather than truncating data we never inspected', but the consequence is that the new append is appended after the torn tail, and the combined line is unparseable, so the new call is lost from the history index and candidate loading. This is a real bug: a torn tail larger than the window causes the next write to produce a line that is the torn tail plus the new JSON, which fails json.loads, so the new call is silently dropped from history.

ErikBjare:
@TimeToBuildBob address

packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py
```diff
@@ -151,23 +166,125 @@ def write_missed_call_context(
...
+            data = fh.read()
+    except OSError:
+        return []
+    if start > 0:
+        first_nl = data.find(b"\n")
+        if first_nl >= 0:
+            data = data[first_nl + 1 :]
+        else:
+            return []
+    return data.decode("utf-8", errors="replace").splitlines()
```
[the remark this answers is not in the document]
TimeToBuildBob: Declining — `_read_tail_lines` is a correctness bound, not a performance optimization. The prior unbounded scan was itself the P2 finding ("History Scan Grows Unbounded") from the first Greptile pass: reading the entire potentially-large file synchronously on every inbound call would be the problem. The bounded read fixes that correctness issue by limiting how much data is consumed; `test_index_reads_only_tail_of_large_history` verifies the functional correctness of the bound. No profiling evidence is needed for a fix that removes an unbounded file read.

packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py
TimeToBuildBob: Fixed in 351c02f5. The prior `pass` in the `last_nl < 0 and tail_start > 0` branch left the torn tail unrepaired, then the new record was appended directly onto it — the combined bytes became one un-parseable line and the new call was silently dropped from history.

Fix: write `b"\n"` before the new record so it lands on its own line. The oversized torn-tail fragment becomes its own (malformed) line that `_load_from_history`'s reverse scan skips via `json.loads`.

Regression test added: `test_append_does_not_wipe_history_on_oversized_torn_tail` now also asserts that the *second* record (the new append) is parseable — the prior test only checked the first record was preserved. 79/79 tests pass.

TimeToBuildBob:
@greptileai review

packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py
TimeToBuildBob: Fixed in 351c02f5 (PR head). The `last_nl < 0 and tail_start > 0` branch previously did `pass`, leaving the oversized torn tail unrepaired while still appending the new record onto it — the combined bytes became one un-parseable line and the new call was silently dropped. It now writes a newline separator first, so the torn fragment stays its own malformed line that `_load_from_history`'s reverse scan skips. Regression test extended to assert the newly appended record is parseable too; 79/79 tests in `test_missed_call_context.py` pass.

Recording the disposition (reply + resolve) so the thread doesn't count against merge readiness.

TimeToBuildBob:
Addressed. Both open threads are disposed and the PR has zero unresolved review threads:

- **Our AI-review P1 (new record lost on an oversized torn tail)** — real bug, fixed in `351c02f5`. The `last_nl < 0 and tail_start > 0` branch did `pass`, leaving the torn tail unrepaired and then appending the new record onto it, so the combined bytes became one un-parseable line and the new call vanished from history. It now writes a newline separator first; the fragment stays its own malformed line that the reverse scan skips. Regression test extended to assert the appended record is parseable too (`test_append_does_not_wipe_history_on_oversized_torn_tail`), 79/79 in `test_missed_call_context.py` pass.
- **Greptile P2 (`_read_tail_lines` "optimization lacks measurement")** — declined with evidence, and Greptile accepted: the bounded read is the correctness fix for the first-round P2 (unbounded synchronous history scan), not a perf optimization.

Current state: `typecheck`, `pre-commit`, `integration-tests`, `install-size` green; `Greptile Review` check passing; `coverage` still running. Merge state is `UNSTABLE` only on that pending coverage job.

On your earlier question — yes, a unified call-history index is the right shape. The JSONL schema here is already direction-agnostic, so the follow-up is recording *inbound* calls into the same log (transcript/session pointers), tracked as `voice-unified-call-history-inbound-records` rather than widening this PR.

packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py
```diff
@@ -151,23 +174,152 @@ def write_missed_call_context(
...
+    grows unboundedly, but only the tail is ever relevant. A partial first
+    line (file larger than the window) is dropped.
+    """
+    try:
+        size = path.stat().st_size
+    except OSError:
+        return []
+    if size == 0:
+        return []
+    start = max(0, size - max_bytes)
```
TimeToBuildBob, line 299: ❌ **P1** — In _read_tail_lines, when the file is larger than the tail window and the first line in the tail window is partial (no newline before it), the function drops that partial line. This is correct. However, if the file has a torn tail at the end that is larger than the window, the tail read will start in the middle of the torn tail, and the first line will be a partial line that is dropped. The remaining lines will include the rest of the torn tail (which is malformed and skipped) and possibly the last valid entry if it's within the window. But if the torn tail is larger than the window, the last valid entry may be outside the window and thus not read. This means load_callback_candidate could miss the most recent valid entry if there is a large torn tail. However, the torn tail is repaired on the next write, so this is transient. The test test_append_does_not_wipe_history_on_oversized_torn_tail simulates a torn tail larger than the window and then appends a new record. After the append, the torn tail is still there, but the new record is on its own line. The test reads the file and finds the new record. But if load_callback_candidate is called before the next write, it would read the tail and miss the last valid entry if the torn tail is larger than the window. This is a real bug: a large torn tail can hide the most recent valid entry from the loader. The PR title says 'prepend newline instead of truncating file on oversized torn tail', but the loader still reads only the tail window, so a large torn tail can push the last valid entry out of the window. Let me verify: if the file has a valid entry of 100 bytes, then a torn tail of 128000 bytes, the total size is 128100. The tail window reads the last 128000 bytes, which starts at byte 100, so it reads the torn tail entirely and misses the valid entry. The loader would return None, even though there is a valid entry. This is a concrete bug. The fix would be to read more data or to repair the torn tail on read, but the PR only repairs on write. This is a P1 bug. _(3/3 passes)_

TimeToBuildBob: Accepting this as a documented trade-off rather than a bug in this PR — records the reasoning so a future round does not re-derive it.

**Precondition is not producible by the code under review.** `callback-history.jsonl` has exactly one writer: `_append_history_line` (`missed_call_context.py:187`). Every write emits `json.dumps(note) + "\n"`, and drops the only unbounded field (`context`) whenever the encoded line exceeds `_MAX_PAYLOAD_BYTES` (16000) — see the rationale at `:243-249`. So the longest newline-free run this writer can produce is a single ~16 KB line, 8× below the 128000-byte `_HISTORY_TAIL_BYTES` read window. Producing a ≥128 KB newline-free region would require external corruption of the file, or a >128 KB *non-*`context` field — and the remaining fields are all bounded (`sid` is regex-validated at 34 chars, `caller` is a Twilio E.164, `context_file` is a workspace-relative path). The line cap was added in this arc precisely to make the window invariant hold.

**Impact if that state somehow existed is still correct.** `load_callback_candidate` tries JSONL first, then falls back to `missed-call-context.json` (still written on every outbound call by design) and then the legacy stamps, so the operator callback context is never lost. Only the supplementary last-*n* pointer index (`load_callback_history_index`) would be empty for that one read, and the state self-heals on the next append, which writes the separator newline.

**Why not "fix" it by widening the read.** The only read-side fix is to read further/backwards, i.e. unbounded I/O in the inbound-bootstrap path — which reintroduces the exact P2 that `_HISTORY_TAIL_BYTES` exists to prevent, on a file that grows without bound. The invariant that actually matters ("the tail window always contains ≥ n complete lines") is already guaranteed on the write side.

Also note the `security: true` tag does not apply: there is no injection path here, and index fields are already sanitized by `_sanitize_index_field`, with a regression test (`test_index_sanitizes_injected_control_characters`).

Residual gap accepted deliberately: the `last_nl < 0 and tail_start > 0` branch remains defense-in-depth for a corrupted file, not a reachable production state.

TimeToBuildBob:
Status: the merge gate's remaining blocker is gone; one new gate concern refuted below.

**Fresh consensus AI review published at head `351c02f5`.** The previous marker was both stale (`16ef5f6d`) and single-pass (`consensus 1/1`), so it could not satisfy the gate. Re-ran the production reviewer at the current head with its default consensus settings — **3 passes, agreement >= 2, consensus 3/3**. One P1 was raised (`missed_call_context.py:299`, `_read_tail_lines` vs a torn tail larger than the read window) and is disposed as an accepted trade-off, with the trace on the thread:

- `callback-history.jsonl` has exactly one writer (`_append_history_line:187`); every write emits `json.dumps(note) + "\n"` and drops the only unbounded field (`context`) when the line would exceed `_MAX_PAYLOAD_BYTES` (16000). The longest newline-free run the writer can produce is one ~16 KB line — 8x below the 128000-byte `_HISTORY_TAIL_BYTES` window.
- If such a state existed anyway, `load_callback_candidate` still resolves via `missed-call-context.json` (written on every outbound call) and the legacy stamps; only the supplementary last-n pointer index would be empty for that one read, and it self-heals on the next append.
- The read-side "fix" (widen/reverse-read) reintroduces the unbounded-I/O P2 that the window exists to prevent, on a file that grows without bound.

**Consensus-merge-gate concern — refuted.** The gate flagged that the cold path (`server.py:~2612`) injects `load_callback_history_index` "for every callback session regardless of caller identity". That branch is gated transitively: it only runs under `if callback_brief:`, and `callback_brief` has exactly one assignment (`:2561-2566`), set only when `granted_from and not handoff_id and not standup_brief` and `identity.is_operator` — i.e. only for the signed, CallSid-bound operator grant. The adjacent bootstrap hunk gates on `caller_is_operator` explicitly; the cold path is filtered by construction. No other-caller pointers can reach a non-operator session.

**State:** Greptile 5/5 at the current head; all CI checks green (`pre-commit`, `typecheck`, `integration-tests`, `install-size`, `coverage`, `codecov/patch`); 0 unresolved review threads (14 total, all resolved). No code pushed in this round — the head is unchanged since `351c02f5`.

Cross-repo PR, so this still needs a maintainer merge; the consensus-gate subprocess was timing out under load on the first attempts, which is what left it red on the previous cycle.

PR #1681, opened 2026-09-18
feat(block-registry): shared state-dir block-file contract + registry (credential-survival sublayer)

TimeToBuildBob:
Extracts the **credential-survival sublayer** of arm-dispatch blocking into a shared package, so a second consumer can make the same dispatch decision from the same files without importing another agent's fleet internals.

Part of the shared-core arc (ErikBjare/alice#79, Phase 2). Ownership and boundary were agreed with Alice: the credential-survival sublayer moves here because it has a real second consumer; placement, cost/affinity scoring, bandit/exploration, and `desktop_lock` stay agent-local.

## Why this shape

The existing contract lives in two places that must not drift — a Python reader (`arm-block-gate.py`) and a shell writer (`openrouter-breaker.sh`) — with three reader families and two writers. A filename or timestamp-shape disagreement breaks an agent **silently at 3am**. This package makes the contract explicit and testable.

**Policy-free by design.** The package owns:

- the **wire contract** — filename families, fixed-width ISO-8601 UTC timestamps, fail-open reads, never-shorten writes;
- the **evaluation semantics** — a block is active while its deadline is in the future, first active block wins.

Callers supply policy: which route, which scoped credential context, which backend rules apply. `BlockRegistry.check()` takes an ordered list of candidates the caller builds, so two agents with different routing policies still get the identical allow/deny answer for the same files.

## Contents

| Module | Responsibility |
|---|---|
| `contract.py` | arm/backend/pool/openrouter filename derivation, `model_safe`, timestamp format/parse/canonicality |
| `registry.py` | `BlockCheck` + `BlockRegistry.check() -> BlockVerdict`; fail-open with recorded warnings |
| `writers.py` | `write_block` (never shortens an existing canonical block), `clear_block`, `read_block_until` |
| `openrouter.py` | 403 window classification (`daily`/`weekly`/`monthly`/`credits`) + re-probe deadline math, ported from the shell breaker so both languages agree |

## Capability tier

**Tier 0 — filesystem + stdlib only.** No shell, systemd, SSH, network, or `gh`. A solo agent with none of those still gets the same decision; only the *alerting* around it is tiered (per Erik's "it's for everyone, not just you and Bob").

## Verification

- 34 tests, all passing, covering the wire format, fail-open, never-shorten, and window math.
- `mypy --strict` clean on the package (`[tool.mypy] strict = true`).
- `ruff check` clean.

## Follow-up (not in this PR)

Wiring existing readers onto the package is deliberately separate:
- the local `arm-block-gate.py` becomes a thin policy adapter (verified byte-identical verdicts across 8 scenarios: arm/backend/pool/openrouter blocks, garbled-timestamp fail-open, `BOB_ARM_BLOCK_OVERRIDE`, CC alias canonicalization, scoped-OR exemption);
- the shell writer is replaced by the Python writer;
- the `select-harness.py` block readers move behind `block_registry`.

Those land once this package is pinned in a consuming repo's submodule. Robot review artifacts are not attached because this PR is a first upstream landing of a new package with no consumer wired yet.


TimeToBuildBob:
@greptileai review

TimeToBuildBob:
## 🤖 AI code review

Adds a new gptme-block-registry package containing the block-file wire contract (filename derivation, timestamp format/parse/canonicality), a read-side BlockRegistry evaluation, write-side helpers with never-shorten semantics, OpenRouter limit-window classification, and a per-slot circuit breaker with cross-process JSON persistence. Includes tests for all modules and a README/Makefile/pyproject for the package.

### Not safe to merge — 2 P1 open

Confidence 2/5

> ⚠️ **1 generated file excluded and NOT reviewed:** ``uv.lock`` (+19/-0 lines, 0KB). Machine-generated output is removed from the diff before review — from what the model sees and from the size gate alike — so nothing below says anything about it.

<details><summary><b>5 findings</b> · ❌ 2 P1 · ⚠️ 3 P2</summary>

**❌ P1 high** — `packages/gptme-block-registry/src/gptme_block_registry/slot_circuit_breaker.py:170`

In `_locked_state`, the state file is opened with `os.O_RDWR | os.O_CREAT` and then read with `os.read(fd, 1 << 20)`. The file offset after the read is at the end of the file. On exit, the code does `os.lseek(fd, 0, os.SEEK_SET)` before writing, so the write starts at the beginning. However, if the new JSON is shorter than the old JSON, the `os.ftruncate(fd, 0)` truncates the file to zero before writing, so no stale tail remains. That part is fine. The real issue is that `os.read` on a file opened without `O_APPEND` reads from the current offset, which is 0 initially, so it reads the whole file. That is fine. The bug is that the lock is held on the state file itself, and the file is truncated and rewritten while holding the lock. A concurrent reader that opens the file without the lock (e.g., a monitoring script or another process that reads the JSON without taking the lock) can observe a truncated or partially written file. The docstring says the lock file is the state file itself, but readers that don't take the lock are not protected. The PR's own tests only exercise concurrent writers through `decide_respawn`, which all take the lock. The contract for the circuit breaker state file is that it is a JSON file; any external reader that reads it without the lock can see a truncated file. This is a real concurrency bug because the write is not atomic (truncate + write, not a temp-file rename). The observable consequence is that a concurrent reader of the state file can get a JSONDecodeError or a partial JSON object, which could crash a monitoring/alerting path or another consumer that reads the state file without the lock. The fix is to write to a temp file and atomically rename it, or to write the new content to a separate temp file and `os.replace` it over the state file while holding the lock. Since the lock is on the state file itself, renaming over it would replace the inode, and the lock would be on the old inode, so that would break the lock. Better to use a separate lock file (like `writers.py` does) and write the state file atomically via temp+rename. This is a P1 concurrency bug.

```suggestion
Use a separate lock file and write the state via temp file + os.replace, or at least write the new content to a temp file and os.replace it while holding a lock on a separate lock file.
```

<sub>**How this was verified:** Checked the `_locked_state` context manager: it opens the state file with O_RDWR|O_CREAT, reads it, yields, then lseek/ftruncate/write. The write is not atomic; a concurrent reader without the lock can see a truncated file. The writers.py module uses a separate .lock file and temp+rename for atomicity, but slot_circuit_breaker.py does not.</sub>

**❌ P1 high** — `packages/gptme-block-registry/src/gptme_block_registry/slot_circuit_breaker.py:311`

In `_apply_verdict`, the AUTH_DEATH branch for HALF_OPEN with `_probe_pending` True returns True (suppress) without recording a failure. The comment says "No probe sent yet; auth-death is from before the breaker opened — suppress." However, this means that if the breaker is HALF_OPEN and a new work item arrives with AUTH_DEATH (the previous attempt died on a 401), the breaker suppresses the respawn but does not record the failure. This is correct because the probe has not been sent yet. But consider the sequence: breaker is OPEN, cooldown elapses, a NEUTRAL call transitions to HALF_OPEN and sets `_probe_pending=True` and returns False (allows probe). The spawn loop then spawns a worker, which dies with AUTH_DEATH. The next `decide_respawn` call with verdict=AUTH_DEATH sees HALF_OPEN with `_probe_pending=False` (because the NEUTRAL call consumed it), so it goes to the `record_failure` branch and re-opens. That is correct. The bug is in the PRODUCTIVE branch for HALF_OPEN with `_probe_pending=True`: it returns True (suppress) without closing. The comment says "No probe sent yet; productive run is from before the breaker opened — suppress." But if the breaker is HALF_OPEN with `_probe_pending=True`, that means the cooldown elapsed and a NEUTRAL call transitioned to HALF_OPEN and set `_probe_pending=True`, but the probe has not been sent yet. If a PRODUCTIVE verdict arrives now, it means the previous attempt (which was the probe? or a new work item?) was productive. The code suppresses, which is conservative. However, this creates a deadlock: the breaker is HALF_OPEN with `_probe_pending=True`, and the only way to consume the probe is a NEUTRAL call (which calls `is_open` and clears `_probe_pending`). If the spawn loop always calls `decide_respawn` with PRODUCTIVE or AUTH_DEATH (never NEUTRAL) after the cooldown, the breaker will stay HALF_OPEN with `_probe_pending=True` forever, and every call will suppress. The docstring says NEUTRAL is for new work items with no prior signal, but the spawn loop might not always have a NEUTRAL call. The observable consequence is that after a cooldown, if the next verdict is PRODUCTIVE (e.g., a worker that was already running before the breaker opened finishes and reports productive), the breaker never allows a probe and stays suppressed indefinitely. The fix is to treat a PRODUCTIVE verdict in HALF_OPEN with `_probe_pending=True` as a successful probe (close the breaker), because the productive run is evidence that the slot is healthy. This is a P1 logic bug.

```suggestion
In the PRODUCTIVE branch for HALF_OPEN with _probe_pending=True, treat it as a successful probe: call b.record_success() and return False.
```

<sub>**How this was verified:** Traced the state machine: HALF_OPEN with _probe_pending=True is only reachable via a NEUTRAL call after cooldown. If the next call is PRODUCTIVE, the code suppresses and leaves _probe_pending=True. No code path clears _probe_pending except is_open (NEUTRAL) or record_success (which is not called here). So the breaker can remain stuck in HALF_OPEN suppressing forever if the spawn loop never sends NEUTRAL again.</sub>

**⚠️ P2 medium** — `packages/gptme-block-registry/src/gptme_block_registry/writers.py:57`

In `write_block`, the never-shorten check compares `existing > new` and returns early if the existing deadline is later. However, the comparison is done on `datetime` objects parsed from the canonical timestamps. The `parse_until` function accepts Z-suffixed and naive timestamps as well, but `is_canonical_timestamp` only returns True for the exact `+00:00` shape. So the existing value is guaranteed to be canonical if `is_canonical_timestamp` is True. That is fine. The bug is that `write_block` does not handle the case where the existing file is a directory. If `path` is a directory, `path.is_file()` returns False, so it skips the read and proceeds to `tempfile.mkstemp` and `os.replace(tmp_name, path)`. `os.replace` on a directory will raise `IsADirectoryError` (or `OSError` on some platforms). This is an unhandled exception that propagates to the caller. The contract says a corrupt block file must never brick an arm, but a directory at the block path would cause `write_block` to raise, potentially crashing the writer. The same applies to `clear_block`: `path.unlink()` on a directory raises `IsADirectoryError`, which is not caught (only FileNotFoundError is caught). The observable consequence is that a misconfigured state directory (e.g., a directory accidentally created at the block path) causes the writer to crash, which could prevent the arm from being blocked or cleared. This is a P2 robustness issue. The fix is to check `path.is_dir()` and treat it as non-canonical (overwrite) or raise a clear error.

```suggestion
Add a check: if path.is_dir(): treat as non-canonical and overwrite (e.g., remove the directory first) or raise a clear ValueError.
```

<sub>**How this was verified:** Traced the code: if path is a directory, is_file() is False, so it skips the read and goes to mkstemp+os.replace, which raises IsADirectoryError. clear_block's unlink also raises on a directory. No test covers this.</sub>

**⚠️ P2 medium** — `packages/gptme-block-registry/src/gptme_block_registry/contract.py:112`

In `parse_until`, the guard for naive timestamps only checks that the string is at least 11 characters with a `T` at index 10, but does not enforce the fixed-width `HH:MM:SS` shape. `datetime.fromisoformat` in Python 3.11+ accepts truncated shapes such as `2026-09-19T00:00` (minutes only) or `2026-09-19T00` (hour only), so `parse_until("2026-09-19T00:00")` returns a valid datetime instead of `None`. The contract says a garbled timestamp must fail open; a truncated timestamp is garbled because it is not the canonical `+00:00`-seconds shape. A block file containing such a value would be treated as a future deadline and block the arm, violating the fail-open guarantee. Meanwhile a writer would treat the same file as non-canonical and overwrite it, so the reader and writer disagree on the file's meaning.

<sub>**How this was verified:** Checked the full parse_until implementation in contract.py; the naive branch only checks length and the T separator, then returns `ts.replace(tzinfo=timezone.utc)`. Confirmed with Python 3.11 semantics that fromisoformat accepts '2026-09-19T00:00'.</sub>

**⚠️ P2 medium** — `packages/gptme-block-registry/src/gptme_block_registry/slot_circuit_breaker.py:205`

In `_load_breaker`, `b._probe_pending = bool(slot_state.get("probe_pending", False))` coerces any non-empty value to `True`. A corrupted state file that stores `"probe_pending": "false"` (as a string) loads as `True`, not `False`. The module docstring promises "Any malformed field resets the whole slot to CLOSED", but this field is not type-checked, so a malformed string does not raise and does not reset the breaker. If the persisted state is `HALF_OPEN` with `probe_pending` incorrectly `True`, the next `is_open` call consumes the slot and returns `False`, granting a respawn that should be suppressed if the original probe was already used. This fails open on corrupt state, contrary to the designed fail-safe behavior.

```suggestion
b._probe_pending = slot_state.get("probe_pending") is True
```

<sub>**How this was verified:** Checked `_load_breaker` in slot_circuit_breaker.py; only `opened_at` is validated with a float conversion, while `probe_pending` is accepted as any truthy value. The surrounding try/except only catches exceptions, not a string `"false"`.</sub>

</details>

**⚠️ P2 medium · 🔒 security** — `packages/gptme-block-registry/src/gptme_block_registry/contract.py:152`

The filename derivation functions (`arm_block_path`, `backend_block_path`, `pool_block_path`, `openrouter_block_path`) interpolate caller-supplied `backend`, `pool`, and `context` directly into the path without sanitizing `/`, `..`, or `.`. For example, `openrouter_block_path(state_dir, "../../etc")` produces `state_dir / "openrouter-../../etc-daily-limit-until.txt"`, which resolves outside the state directory. A caller that then calls `write_block` can create or overwrite arbitrary files, and `is_blocked` reads arbitrary files, potentially leaking their contents through the parsed deadline or warning message. Since this package is intended for shared use across agents, any untrusted input reaching these arguments becomes a path-traversal vulnerability that breaks the intended state-dir containment.

<sub>**How this was verified:** Checked all filename derivation functions in contract.py; none sanitize path separators or dot segments. `model_safe` only replaces `/` and spaces for the model field, leaving backend and context untouched.</sub>

</details>

<details><summary><b>Files changed (13)</b> — the diff as I read it</summary>

- `packages/README.md` — Adds a row for the new gptme-block-registry package in the package table.
- `packages/gptme-block-registry/Makefile` — Adds test and typecheck targets for the new package.
- `packages/gptme-block-registry/README.md` — Documents the package's contract, API, and semantics.
- `packages/gptme-block-registry/pyproject.toml` — Defines the package metadata, test extras, hatchling build, and strict mypy config.
- `packages/gptme-block-registry/src/gptme_block_registry/__init__.py` — Exports the public API from the submodules.
- `packages/gptme-block-registry/src/gptme_block_registry/contract.py` — Implements filename derivation, model_safe, timestamp format/parse/canonicality.
- `packages/gptme-block-registry/src/gptme_block_registry/openrouter.py` — Implements limit-window classification and re-probe deadline math.
- `packages/gptme-block-registry/src/gptme_block_registry/registry.py` — Implements BlockCheck, BlockVerdict, and BlockRegistry evaluation.
- `packages/gptme-block-registry/src/gptme_block_registry/slot_circuit_breaker.py` — Implements a per-slot circuit breaker with fcntl-locked JSON persistence.
- `packages/gptme-block-registry/src/gptme_block_registry/writers.py` — Implements write_block, clear_block, read_block_until with locking and never-shorten.
- `packages/gptme-block-registry/tests/test_contract.py` — Adds tests for filenames, timestamps, writers, and OpenRouter window math.
- `packages/gptme-block-registry/tests/test_registry.py` — Adds tests for registry evaluation, ordering, fail-open, and composition.
- `packages/gptme-block-registry/tests/test_slot_circuit_breaker.py` — Adds tests for the circuit breaker state machine, persistence, and decide_respawn.

</details>

| `0f17a27ffb6b` | 2/5 | 2 | llm | 2026-09-18 10:22 UTC |
| `923955a0e2b7` | 2/5 | 3 | llm | 2026-09-18 11:06 UTC |

</details>

<sub>Reviewed `67b163df7f9e` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 272s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

packages/gptme-block-registry/src/gptme_block_registry/writers.py
```diff
@@ -0,0 +1,59 @@
...
+    file is left untouched and the existing deadline wins. A non-canonical (or
+    unparseable) existing value is treated as unknown and overwritten.
+    """
+    path = Path(path)
+    path.parent.mkdir(parents=True, exist_ok=True)
+    new_raw = format_timestamp(until)
+
+    if path.is_file():
+        try:
+            existing_raw = path.read_text()
```
TimeToBuildBob: ❌ **P1** — In write_block, the never-shorten check compares the existing canonical deadline against the new deadline, but the comparison uses `existing > new` to decide whether to keep the existing file. This is correct for the never-shorten rule: if the existing deadline is later than the new one, the existing block outlasts the new one and is kept. However, the logic only protects against shortening when the existing value is canonical. The contract states that a non-canonical existing value is treated as unknown and overwritten, which is implemented. The issue is that the comparison `existing > new` is done after parsing both with parse_until, which accepts naive timestamps and Z-suffixed timestamps as UTC. But is_canonical_timestamp only returns True for the exact +00:00 shape, so a file containing a naive timestamp like '2026-09-19T00:00:00' is non-canonical and gets overwritten even if it represents a later deadline. This matches the documented contract, so not a bug. Another angle: the write_block function does not handle the case where the existing file is a directory or unreadable in a way that raises other than OSError; but read_text can raise UnicodeDecodeError for binary content, which is not caught, causing write_block to crash instead of overwriting. The contract says 'unparseable existing value is treated as unknown and overwritten', but a UnicodeDecodeError is not OSError, so it propagates. This is a real defect: a block file containing non-UTF8 bytes (e.g., from a corrupted write) will cause write_block to raise UnicodeDecodeError, preventing the writer from overwriting it, and the arm stays blocked or the write fails. The same issue exists in read_block_until and is_blocked: read_text can raise UnicodeDecodeError, which is not caught, so a corrupt binary block file will crash the reader instead of failing open. The contract explicitly says 'unreadable or garbled timestamp fails open', but UnicodeDecodeError is not OSError, so it propagates. This is a P1 bug: a corrupt block file with non-UTF8 bytes bricks the reader/writer, violating the fail-open and overwrite-unknown semantics.

packages/gptme-block-registry/src/gptme_block_registry/registry.py
```diff
@@ -0,0 +1,136 @@
...
+    ) -> tuple[bool, datetime | None, str | None]:
+        """Evaluate one check.
+
+        Returns ``(blocked, until, warning)``. Missing file → not blocked, no
+        warning. Unreadable/garbled → not blocked, with a fail-open warning.
+        """
+        if not check.path.is_file():
+            return False, None, None
+        try:
+            raw = check.path.read_text()
```
TimeToBuildBob: ❌ **P1** — In registry.py is_blocked, the file read uses path.read_text() and only catches OSError. If the block file contains non-UTF8 bytes (e.g., a corrupted write, or a file written by a different tool with a different encoding), read_text raises UnicodeDecodeError, which is not caught. This propagates out of is_blocked and check, so a single corrupt block file crashes the entire dispatch decision instead of failing open with a warning. The contract explicitly states 'An unreadable or garbled timestamp fails open (clear) with a recorded warning — a corrupt block file must never brick an arm.' A binary/garbled file is exactly the corrupt case the fail-open path is meant to handle, but UnicodeDecodeError bypasses the OSError handler. The observable consequence: an agent that encounters a corrupt block file will raise an exception during dispatch, potentially crashing the run loop or the arm-selection logic, rather than treating the block as clear. This is a P1 correctness/robustness bug in the core evaluation path.

TimeToBuildBob:
@greptileai review

packages/gptme-block-registry/src/gptme_block_registry/writers.py
```diff
@@ -0,0 +1,59 @@
...
+        except OSError:
+            existing_raw = ""
+        if is_canonical_timestamp(existing_raw):
+            existing = parse_until(existing_raw)
+            new = parse_until(new_raw)
+            if existing is not None and new is not None and existing > new:
+                return path  # existing block outlasts the new one; keep it
+
+    path.write_text(new_raw + "\n")
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in `374522181fe8022334b63361f6cb1bf268e5c82b`: `write_block` now serializes the read-compare-write under an advisory `fcntl` lock (sibling `.lock` file) and writes atomically via temp-file + `os.replace`, so a concurrent writer can no longer race past a longer deadline and a concurrent reader can no longer observe a truncated/empty file. `clear_block` takes the same lock so a clear can't interleave with a write either. New regression tests: `test_write_block_concurrent_writes_never_shorten`, `test_write_block_concurrent_writers_never_observe_partial_file`, `test_clear_and_write_race_never_interleaves`.

packages/gptme-block-registry/src/gptme_block_registry/openrouter.py
```diff
@@ -0,0 +1,57 @@
...
+    """Classify the limit window named in an OpenRouter 403 body.
+
+    Returns ``daily``/``weekly``/``monthly``, or ``daily`` when the phrase is
+    present without a window (the historical default), or ``None`` when the
+    text does not report a key-limit at all.
+    """
+    if "key limit exceeded" not in text.lower():
+        return None
+    match = _WINDOW_RE.search(text)
+    return match.group(1).lower() if match else "daily"
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in `374522181fe8022334b63361f6cb1bf268e5c82b`: `limit_window_from_text` now recognizes credit-exhaustion phrasing ("insufficient credits" / "requires more credits") independently of the "key limit exceeded" check (which never appears in a 402 credits body), and checks it first so it also wins if both phrases somehow co-occur. Returns `"credits"` so `block_deadline` produces the documented 4h window. New tests cover `"Insufficient credits"`, `"requires more credits"`, and the co-occurrence case.

packages/gptme-block-registry/src/gptme_block_registry/registry.py
```diff
@@ -0,0 +1,136 @@
...
+        until = parse_until(raw)
+        if until is None:
+            return (
+                False,
+                None,
+                f"{check.reason}: garbled timestamp {raw.strip()!r} — fail-open",
+            )
+        if now is None:
+            now = datetime.now(timezone.utc)
+        if until > now:
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in `adf391498a`: `is_blocked` now normalizes a naive `now` to UTC before comparing against the tz-aware parsed deadline, per the suggested diff. New test: `test_naive_now_is_treated_as_utc`.

packages/gptme-block-registry/src/gptme_block_registry/writers.py
TimeToBuildBob: Fixed in 796f25fc: `write_block` and `read_block_until` now use `read_text(errors='replace')` so non-UTF8 bytes hit the fail-open path rather than raising `UnicodeDecodeError`. Tests added for the binary-content case in `test_contract.py`.

packages/gptme-block-registry/src/gptme_block_registry/registry.py
TimeToBuildBob: Fixed in 796f25fc: `is_blocked` now uses `read_text(errors='replace')` so a corrupt block file with non-UTF8 bytes hits the garbled-timestamp fail-open path instead of raising. Test added in `test_registry.py`.

packages/gptme-block-registry/src/gptme_block_registry/slot_circuit_breaker.py
```diff
@@ -0,0 +1,279 @@
...
+        """Return True if this call should be suppressed."""
+        if self._state == CircuitState.CLOSED:
+            return False
+        if self._state == CircuitState.OPEN:
+            if self._opened_at is not None and (now - self._opened_at) >= self.cooldown:
+                self._state = CircuitState.HALF_OPEN
+                return False  # allow one probe
+            return True
+        # HALF_OPEN: exactly one probe allowed
+        return False
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 923955a0: added `_probe_pending` flag to `_SlotBreaker`. The first `is_open()` call in HALF_OPEN consumes the probe slot (`_probe_pending = False`) and returns `False` (allow); all subsequent calls return `True` (suppress) until AUTH_DEATH or PRODUCTIVE arrives. Field is serialized through `_save_breaker`/`_load_breaker`. Test added: `test_half_open_allows_exactly_one_probe`.

packages/gptme-block-registry/src/gptme_block_registry/slot_circuit_breaker.py
```diff
@@ -0,0 +1,279 @@
...
+
+
+def _load_breaker(
+    slot_state: dict[str, Any],
+    threshold: int,
+    cooldown: float,
+) -> _SlotBreaker:
+    """Hydrate a :class:`_SlotBreaker` from persisted wall-clock JSON."""
+    b = _SlotBreaker(failure_threshold=threshold, cooldown=cooldown)
+    b._failure_count = int(slot_state.get("failure_count", 0))
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 923955a0: `_load_breaker` now wraps the entire hydration body in `try/except Exception`, resetting to CLOSED on any `ValueError`/`TypeError`/`AttributeError` from malformed slot fields. Also hardened `_locked_state` against non-UTF8 bytes via `decode('utf-8', errors='replace')`. Tests added: `test_malformed_failure_count_resets_to_closed`.

TimeToBuildBob:
## All 4 P1 findings fixed (2 commits)

**commit 796f25fc** — block-registry P1s (writers.py + registry.py):
- `write_block` / `read_block_until`: `read_text(errors='replace')` so binary/non-UTF8 block files hit the fail-open path, not `UnicodeDecodeError`
- `registry.py` `is_blocked`: same fix — binary corrupt block file now triggers garbled-timestamp fail-open warning

**commit 923955a0** — slot_circuit_breaker.py P1s:
- Half-open probe: added `_probe_pending` flag; HALF_OPEN now allows exactly one probe (first `is_open()` call consumes the slot, all subsequent calls suppress until verdict)
- Malformed state: `_load_breaker` wraps hydration in broad `try/except`; any `ValueError`/`TypeError` resets the slot to CLOSED instead of crashing dispatch

72 tests pass, all pre-commit hooks green.

TimeToBuildBob:
@greptileai review

packages/gptme-block-registry/src/gptme_block_registry/contract.py
```diff
@@ -0,0 +1,143 @@
...
+    """Parse a deadline; ``None`` means garbled/empty (reader fails open).
+
+    Naive timestamps are read as UTC rather than rejected — the wire shape is
+    UTC and a missing offset is a writer bug we can still interpret safely.
+    """
+    stripped = raw.strip()
+    if not stripped:
+        return None
+    try:
+        ts = datetime.fromisoformat(stripped.replace("Z", "+00:00"))
```
TimeToBuildBob: ❌ **P1** — In `contract.py`, `is_canonical_timestamp` uses `datetime.fromisoformat(stripped.replace("Z", "+00:00"))`. This means a string like `2026-09-19T00:00:00+00:00` is canonical, but also a string like `2026-09-19T00:00:00+00:00.123` would parse and then `format_timestamp(parsed)` would produce `2026-09-19T00:00:00+00:00` (dropping microseconds), so it would NOT equal the stripped string, thus not canonical. That is correct. However, `fromisoformat` in Python 3.10 accepts a space instead of 'T'? Actually `fromisoformat` accepts a space as a separator in some versions? In Python 3.10, it accepts a space? The test `test_canonical_timestamp_detection` asserts that `2026-09-19 00:00:00+00:00` is NOT canonical. But `fromisoformat` in Python 3.10 does accept a space? Let me check: Python 3.10's `datetime.fromisoformat` accepts a space as a separator? Actually, the docs say it accepts any single character separator? In Python 3.10, it accepts a space? I recall that `fromisoformat` in 3.10 is more lenient and accepts a space. If so, `is_canonical_timestamp` would parse `2026-09-19 00:00:00+00:00` and then `format_timestamp` would produce `2026-09-19T00:00:00+00:00`, which is not equal to the stripped string, so it returns False. So the test passes. But the issue is that `parse_until` also uses `fromisoformat` and would accept a space, and then treat it as a valid timestamp. That means a non-canonical timestamp like `2026-09-19 00:00:00+00:00` would be parsed and used for blocking, even though it is not canonical. The contract says only canonical shapes may be compared lexicographically or trusted as a floor, but `parse_until` is used by readers to determine if a block is active. If a writer (e.g., a shell script) writes a space-separated timestamp, the reader would parse it and block, even though it is not canonical. This is actually a fail-open? No, it would block, which is fail-closed. The contract says a non-canonical existing value is treated as unknown and overwritten by writers, but readers still parse it. This is a potential inconsistency: a non-canonical timestamp that is parseable will be honored by the reader, even though the writer would overwrite it. This could lead to a block being active when it should be considered garbled and fail-open. For example, if a buggy writer writes `2026-09-19 00:00:00+00:00` (space instead of T), the reader will parse it and block until that time, even though the contract says non-canonical shapes are not trusted. The test `test_parse_until_accepts_z_and_naive` does not cover space-separated. The test `test_canonical_timestamp_detection` asserts that space-separated is not canonical, but `parse_until` would still parse it. This is a real bug: `parse_until` should reject non-canonical shapes if the contract says only canonical shapes are valid. The docstring for `parse_until` says 'Naive timestamps are read as UTC rather than rejected' but does not mention space-separated. The contract says 'Fixed-width ISO-8601 UTC with a +00:00 offset' and 'Lexicographic comparison is valid only for this exact shape'. The reader should probably only accept the exact canonical shape, or at least reject space-separated. This is a correctness issue: a malformed timestamp that is parseable by `fromisoformat` but not canonical will be treated as a valid deadline, potentially blocking an arm when it should fail open. I'll report this as a bug.

packages/gptme-block-registry/src/gptme_block_registry/slot_circuit_breaker.py
```diff
@@ -0,0 +1,297 @@
...
+
+    if verdict == AUTH_DEATH:
+        if b.is_open(now):
+            # Already / still open — record nothing, suppress.
+            return True
+        b.record_failure(now)
+        return b.is_open(now)
+
+    # PRODUCTIVE
+    if b.is_open(now):
```
TimeToBuildBob: ❌ **P1** — In `_apply_verdict`, the PRODUCTIVE branch calls `b.is_open(now)` first. If the breaker is OPEN and the cooldown has elapsed, `is_open` transitions it to HALF_OPEN and sets `_probe_pending = True`, then returns False (probe allowed). The PRODUCTIVE branch then sees `is_open` returned False, so it does NOT suppress, and calls `b.record_success()`, which closes the breaker. This means a PRODUCTIVE verdict arriving after the cooldown has elapsed will close the breaker even though no probe was actually sent. The comment says 'the productive run happened *before* this call; closing on its behalf would be premature', but the code only suppresses when `is_open` returns True. When the cooldown has elapsed, `is_open` returns False and the breaker is closed. This is a logic error: a PRODUCTIVE verdict should only close the breaker if a probe was actually allowed and succeeded, but here it closes it merely because the cooldown elapsed. The consequence is that a slot that was open and whose cooldown elapsed will be closed by a PRODUCTIVE verdict from a run that started before the breaker opened, allowing immediate respawns without a real probe. The test `test_productive_closes_after_cooldown` only checks the case where cooldown has NOT elapsed, so it misses this. The fix is to check the state before calling `is_open`, or to treat PRODUCTIVE while OPEN as suppress regardless of cooldown, and only allow PRODUCTIVE to close when in HALF_OPEN after a probe.

packages/gptme-block-registry/tests/test_slot_circuit_breaker.py
```diff
@@ -0,0 +1,296 @@
...
+        assert b.state == CircuitState.CLOSED
+        assert b.failure_count == 0
+
+    def test_malformed_failure_count_resets_to_closed(self) -> None:
+        """A non-int failure_count (e.g. 'invalid') must reset the slot, not raise."""
+        b = _load_breaker({"failure_count": "invalid", "state": "OPEN"}, 3, 300.0)
+        assert b.state == CircuitState.CLOSED
+        assert b.failure_count == 0
+
+    def test_malformed_slot_value_resets_to_closed(self) -> None:
```
TimeToBuildBob, line 146: ⚠️ **P2** — In `test_slot_circuit_breaker.py`, `test_malformed_slot_value_resets_to_closed` has a docstring that says 'A slot value that isn't a dict (e.g. bare int) must reset, not raise.' But the test actually passes a dict with a bad `opened_at`, not a non-dict slot value. The test name and docstring are misleading. The test asserts that state is preserved (OPEN) and `_opened_at` is None. This is a test-defect because the test does not test what it claims to test. The comment in the test even acknowledges that `_load_breaker` receives only the slot-level dict, and the upstream parse resets a non-dict top-level. So the test is not actually testing a non-dict slot value. This is a vacuous test for the stated scenario. I'll report this as a test-defect.

TimeToBuildBob:
## P1 fixes in `78afb2f7`

**contract.py — `parse_until` now rejects non-canonical shapes**

`fromisoformat` accepts space-separated and non-UTC-offset timestamps that are not canonical per the contract. These parseable-but-non-canonical shapes now fail open (return `None`): only the canonical `+00:00` UTC shape, Z-suffix, and T-separated naive timestamps are accepted. Added 4 parametrized regression tests (`test_parse_until_rejects_non_canonical_shapes`).

**slot_circuit_breaker.py — `_apply_verdict` no longer routes through `is_open()`**

Root cause: `is_open()` has a side-effect (OPEN→HALF_OPEN transition + `_probe_pending=True`). Two bugs resulted:
1. PRODUCTIVE+OPEN+cooldown-elapsed: `is_open()` returned False after transitioning to HALF_OPEN, falling through to `record_success()` — breaker closed without a real probe.
2. PRODUCTIVE/AUTH_DEATH+HALF_OPEN+probe-consumed: `is_open()` returned True (suppress), verdict was never processed — breaker stayed HALF_OPEN indefinitely.

Fix: capture `state = b._state` before any `is_open()` calls; route HALF_OPEN+probe-consumed verdicts directly to `record_success()`/`record_failure()`. Added 3 new tests.

packages/gptme-block-registry/src/gptme_block_registry/slot_circuit_breaker.py
```diff
@@ -0,0 +1,297 @@
...
+            if self._opened_at is not None and (now - self._opened_at) >= self.cooldown:
+                self._state = CircuitState.HALF_OPEN
+                self._probe_pending = True
+            else:
+                return True
+        # HALF_OPEN: exactly one probe allowed — consume the slot on first call
+        if self._probe_pending:
+            self._probe_pending = False
+            return False
+        return True
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in `78afb2f7`: `_apply_verdict` now captures `state = b._state` before calling `is_open()`. HALF_OPEN+probe-consumed is now routed directly to `record_success()` or `record_failure()` — the path that was suppressed was PRODUCTIVE/AUTH_DEATH arriving after the probe slot was consumed; `is_open()` returned True (suppress) so the verdict was dropped. 3 new tests cover the two previously-infinite-loop cases: `test_productive_closes_after_probe_in_half_open`, `test_auth_death_reopens_after_probe_in_half_open`, `test_half_open_with_probe_pending_stays_open_on_early_verdict`.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

packages/gptme-block-registry/src/gptme_block_registry/slot_circuit_breaker.py
```diff
@@ -0,0 +1,320 @@
...
+        fcntl.flock(fd, fcntl.LOCK_EX)
+        raw = os.read(fd, 1 << 20).decode("utf-8", errors="replace") or "{}"
+        try:
+            data = json.loads(raw)
+            if not isinstance(data, dict):
+                data = {}
+        except (json.JSONDecodeError, ValueError):
+            data = {}
+        yield data
+        out = json.dumps(data, indent=2, sort_keys=True).encode("utf-8")
```
TimeToBuildBob: ❌ **P1** — In `_locked_state`, the state file is opened with `os.O_RDWR | os.O_CREAT` and then read with `os.read(fd, 1 << 20)`. The file offset after the read is at the end of the file. On exit, the code does `os.lseek(fd, 0, os.SEEK_SET)` before writing, so the write starts at the beginning. However, if the new JSON is shorter than the old JSON, the `os.ftruncate(fd, 0)` truncates the file to zero before writing, so no stale tail remains. That part is fine. The real issue is that `os.read` on a file opened without `O_APPEND` reads from the current offset, which is 0 initially, so it reads the whole file. That is fine. The bug is that the lock is held on the state file itself, and the file is truncated and rewritten while holding the lock. A concurrent reader that opens the file without the lock (e.g., a monitoring script or another process that reads the JSON without taking the lock) can observe a truncated or partially written file. The docstring says the lock file is the state file itself, but readers that don't take the lock are not protected. The PR's own tests only exercise concurrent writers through `decide_respawn`, which all take the lock. The contract for the circuit breaker state file is that it is a JSON file; any external reader that reads it without the lock can see a truncated file. This is a real concurrency bug because the write is not atomic (truncate + write, not a temp-file rename). The observable consequence is that a concurrent reader of the state file can get a JSONDecodeError or a partial JSON object, which could crash a monitoring/alerting path or another consumer that reads the state file without the lock. The fix is to write to a temp file and atomically rename it, or to write the new content to a separate temp file and `os.replace` it over the state file while holding the lock. Since the lock is on the state file itself, renaming over it would replace the inode, and the lock would be on the old inode, so that would break the lock. Better to use a separate lock file (like `writers.py` does) and write the state file atomically via temp+rename. This is a P1 concurrency bug.

packages/gptme-block-registry/src/gptme_block_registry/writers.py
```diff
@@ -0,0 +1,105 @@
...
+    The read-compare-write is serialized by an advisory lock, and the write
+    itself is atomic (temp file + rename) so a concurrent reader never observes
+    a truncated or partial timestamp.
+    """
+    path = Path(path)
+    path.parent.mkdir(parents=True, exist_ok=True)
+    new_raw = format_timestamp(until)
+
+    with _locked(path):
+        if path.is_file():
```
TimeToBuildBob, line 61: ⚠️ **P2** — In `write_block`, the never-shorten check compares `existing > new` and returns early if the existing deadline is later. However, the comparison is done on `datetime` objects parsed from the canonical timestamps. The `parse_until` function accepts Z-suffixed and naive timestamps as well, but `is_canonical_timestamp` only returns True for the exact `+00:00` shape. So the existing value is guaranteed to be canonical if `is_canonical_timestamp` is True. That is fine. The bug is that `write_block` does not handle the case where the existing file is a directory. If `path` is a directory, `path.is_file()` returns False, so it skips the read and proceeds to `tempfile.mkstemp` and `os.replace(tmp_name, path)`. `os.replace` on a directory will raise `IsADirectoryError` (or `OSError` on some platforms). This is an unhandled exception that propagates to the caller. The contract says a corrupt block file must never brick an arm, but a directory at the block path would cause `write_block` to raise, potentially crashing the writer. The same applies to `clear_block`: `path.unlink()` on a directory raises `IsADirectoryError`, which is not caught (only FileNotFoundError is caught). The observable consequence is that a misconfigured state directory (e.g., a directory accidentally created at the block path) causes the writer to crash, which could prevent the arm from being blocked or cleared. This is a P2 robustness issue. The fix is to check `path.is_dir()` and treat it as non-canonical (overwrite) or raise a clear error.

packages/gptme-block-registry/src/gptme_block_registry/contract.py
```diff
@@ -0,0 +1,152 @@
...
+    if not stripped:
+        return None
+    normalised = stripped.replace("Z", "+00:00")
+    try:
+        ts = datetime.fromisoformat(normalised)
+    except ValueError:
+        return None
+    if ts.tzinfo is None:
+        # Naive: only accept the T-separated ISO shape, not a space separator.
+        if len(stripped) < 11 or stripped[10] != "T":
```
TimeToBuildBob: ⚠️ **P2** — In `parse_until`, the guard for naive timestamps only checks that the string is at least 11 characters with a `T` at index 10, but does not enforce the fixed-width `HH:MM:SS` shape. `datetime.fromisoformat` in Python 3.11+ accepts truncated shapes such as `2026-09-19T00:00` (minutes only) or `2026-09-19T00` (hour only), so `parse_until("2026-09-19T00:00")` returns a valid datetime instead of `None`. The contract says a garbled timestamp must fail open; a truncated timestamp is garbled because it is not the canonical `+00:00`-seconds shape. A block file containing such a value would be treated as a future deadline and block the arm, violating the fail-open guarantee. Meanwhile a writer would treat the same file as non-canonical and overwrite it, so the reader and writer disagree on the file's meaning.

packages/gptme-block-registry/src/gptme_block_registry/slot_circuit_breaker.py
```diff
@@ -0,0 +1,320 @@
...
+        opened_at = slot_state.get("opened_at")
+        if (
+            b._state in (CircuitState.OPEN, CircuitState.HALF_OPEN)
+            and opened_at is not None
+        ):
+            try:
+                b._opened_at = float(opened_at)
+            except (TypeError, ValueError):
+                b._opened_at = None
+        b._probe_pending = bool(slot_state.get("probe_pending", False))
```
TimeToBuildBob: ⚠️ **P2** — In `_load_breaker`, `b._probe_pending = bool(slot_state.get("probe_pending", False))` coerces any non-empty value to `True`. A corrupted state file that stores `"probe_pending": "false"` (as a string) loads as `True`, not `False`. The module docstring promises "Any malformed field resets the whole slot to CLOSED", but this field is not type-checked, so a malformed string does not raise and does not reset the breaker. If the persisted state is `HALF_OPEN` with `probe_pending` incorrectly `True`, the next `is_open` call consumes the slot and returns `False`, granting a respawn that should be suppressed if the original probe was already used. This fails open on corrupt state, contrary to the designed fail-safe behavior.

packages/gptme-block-registry/src/gptme_block_registry/slot_circuit_breaker.py
TimeToBuildBob: Fixed in `08b0d38b`: `_locked_state` now locks a sibling `.lock` file (like `writers.py`'s `_locked`) and replaces the state file via temp+rename instead of truncate-in-place, so a reader that opens the file without the lock never observes a truncated/partial JSON. Added a regression test asserting the inode changes on write and the `.lock` sibling exists.

packages/gptme-block-registry/src/gptme_block_registry/slot_circuit_breaker.py
TimeToBuildBob: Fixed in `08b0d38b`: `_load_breaker` now type-checks `probe_pending` (raises → caught by the outer reset-to-CLOSED handler) instead of `bool()`-coercing it, so a corrupted string value like `"false"` no longer silently evaluates truthy. Added a regression test.

packages/gptme-block-registry/src/gptme_block_registry/writers.py
TimeToBuildBob: Fixed in `08b0d38b`: `write_block`/`clear_block` now raise a clear `IsADirectoryError` when the block path is a directory, instead of letting `os.replace`/`path.unlink` fail with an opaque low-level error. Added regression tests for both.

packages/gptme-block-registry/src/gptme_block_registry/contract.py
TimeToBuildBob: Fixed in `08b0d38b`: `parse_until`'s naive-timestamp branch now requires the exact fixed-width `HH:MM:SS` shape (`stripped == ts.isoformat(timespec="seconds")`) instead of only checking for a `T` at index 10, so truncated shapes like `"2026-09-19T00:00"` that `datetime.fromisoformat` accepts on 3.11+ are rejected rather than silently parsed. Added regression test cases.

TimeToBuildBob:
@greptileai review

PR #1683, opened 2026-09-18
test(quota-gate): contract tests for the check-usage → quota-gate JSON seam

TimeToLearnAlice:
## What

Adds `tests/test_quota_gate_contract.py` — contract tests for the seam between
`scripts/check-claude-usage.sh --json` and `scripts/quota-gate.sh`.

## Why

`quota-gate.sh` is sourced by every agent's autonomous run script before a
session and reads specific top-level JSON keys from the usage probe
(`five_hour.utilization`, `seven_day.utilization`, `seven_day_sonnet.utilization`,
`_pacing`). Nothing defended that seam: if a rename in *either* file drifts the
key names apart, the gate silently reads `0` for every utilization and **never
blocks** — the exact 3am failure where an agent burns through quota because the
guard quietly became a no-op.

This is the "contract tests for the shared shell surface" item from the
shared-core convergence arc (Phase 0): the undefended seams where a contrib
rename breaks every consumer silently.

## What's pinned

1. **Dual mode** — sourcing defines the function without running the gate; the
   run-on-execute guard only fires when executed directly.
2. **Exit contract** — 0 (proceed) below threshold, 1 (skip) at/over it, and
   fail-*open* (0) when the probe is missing or emits garbage (a broken quota
   probe must never wedge the lane shut). Covers the env-threshold override and
   the sonnet-uses-`seven_day_sonnet` branch.
3. **Key binding** — the JSON keys the gate reads are asserted to exist in both
   scripts statically, so a rename in either file fails this test rather than a
   live run.

9 tests, ~0.8s, no network / no real `claude` invocation (the probe is faked via
a temp script and `QUOTA_CHECK_SCRIPT` override).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


TimeToLearnAlice:
@greptileai review

TimeToLearnAlice:
Addressed all three Greptile P2 findings (commit 186ce97d):

- **Key binding misses producer**: Now checks `check-claude-usage-parser.py` (actual JSON emitter) + `_pacing` key + uses quoted-key form to prevent false matches.
- **Direct execution untested**: Added two direct-execution tests. Required a one-line fix in `quota-gate.sh` to honor a pre-set `QUOTA_CHECK_SCRIPT` env var (`${:-}` default), which also benefits operators.
- **Threshold equality untested**: Three new boundary tests pin the strict-less-than contract (utilization == 0.90 → SKIP).

14/14 tests pass. All 3 Greptile threads resolved. Waiting for Greptile re-review — ready for maintainer merge once 5/5.

TimeToLearnAlice:

ErikBjare:
@TimeToLearnAlice conflicts

TimeToLearnAlice:
Thanks for resolving and merging, Erik. The conflict was self-inflicted: I cut this branch fresh off `origin/master` (deliberately, since our submodule pin is ~205 behind) so it drifted against master under review. Going forward I'll rebase these small contract-test PRs onto master right before flagging them merge-ready, so they land clean without a manual resolve on your end.

PR #1685, opened 2026-09-18
feat(precommit): agent-neutral contrib fork-marker validator

TimeToLearnAlice:
## What

Generalizes Bob's `validate-contrib-fork-marker` hook (ErikBjare/bob, added 2026-08-06) into contrib as an **agent-neutral** governance primitive.

A file in an agent's `scripts/` that shares a path with `gptme-contrib/scripts/<same-path>` must carry a fork marker in its first 30 lines. This makes an intentional local fork *declared*, so a contrib fix to the same path can't silently fail to reach the agent (the silent-no-op class Bob hit: a session fixes `scripts/…`, marks the task done, but the live hook sources contrib's copy which still has the bug — no marker warned the editing session an original existed).

## Why this belongs in contrib

This is the governance primitive for the shared-core convergence arc ([alice#79](https://github.com/ErikBjare/alice/issues/79)): the thing that distinguishes intentional specialization from silent drift. It already works in production on Bob's side; lifting it to contrib lets **any** forked agent adopt it, which is the whole point of the arc — the enforcer of "consume contrib" should itself be consumable.

## Agent-neutral design

- **Default marker: `AGENT-LOCAL`** — works out of the box for a fresh fork, no config.
- **Per-agent override: `CONTRIB_FORK_MARKER`** (e.g. `BOB-LOCAL`) — keeps Bob's existing markers valid without a mass re-marking.
- The universal `AGENT-LOCAL` token is **always** accepted alongside any override, so a file marked for one agent stays valid if the tree is reused/forked by another.
- Case-insensitive; checks first 30 lines; no-op when `gptme-contrib/scripts/` isn't present (e.g. contrib's own CI).

## Adoption

```yaml
- id: validate-contrib-fork-marker
  name: Require AGENT-LOCAL marker on scripts forked from gptme-contrib
  entry: python3 gptme-contrib/scripts/precommit/validators/validate_contrib_fork_marker.py
  language: system
  files: ^scripts/
  exclude: ^gptme-contrib/
  pass_filenames: true
```

## Tests

`tests/test_validate_contrib_fork_marker.py` — 9 tests, ~0.05s, no network: no-counterpart pass, missing-marker fail, universal marker, per-agent override, override-still-accepts-universal, case-insensitivity, marker-past-window fail, no-submodule no-op, no-args.

`main(argv)` and `get_repo_root()` are refactored to be patchable (matches contrib's other validator tests).


TimeToLearnAlice:
@greptileai review

TimeToLearnAlice:
Review heads-up: the latest commit on this branch (`e30d7de0`) also contains `scripts/precommit/validators/check_symlinks.py` (+388) and `tests/test_check_symlinks.py` (+149), which are **not** fork-marker work. They're the shared-core Phase 0 divergence-scanner port — moving the template's `check-symlinks.py` into contrib so any agent (not just the template) can run default-mode. They were swept into this commit by a concurrent Claude session sharing the same working tree (root-caused in ErikBjare/alice#83). The files are sound: 13 tests pass and default-mode output is behaviourally identical to the template original. They're just misfiled here rather than in a dedicated PR. Not splitting them out with a force-push while another session may still be pushing to this branch; happy to do so once the churn settles.

TimeToLearnAlice:

TimeToLearnAlice:
Force-pushed to drop the `check_symlinks.py` scanner that had landed here misfiled via a concurrent-session working-tree collision (ErikBjare/alice#83). This PR now carries only the agent-neutral fork-marker validator + its tests. The scanner is split out into #1687.

PR #1686, opened 2026-09-18
feat(notify): pluggable notify-my-principal seam with Tier-0 fallback

TimeToLearnAlice:
## What

A pluggable **notify-my-principal** seam for shared-core scripts: `scripts/principal_notify.py` + 9 tests.

Part of the shared-core convergence arc ([alice#79](https://github.com/ErikBjare/alice/issues/79)). Core scripts like `self-outage-check.py` need to *escalate out-of-band to the human who owns the agent* — but "escalate" was hardwired to `gh issue create`. For **any forked agent without `gh`**, that alarm silently no-ops — the exact failure it exists to prevent, now with better cover.

## Design

- **Tier 0 — universal.** `notify_principal()` always writes a local alert file that `context.sh` surfaces. Requires only filesystem + python3. Never a silent no-op: if no Tier-1 backend delivers, the local alert still exists and the result is marked `degraded=True`.
- **Tier 1 — pluggable channel.** `github` / `pushover` / `telegram` today, open registry, selected via `PRINCIPAL_NOTIFY_BACKENDS`. **stdlib-only** (subprocess + urllib) — it must run when the model/auth is dead, so no `requests`/gptme imports.
- **Identity rule.** The `github` backend refuses to act when `gh` authenticates as the *principal* (Gordon's PAT-as-Erik anti-pattern) and degrades to Tier-0, so an alarm never appears to come from the person it's meant to alert.

## Why it's for everyone

The audience is every agent forked from the template, not just Alice/Bob/Gordon. A solo user's fork with pushover configured (and no `gh`) now gets alerted when its loop goes dark — instead of a silent no-op.

## Tests

9 tests: Tier-0-always-fires, degraded semantics, backend isolation (a throwing backend can't take out the alarm), the identity guard (refuses-as-principal + files-as-agent), CLI dry-run, env parsing.

First consumer (`self-outage-check.py`) is wired in Alice's repo — defensive, env-gated fan-out to non-github Tier-1 channels; the proven `gh` issue lifecycle is deliberately left in place.


TimeToLearnAlice:
@greptileai review

TimeToLearnAlice:

PR #1687, opened 2026-09-18
feat(precommit): port symlink-hygiene scanner to contrib

TimeToLearnAlice:
Ports the agent-template's divergence scanner into gptme-contrib so **any** agent (not just the template) can run default-mode against its own workspace to find scripts that should be symlinks to contrib but have drifted into verbatim copies.

## What it does
- **default-mode** (for an established agent or the template): flags verbatim copies (hash match → should-be-symlink) and name-drift (same name, different content → likely drifted).
- **forked-mode** (fresh forks / template CI): strict "every script is a symlink or allowlisted" invariant.

## Notes for review
- forked-mode allowlist moved out of the shared script into an optional `--allowlist-file` (built-in defaults kept for backward compat), so a consuming agent can bring its own allowlist without editing the shared file.
- `main(argv)` made patchable; 13 tests in `tests/test_check_symlinks.py`, no network.
- default-mode output verified behaviourally identical to the template's original against a real established-agent workspace.

Part of the shared-core convergence arc (ErikBjare/alice#79, Phase 0 — "make divergence visible").

**Split out of #1685**, where this scanner landed misfiled due to a concurrent-session working-tree collision (ErikBjare/alice#83). #1685 has been rewritten to carry only the fork-marker validator.


TimeToLearnAlice:
@greptileai review

PR #1692, opened 2026-09-18
fix(voice): anchor MAX_CONTEXT_AGE to placed_at, not callback time

TimeToBuildBob:
## Summary

- `_serialize_context()` and `_load_legacy()` in `missed_call_context.py` were checking whether the standup brief was "fresh enough" by comparing `current_time - generated_at <= MAX_CONTEXT_AGE (4h)`
- The brief is generated ~2.5h before the standup call is placed (daily-brief-generator runs at ~05:30; standup placed at ~08:00)
- Any callback >1.5h after the standup would exceed the 4h budget, even though the brief was still valid for that standup context
- Root cause of the 2026-09-18 regression: callback at ~10:23, brief age from callback = ~4h52min > MAX_CONTEXT_AGE → brief not injected → Bob greeted generically instead of serving the prepared standup

## Fix

Anchor the age check to `placed_at` (when the standup call was placed), not `current` (when the callback arrives):

- `generated.date() == placed.date() and timedelta(0) <= placed - generated <= MAX_CONTEXT_AGE`
- This validates the brief was generated on the same morning as the standup, within the 4h pre-standup generation window — not the callback timing

Both code paths fixed: `_serialize_context()` (JSONL / note file path) and `_load_legacy()` (stamp + brief file path).

## Test plan

- Added `test_trusted_callback_at_2h_receives_prepared_brief` regression test simulating the exact 2026-09-18 failure scenario (placed 2h23m ago, brief 4h53m old)
- Existing `test_trusted_callback_at_58_minutes_receives_prepared_brief` (2026-09-17 regression) still passes
- All 133 tests passing (46 standup_callback + 87 missed_call_context)


packages/gptme-voice/tests/test_standup_callback.py
```diff
@@ -406,3 +406,28 @@ def test_trusted_callback_at_58_minutes_receives_prepared_brief(callback_case):
...
+    stamp["date"] = placed.date().isoformat()
+    brief["generated_at"] = (placed - timedelta(hours=2, minutes=30)).isoformat()
+    (state / "standup-brief.json").write_text(json.dumps(brief))
+    (state / "voice-calls/last-standup-call-sid.txt").write_text(json.dumps(stamp))
+    cfg = run()
+    assert (
+        MARKER in cfg.instructions
+    ), "2h+ same-morning callback must inject the prepared standup brief"
+    assert "callback" in cfg.initial_response_instructions.lower()
+    assert len(requests) == 1
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 9f387105: the regression test is now parametrized over all three loaders (legacy stamp, `missed-call-context.json`, `callback-history.jsonl`). Verified all three variants fail against the pre-fix source and pass with the fix. Also added a skip when the scenario would span UTC midnight (loaders require a same-day brief).

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
## 🤖 AI code review

This PR changes the freshness check for standup-brief context in missed_call_context.py so that the brief's age is measured from when the outbound call was placed (placed_at) rather than from the callback time (current). Both _serialize_context() and _load_legacy() are updated to require generated.date() == placed.date() and placed - generated &lt;= MAX_CONTEXT_AGE, while the callback window check (current - placed &lt;= CALLBACK_WINDOW) remains anchored to current. The PR also adds regression tests covering the 2h callback case across all three loaders and guard tests for next-day and past-callback-window scenarios.

Confidence 5/5

- `packages/gptme-voice/src/gptme_voice/realtime/missed_call_context.py` — Changes _serialize_context and _load_legacy to anchor the brief age check to placed_at instead of current time, removing the generated &lt;= placed comparison.
- `packages/gptme-voice/tests/test_standup_callback.py` — Adds parametrized regression tests for 2h callback across legacy/note/history loaders and guard tests for next-day and past-callback-window callbacks.

</details>

| `9f3871055e84` | 4/5 | 1 | llm | 2026-09-18 20:06 UTC |
| `5c9fe13d36f0` | 4/5 | 1 | llm | 2026-09-18 20:19 UTC |

</details>

<sub>Reviewed `3c39be2414e1` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 25s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

packages/gptme-voice/tests/test_standup_callback.py
```diff
@@ -406,3 +407,47 @@ def test_trusted_callback_at_58_minutes_receives_prepared_brief(callback_case):
...
+
+    Covers all three loaders (legacy stamp, missed-call-context.json,
+    callback-history.jsonl): the 2026-09-18 failure had the brief generated
+    ~4h52m before the callback but ~2h30m before the call was placed.
+    """
+    run, _, _, requests, state, stamp, brief = callback_case
+    now = datetime.now(timezone.utc)
+    placed = now - timedelta(hours=2, minutes=23)
+    generated = placed - timedelta(hours=2, minutes=30)
+    if generated.date() != now.date():
```
TimeToBuildBob: ⚠️ **P2** — The new test's parametrized 'note' and 'history' sources write a note with context_file pointing at 'state/standup-brief.json', but the fixture's standup-brief.json is the mutable file that the production write path snapshots into an immutable per-call copy. The test does not exercise the snapshot mechanism; it directly writes the brief file and the note references it. More importantly, the test's assertion `now - generated > MAX_CONTEXT_AGE >= placed - generated` is checked before the test writes the files, but the test does not verify that the brief is actually older than MAX_CONTEXT_AGE relative to the callback in a way that would have failed under the old code. The old code compared `current - generated <= MAX_CONTEXT_AGE`; with placed = now - 2h23m and generated = placed - 2h30m = now - 4h53m, the old condition would be false (4h53m > 4h), so the test would fail on the old code. That part is fine. However, the test's skip condition `if generated.date() != now.date()` means the test is skipped whenever the scenario crosses UTC midnight, which is a non-hermetic dependency on the current wall-clock time; the test can silently pass or skip depending on when it runs. This is a test-defect because the test's validity depends on the environment's current date, and the skip is not deterministic. _(2/3 passes — pass 3 looked and disagreed)_

TimeToBuildBob: Fixed in 5c9fe13d: the test no longer depends on wall-clock time. It now replays the 2026-09-18 timeline on a fixed clock (brief 05:30Z, placed 08:00Z, callback 10:23Z) through `load_callback_candidate(now=...)` for all three loaders, and the UTC-midnight skip is gone. Confirmed all three variants fail against the pre-fix `missed_call_context.py` and pass with the fix (135 tests passing).

TimeToBuildBob:
@greptileai review

packages/gptme-voice/tests/test_standup_callback.py
```diff
@@ -406,3 +407,45 @@ def test_trusted_callback_at_58_minutes_receives_prepared_brief(callback_case):
...
+            "date": "2026-09-18",
+            "placed_at": placed.isoformat(),
+            "caller": PHONE,
+            "context_file": "state/standup-brief.json",
+        }
+        target = (
+            "missed-call-context.json" if source == "note" else "callback-history.jsonl"
+        )
+        (voice / target).write_text(json.dumps(note) + "\n")
+    candidate = load_callback_candidate(str(state.parent), now=now, caller=PHONE)
```
TimeToBuildBob, line 449: ⚠️ **P2** — The new test test_trusted_callback_at_2h_receives_prepared_brief does not actually exercise the callback path through the Twilio trust boundary; it calls load_callback_candidate directly. The fixture's run() path (which goes through handle_twilio_websocket, validates the body grant, and then calls load_callback_brief) is bypassed. This means the test does not verify that the full server flow injects the brief into the session instructions, only that the local candidate loader returns a payload. The existing 58-minute test does go through run(), so the new test is not a full regression test for the reported failure (Bob greeted generically instead of serving the prepared standup). The test would pass even if the server-side integration were broken, as long as the loader returns a candidate. This is a test-defect because it stubs the wrong boundary: it tests the loader in isolation rather than the end-to-end callback behavior that regressed. _(2/3 passes — pass 1 looked and disagreed)_

TimeToBuildBob: Declining this one: the loader is the correct boundary for this change.

- The PR only changes the freshness predicate inside `missed_call_context.py`. The server flow that turns a candidate into session instructions is untouched, and is already covered end to end by `test_trusted_callback_at_58_minutes_receives_prepared_brief` and `test_signed_webhook_preserves_callback_evidence_before_stream`.
- `run()` reads the real wall clock, so replaying the 2026-09-18 timeline (brief 05:30Z, placed 08:00Z, callback 10:23Z) through it needs `datetime` monkeypatching, or relative timestamps. The previous review round flagged relative timestamps as non-hermetic, so the fixed-clock `load_callback_candidate(now=...)` form is what that finding asked for. The file already uses this pattern in `test_callback_crossing_utc_midnight_is_not_same_day`.

Related, in 3c39be24: the merge-gate consensus raised that dropping the `current` check in `_serialize_context` could revive a stale brief on a later day. It can't (`_validate_note` still requires same-day and `current - placed <= CALLBACK_WINDOW` before any call to it), but nothing pinned that. I added `test_placed_at_anchor_does_not_revive_expired_callback` (next-day and past-window callbacks on the note and history loaders); all four cases fail if the `_validate_note` window check is removed.

TimeToBuildBob:
@greptileai review

PR #1698, opened 2026-09-22
feat(lib): add reusable log/disk hygiene mechanism for long-running agents

TimeToLearnAlice:
## What

Adds `scripts/lib/log-hygiene.sh` — mechanism-only disk/log hygiene functions for long-running agents — plus 4 tests.

- **`hygiene_vacuum_user_journal <size> <time> [dry_run]`** — caps the systemd *user* journal by size and age. No root; never touches the system journal. Degrades to a visible no-op when `journalctl` is absent (non-systemd hosts) — Tier-0 safe.
- **`hygiene_prune_old_files <dir> <glob> <keep_days> [dry_run]`** — generic mtime pruning of ephemeral state, glob-scoped, tallies what it removed.

## Why

contrib currently ships **no** log/disk hygiene: no journald vacuum, no ephemeral-state pruning. On a multi-month VM this is a guaranteed, boring, total outage that *every* forked agent will eventually hit. Alice has run this mechanism on a daily cleanup timer; this lifts it into the shared layer so every fork inherits it instead of rediscovering the problem.

## Design

- **Mechanism-only**: every function takes its target as an argument and reads no agent-specific paths. Callers supply *policy* (which dirs, which retention).
- **Safety principle in the header**: these only ever touch ephemeral operational state (gate results, per-session temp caches) and OS journals — never trajectories or session records. Deleting session history is unrecoverable.
- **Capability-tiered**: Tier-0 filesystem for pruning; journal vacuum degrades visibly when the service manager isn't systemd.

## Tests

4 tests pin the properties that matter (a regression here is invisible until a disk fills): correct age cutoff, dry-run touches nothing, glob scoping, missing-dir safe. shellcheck + ruff/mypy clean.

Part of the shared-core-convergence arc (ErikBjare/alice#79).


TimeToLearnAlice:
Pushed a fix to this branch (b3f2cd8e) before merge: `hygiene_vacuum_user_journal` was a **silent no-op** on hosts where `journalctl --user` maps to a root-owned system journal (`Storage=persistent` under `/var/log/journal`). The vacuum fails with "Permission denied", but the function swallowed stderr and reported `before==after` as if it had capped the journal — a Tier-1 action that silently does nothing.

Caught it running the function live against a real 1.2G journal (cap 200M, reclaimed 0). It now captures the vacuum stderr and WARNs that capping needs root (`SystemMaxUse`) — a host task, not an agent one. Hosts where the user genuinely owns their journal are unaffected (no false alarm). +2 tests via a stubbed `journalctl`. shellcheck + full pre-commit green.

TimeToLearnAlice:

TimeToLearnAlice:

TimeToLearnAlice:

ErikBjare:
@TimeToBuildBob Make a PR amending it with your own improvements, or adapt it to a better/more complete approach if you have one, then start using the upstreamed versions.

TimeToBuildBob:
Amendment up at gptme/gptme-contrib#1708: adds `hygiene_prune_uv_cache [dry_run]`.

Key design choice: `uv cache prune --force`, not `uv cache clean` or plain `prune`. On a live agent VM the cache lock is held permanently by the service manager — the non-force variants time out after 60s without reclaiming anything. `--force` removes only unreachable entries without the exclusive lock; safe at any time.

